PR #7, opened 2016-05-27
Bug fix

mguijarr:
the code introduced by commit 3d24e is breaking compatibility with
the Hardware Repository server (still in use on some ESRF beamlines,
mainly because it supports *saving* of XML files when configuration
information is modified modified by hardware objects, eg. when adjusting
beamstop IN position...), so I put back the old code.

By the way I don't understand why self.serverAddress should always be
a list or a tuple ? The documentation says it can be a host:port string
(which is the case when used with Hardware Repository Server).


mguijarr:
*poke*

PR #10, opened 2016-09-19
2016.06.19. Added feature to use nested xmls

IvarsKarpics:
Added feature to redirect a xml to different xml. Feature allows to define a referemce (**hwr_import href**) to different xml. For example

* detector-mockup.xml:

```
<hwr_import href="/detector-mockup-pilatus"/>
```

* detector-mockup-pilatus.xml:

```
<equipment class = "DetectorMockup">
  <tempThreshold>33.5</tempThreshold>
  <humidityThreshold>20.0</humidityThreshold>
  <tolerance>0.2</tolerance>
  <type>pilatus</type>
  <model>6M_F</model>
  <manufacturer>DECTRIS</manufacturer>
  <px>0.172</px>
  <py>0.172</py>
  <hasShutterless>True</hasShutterless>
  <fileSuffix>cbf</fileSuffix>
</equipment>
 ```

If from the gui or hardware objects a **detector-mockup** is requested then **detector-mockup-pilatus** will be loaded.
This idea came up by switching detectors at beamline. Previously the only way was to rewrite xml or redefine all references. This makes life a little bit easier.

### DIFF
--- a/HardwareObjectFileParser.py
+++ b/HardwareObjectFileParser.py
@@ -23,7 +23,7 @@ def parse(filename, name):
         currentXML = f.read()
     except:
         currentXML = None
-        
+       
     xml.sax.parse(filename, curHandler)
     
     return curHandler.getHardwareObject()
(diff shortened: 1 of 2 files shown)


mguijarr:
It complements well the existing xml references feature.

PR #13, opened 2017-01-24
TangoShutter (dual command - read-write attribute)

vrey01:
The main goal of this commit is to change TangoShutter to either use commands configured in xml (as before) or if those are not configure, use the State channel as a read/write attribute.

### DIFF
--- a/HardwareObjects/TangoShutter.py
+++ b/HardwareObjects/TangoShutter.py
@@ -1,90 +1,215 @@
-'''Tango Shutter Hardware Object
-Example XML:
+# -*- coding: utf-8 -*-
+'''
+File:  TangoShutter.py
+
+Description:
+---------------------------------------------------------------
+Hardware Object to provide Shutter functionality through Tango
+or other compatible system. In fact any combination of HardwareRepository
+commands/states will work.
+
+Two possible situations are supported by this hardware object:
+    - One state attribute
+    - Two action commands (in/out, open/close, insert/extract..)
+
+ or 
+    - One read/write attribute with two states
+
+Signals
+---------------------------------------------------------------
+This hardware object will emit signals:
+
+    stateChanged(new_state)
+
+The `new_state` will be one string out of:
+
+    'closed',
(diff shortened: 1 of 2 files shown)


HardwareObjects/TangoShutter.py
```diff
@@ -1,90 +1,215 @@
...
-        #
-        self.shutterStateValue = value
-        # print "emit : ",TangoShutter.shutterState[self.shutterStateValue]
-        self.emit('shutterStateChanged', (self.getShutterState(),))
+        self.state_value_str = self._convert_state_to_str(value)
+        self.emit('shutterStateChanged', (self.state_value_str,))
 
+    def _convert_state_to_str(self, value):
+        state = str(value).upper()
+        if state in self.shutterState:
```
mguijarr: Could be `state = self.shutterState.get(state, "unknown")`

HardwareRepository.py
```diff
@@ -119,7 +119,7 @@ def require(self, mnemonicsList):
             mnemonics = ",".join([repr(mne) for mne in mnemonicsList])
             if len(mnemonics) > 0:
                 self.requiredHardwareObjects = SpecWaitObject.waitReply(self.server, 'send_msg_cmd_with_return' , ('xml_getall(%s)' % mnemonics, ), timeout = 3)
-                logging.getLogger("HWR").debug("Getting all the hardware objects took %s ms." % ((time.time()-t0)*1000))
+                logging.getLogger("HWR").debug("Getting all the hardware objects (%s) took %s ms." % (len(self.requireHardwareObjects), (time.time()-t0)*1000))
```
mguijarr: Would "Getting %s hardware objects took %s ms." be better?

mguijarr:
Thanks for this :)

I made 2 little comments, a bit pedantic I have to say ! So if you have time to change according to the comments it's perfect otherwise I will merge the code tomorrow at least in anyways.

vrey01:
Both comments adopted. :-)

PR #15, opened 2017-02-08
Various bug fixes

mguijarr:
Energy.py: bug fix: truth value of a hardware object can be surprising, better check with None
InstanceServer.py: bug fix: takes qt3 into account, what if QtImport doesn't exist?
HardwareRepository.py: fixed typo
TacoDevice_MTSafe.py: bug fix: reverted Python 3 change


mguijarr:
This PR also reverts some changes introduced by Python 3 adaptations almost 1 year ago, that are in fact causing problems today (like __nonzero__ being removed)

PR #22, opened 2017-06-19
[master] 2017.06.19. Started to clean up directories

IvarsKarpics:
Introduced directories:

abstract
core
generic
mockup


mguijarr:
Hello :)

I wonder if we should do this now, since it would make it more difficult to compare across branches.

I would suggest to wait a bit...

What do others think ?

meguiraun:
hi,

For me, the comparison will be a pain no matter when... But I see that at your place you will have more of that pain than the rest of us. Do you expect pr-s coming from you site soon?

I do not mind waiting some time, but with a clear time frame, otherwise we will meet again in England with this still pending, and we basically all agree on this proposal (I think).

mikel

mguijarr:
I will be able to work on Hardware Objects improvements/cleaning/refactoring starting from September. Not before, unfortunately. Meanwhile, it doesn't prevent anyone from adding tests, if possible, and abstract classes and mockups but I would recommend not changing the file structure yet.

IvarsKarpics:
Ok, lets do changes in structure when the branches are in sync

PR #26, opened 2017-07-04
small additions [2.2]

meguiraun:
small additions

### DIFF
--- a/HardwareObjects/Shapes.py
+++ b/HardwareObjects/Shapes.py
@@ -241,6 +241,8 @@ class Shape(object):
         self.screen_coord = screen_coord
         self.selected = False
         self.refs = []
+        self.overlay_level = 1
+        self.image_ratio = 1
 
         self.add_cp_from_mp(mpos_list)
 
@@ -360,6 +362,18 @@ class Grid(Shape):
 
     def get_centred_position(self):
         return self.cp_list[0]
+
+    def get_grid_range(self):
+        return (float(self.cell_width * (self.num_cols - 1)), \
+                float(self.cell_height * (self.num_rows - 1)))
+
+    def get_num_lines(self):
+        if self.cell_count_fun == "zig-zag":
+            return self.num_rows
+        elif self.cell_count_fun == "inverse-zig-zag":
+            return self.num_cols
+        else:
+            return self.num_rows
 
     def set_id(self, id_num):
         Shape.set_id(self, id_num)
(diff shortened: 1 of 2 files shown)


HardwareObjects/Shapes.py
```diff
@@ -241,6 +241,8 @@ def __init__(self, mpos_list=[], screen_coord=(-1, -1)):
         self.screen_coord = screen_coord
         self.selected = False
         self.refs = []
+        self.overlay_level = 1
+        self.image_ratio = 1
```
marcus-oscarsson: What does image_ratio refer to here, is it the scaling on the client ?. In that case, that scale is a pure client side concept
the server should not care for the scale of the grid.

HardwareObjects/Shapes.py
```diff
@@ -361,6 +363,18 @@ def __init__(self, mpos_list, screen_coord):
     def get_centred_position(self):
         return self.cp_list[0]
 
+    def get_grid_range(self):
```
marcus-oscarsson: Is this not the same as the grid width and grid height ?

marcus-oscarsson:
Hey thanks for the good work,

I just added two questions above that I was not entirely sure of, its otherwise fine 

Marcus

HardwareObjects/Shapes.py
meguiraun: not exactly, and I feel more elegant to retrieve both via a method

HardwareObjects/Shapes.py
meguiraun: I added here for convenience, but also because storing all the info here makes the retrieval of all the shape attributes simpler, one does not need to mix the shape object with the sampleview.imageratio (or whatever) in the state

meguiraun:
I submit this pr since these changes are needed for the mesh scan PR of mxcube3, I bet there will be more additions/changes later on but without these the other PR will not work.

HardwareObjects/Shapes.py
marcus-oscarsson: Hm, it is perhaps practical but the problem is, if it is the scale of the grid, that it wont, for instance, work with several clients because each client might have a different scale. So its not just that its not elegant its simply that I don't think it will work as we like it, or did i misunderstand what image_ratio does ?

meguiraun: You understood perfectly. I was not thinking on several clients... grrr... closing this one.

HardwareObjects/Shapes.py
marcus-oscarsson: I'm trying to understand what the difference is, we can discuss it on our meeting later if you like. Id just like the API of the HardwareObjects to be as clean as possible.

PR #28, opened 2017-07-04
[2.2] Added missing class OpticalCentring (taken from master) refered to from queue_entry.py

marcus-oscarsson:
Hi,

The OpticalCentring class was missing in queue_model_objects_v1.py, I guess its an artifact from a previous merge process. In anyways its the same as for master.

Cheers,
Marcus


marcus-oscarsson:
Sorry, made a small mistake, this class should NOT exist in 2.2

PR #30, opened 2017-07-13
[2.2] Added signals for interleaved entry execution progress

marcus-oscarsson:
Hi,

Just added some signals to be able to track interleaved queue entry execution

Marcus


marcus-oscarsson:
Thanks !

Let me know if you think it could be done in a better way ?

Marcus

IvarsKarpics:
Hi, 

Location of the signals looks fine. We can keep as it is.

Cheers,
Ivars

PR #32, opened 2017-07-18
[2.2] Characterization related methods

marcus-oscarsson:
Hey !

 - Adding a easier way to reference the queue_model

 - Added the possibility to set originating queue_id for nodes that was generated 
by third party software and have another origin than the node directly above it in the tree

Cheers,
Marcus


marcus-oscarsson:
Wow, that was amazingly fast !!!

Thanks !

PR #36, opened 2017-07-26
deperecation fix

beteva:
Fix deprecated exception.message usage


mguijarr:
The PR cannot be merged as it is - it seems like the PR branch is too different from current master branch. My advice is to redo it: `git checkout -b deprecation_fixes` from a freshly pulled master ; then, cherry-pick commits from 2.2 or make new ones. Then, push the branch to make a new PR. 
Thanks a lot ! :)

PR #37, opened 2017-07-27
Small additions [2.2]

meguiraun:


Command/Tango.py
```diff
@@ -63,7 +63,11 @@ def __call__(self, *args, **kwargs):
 
     def abort(self):
         pass
-        
+
+    def setTimeout(self, timeout):
+        if self.device is None:
+            self.init_device()
+            self.device.set_timeout_millis(timeout)
```
mguijarr: Hello :)
This sets the timeout for all commands, not just the targeted one.
There is a possibility to bypass this Tango limitation using PyTango.gevent - I think it would be more correct to have `setTimeout` in `TangoCommand` only if it affects the specified command timeout.
Otherwise what about renaming it `setDeviceTimeout` ?

CommandContainer.py
```diff
@@ -263,7 +263,7 @@ def addChannel(self, attributesDict, channel, addNow=True):
             uribase = attributesDict['taurusname']
 
             try:
-              from Command.Sardana import SardanaChannel
+              from Command.SardanaCommand import SardanaChannel
```
mguijarr: Can you tell more about the conflict that is fixed by this change?
What is wrong with calling the Sardana command file `Sardana.py` ?

meguiraun: My filesystem does not distinguish the capital letter between Sardana.py (for mxcube) and sardana.py (main sardana code in /usr/lib/....). I think this change does not affect anyone since it is kept ~internal.

Command/Tango.py
meguiraun: yep, you are right. I will rename the method

CommandContainer.py
mguijarr: Thanks for the explanation.
From my understanding it means you have another package `command` (or Command) that also has a `sardana` (or Sardana) file. Did you try `from .Command.Sardana import SardanaChannel` (with an added dot `.`) in order to make a relative import ?
Otherwise it means we made a mess with Python paths :( And then we can merge your proposed change but I would also like to fix the paths.

meguiraun: I do not have any `command/Command` package anywhere in the system (at least that I can find or import).

I saw the problem in this import line in the `Sardana.py`
`from sardana.taurus.core.tango.sardana import registerExtensions`, it was actually trying to import from the 'sardana' named file in its current directory ignoring the capital case, and clearly not finding taurus attr etc. I think python paths are fine, this looks to me like a weird case because of my system.

mguijarr: Can you try the following at the top of the `Sardana.py` file:

`from __future__ import absolute_import`

?

It should make the current code working without anything else to change.

meguiraun:
let's see now

PR #40, opened 2017-07-28
[master] 2017.07.28. Reload hardware objects

IvarsKarpics:
Added method to reload hardware objects.
Method is based on reimport package (sudo pip install reimport). It detects files that have been changed. Then these hwobj are disconnected from gui, reimported and connected back to gui.
After connection to gui `update_values` method is called to refresh gui.
Before reimporting all gevent greenlets needs to be stoped and in init relaunched.

This needs more testing

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -299,11 +299,14 @@ class HardwareObjectNode:
         """
         return
 
+    def clear_gevent(self):
+        pass
+
 class HardwareObject(HardwareObjectNode, CommandContainer):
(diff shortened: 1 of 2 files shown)


IvarsKarpics:
Thanks @marcus-oscarsson  for the merge. 
This reloading works fine for a "simple" hardware objects like motors, aperture, shutter, etc. For more complex hwobj like Session, Queue this is not so easy because they rely on several other hwobj and reimporting requires some more actions. 
I will continue to test this and inform how it goes.

Ivars

PR #42, opened 2017-08-04
Changes related mostly with support for ISARA SampleChanger and SampleChangerMaintenance features for mxcube3

vrey01:
Includes different small fixes commits plus

- TangoLimaVideoDevice - independent of Qt (use opencv instead for image encoding)
- SampleChanger (ISARA) and changes for mxcube3

@meguiraun

### DIFF
--- a/HardwareObjects/GenericVideoDevice.py
+++ b/HardwareObjects/GenericVideoDevice.py
@@ -1,0 +1,400 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+[Name]
+GenericVideo
+
+[Description]
+This module declares class GenericVideo.
+
+This class is not meant to be instanced directly but as 
+the base class for classes providing access to Video in MXCuBE
+
+
+"""
+
+import abc
+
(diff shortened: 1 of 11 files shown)


HardwareObjects/GenericVideoDevice.py
```diff
@@ -0,0 +1,400 @@
...
+
+            if self.cam_mirror is not None:
+                qimage = qimage.mirrored(self.cam_mirror[0], self.cam_mirror[1])     
+
+            qpixmap = QPixmap(qimage)
+            self.emit("imageReceived", qpixmap)
+            return qimage
+
+    def get_jpg_image(self):
+        """ for now this function allows to deal with prosilica or any RGB encoded video data"""
```
marcus-oscarsson: Pay attention to double comment, if not made attentionaly ?

HardwareObjects/GenericVideoDevice.py
```diff
@@ -0,0 +1,400 @@
...
+            
+        else:
+            if bw:
+                return qimage.convertToFormat(QImage.Format_Mono)
+            else:
+                return qimage
+
+    def get_scaling_factor(self):
+        """
+        Descript. :
```
marcus-oscarsson: Descript. does not mean anything in sphinx (at least not to my knowledge),  I would suggest to remove it or replace it with something that tells us what the method does. As far as I know returns should also be :returns: .... if we are not using some alternative sphinx syntax ?

HardwareObjects/MAXIV/MaxIVSession.py
```diff
@@ -0,0 +1,46 @@
+"""
```
marcus-oscarsson: Well done, this is exactly how Session was meant to be used. Thanks !!!

marcus-oscarsson:
Hey !

Thanks for the nice work, do you think you could take a look at the comments ? Ill try to take another look next week and merge 

Marcus

mguijarr:
Hello :)

It would be much easier for me if the PR could be split into several smaller ones, one for each thing it is modifying/adding instead of a bunch of commits.

Thanks in advance !

meguiraun:
hej @mguijarr, is there any commit in particular that you have trouble with? So I can reduce the number of PRs, e.g. single PR for the ok commits and another one for the one that you might want to discuss in detail.

mguijarr:
The PRs should be grouped by feature or by bug fix as much as possible, as changes do not only concern site specific files. Otherwise it makes review much more difficult.

meguiraun:
see https://github.com/mxcube/HardwareRepository/pull/48

PR #48, opened 2017-08-18
Fixes 2.2

meguiraun:
@vrey01 's original pr

### DIFF
--- a/Command/Tango.py
+++ b/Command/Tango.py
@@ -4,6 +4,7 @@ import weakref
 import new
 import time
 import types
+import numpy
 import gevent
 import gevent.event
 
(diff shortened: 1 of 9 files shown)


mguijarr:
Hi @meguiraun @vrey01 

Sorry to bother again with the PR - however, is it possible to split the PR in pieces:

1) one for issues / bug fixes = commit 7d79eda and commit 3f650b, eventually commit 9312c7d
2) one for the new TangoLimaVideoDevice & GenericVideoDevice
3) one (or two) with Sample Changer related changes

Indeed, I would happily accept the first PR, then I would like someone like @IvarsKarpics to check the second PR (for me it's fine, but in any case it has nothing to do with bug fixes since it is adding new files,
so it's good to have it as another PR).

Regarding sample changer changes, I have no idea if CATS-related changes are ok for @hellmig or
@jordiandreu ; I do not have a CATS here, so I would like someone else to agree with changes...
There is no point making review if the reviewer cannot really have an opinion on the proposed code.

Moreover, it would be good to not have merge commits in PRs.

Thanks in advance.

meguiraun:
closing this since @vrey01  already did separate PRs

PR #57, opened 2017-08-29
Handle chi angle in MiniDiff, Microdiff and new MD3UP diffractometer

mguijarr:
chi is the goniometer orientation axis (0 degree means : like a MD2, -90 degrees means it's a MD3 UP).

### DIFF
--- a/HardwareObjects/MD3UP.py
+++ b/HardwareObjects/MD3UP.py
@@ -51,10 +51,8 @@ class MD3UP(Microdiff.Microdiff):
         scan_params = "1\t%0.3f\t%0.3f\t%0.4f\t1"% (start, (end-start), exptime)
         scan = self.addCommand({"type":"exporter", "exporter_address":self.exporter_addr, "name":"start_scan" }, "startScanEx")
         scan(scan_params)
-        print "scan started at ----------->", time.time()
         if wait:
             self._wait_ready(300) #timeout of 5 min
-            print "finished at ---------->", time.time()
 
     def oscilScan4d(self, start, end, exptime,  motors_pos, wait=False):
         if self.in_plate_mode():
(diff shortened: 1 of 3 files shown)


HardwareObjects/MD3UP.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+        scan = self.addCommand({"type":"exporter", "exporter_address":self.exporter_addr, "name":"start_raster_scan" }, "startRasterScanEx")
+
+        scan(scan_params)
+        print "scan started at ----------->", time.time()
+        if wait:
+            self._wait_ready(1800) #timeout of 30 min
+            print "finished at ---------->", time.time()
+
```
meguiraun: small thing, it would be better to have log messages instead of prints

HardwareObjects/MD3UP.py
```diff
@@ -0,0 +1,138 @@
...
+        #self.centringVertical.syncMoveRelative((mesh_range['horizontal_range'])/2)
+        #self.centringPhiy.syncMoveRelative(-(mesh_range['vertical_range'])/2)
+        #scan_params = "%0.3f\t" % mesh_range['horizontal_range']
+        #scan_params += "%0.3f\t" % -mesh_range['vertical_range']
+        #scan_params += "%d\t" % mesh_num_lines
+        #scan_params += "%d\t" % (mesh_total_nb_frames/mesh_num_lines)
+        #scan_params += "%r" % True   # TODO
+        #scan = self.addCommand({"type":"exporter", "exporter_address":self.exporter_addr, "name":"start_raster_scan" }, "startRasterScan")
+        #import pdb; pdb.set_trace()
+
```
meguiraun: these comments makes difficult to read, unless you really want to keep them there

HardwareObjects/MD3UP.py
mguijarr: I am going to remove the prints.

PR #59, opened 2017-08-29
TangoLimaVideo.py - support for RGB24

mguijarr:
* 'video_mode' property to be set in XML file
* fix a bug with video_live attribute to prevent server crash
* added example config. in docstring


meguiraun:
pr to master?

mguijarr:
Thanks for the merge :)
I should make a lot of PRs to master but I don't have time for the moment. Branches have diverged quite a lot, I postpone this for later.

meguiraun:
yes, I totally understand, I just made the comment so we do not forget... it will be a pain of merge!

PR #60, opened 2017-08-29
ESRF specific code - TO BE MERGED LAST

mguijarr:


HardwareObjects/MicrodiffHutchTrigger.py
```diff
@@ -77,14 +84,14 @@ def macro(self, entering_hutch, old={"dtox":None}):
         udiff_ctrl = self.getObjectByRole("predefined")
         ctrl_obj = self.getObjectByRole("controller")
         if not entering_hutch:
-            ctrl_obj.detcover.set_out()
             if old["dtox"] is not None:
                 print "Moving %s to %g" % (dtox.name(), old["dtox"])
                 dtox.move(old["dtox"])
+            self.flex_device.eval("flex.user_port(0)")
```
meguiraun: this will fail if `self.flex_device` is not defined, another try/except?

mguijarr: Well, it's ok it's for our internal use at ESRF - flex_device is defined.

meguiraun: ok then

PR #64, opened 2017-08-29
Cats changes

vrey01:
Add a new available signal in GenericSampleChanger.py for SampleChanger objects to emit

Modify heavily Cats90.py - NOTE:  the version in the repository previous to this commit is completely forced-adapted for its use in old MaxLab 911 beamline.  It could not be used anywhere else.   This commit tries to restore generality so that code can be used in other sites.  At the same time it introduces support for ISARA models and adapted general state handling for Cats (as it is used already in other beamlines) that takes into account Powered, Lid state and other DS state information

The CatsMaint.py also adds support for ISARA, plus a number of new methods aiming at a generic SampleChanger-Maintenance object to be used in mxcube3.  If this shows to be of general use (with other types of sample changers), it would be advisable to add a generic class for maintenance type hardware objects.

### DIFF
--- a/HardwareObjects/sample_changer/Cats90.py
+++ b/HardwareObjects/sample_changer/Cats90.py
@@ -3,23 +3,81 @@ CATS sample changer hardware object.
 
 Implements the abstract interface of the GenericSampleChanger for the CATS
 sample changer model.
+
 Derived from Alexandre Gobbo's implementation for the EMBL SC3 sample changer.
 Derived from Michael Hellmig's implementation for the BESSY CATS sample changer
- -fix the Abort Bug
- -enable setondiff for the catsmaint object 
- -fix the bug of MD2 jam during exchange or unload
+
+History:
+   - adds support for ISARA Model
+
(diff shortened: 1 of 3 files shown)


mguijarr:
Code is merged now :)

One remark about your comment, for information to everybody interested in HardwareRepository in general:

> NOTE: the version in the repository previous to this commit is completely forced-adapted for its use in
> old MaxLab 911 beamline. It could not be used anywhere else. This commit tries to restore generality so that code can be used in other sites

I am glad you fixed the code to make it general again. Thanks a lot.

Now, a situation like this should never happen again!!! We (collectively) need to be more conservative and more careful, since changes at one site can break other sites in bad ways. Hence, the importance of doing code reviews (like we try to do now), or testing.

Finally, I merged all this without testing (no CATS or ISARA here), so I am breaking my own rule ; this is because you explicitly mention it improves the situation regarding generality and because I believe you did it carefully from past pull requests. But ideally, no PR should be accepted without code review or
testing (especially when it changes code related to devices or beamline equipment).

PR #66, opened 2017-08-31
GenericVideoDevice allows for non-qt encoding

vrey01:
Add useqt property from xml to decide whether to use qt (qimage) or not for video encoding

Add scale property from xml so that classes can use on reading of images
Fix a bug where if cam_type was not specified, the wrong variable was assigned the default_cam_type

### DIFF
--- a/HardwareObjects/GenericVideoDevice.py
+++ b/HardwareObjects/GenericVideoDevice.py
@@ -73,7 +73,7 @@ class GenericVideoDevice(Device):
         self.default_poll_interval = None
 
     def init(self):
-        modulenames = ['PyQt', 'PyQt5', 'PyQt4']
+        modulenames = ['qt', 'PyQt5', 'PyQt4']
 
         if any(mod in sys.modules for mod in modulenames):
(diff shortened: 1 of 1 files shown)


HardwareObjects/GenericVideoDevice.py
```diff
@@ -75,6 +75,15 @@ def init(self):
...
 
         # Read values from XML
+        #    set useqt in  xml to False if you do not want to use in qt app
+        self.for_qt = self.getProperty("useqt")
+
+        if self.for_qt in [False,"No", "no"]:
+            self.for_qt = False
+        else:
+            self.for_qt = True
+            from QtImport import QImage, QPixmap
```
mguijarr: Would it be a good idea to have this set automatically? Like, if the code is executed within a Qt app (Qt module(s) is(are) already imported) `for_qt=True`, otherwise `for_qt=False` ?
I don't know if there are cases where we are within a Qt app, but still don't want to use qt or the opposite, if we are not in a Qt app we probably don't want to use it.

PR #71, opened 2017-09-08
ISPyBClient2Mockup: anything else than 'idtest000' is considered a wrong login id

mguijarr:
Why making it more complicated?


meguiraun:
to the list of 'to be added to master'

Issue #76, opened 2017-09-18
Improved XMLRPC communication security


olofsvensson:
The current implementation of the XMLRPC communication is very insecure in the sense that any client can connect to the server and execute any command. 

This has up till now not be a problem here at the ESRF. However, last week it started to be a problem when more than one workflow was communicating with the same server. Normally this should not have been possible because an already running workflow should be aborted when a new workflow starts, but because mxcube was restarted between launching of workflows the already running workflow was not aborted.

I would like to improve the security of the XMLRPC server by introducing a token in the communication. The idea is that a random token should be generated at the start of the workflow execution, and this token should then be passed by the workflow in all the calls to the XMLRPC methods. If the token passed by the XMLRPC method doesn't match the one generated at the start of the workflow the method will not be executed.

In order to allow non-workflow clients to access the method a new method will be added than can generate and return a new token. However, as this method can compromise security administrator login credentials will be added as parameters.

In order to allow backward compatibility the XMLRPC server will still work as before if no token is provided. This backward compatibility can later be revoked by simply not allowing method calls without tokens.

mguijarr:
@olofsvensson thanks a lot for this idea.

What about http://code.activestate.com/recipes/501148-xmlrpc-serverclient-which-does-cookie-handling-and/ ?

olofsvensson:
It looks indeed interesting! I would indeed prefer a solution where the authentication is done without changing the signature of the XMLRPC methods.

However, the example above will not fully solve my problem because several workflows on the same machine will be allowed to execute methods at the same time. Maybe the code can be changed to accept only one cookie at a given instance? Even so, the cookie generation is initialised on the client side, which would mean that any successfully authenticated client can re-initialise the cookie during an execution of another workflow...

olofsvensson:
I think I have a solution, please have a look on the commit https://github.com/olofsvensson/HardwareRepository/commit/d7cd536901552db36e7ff3b561a588c73de2d860
The connection from the workflow HO to the XMLRPC server can certainly be improved, comments are welcome.

mguijarr:
Thanks Olof.

PR #83, opened 2017-09-21
Improved XMLRPC security.

olofsvensson:
This is the PR for issue #76. It implements secure XMLRPC communication based on a token. 

The security is disabled by default, and can be enabled by adding "`<enforceUseOfToken>True</enforceUseOfToken>`" to the XMLRPC configuration XML file.

The token is passed from the EdnaWorkflow HO to the class SecureXMLRpcRequestHandler via a static call. It's working but maybe can be made more elegantly?

I have tested this code on MASSIF 1 on 2017/09/20, and it's working as expected.

If this PR is accepted I'll add similar PRs for the branches 2.1 and 2.2.

### DIFF
--- a/HardwareObjects/EdnaWorkflow.py
+++ b/HardwareObjects/EdnaWorkflow.py
@@ -6,9 +6,11 @@ import gevent
 import pprint
 import httplib
 import logging
+import binascii
 #import threading
 from lxml import etree
 import types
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Thanks for all this nice work Olof !

PR #84, opened 2017-09-21
[master] 2017.09.21. Bugfix for GenericVideoDevice

IvarsKarpics:
Added missing useqt attribute

### DIFF
--- a/HardwareObjects/GenericVideoDevice.py
+++ b/HardwareObjects/GenericVideoDevice.py
@@ -57,6 +57,7 @@ class GenericVideoDevice(Device):
     def __init__(self, name):
         Device.__init__(self,name)
 
+        self.useq = None
         self.cam_mirror = None
         self.cam_encoding = None
         self.cam_gain = None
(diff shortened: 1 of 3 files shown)


HardwareObjects/GenericVideoDevice.py
```diff
@@ -57,6 +57,7 @@ class GenericVideoDevice(Device):
     def __init__(self, name):
         Device.__init__(self,name)
 
+        self.useq = None
```
mguijarr: Be careful there is a typo here: seems to be `self.useqt`

mguijarr:
About `useqt`: in fact we can detect if Qt is in use or not (are Qt modules imported ?) so we may not need this at all. 
We removed the `useqt` flag in 2.2 for MXCuBE 3 ; we can probably make a PR to have both versions in master and 2.2 the same if people are interested in getting rid of `useqt`.
What do you think ?

IvarsKarpics:
Ups... Thank you for pointing out the typo.
Yes you are right. I will cherry pick the commit regarding useqt and submit a new PR.
I will close this PR

PR #88, opened 2017-09-22
[master] 2017.09.22. Cherry-pick commit from 2.2

IvarsKarpics:
Cherry-picked commit from 2.2 to detect if GenericVideoDevice is using qt or not
Fixed variable name


IvarsKarpics:
Detection of qt was already committed in master. Changed `for_qt` to `useqt`, because `use_qt` is used in the code.

Issue #82, opened 2017-09-21
BaseHardwareObjects: Objects exposing internal data structures


rhfogh:
I got bitten by a lack of encapsulation in HardwareObjectNode
The HardwareObjectNode.getProperties function returns the actual, internal dictionary, so that modifying the result of the  function call means you are modifying the object. This was unexpected  and, I would say, fragile. Normally you would expect that the result of a function call could be modified without modifying the original object (encapsulation). So:

- Is this a bug or a feature? 
- What is the MXCuBE policy on encapsulation? For function results? for attributes? for _attributes?
- Is there code that depends on this behaviour?
- Can or should I fix this and add encapsulation, either here, or globally?

IvarsKarpics:
Hi @rhfogh ,

I would say this is an issue (topic) that needs a continues work. Method like `getProperties` should not return internal objects, so my answer is: this is a bug.
Thank you bringing this up and you are welcomed to start improving this. A PR would be highly appreciated!

mguijarr:
Python dictionaries are handled "by reference". So every time a dictionary is returned (or a list, for example - this is true for all mutable objects), the return value is really the object, not a copy.

As far as I understand you modified (by mistake ?) the values returned by a 'getProperties' call that returned a dictionary - indeed, we can expect the framework to be more 'protective' and to not expose objects knowing they can be modified by framework users. Returning a copy (beware: a deep copy !) would make the framework more robust.

So:
* it's more a bug than a feature
* encapsulation rules in MXCuBE are the same as in Python, it is left to the good collaboration of programmers vs users
* I don't think code depends on this behaviour
* A pull request is welcome to fix the problem

Thanks !

rhfogh:
Not sure I agree with this fix - and I am starting to look at an alternative pull request myself. Anyway:

For one thing, you should be returning a copy, as you say.

But more fundamentally, I do not think __attributes are the solution here. The '__' is only really needed to prevent an attribute from being referenced in subclasses, and should, I think, be limited to this case. For normal internal attributes a single underscore prefix should be enough (e.g. '_attribute') to signal that this is an internal attribute.  Python programmers are supposed to be grown-up enough to avoid doing the wrong thing, even if there is no one around to say 'stop'. I would propose this kind of hierarchy:

[...]

PR #93, opened 2017-09-27
[master] 2017.09.27. First fix for #82

IvarsKarpics:
First fixt for issue #82
From now all properties has to be retrieved with getProperty()

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -46,7 +46,8 @@ class PropertySet(dict):
 class HardwareObjectNode:
     def __init__(self, nodeName):
         """Constructor"""
-        self.__dict__['_propertySet'] = PropertySet()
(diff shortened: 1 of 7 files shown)


BaseHardwareObjects.py
```diff
@@ -275,15 +276,15 @@ def setProperty(self, name, value):
...
+        self.__propertySet.setPropertyPath(name, self._path+'/'+str(name))
         
 
     def getProperty(self, name, default_value=None):
-        return self._propertySet.get(str(name), default_value)
+        return self.__propertySet.get(str(name), default_value)
 
     def getProperties(self):
-        return self._propertySet
+        return self.__propertySet
```
IvarsKarpics: Probably here should we return `copy(self.__propertySet)` ?

PR #94, opened 2017-09-29
Encapsulated some internal lists / dicts 

rhfogh:
Changed cases where a get function returned internal lists or dicts to return a copy.

NB1 This is not tested - I do not have a test set-up for master

NB2 Two clear cases were left alone as being too complicated for em to deal with:
BeamInfo.get_beam_info and predefinedPositionsNamesList

Also changed some out-of-place tab indentations to spaces and some CR-LF to CR

Looks like I combined two commits into this one (sooy). The other one is:

I kept coming across files where the indentation had an inconsistent mix of tabs and spaces. The present pull request replaces all tabs globally with 8*' ' (eight spaces).

NB The changes are eyeballed but NOT tested (I do not have a test set-up for master).

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -283,7 +283,7 @@ class HardwareObjectNode:
         return self._propertySet.get(str(name), default_value)
 
     def getProperties(self):
-        return self._propertySet
+        return self._propertySet.copy()
             
     def update_values(self):
         """Method called from Qt bricks to ensure that bricks have values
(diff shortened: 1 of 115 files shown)


rhfogh:
I should add that there are two cases that I left alone as being too complicated for outsiders to emss with:
BeamInfo.get_beam_info
and the
predefinedPositionsNamesList

mguijarr:
This is a good idea to encapsulate internal lists and dicts however this PR touches 115 files and, for this reason, cannot be merged.

PR #96, opened 2017-10-05
Ispyb

meguiraun:
- usage of a proxy (default to {})
- capitalize some words in order to unify to some other methods

### DIFF
--- a/HardwareObjects/ISPyBClient2.py
+++ b/HardwareObjects/ISPyBClient2.py
@@ -747,12 +747,9 @@ class ISPyBClient2(HardwareObject):
                 if detector:
                     detector_id = detector.detectorId
                     data_collection.detectorId = detector_id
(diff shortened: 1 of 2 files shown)


HardwareObjects/ISPyBClient2.py
```diff
@@ -731,10 +747,12 @@ def _store_data_collection(self, mx_collection, beamline_setup = None):
                 if detector:
                     detector_id = detector.detectorId
                     data_collection.detectorId = detector_id
-
+	    logging.getLogger('ispyb_client').info("ISPYB NOT WORKING>>> mx_collection: %s" %mx_collection)
```
marcus-oscarsson: Hey, just a detail. These log messages are a bit inconsistent, war are mostly using [ISPyB] .... and here there are two messages with  >>>, we can perhaps try to unify them a bit ?

meguiraun: sorry, I missed those lines. Removed, they were intended only for my eyes...

PR #101, opened 2017-10-26
[2.2] Center on collect

marcus-oscarsson:
Hi

  Assigning position from centering procedure to collection.

Marcus


meguiraun:
same to master?

Issue #135, opened 2018-02-15
Shape get_centred_position v. get_centred_positions


rhfogh:
Some shapes have a get_centred_position function, some have a get_centred_positions function.
I had to put this code into SampleCentringQueueEntry.execute to avoid a crash:

        shapes = list(self.shape_history.get_selected_shapes())
        if shapes:
            pos = shapes[0]
            if hasattr(pos, 'get_centred_position'):
                cpos = pos.get_centred_position()
            else:
                cpos = pos.get_centred_positions()[0]

For now that is fine (and I'll check it in in the big merge coming up) , but could this be rationalised by either 

1) having only get_centred_positions in all files or

2) having get_centred_position everywhere, returning the first centred_position of the list

I do not understand shapes well enough to muck about with them, which is why I am not proposing a refactoring myself.

marcus-oscarsson:
Hi Rasmus,

Yes that is indeed not very good, are you using the shapes from Shapes.py or ShapeHistor.py, I noticed that the Line object and the base Shape object is missing the definition of get_centred_position() in Shapes.py. Ill add that in my next PR to HR,

   The intent of get_centred_position, in Shapes.py is to be a nicer way to do:
       pos.get_centred_positions()[0]

Shapes.py is used in MXCuBE3 and ShapeHisotry is used in Qt3 and I think parts of it or all of it was replaced by Qt4_GraphicsManager in Qt4/Qt5.

Ill do the necessary fix for Shapes.py, however shape.get_centred_positions()[0] should always work
for Shapes.py even today ...

Marcus

IvarsKarpics:
Hi,

Thanks @rhfogh  for pointing this out. We should have one method: get_centred_positions. I will update the Qt4/5 part to use only get_centred_positions.

Ivars

PR #136, opened 2018-02-16
[2.2] Shapes fix

marcus-oscarsson:
Hi

 - Taking care of an issue pointed out by @rhfogh in #135
 -  Using the labels Point, Line and Grid (by default) instead of P, L, and G mainly to avoid confusion with space groups in the case of Points, i.e. P1, P2 and so on.

Marcus


meguiraun:
can you do the same for master?

PR #138, opened 2018-02-16
[master] Redesigned AbstractMotor

IvarsKarpics:
Comments, suggestions? ...


IvarsKarpics:
Few methods are removed to keep the AbstractMotor as simple as possible. If there is a reasonable need, we can add other method.

marcus-oscarsson:
Hi Ivars,

Thanks for this, It depends a bit on the context, is it just for master or also for the new refactoring
that we are currently discussing.

Regardless I think you have a small bug in is_ready and set_ready, set_ready seems to actually do something while is_ready always returns True, which might cause some issues I imagine ?

Otherwise I don't really have anything in particular, perhaps that the constants you define are in camel case I would personally prefer upper case with underscore if its global constants, at least if we go for
PEP8 or similar.

Marcus

IvarsKarpics:
Hi @marcus-oscarsson ,

Thanks for the comments. I updated files:
- removed is_ready that always returns True.
- constants in upper case.
I see this as a not finished work of Mockups and Abstract*. For me first pillar of a clear API is clean and clear abstraction. We should have Abstract* classes as clean and compact as possible. all other things will go to site specific implementations.

marcus-oscarsson:
Thanks okey, great  !. Looks good to me, I'm ready to accept this one if you don't have any objections ?, although there is a conflict (I guess you noticed ? )

IvarsKarpics:
There are typos and few bugs. Fix will come soon...

PR #141, opened 2018-02-20
[master] Small fixes for abstrac and mockup

IvarsKarpics:
* HardwareObjects/AbstractCollect.py: emit progressInit signal in abstractCollect and not in Mockup
	* HardwareObjects/CollectMockup.py
	* HardwareObjects/DiffractometerMockup.py: added set_phase and fixed centering
	* new FluxMockup and PlateManipulatorMockup

### DIFF
--- a/HardwareObjects/FluxMockup.py
+++ b/HardwareObjects/FluxMockup.py
@@ -1,0 +1,115 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


HardwareObjects/FluxMockup.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def get_flux_info(self):
+        return self.flux_info
+
+    def update_flux_value(self):
+        self.flux_info = {"flux" : self.flux_value,
+                          "beam_info": self.beam_info,
+                          "transmission": self.transmission}
+        self.emit('fluxChanged', self.flux_info)
+        """
```
marcus-oscarsson: Its maybe but this entire block seems to be commented ?

IvarsKarpics: Hi @marcus-oscarsson ,

Yes, indeed. The original idea was to recalculate flux when transmission or beam size has been changed. However it is not so straight forward, because rarely beam is homogeneous and flux density is not even. So for time being this block can be commented out and if someone needs it can customize it with some variable and if block.

Cheers,
Ivars

marcus-oscarsson: Aha, I see okey

IvarsKarpics: Thanks...

PR #144, opened 2018-02-22
GphL additions

rhfogh:
Part of series: 3 - merging the GPhL Workflow/emulation branch into master

This PR contains new files added for the GPhL workflow.

There are two main functionalities:
- Interface to the external GPhL workflow programs
- CollectEmulator that (using external GPhL programs) serves as a CollectMockup alternative
that simulates actual images based on teh current collection parameters and one of several test sample records.

Further pull requests to come in the mxcube repository


marcus-oscarsson:
Hey again :),

Well same comment as the PRs on the mxcube repository. Thanks for all this work that you guys have done, you deserve a pint this afternoon ;)

Marcus

HardwareObjects/General.py
```diff
@@ -0,0 +1,92 @@
+#! /usr/bin/env python
```
IvarsKarpics: Hmm, somehow name General is not so descriptive. From the name it should be more or less clear what the hwobj does. If there is no better name we can keep also General.

rhfogh: I felt the need for a single place to put general constants and utility functions not related to any particular hardware object or area. If somebody can think of a better name than 'General' that would be fine. Or maybe it should not be a hardware object at all, but be put somewhere else? Any Ideas?

IvarsKarpics: Maybe it can stay in the HardwareRepository directory. There is now a file TaskUtils. Could this be renamed to GeneralUtils.py?

marcus-oscarsson: Hey, its maybe good to avoid names that are too general and General is a name like that :). Following the existing logic putting it in HardwareRepository would perhaps be better and give it a more specific name. It looks to be that most functions deals with lookups or conversions so maybe ConvertUtils or LookupUtils.

Further I'm personally not a fan of camel case module names, I think we are slowly moving towards a more pep8 compatible code which would mean all lowercase names in modules.

I also noticed that you mix camel and snake case in the functions in this module. It does not really matter for the moment and since its on master Ill let you guys decide how you like it :)

Cheers,
Marcus

rhfogh: - I am converting the camelCase to snake_case. What is the standard procedure here: Should I withdraw the Pull Request and replace it with a fixed one, or should I wait till it is approved and add a new one with the fixes?

- My idea was to have a single, central location for constants (like h_over_e), standard sets of states, general utility functions etc. and encourage people to put them there, and use those already made, instead of making their own separate copies. Using a specific name like TaskUtils or LookupUtils would defeat that purpose, and force me to split the contents of General.  How about GeneralUtils, maybe in HardwareRepository? 

- PEP 8 says:
"Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged."
Changing to PEP8 rules for functions is one thing, but I do not think it makes sense for module and package names. All the modules have CamelCase names at the moment (BlissFramework etc.), so do almost all packages, and the vast majority of those cannot be renamed to follow the recommendation (doorinterlockmockup.py anyone?)
I am up for systematically renaming all modules and packages in the code, if people think it is worth it, but is it not better to stick to CamelCase in this case, otherwise?

marcus-oscarsson: Hey,

- You don't need to withdraw it just do your changes and push to your branch and it will be updated automatically

- As I said since its master I have no specific opinion, I always think its better if we easily can see can deduce something from a name, so that its self documenting. Naming a file to GeneralUtils can mean anything and will likely grow to include all kinds of things. Just my thoughts but I'm fine with the name if you guys like it.

- Well, as I said we are slowly moving towards it means what parts you'd like to adapt is up to you. I try to conform to PEP8 as much as reasonably can. My hope is that we can reuse a big part of the code base in some way or the other for our big refactoring that's coming up, so that would be another reason for trying to be more PEP8 compatible. As I said before its not all that important to me how we do on master right now, since master is very Qt4 specific so Ill let you guys decide on that.

PEP8 says "Underscores can be used in the module name if it improves readability" so for 
doorinterlockmockup.py that would probably mean doorinterlock_mockup or door_interlock_mockup. One could even imagine a sub package for mockups so that it would be
from mockups import doorinterlock

In anyways, this is mostly my thoughts and opinions !

Cheers,
Marcus

IvarsKarpics: Could it be called `ConvertUtils`? Most of the functions convert something. Enumerator `States` could go to `queue_model_enumerables_v1.py`

rhfogh:
It could be called ConvertUtils - but where would h_over_e go?

My point was to propose the idea that there should be a central location to put, and look for, general constants, states, and utilities. So that if you needed a constant like 12.3984 or a function like to_ascii, you would know where to go. I think that it would be one way to avoid duplication of code, and of state enumerations, and could avoid having all those hardcoded values all over the place. If  people think there is merit in that idea, having a ConvertUtils package does not really answer the case. If not, it is likely that no one but me will ever use this - in which case I could put it all in a GPhL-specific location like GphlMessages.

Maybe I should do that anyway, and we can discuss another time if we want a central location for constants?

marcus-oscarsson:
I think that's a really good idea, although mathematical functions and string or other types of conversion should probably be separated. We could look at  h_over_e as a "utility" for conversion from energy to wave length for instance, thus it can go in something called ConvertUtils for the time being.

rhfogh:
OK, then I'll move the file to HardwareRepository/ConvertUtils, and move the states to queue_model_enumerables. Back soon

IvarsKarpics:
Nice discussion :) This is how a co-development should be!

marcus-oscarsson:
Yeah, thanks guys :) :+1:

rhfogh:
Changes are now in.

Thanks for the quick and constructive handling of all this.

HardwareObjects/queue_entry.py
```diff
@@ -28,14 +28,9 @@
 import os
 import autoprocessing
 
-import edna_test_data
-from XSDataMXCuBEv1_3 import XSDataInputMXCuBE, XSDataResultMXCuBE
```
IvarsKarpics: This we need to test Charaterisation. Could you add them back?

IvarsKarpics:
I will accept the PR and test.

rhfogh:
Thanks, Ivars.

My IDE removed those imports automatically as unused; I did go through to check the changes but thought 'what the heck'. Next time I shall leave them in whatever my IDE says.

HardwareObjects/General.py
beteva: Hello,
agrre with @IvarsKarpics - this file should not be called General. Especially there are some imports (like enum) or some definitions (like the states), which are not that general but more hardware specific. ConverUtils sound good to me.

beteva:
Please, do not move the states to any file. They are no so generic.

PR #146, opened 2018-02-26
[master] Bugs in Motor classes

IvarsKarpics:
* Fixed few typos in the AbstractMotor
* Removed MicrodiffMotorMockup. MotorMockup can be used insted. Xmls will be updated asap


HardwareObjects/AbstractMotor.py
```diff
@@ -60,25 +60,31 @@ def tostring(state):
         return MotorStates.STATE_DESC.get(state, "Unknown")
 
 
-
 class AbstractMotor(object):
     __metaclass__ = abc.ABCMeta
 
     def __init__(self):
-        self.__state = MotorStates.INITIALIZING
+        self.motor_states = MotorStates()
```
IvarsKarpics: This avoids creating a new MotorStates instance each time when state is requested. Allows also to compare states in the gui level.

PR #147, opened 2018-03-05
Fixed typo - added ':' in dictionary specification

rhfogh:


beteva:
Dear all,
the above fix clearly shows that the code has not been test before putting it to git and it has been accepted without looking in the code. This is scary!!
Could we think of something like pass the file through pep8 (or similar), which will show the typos or the misalignment and **more important - test and check before submitting to the repository.**
Antonia

rhfogh:
Dear All,

I had not thought beyond that particular bug, but I agree with Antonia.

One solution would be to set up a checker, like pylint, and then have people run it. The hard bit is that we need to agree on what warnings and errors we want to take time to fix, and configure the linter to return only those. If there are too many warnings, new ones will just be ignored.

I tried to run the checker in Pycharm (my IDE - highly recommended) across the entire master branch, at a sort of generic medium warning level, just to get an idea. 
That gave me 17700 'line too long', 10000+ 'Unresolved references', about 6000 other warnings, and 4200 weak warnings. Some highlights are:
- 39 syntax errors
- 468 incorrect call arguments
- 381 inconsistent indentations (tabs and spaces mixed)
- 102 files using tabs for indentation
- 57 cases of overriding functions with non-matching parameters
- 19 'trying to call non-callable object'
- 16 incorrect assignments to or from a tuple
- 1662 'Too broad exception clauses'

Likely some of these are actually OK (I did not try to check), but you can see what we would have to deal with.

IvarsKarpics:
Hmm, I submitted a PR which fixes the bug 9 days ago. But it still pending...
Do we have any PR acceptance/merge policy?

PR #148, opened 2018-03-14
[2.2] XRF Fixes and some minor bug fixes

marcus-oscarsson:
Hi,

Fixes for the XRF scan, snapshots and ISPyBClient2 (mostly made by Antonia who are on vacation for the moment)

Marcus


meguiraun:
it was not big so ok, but it was a mix of esrf specific and generic code.

PR #151, opened 2018-03-15
Maxiv updates

meguiraun:


HardwareObjects/MAXIV/BIOMAXEnergy.py
```diff
@@ -0,0 +1,69 @@
...
+        return self.default_en
+
+    def getEnergyLimits(self):
+        if not self.tunable:
+            return None
+
+        if self.energy_motor is not None:
+            try:
+                self.en_lims = self.energy_motor.getLimits()
+		self.en_lims = (float(self.en_lims[0])/1000, float(self.en_lims[1])/1000)
```
marcus-oscarsson: Looks like you have any indentation miss here

HardwareObjects/MAXIV/BIOMAXKafka.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def init(self):
+        """
+        Init method declared by HardwareObject.
+        """
+        self.kafka_server = self.getProperty('kafka_server')
+        self.topic = self.getProperty('topic')
+        self.session_hwobj = self.getObjectByRole('session')
+        self.beamline_name = self.session_hwobj.beamline_name
+	self.file = open('/tmp/kafka_errors.txt', 'a')
```
marcus-oscarsson: Indentation ...

HardwareObjects/MAXIV/BIOMAXResolution.py
```diff
@@ -1,44 +1,64 @@
...
         
+    def res2dist(self, res=None):
+        current_wavelength = self.getWavelength()
+
+        if res is None:
+            res = self.currentResolution
+
+        try:
+            ttheta = 2*math.asin(current_wavelength / (2*res))
+	    return self.det_radius / math.tan(ttheta)
```
marcus-oscarsson: indent ...

HardwareObjects/MAXIV/BIOMAXResolution.py
```diff
@@ -1,44 +1,64 @@
...
+        except Exception as ex:
+	    print ex
```
marcus-oscarsson: indet...

HardwareObjects/MAXIV/BIOMAXTransmission.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class BIOMAXTransmission(Equipment):
+
+    def init(self):
+        self.ready_event = gevent.event.Event()
+        self.transmission_motor = None
+        self.moving = None
+        self.limits = [0, 100]
+	self.threhold = 5
```
marcus-oscarsson: indent ...

HardwareObjects/MAXIV/MachInfo.py
```diff
@@ -0,0 +1,151 @@
...
+class MachInfo(Equipment):
+    default_current = 0
+    default_lifetime = 0
+    default_message = ""
+    default_topup_remaining = 0
+
+    def __init__(self, *args):
+        Equipment.__init__(self, *args)
+        default_current = 0
+	default_lifetime = 0
```
marcus-oscarsson: indent

meguiraun:
shit! i really messed up with the merging! thanks

marcus-oscarsson:
Hey !

I noticed some indentation mistakes, you should probably take a look at those. I just marked the ones that I found when I skimmed through the files. I did not look in details since this is your specific implementation.

Marcus

HardwareObjects/MAXIV/BIOMAXMD3.py
```diff
@@ -448,6 +448,12 @@ def set_phase(self, phase, wait=False, timeout=None):
     # def move_sync_motors(self, motors_dict, wait=False, timeout=None):
     def move_sync_motors(self, motors_dict, wait=True, timeout=30):
         argin = ""
+	try:
+	    motors_dict.pop('kappa')
+	    motors_dict.pop('kappa_phi')
+            logging.getLogger('HWR').info('[BIOMAXMD3] Removing kappa and kappa_phi motors.')
```
rhfogh: At GPhL we have severe problems with this - in fact we have spent several beamline test sessions and much debugging time dealing with similar code at ESRF ID30B.

1) Can you explain what the purpose of this is? We never found that out at ID30B - and it would help figure out how one might deal with the situation.

2) It is highly unexpected that a command in a supertype that takes a set of motors and moves to the position of all the motors is changed in a subtype so that some of the motors are simply ignored. One might almost call it breach of contract. In this case, it seems to be the case that move_sync_motors is ultimately called from AbstractCollect.do_collect, or rather the implementation in the subtypes, which means that kappa and phi positions passed into the dc_parameters_list are ignored without explanation. 

3) The function is changing the contents of the input dictionary. That is a problem waiting to happen - suppose I pass a motor positions dictionary (possibly indirectly) to this function, and then use it again without knowing it has been changed under me?

4) Our use case is that we are setting up multiple collection sweeps, at different kappa, kappa_phi positions,  from a single invocation of our workflow, We do it in the obvious way, by putting the relevant motor positions into the appropriate places in the dc_parameters_list and expecting the code to follow our instructions. If individual beamlines implement their diffractometer so that move_motors ignores some of the motors passed in, how can we possibly make site-independent code? Or, for that matter, even find out what we need to do?

Any help with getting to the bottom of this will be greatly appreciated

meguiraun:
thanks @marcus-oscarsson for the review!

HardwareObjects/MAXIV/BIOMAXEnergy.py
```diff
@@ -0,0 +1,70 @@
+import sys
+import time
+import logging
+import math
+from HardwareRepository.BaseHardwareObjects import Equipment
+from HardwareRepository.TaskUtils import *
```
rhfogh: import * is deprecated in PEP-8, as it confuses readers and IDEs, makes refactoring harder, and risks hard-to-discover name clashes. Could it be done with 
import TaskUtils
[...]
TaskUtils.someFunction(...)
instead?

HardwareObjects/MAXIV/BIOMAXMD3.py
meguiraun: the reason is that we are not using the minikappa at this moment, and thus writing to those motors fails. Maybe I should check the status of the motor before trying to move instead, but this was done probably in a rush a while ago.

In mxcube3 `mxcube/api/v0.1/diffractometer/movables/state` API gives you the status of all the motors of the diffractometer. One could use that info you check if a motor is usable or not. There is a `centring_motors_list`.

In short, quick fix done while the beamline staff was breathing on my neck...

HardwareObjects/MAXIV/BIOMAXEnergy.py
meguiraun: I take note on this.

rhfogh: Yeah, I know, it is indeed in a lot of places. I only commented because it was a new file.

meguiraun:
hi guys, can anyone merge this?

marcus-oscarsson:
Sure, I thought that you maybe wanted to fix the issue Rasmus pointed out, but I can let this one pass since its your specific code.

PR #155, opened 2018-03-22
Added/refactored queue_model_enumerables

rhfogh:
I need (some of) the additional information, and thought it worth having a cleaner and more complete list of space groups.

NB The pull request, as it stands, will change the order in which space group names appear in the space group selection pulldowns (but not the content of the list, obviously). If you prefer to avoid this change, just comment and reject the pull request, and I shall make sure the order is unchanged.


marcus-oscarsson:
It looks good in general, I'm checking with the other people here to see what they think. The input that
I got so far is that not all of these space groups are actually needed when we work with protein structures, and that's why the list was limited to protein space groups.  So the question I then have is simply why do we need other space groups ?

Marcus

rhfogh:
We need at least one - we have a test crystal for diffractometer calibration in some places that is in space group 230 (Ia-3d). 

Whether we need all of them, or the laue groups, is maybe uncertain. But it was actually easier to include the lot than to filter out some of them, and they eat no bread, as they say.  These are standard reference data, that are not going to change, and they are copy-pasted from a tested input file in GPhL software so there is no question of typos (I have found a couple of those in the previous reference data). I can start pruning the list if you think it is preferable, but that would mean having to add things back one by one if they are ever needed.

BTW the CRYSTAL_SYSTEM_MAP dictionary is actually not needed, so one way or the other I shall remove that again at the first suitable opportunity

marcus-oscarsson:
I see I think we are more concerned about the options that are displayed to our users, I think that most users would ask us why we have a fairly big number of space groups in the space group drop-down that are never used.

So we can add Ia-3d, preferably to the end of the list, if you really need it. Do you think that you will use the other non protein related space groups in the near future and in that case for what ?

There is the possibility to add a custom dropdown if its used in one of your workflows, Olof does this for other types of enumerables. 

Marcus

rhfogh:
I may have been unclear here. We never had any need for showing additional space groups in the selection dropdowns. We just needed the data available internally (so we could convert from space group name 'Ia-3d' to the number 230), and as a side effect I derived the space group pulldown list (with the same contents, bar a typo) from the full data dictionary, instead of having a parallel typed list.

If we ever need them I shall indeed make a custom dropdown instead of forcing our needs on everybody else. It just seems nice to have the reference data available centrally.

marcus-oscarsson:
Okey, thanks for clarifying that, I think I misunderstood your first message. I guess its better if the order of the list is maintained but its not essential as long as the content is the same.

PR #160, opened 2018-04-06
[master] AbstractAperture and ApertureMockup

IvarsKarpics:


HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,177 @@
...
+                eval(self.getProperty("diameter_size_list"))
+        except:
+            logging.getLogger("HWR").error(\
+                "Aperture: no diameter size list defined")
+
+        try:
+            self.__position_list = eval(self.getProperty("position_list"))
+        except:
+            logging.getLogger("HWR").error(\
+                "Aperture: no diameter size list defined")
```
rhfogh: Should be 'position list'

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,177 @@
...
+
+class AbstractAperture(HardwareObject):
+
+    def __init__(self, name):
+        HardwareObject.__init__(self, name)
+
+        self.__current_position_name = None
+        self.__current_diameter_index = None
+        self.__diameter_size_list = []
+        self.__position_list = []
```
rhfogh: Would it not be better to use a single underscore?

The double underscore has two functions: 
1)- to make attributes that can be accessed from inside the class but NOT from code in its subclasses
2)- to make it so irritating to access the attributes directly that people shy away from it.

I doubt there is any need for 1) and as for 2) is it not better to rely on people to do the right thing rather than to put up artificial barriers? if the attribute is called self._diameter_size_list that is enough to tell other programmers that this is an internal attribute that should not be messed with, without making it unduly hard for them to do so if they need to (from a subclass, for instance).

IvarsKarpics: Fine for me

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        Sets active diameter index
+
+        Keyword Args:
+            diameter_index (int): selected diameter index
+
+        Emits:
+            diameterIndexChanged (int, float): current index, diameter in mm
+        """
+        self.__current_diameter_index = diameter_index
```
rhfogh: Why no test for index value, as is done for set_position_index?

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,177 @@
...
+        Returns:
+            int: current diameter index
+        """
+        return self.__current_diameter_index
+
+    def set_diameter_index(self, diameter_index):
+        """
+        Sets active diameter index
+
+        Keyword Args:
```
rhfogh: Should be just 'Args'

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,177 @@
...
+        Returns:
+            bool: True if aperture is in the beam, otherwise returns false
+        """
+        return self.__current_position_name != "BEAM"
+
+    def update_values(self):
+        """
+        Reemits all signals
+        """
+        self.emit('positionChanged', self.__current_position_name)
```
rhfogh: Why is it 'diameterIndexChanged', and not just 'diameterChanged' (like it is for 'positionChanged')?

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,181 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
```
rhfogh: We should be using LGPL, Lesser GPL from now on, no?

HardwareObjects/ApertureMockup.py
```diff
@@ -17,87 +17,30 @@
 #  You should have received a copy of the GNU General Public License
```
rhfogh: LGPL right?

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,181 @@
...
+            logging.getLogger("HWR").warning(\
+                "Aperture: Selected position is not in the position list")
+
+    # Other methods -----------------------------------------------------------
+
+    def set_in(self):
+        """
+        Sets aperture in the beam
+        """
+        self.set_position("BEAM")
```
rhfogh: Should this not be set_position_name?

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def set_out(self):
+        """
+        Removes aperture from the beam
+        """
+        self.set_position("OFF")
```
rhfogh: Should this not be set_position_name?

rhfogh:
Hi Ivars,
For now this is likely needed, but for long-term I wonder whether one should consider simplifying the position part - you can set and get both position name and index (well, OK, but do you need both?) with fully configurable position names. But you can also use the set_in, set_out is_out functions, that rely on hardcoded position names 'BEAM' and 'OUT'. Might it not make more sense to have either the hardcoded 'BEAM'/'OUT' names or the configurable names, but not both?

HardwareObjects/ApertureMockup.py
IvarsKarpics: True. See 500f19024cafbe6080cdf94c9564906099c05138

HardwareObjects/AbstractAperture.py
```diff
@@ -0,0 +1,186 @@
...
+
+class AbstractAperture(HardwareObject):
+
+    def __init__(self, name):
+        HardwareObject.__init__(self, name)
+
+        self._current_position_name = None
+        self._current_diameter_index = None
+        self._diameter_size_list = []
+        self._position_list = []
```
rhfogh: These lists are exported unchanged, which means that getting the list and then modifying it will modify the internals of the object. Would it not be better to either have these as tuples (which being frozen can be freely exported) or to export a copy?

Sorry for not seeing these things at once. Rasmus

IvarsKarpics: Yes, it makes sense. I will fix it

PR #165, opened 2018-04-27
[master] EMBL hwobjs

IvarsKarpics:


HardwareObjects/EMBL/EMBLAperture.py
```diff
@@ -5,42 +5,38 @@
...
 
 
 __credits__ = ["EMBL Hamburg"]
-__version__ = "2.3."
 __category__ = "General"
+__license__ = "LGPLv3"
 
 
-POSITIONS = ("BEAM", "OFF", "PARK")
+DEFAULT_POSITION_LIST = ("BEAM", "OFF", "PARK")
```
rhfogh: Must be changed lower down ('POSITIONS' still found in file)

HardwareObjects/EMBL/EMBLAperture.py
```diff
@@ -50,44 +46,47 @@ def __init__(self, name):
...
+            self._diameter_size_list = [10, 20]
 
         self.chan_diameter_index = \
             self.getChannelObject('CurrentApertureDiameterIndex')
         if self.chan_diameter_index is not None:
-            self.current_diameter_index = self.chan_diameter_index.getValue()
-            self.diameter_index_changed(self.current_diameter_index)
+            self._current_diameter_index = self.chan_diameter_index.getValue()
+            self.diameter_index_changed(self._current_diameter_index)
             self.chan_diameter_index.connectSignal(\
```
rhfogh: Redundant backslash

HardwareObjects/EMBL/EMBLEnergyScan.py
```diff
@@ -262,8 +262,8 @@ def scanCommandFinished(self, *args):
 
     def doChooch(self, elt, edge, scan_directory, archive_directory, prefix):
         symbol = "_".join((elt, edge))
-        scan_file_prefix = os.path.join(scan_directory, prefix)
-        archive_file_prefix = os.path.join(archive_directory, prefix)
+        scan_file_prefix = str(os.path.join(scan_directory, prefix))
```
rhfogh: WHy is that needed? Can os.path.join return anything but a string?

IvarsKarpics: If a user changes directory via gui, then attribute is QString. But os.path.join throws error if argument is QString

HardwareObjects/EMBL/EMBLBeamlineTest.py
```diff
@@ -1067,7 +1079,7 @@ def center_beam_task(self):
 
                 if abs(delta_hor) > 0.001:
                     log.info("Beam centering: Moving horizontal by %.4f" % delta_hor)
-                    self.horizontal_motor_hwobj.move_relative(delta_hor, timeout=5)
+                    self.horizontal_motor_hwobj.mov_relative(delta_hor, timeout=5)
```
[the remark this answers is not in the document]
IvarsKarpics: Fixed

HardwareObjects/EMBL/EMBLAperture.py
IvarsKarpics: Will be fixed

HardwareObjects/EMBL/EMBLAperture.py
IvarsKarpics: Hmm, whats wrong here?

rhfogh: Ah, the  comment should actually refer to line 64. 
Backslash line continuations are allowed but discouraged, since their meaning changes if there is a space (invisibly) after the backslash. The recommended style is to avoid backslashes and use open paranthesis (or {, [. ) for line continuations. Admittedly the backslash is sometimes the least bad option when you have to fit a complex expression into 80 characters withotu too many contortions. But in this case you can (and therefore should) simply remove it, since the open parenthesis will ensure that line 65 (etc.) are read together with line 64.

Issue #169, opened 2018-06-04
Ungeneric code in GenericDiffractometer?


rhfogh:
When opening DiffractometerMockup you get error messages in the user interface display:
Unable to add channel <ChannemaName>.

The root is that the GenericDiffreactometer class contains the list 

    CHANNELS_NAME = ["CoaxCamScaleX",
                     "CoaxCamScaleY",
                     "State",
                     "CurrentPhase",
                     "TransferMode",
                     "SampleIsLoaded",
                     "HeadType"]

tries to add each channel, and gives a UI error message for each when it cannot. My latest pull request has stopped the program from re-raising a previous and irrelevant error when this happens. But even so, surely these channels are not universal enough to belong in GenericDiffractometer? And when the program can run without the channels (at another site, or uin mock mode) the fact taht they cannot be connected to surely should not be put in the UI window?

PR #170, opened 2018-06-04
Cleaned error output and reduced missing channel visibility.

rhfogh:
NB The log output for missing channels is still excessively visible.
See issue #169 

This change avoids an unnecessary error raised (which is anyway covered by except: pass) and reduces teh log level from exception to error.

### DIFF
--- a/HardwareObjects/GenericDiffractometer.py
+++ b/HardwareObjects/GenericDiffractometer.py
@@ -248,10 +248,12 @@ class GenericDiffractometer(HardwareObject):
         self.back_light_swtich = self.getObjectByRole('backlightswtich')
 
         # Channels -----------------------------------------------------------
-        try:
-           self.used_channels_list = eval(self.getProperty("used_channels"))
-        except:
-           pass # used the default value
+        ss = self.getProperty("used_channels")
+        if ss:
+            try:
+               self.used_channels_list = eval(ss)
+            except:
+               pass # used the default value
 
         for channel_name in self.used_channels_list:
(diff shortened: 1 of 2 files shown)


HardwareObjects/GenericDiffractometer.py
```diff
@@ -248,10 +248,12 @@ def init(self):
         self.back_light_swtich = self.getObjectByRole('backlightswtich')
 
         # Channels -----------------------------------------------------------
-        try:
-           self.used_channels_list = eval(self.getProperty("used_channels"))
-        except:
-           pass # used the default value
+        ss = self.getProperty("used_channels")
```
IvarsKarpics: Opaa, I would be careful with using ss in the code :) Also it is not so descriptive. Maybe __used_channels instead of ss?

rhfogh: What is the problem - does it make people think of Hitler?

It could be renamed, of course, but the point of that variable name is that it is a temporary string variable that has  no significance outside the immediate context (I would use oo for an object, ll for a list, tt for a tuple, xx for a float, but that is personal style). Much like a loop counter, really. If we call it __used_channels it looks like it is official, important, and people assume that it cannot be changed without carefully checking for where else it might be used. Which I would find misleading. GIven that, what could I name it?

IvarsKarpics: There is no problem (it was a bad joke :) Lets leave it as it is.

rhfogh:
Thanks.

(and sorry for missing the joke).

Rasmus

PR #181, opened 2018-06-08
ISPyBClient2.py and SOLEILISPyBClient

MartinSavko:
-- changing leading '__' to single underscore
    self._shipping
    self._collection
    self._tools_ws
    self._translations
    self._disabled

This is to allow for easier subclassing -- attributes are not easily accessible in children when subclassing due to special treatment defined in BaseHardwareObjects.

Or may there be a workaround I have not thought about ?

### DIFF
--- a/HardwareObjects/ISPyBClient2.py
+++ b/HardwareObjects/ISPyBClient2.py
@@ -20,6 +20,7 @@ from datetime import datetime
 from collections import namedtuple
 from pprint import pformat
 
+import traceback
 
 # Production web-services:    http://160.103.210.1:8080/ispyb-ejb3/ispybWS/
 # Test web-services:          http://160.103.210.4:8080/ispyb-ejb3/ispybWS/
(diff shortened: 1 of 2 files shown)


IvarsKarpics:
Is there a particular reason to subclass ISPyBClient2.py (SOLEILISPyBclient)? I should suggest to use one class (ISPyBClient2), so we all have the same interface to ispyb. Is there something missing?

rhfogh:
No opinion on the need for subclassing here. But on general principles, variables like __xyz should only (IMHO) be used for attributes that must be shielded from access by subclasses, since that is the particular function of the '__' prefix. Variables that are merely internal should have a single underscore. That avoids making things unnecessarily difficult if someone does need to access them, and leaves it to the good sense of other developers to not mess with object internals.

IvarsKarpics:
Hi @MartinSavko ,

Thanks for the explanation. If no one is against these changes in ISPyBClient2 then I will merge the PR.

PR #183, opened 2018-06-08
Microdiff motor modifs

MartinSavko:
-- look for channel and command definition in the configuration file first, fall back to in-place definition if missing (this makes it possible to use the same class for different protocols: e.g. tango, exporter, tango_events ...)
-- modifications to state related methods
-- state no 2 does not seem to correspond to UNUSABLE -- there may be a better way around this ...
-- get_limits() == getLimits()

### DIFF
--- a/HardwareObjects/MicrodiffMotor.py
+++ b/HardwareObjects/MicrodiffMotor.py
@@ -57,18 +57,56 @@ class MicrodiffMotor(AbstractMotor):
 
         self.motorState = MicrodiffMotor.NOTINITIALIZED
         
-        self.position_attr = self.getChannelObject("%s%s" % (self.motor_name, self.motor_pos_attr_suffix))
-        self.state_attr = self.getChannelObject("%s%s" % (self.motor_name, self.motor_state_attr_suffix))
+        try:
+            self.position_attr = self.getChannelObject("%s%s" % (self.motor_name, self.motor_pos_attr_suffix))
+        except:
+            self.position_attr = self.addChannel({"type": "exporter",
+                                                  "name": "%sPosition" %self.motor_name},
+                                                  self.motor_name + self.motor_pos_attr_suffix)
         
         if self.position_attr is not None:
+            try:
+                self.state_attr = self.getChannelObject("%s%s" % (self.motor_name, self.motor_state_attr_suffix))
+            except:
+                self.state_attr = self.addChannel({"type": "exporter",
+                                                   "name": "%sState" %self.motor_name},
+                                                   self.motor_name + self.motor_state_attr_suffix)
+                
             self.position_attr.connectSignal("update", self.motorPositionChanged)
             self.state_attr.connectSignal("update", self.motorStateChanged)
             
-            self._motor_abort = self.getCommandObject("abort")
-            self.get_dynamic_limits_cmd = self.getCommandObject("get%sDynamicLimits" % self.motor_name)
-            self.get_limits_cmd = self.getCommandObject("getMotorLimits")
-            self.get_max_speed_cmd = self.getCommandObject("getMotorMaxSpeed")
-            self.home_cmd = self.getCommandObject("homing")
+            try:
+                self.motors_state_attr = self.getChannelObject("motor_states")
+            except:
+                self.motors_state_attr = self.addChannel({"type":"exporter", "name":"motor_states"}, "MotorStates")
+            self.motors_state_attr.connectSignal("update", self.updateMotorState)
+            
+            try:
+                self._motor_abort = self.getCommandObject("abort")
+            except:
+                self._motor_abort = self.addCommand( {"type":"exporter", "name":"abort" }, "abort")
+                
+            try:
+                self.get_dynamic_limits_cmd = self.getCommandObject("get%sDynamicLimits" % self.motor_name)
+            except:
+                self.get_dynamic_limits_cmd = self.addCommand({"type": "exporter",
+                                                               "name": "get%sDynamicLimits" % self.motor_name},
+                                                               "getMotorDynamicLimits")
+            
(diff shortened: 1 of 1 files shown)


HardwareObjects/MicrodiffMotor.py
```diff
@@ -49,24 +56,58 @@ def init(self):
...
 
         self.motorState = MicrodiffMotor.NOTINITIALIZED
-
-        self.position_attr = self.addChannel({"type": "exporter", 
-                                              "name": "%sPosition" %self.motor_name}, 
-                                               self.motor_name + self.motor_pos_attr_suffix)
+        
+        try:
+            self.position_attr = self.getChannelObject("%s%s" % (self.motor_name, self.motor_pos_attr_suffix))
+        except:
```
IvarsKarpics: except part will never be executed because getChannelObject will return None if loading of a channel fails. It should look like:
```
self.position_attr = self.getChannelObject("%s%s" % (self.motor_name, self.motor_pos_attr_suffix))
if not self.position_attr:
    self.position_attr = self.addChannel({"type": "exporter",
                                                   "name": "%sPosition" %self.motor_name},
                                                   self.motor_name + self.motor_pos_attr_suffix) 
```
Same for state_attr, motors_state_attr, etc..

HardwareObjects/MicrodiffMotor.py
```diff
@@ -20,7 +21,7 @@ class MD2TimeoutError(Exception):
 """
 
 class MicrodiffMotor(AbstractMotor):
-    (NOTINITIALIZED, UNUSABLE, READY, MOVESTARTED, MOVING, ONLIMIT) = (0,1,2,3,4,5)
+    (NOTINITIALIZED, UNUSABLE, READY, MOVESTARTED, MOVING, ONLIMIT) = (0,1,3,4,5,6)
```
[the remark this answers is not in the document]
MartinSavko: When testing the MD2 motors in exporter mode, all motor bricks were going immediately to unusable state (grayed out). With the above change it behaved as expected.

The problem may be in mismatch between hardware object and the brick and may be more appropriately solved in the brick. I will revert the change in the PR.

HardwareObjects/MicrodiffMotor.py
```diff
@@ -57,56 +57,58 @@ def init(self):
...
-            try:
-                self._motor_abort = self.getCommandObject("abort")
-            except:
-                self._motor_abort = self.addCommand( {"type":"exporter", "name":"abort" }, "abort")
-                
-            try:
-                self.get_dynamic_limits_cmd = self.getCommandObject("get%sDynamicLimits" % self.motor_name)
-            except:
+            self._motor_abort = self.getCommandObject("abort")
+            if not self._motro_abort:
```
IvarsKarpics: Typo (self._motro_abort)

HardwareObjects/MicrodiffMotor.py
```diff
@@ -80,32 +123,49 @@ def connectNotify(self, signal):
...
-            return MicrodiffMotor.NOTINITIALIZED
+            if self.state_attr.getValue() in MicrodiffMotor.EXPORTER_TO_MOTOR_STATE:
+                self.motorState = MicrodiffMotor.EXPORTER_TO_MOTOR_STATE[self.state_attr.getValue()]
+            else:
+                self.motorState = MicrodiffMotor.TANGO_TO_MOTOR_STATE[self.state_attr.getValue().name]
+            self.motorStateChanged(self.motorState)
+                #self.updateMotorState(self.motors_state_attr.getValue())
         return self.motorState
     
+    def get_state(self):
```
IvarsKarpics: This is a bit strange, because in the AbstractMotor:
```
def getState(self):
        #TODO remove this method
        #print ("Deprecation warning: Instead of getState please call get_state")
        return self.get_state()
```
Maybe better rename methof getState to get_state to avoid confusion. The same applies to getPosition, getLimits

HardwareObjects/MicrodiffMotor.py
```diff
@@ -80,32 +123,49 @@ def connectNotify(self, signal):
...
         if self.motorState == MicrodiffMotor.NOTINITIALIZED:
-          try:
-            self.updateMotorState(self.motors_state_attr.getValue())
-          except:
-            return MicrodiffMotor.NOTINITIALIZED
+            if self.state_attr.getValue() in MicrodiffMotor.EXPORTER_TO_MOTOR_STATE:
+                self.motorState = MicrodiffMotor.EXPORTER_TO_MOTOR_STATE[self.state_attr.getValue()]
+            else:
+                self.motorState = MicrodiffMotor.TANGO_TO_MOTOR_STATE[self.state_attr.getValue().name]
+            self.motorStateChanged(self.motorState)
```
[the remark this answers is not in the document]
MartinSavko: Looking at the code it does not make much sense to call self.get_state() within motorStateChanged() method as the state parameter is compulsory. Removing the self.get_state() line in that method would avoid infinite loop.

Is there a reason for special treatment of MicrodiffMotor.NOTINITIALIZED in get_state() ? Would it be ok we executed the check in all states ?

```python
def get_state(self):
    if self.state_attr.getValue() in MicrodiffMotor.EXPORTER_TO_MOTOR_STATE:
        self.motorState = MicrodiffMotor.EXPORTER_TO_MOTOR_STATE[self.state_attr.getValue()]
    else:
         self.motorState = MicrodiffMotor.TANGO_TO_MOTOR_STATE[self.state_attr.getValue().name]
    self.motorStateChanged(self.motorState)
    return self.motorState

```

HardwareObjects/MicrodiffMotor.py
MartinSavko: Agree.

PR #184, opened 2018-06-08
Microdiff zoom modifs

MartinSavko:
-- inheriting from MicrodiffMotor instead of MD2Motor.
-- look for channel in the configuration file first, fall back to in-place definition if missing

### DIFF
--- a/HardwareObjects/MicrodiffZoom.py
+++ b/HardwareObjects/MicrodiffZoom.py
@@ -1,23 +1,24 @@
-from MD2Motor import MD2Motor
+from MicrodiffMotor import MicrodiffMotor
 import logging
 import math
 
-class MicrodiffZoom(MD2Motor):
+class MicrodiffZoom(MicrodiffMotor):
     def __init__(self, name):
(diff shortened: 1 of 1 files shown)


HardwareObjects/MicrodiffZoom.py
```diff
@@ -1,23 +1,26 @@
...
         self.motor_name = "Zoom"
         self.motor_pos_attr_suffix = "Position"
         self._last_position_name = None
- 
-        MD2Motor.init(self)
-
-        self.predefined_position_attr = self.addChannel({"type":"exporter", "name":"predefined_position"  }, "CoaxialCameraZoomValue")
-
+        
+        try:
```
IvarsKarpics: Except will never be executed because getChannelObject will return None if loading of a channel fails. This should look like:
```
self.predefined_position_attr = self.getChannelObject("predefined_position")
if not self.predefined_position_attr:
    self.predefined_position_attr = self.addChannel({"type":"exporter", "name":"predefined_position" }, "CoaxialCameraZoomValue")
```

PR #193, opened 2018-06-13
Microdiff motor state

MartinSavko:
-- fix typo if not self.get_max_spped_cmd: -> if not self.get_max_speed_cmd:
 -- add translate directory to translate microdiff states to states defined
    in AbstractMotor -- using so far unused self.motor_states instance.
    The reason we need this is that integer representation of microdiff state
    do not correspond to the integer representation of AbstractMotor (and bricks)
    in general.
    For instance when MD2 motors are in Ready state self.state_attr.getValue()
    returns 'Ready', but MicrodiffMotor.READY is 2  which designates state 'OFF' 
    of the AbstractMotor.
 -- state is currently updated via two signals: 'update' and 'stateChanged'
    there was inconsistency between the two:
    the call defined in connectNotify was calling self.motorStateChanged()
    using self.get_state() as parameter which returns number while 'update'
    handling defined in init() was connected to self.state_attr.getValue()
    which returns a string.
    motorStateChanged() now checks for type of state variable and translates
    state to number if it is a string.
 -- I wonder whether we need both signals as they seem to emitted always
    at the same time ...

### DIFF
--- a/HardwareObjects/MicrodiffMotor.py
+++ b/HardwareObjects/MicrodiffMotor.py
@@ -127,7 +127,7 @@ class MicrodiffMotor(AbstractMotor):
             self.motorLimitsChanged()  
  
     def updateState(self):
-        self.setIsReady(self.motorState > MicrodiffMotor.UNUSABLE)
+        self.setIsReady(self._get_state() > MicrodiffMotor.UNUSABLE)
 
     def setIsReady(self, value):
         if value == True:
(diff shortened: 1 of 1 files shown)


HardwareObjects/MicrodiffMotor.py
```diff
@@ -144,20 +150,18 @@ def updateMotorState(self, motor_states):
...
 
     def get_state(self):
-        if self.motorState == MicrodiffMotor.NOTINITIALIZED:
-            if self.state_attr.getValue() in MicrodiffMotor.EXPORTER_TO_MOTOR_STATE:
-                self.motorState = MicrodiffMotor.EXPORTER_TO_MOTOR_STATE[self.state_attr.getValue()]
-            else:
-                self.motorState = MicrodiffMotor.TANGO_TO_MOTOR_STATE[self.state_attr.getValue().name]
-            self.motorStateChanged(self.motorState)
-                #self.updateMotorState(self.motors_state_attr.getValue())
+        if self.state_attr.getValue() in MicrodiffMotor.EXPORTER_TO_MOTOR_STATE:
```
IvarsKarpics: Instead of calling getValue() 3 times one could read it once and assign value to a temporary variable
`state = self.state_attr.getValue()`

HardwareObjects/MicrodiffMotor.py
```diff
@@ -40,7 +40,13 @@ def __init__(self, name):
         AbstractMotor.__init__(self, name) 
         self.motor_pos_attr_suffix = "Position"
         self.motor_state_attr_suffix = "State"
-    
+        self.translate_state = {MicrodiffMotor.NOTINITIALIZED: self.motor_states.NOTINITIALIZED,
```
IvarsKarpics: Is there a way to avoid this and use self.motor_states from AbstractMotor. In fact I created ExporterMotor that is based on AbstractMotor and handles states correctly. Maybe one could combine them in one or you can try ExporterMotor?
Once we should clean up all those available motors.

marcus-oscarsson:
Did you finally come to a conclusion regarding this ?

marcus-oscarsson:
@IvarsKarpics and @MartinSavko are you happy with this now ? are there still things to be discussed ?

marcus-oscarsson:
PING PONG ?!

PR #194, opened 2018-06-14
[master] EMBL hwobjs

IvarsKarpics:


HardwareObjects/EMBL/EMBLAperture.py
```diff
@@ -59,18 +57,19 @@ def init(self):
         if self.chan_diameter_index is not None:
             self._current_diameter_index = self.chan_diameter_index.getValue()
             self.diameter_index_changed(self._current_diameter_index)
-            self.chan_diameter_index.connectSignal(\
-                 'update', self.diameter_index_changed)
+            self.chan_diameter_index.connectSignal( \
```
rhfogh: I would recommend the following;
the backslash excape is discouraged in the python standard for good reason, and anyway it is not needed.


            self.chan_diameter_index.connectSignal(
                'update', self.diameter_index_changed
            )

HardwareObjects/EMBL/EMBLBeamInfo.py
```diff
@@ -105,225 +90,272 @@ def init(self):
...
         return self.evaluate_beam_info()
-        
+
     def get_beam_size(self):
-        """
-        Descript. : returns beam size in microns
-        Return   : list with two integers
+        """Returns beam size in microns
+
+        :return: list with two integers
```
rhfogh: Should be (tu[ple, not list):

:return: Tuple[int, int]

HardwareObjects/EMBL/EMBLBeamInfo.py
```diff
@@ -105,225 +90,272 @@ def init(self):
...
+
+        :return: [float, float]
         """
         if self.chan_beam_position_hor and self.chan_beam_position_ver:
-            self.beam_position = [self.chan_beam_position_hor.getValue(), \
-	                          self.chan_beam_position_ver.getValue()]
-        return self.beam_position	
+            self.beam_position = [self.chan_beam_position_hor.getValue(),
+                                  self.chan_beam_position_ver.getValue()]
+        return self.beam_position
```
rhfogh: Is it deliberate that if anyone modifies the returned list it wll modify the internal state of teh object? If not this ought to return a copy or a tuple.

HardwareObjects/EMBL/EMBLBeamInfo.py
```diff
@@ -105,225 +90,272 @@ def init(self):
...
-		     self.beam_size_focusing[1]) 
+                     self.beam_size_slits[1],
+                     self.beam_size_focusing[1])
 
         if size_x == 9999 or size_y == 9999:
-            #fix this
             return
         if (abs(size_x - self.beam_info_dict.get("size_x", 0)) > 1e-3 or
-            abs(size_y - self.beam_info_dict.get("size_y", 0)) > 1e-3):	
+                abs(size_y - self.beam_info_dict.get("size_y", 0)) > 1e-3):
```
rhfogh: Should the two 'abs' not be below each other - this looks like the two calls are nested?

HardwareObjects/EMBL/EMBLMachineInfo.py
```diff
@@ -280,7 +300,7 @@ def update_machine_state(self):
 
     def low_level_alarm_changed(self, value):
         """Low level alarm"""
-        self.low_level_alarm = value
+        self.low_level_alarm = value == 0
```
rhfogh: Could value ever be None (or temtpy list etc.)- and what should happen if it is?
Alternative (with different effects) would be  ... = not(value)

HardwareObjects/EMBL/EMBLSafetyShutter.py
```diff
@@ -106,42 +113,84 @@ def disconnected(self):
...
+        :type state: str
+        :return: None
+        """
         self.data_collection_state = state
         self.getShutterState()
 
     def state_open_changed(self, state):
+        """Updates shutter state when shutter open value changes
+
+        :param state: shuuter open state
```
rhfogh: shuuter -> shutter

rhfogh:
Hi Ivars,

I just breezed through the changes, as I obvlously cannot test them. As you can maybe guess, I highly approve of all this clean-up and documentation work ;-)

I picked a few nits, but it is largely a matter of style (see below), and for future reference. So if you do as much or as little as you feel like about my comments (including nothing), I shall merge the result unseen as soon as you let me know it is time.

On style:

- I thought that :result: was for a description in words of what was returned, while :rtype: was for the type. Or is rtype ignored unless there is also a :result:?

- My personal preference is for e.g.  :param int position_index: instead of having a separate line for the type (less typing).

- I think you could get rid of even more backslashes by putting parantheses aroune artihmetic or logical  expressions. 

- Although it is a bit pesky to change, it can be problem if a getter function call returns a list internal to the object, so that modifying the return value modifies the object internals.

HardwareObjects/EMBL/EMBLMachineInfo.py
IvarsKarpics: low_level_alarm appears if the value is 0, otherwise alarm is not raised.

IvarsKarpics:
Hi @rhfogh ,

Thanks for the detailed review. I made changes based on you suggestions.

PR #196, opened 2018-06-14
missing type and value

meguiraun:


HardwareObjects/BeamlineActionsMockup.py
```diff
@@ -25,6 +25,7 @@ def __init__(self, name, cmd, username=None, klass=SimulatedAction):
         CommandObject.__init__(self, name, username)
         self._cmd = klass()
 	self._cmd_execution = None
```
marcus-oscarsson: It was probably like this before, but isnt there a problem with the indentation here ?

marcus-oscarsson:
Thanks, for fixing this one !

I might have spotted an indentation problem, but apart from that all fine !

HardwareObjects/BeamlineActionsMockup.py
meguiraun: I did not see that one!

PR #202, opened 2018-07-03
[master] Cleaned up ParallelProcessing hwobj

IvarsKarpics:
* HardwareObjects/GenericParallelProcessing.py: base class
* HardwareObjects/DozorParallelProcessing.py: parallel processing based on the EDNA Dozor plugin
* HardwareObjects/ParallelProcessingMockup.py: mockup with various scoring scenarious (none, random, linear, first, last...)

### DIFF
--- a/HardwareObjects/DozorParallelProcessing.py
+++ b/HardwareObjects/DozorParallelProcessing.py
@@ -1,0 +1,90 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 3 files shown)


marcus-oscarsson:
Hi Ivars,

Looks really nice, just one question. Why is the base class called GenericParallelProcessing, would not simply ParallelProcessing suffice ?

Marcus

IvarsKarpics:
Hi @marcus-oscarsson ,

I wanted to follow Abstract*, Generic* approach and provide a generic class for parallel on-the-fly processing. Now it is based on EDNA and dozor, but could be extended (formalized) also for other online processing methods.

marcus-oscarsson:
Hi agian,

Okey, I see. I think the original intent with the generic was mostly for procedures were it might make more sense to prefix with Generic. But I might be mistaken, its maybe more a subject for the upcoming meetings. Although I don't really see what Generic in this case would mean or tell us in this context.

I'm happy to merge this, if you are not intending to change it and we will probably discuss it later.

Marcus

IvarsKarpics:
Well the definition of Generic is missing among us. For me, almost all hardware objects can have abstract (single hardware instance) or generic (procedure) representation. I would suggest to merge this and about naming discuss during the developers meeting.

beteva:
Well, are we going to this discussion again - what is generic and what is abstract. We have agreed (on the 10th meeting) that Generic would be for the equipment, Abstract for the sequences. I understand that there may be some procedures (like diffractometer initialisation), but as soon as the main issue is the hardware, we can still stuck to our Generic/Abstract model.  Than with the definition of the API, this won't be an issue at all.

PR #203, opened 2018-07-03
GPhL bug fix and change to session.xml

rhfogh:
GPhL bug fix. session.xml changed default base directory names to fit new GPhL dockers


rhfogh:
Thanks!

PR #207, opened 2018-09-03
[2.2] Shapes and SC

marcus-oscarsson:
Hi,

Some fixes to FlexHCD, Shapes (for MXCuBE3) and ISPyBRestClient

- Improved the sample loading and unloading routines of FlexHCD
- Hiding grids that get out of a certain, configurable,  threshold angle automatically when omega rotates
- Handling workfloStepId's in get_dc of ISPyBRestClient
- And some other minor fixes like typos

Marcus

### DIFF
--- a/HardwareObjects/Shapes.py
+++ b/HardwareObjects/Shapes.py
@@ -429,9 +429,9 @@ class Grid(Shape):
         self.set_id(Grid.SHAPE_COUNT)
 
     def update_position(self, transform):
-        phi_pos = self.shapes_hw_object.diffractometer.phiMotor.getPosition()
+        phi_pos = self.shapes_hw_object.diffractometer.phiMotor.getPosition() % 360
 
-        if abs(self.get_centred_position().phi - phi_pos) > self.hide_threshold:
+        if abs((self.get_centred_position().phi % 360) - phi_pos) > self.hide_threshold:
             self.state = "HIDDEN"
         else:
             super(Grid, self).update_position(transform)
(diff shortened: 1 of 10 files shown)


HardwareObjects/Shapes.py
```diff
@@ -98,6 +102,7 @@ def add_shape(self, shape):
         :type shape: Shape object.
         """
         self.shapes[shape.id] = shape
+        shape.shapes_hw_object = self
```
meguiraun: why adding this? I do not understand

HardwareObjects/Shapes.py
```diff
@@ -414,11 +428,22 @@ def __init__(self, mpos_list, screen_coord):
         self.beam_pos = [1, 1]
         self.beam_width = 0
         self.beam_height = 0
+        self.hide_threshold = 5
 
         self.set_id(Grid.SHAPE_COUNT)
 
+    def update_position(self, transform):
+        phi_pos = self.shapes_hw_object.diffractometer.phiMotor.getPosition() % 360
```
meguiraun: and `self.diffractometer.phiMotor....` instead?

meguiraun: sorry, yes... messed up with the diff. Ignore it

meguiraun:
sorry it took so long for review.

HardwareObjects/Shapes.py
marcus-oscarsson: So that the shape can access the shapes object that is "managing" it,

HardwareObjects/Shapes.py
marcus-oscarsson: So, I can ignore this comment then ?

meguiraun: yes, just merged!

PR #210, opened 2018-09-20
[master] Abstract classes

IvarsKarpics:


HardwareObjects/AbstractCollect.py
```diff
@@ -244,6 +244,7 @@ def do_collect(self, owner):
 
             log.info("Collection: Updating data collection in LIMS")
             self.update_data_collection_in_lims()
+            # ----------------------------------------------------------------
```
marcus-oscarsson: What is this comment for ?

HardwareObjects/AbstractCollect.py
```diff
@@ -186,16 +186,16 @@ def do_collect(self, owner):
...
-            log.info("Collection: Creating directories for raw images and processing files")
+            log.info("Collection: Creating directories for raw images and processing files") 
             self.create_file_directories()
 
-            log.info("Collection: Getting sample info from parameters")
+            log.info("Collection: Getting sample info from parameters") 
             self.get_sample_info()
         
-            #log.info("Collect: Storing sample info in LIMS")
+            #log.info("Collect: Storing sample info in LIMS")
```
marcus-oscarsson: We should maybe remove commented logic, if its not used ?

HardwareObjects/AbstractCollect.py
```diff
@@ -660,7 +657,8 @@ def get_sample_info(self):
         if self.diffractometer_hwobj.in_plate_mode():
             try:
                 plate_location = self.plate_manipulator_hwobj.get_plate_location()
-                self.current_dc_parameters["actualSampleBarcode"] = "Plate %s" % str(plate_location[:40])
+                #self.current_dc_parameters["actualSampleBarcode"] = "Plate %s" % str(plate_location[:30])
```
marcus-oscarsson: Also looks a bit suspicious ?

marcus-oscarsson:
Thanks, looks great in general. I made a few comments on tiny things

HardwareObjects/AbstractCollect.py
IvarsKarpics: To visually separate the collection steps. Can be removed if needed.

marcus-oscarsson: I see, well I guess we will take a look at all this at some point but I think I would prefer that it at least also states whats either above or below the line.

HardwareObjects/AbstractCollect.py
IvarsKarpics: In fact I have to test this... There was some problem in recording the plate location. Thanks for point this out!

marcus-oscarsson: I see, It looked like some kind of work in progress. I also found it a bit strange that actualSampleBarcode is assigned something from plate_location and that does not at all look like a bar code. But that maybe has other explanations ?

IvarsKarpics: Now I remember the issue here. There is no place in ispyb to store plate location (someone may correct me, but at least there is no place accessible via web services) and I used the sample barcode to store this. Maybe to avoid confusion I will remove this code block and contact ispyb people for a better solution.

marcus-oscarsson: Aha, I see :), well I guess its a workaround as good as any ... And a plate cant be represented as a sample which does have location, actually I'm not sure that's better. As you say the best thing is to ask the ISPyB guys. 

Well I don't mind if the code is kept for the moment, you are maybe relaying on it ?, if not I guess its a bit strange that it is the AbstractCollect ... 

We will surely have time to discuss this the coming months, so I'm happy to merge this if you think its okey to keep it like this ?

IvarsKarpics: I would assume that we collect our changes and then during the face-to-face meeting look at abstract classes in general. From my perspective you may merge the PR.

PR #212, opened 2018-09-24
[master] AbstractCollect bugfix

IvarsKarpics:


HardwareObjects/AbstractCollect.py
```diff
@@ -649,6 +649,7 @@ def get_sample_info(self):
 
         if self.diffractometer_hwobj.in_plate_mode():
             #TODO store plate location in lims
+            pass
         elif self.sample_changer_hwobj:
```
marcus-oscarsson: Aha, so this means that we need to disable the sample changer manually when plates are used ? or is this handled automatically somehow

marcus-oscarsson: Hm , took another look at this, looks fine :)

PR #213, opened 2018-09-24
[master] Various fixes

IvarsKarpics:
* Command/Tine.py: added variable force to the getValue method to force value reading
* CommandContainer.py: likewise
* HardwareObjects/Attenuators.py: cleaned up
* HardwareObjects/BeamlineSetup.py: added advance methods to queue_model_objects.DataCollection. This is needed to display processing messages
* HardwareObjects/CentringMath.py. getPosition -> get_position
* HardwareObjects/DataAnalysis.py: fixed log message
* HardwareObjects/ExporterMotor.py
* HardwareObjects/MiniKappaCorrection.py
* HardwareObjects/XMLRPCServer.py: added methods to handel processing results
* HardwareObjects/sample_changer/PlateManipulator.py
* HardwareRepository.py: code cleanup

### DIFF
--- a/Command/Tine.py
+++ b/Command/Tine.py
@@ -154,14 +154,16 @@ class TineChannel(ChannelObject):
             TineChannel.updates.put((weakref.ref(self), value))
             self.oldvalue = value
 
-    def getValue(self):
+    def getValue(self, force=False):
         #logging.getLogger("HWR").debug('TINE channel %s, %s get at val=%s'%(self.tineName,self.attributeName,self.value))
         #if self.tineName == "/P14/BCUIntensity/Device0":
         #   print self.attributeName, self.value           
 
-        if self.value is None:
+        if self.value is None or force:
             try:
-               time.sleep(0.02)
+               #TODO remove this
+               if not force:
+                   time.sleep(0.02)
                self.value = tine.get(self.tineName, self.attributeName, self.timeout)
             except IOError as strerror:
                logging.getLogger("HWR").error("%s" %strerror)
--- a/HardwareRepository.py
+++ b/HardwareRepository.py
@@ -392,16 +392,16 @@ class __HardwareRepositoryClient:
                 if objectName in self.hardwareObjects:
                     ho = self.hardwareObjects[objectName]
                 else:
-                    #start_time = datetime.now()
+                    start_time = datetime.now()
 
                     ho = self.loadHardwareObject(objectName)
 
-                    #end_time = datetime.now()
-                    #time_delta = end_time - start_time
+                    end_time = datetime.now()
+                    time_delta = end_time - start_time
 
-                    #logging.getLogger("HWR").debug(\
-                    #   "Loading of hwobj %s (%s.xml) took %d ms." % \
-                    #   (ho.__module__, objectName[1:], time_delta.microseconds / 1000))
+                    logging.getLogger("HWR").debug(\
+                       "Loading of hwobj %s (%s.xml) took %d ms." % \
+                       (ho.__module__, objectName[1:], time_delta.microseconds / 1000))
                 return ho
         except TypeError as err:
             logging.getLogger("HWR").exception("could not get Hardware Object %s", objectName)
--- a/HardwareRepository.py
+++ b/HardwareRepository.py
@@ -93,6 +93,17 @@ class __HardwareRepositoryClient:
         try:
             self.invalidHardwareObjects = set()
             self.hardwareObjects = weakref.WeakValueDictionary()
+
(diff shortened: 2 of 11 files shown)


marcus-oscarsson:
I'm taking a look at this, It will take a while because its quite big. I will start leaving some comments, Ill let you know when I'm through ...

Command/Tine.py
```diff
@@ -154,14 +154,16 @@ def update(self, value = None):
             TineChannel.updates.put((weakref.ref(self), value))
             self.oldvalue = value
 
-    def getValue(self):
+    def getValue(self, force=False):
```
marcus-oscarsson: TINE specific comments, maybe remove ?

marcus-oscarsson:
Hm, there are some API changes to hardware objects that puzzles me, the new additions does look better. 

```
transmission = self.transmission_hwobj.getAttFactor()
transmission = self.transmission_hwobj.get_value()
axis['motor_HO'].getPosition())
axis['motor_HO'].get_position())
```
It makes me think a little bit about compatibility, is anyone using master as it is today ?, because these looks like breaking changes, or ?

HardwareRepository.py
```diff
@@ -92,19 +93,6 @@ def connect(self):
         try:
             self.invalidHardwareObjects = set()
             self.hardwareObjects = weakref.WeakValueDictionary()
-
-            if type(self.serverAddress)==bytes:
```
marcus-oscarsson: I'm not entirely sure that I know what this does, but why is it removed ?

IvarsKarpics:
Hi @marcus-oscarsson ,

To be honest I do not know whats the best way to proceed. I have lots of things to commit, that need a review (a public onion, lets say). I started to use more abstract classes and with a single commits,PR it is not enough to get a full view.

HardwareRepository.py
```diff
@@ -404,13 +392,16 @@ def getHardwareObject(self, objectName):
...
                     ho = self.loadHardwareObject(objectName)
 
-                #try:
-                #    print (111, self.hardwareObjects, objectName)
-                #    ho = self.hardwareObjects[objectName]
-                #except KeyError:
-                #    ho = self.loadHardwareObject(objectName)
+                    #end_time = datetime.now()
+                    #time_delta = end_time - start_time
+
```
marcus-oscarsson: I think it would be nice to see how long time it takes to load the HO's. If not just remove the code if its not used ....

marcus-oscarsson:
Thanks for all this Ivars, looks good in general. I left a few comments ...

HardwareRepository.py
IvarsKarpics: This was used once in the testing: I had problems with some hwobj. If you or others see it useful I can remove the comments.

marcus-oscarsson:
I see, well okey. Is a good thing that you are using more abstract classes. First think I guess would be to make sure that this works with whats currently in the MXCuBE Qt4. I guess its the case ?

The second thing, which I cant really answer, is how much it impacts the others using master ?

Then just a question that would be interesting for the coming weeks, does this put master in a consistent state so that its whats is used (or are to be) in general for production ?

Apart from that, Im fine with all this if you tell me that this is what you are using or that it works in general with the Qt4 version ?

HardwareRepository.py
marcus-oscarsson: Well, if its there why not use it, can come in handy from time to time :)

HardwareRepository.py
marcus-oscarsson: Aha, I see. Two things:  
  - We are moving away from SPEC and one beamline (ID23-1) is still using it, but I guess we will be  using BLISS by the time all this gets done.
  - I guess its still good to keep SPEC compatibility, why remove it if it was once implemented. I guess some people might find it useful ....

IvarsKarpics:
Well, I am doing my homework :) And all this is running on beamlines. As I am committing many changes I am open to help other people with the merging and conflict resolving.

marcus-oscarsson:
Okey, sounds good ! I will leave some time for someone else to take a look at it and I will merge it later today if don't hear anything else.

I do think its a pity that we are removing SPEC compatibility, I would rather keep it since there should be a working solution, what do you think ?

IvarsKarpics:
Hi @marcus-oscarsson ,

I will revert the spec compatibility and see how to handle that warning message.

marcus-oscarsson:
Okey, that would be great, thanks for doing that. There is maybe another check to do for SPEC and then run that piece of code ?

beteva:
Watch out -we are not using spec, but we are not sure if it is not used by anybody else.

PR #216, opened 2018-09-24
[master] Queue and queue model

IvarsKarpics:
Last PR from the PR series to update the queue and queue_model

### DIFF
--- a/HardwareObjects/QueueManager.py
+++ b/HardwareObjects/QueueManager.py
@@ -83,6 +83,7 @@ class QueueManager(HardwareObject, QueueEntryContainer):
         Starts execution of the queue.
         """
         if not self.is_disabled():
+            self.emit("statusMessage", ("status", "Queue running", "running"))
             self._is_stopped = False
             self._set_in_queue_flag()
             self._root_task = gevent.spawn(self.__execute_task)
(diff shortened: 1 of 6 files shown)


HardwareObjects/QueueManager.py
```diff
@@ -190,15 +191,18 @@ def __execute_entry(self, entry):
...
             else:
                 self.emit('queue_entry_execute_finished', (entry, "Successful"))
+                self.emit('statusMessage', ("status", "", "ready"))
         except queue_entry.QueueSkippEntryException:
             # Queue entry, failed, skipp.
             self.emit('queue_entry_execute_finished', (entry, "Skipped"))
-        #except (queue_entry.QueueExecutionException, Exception) as ex:
-        #    self.emit('queue_entry_execute_finished', (entry, "Failed"))
+        except queue_entry.QueueExecutionException as ex:
+            self.emit('queue_entry_execute_finished', (entry, "Failed"))
```
marcus-oscarsson: It would be great if we could reuse the states already defined in queue entry ['SUCCESS','WARNING', 'FAILED', 'SKIPPED'], but thats maybe for later ....

marcus-oscarsson:
Looks fine to me, lets take a look at symbols and such things in a future refactoring session

IvarsKarpics:
Thanks @marcus-oscarsson!
This was the last PR for the HardwareRepository. Yes, I agree, we should reuse the symbols as much as we can.

PR #218, opened 2018-09-26
[master] Cleanup the loadHardwareObject method

IvarsKarpics:
Cleaned up the loadHardwareObject method
After finishing of gui load print a table with info about all hwobj. like:

```
+ ===================================================================== +
| Xml                      | Class                      | Load time| Comment|
+ ===================================================================== +
| /eh1/diff-omega          | MotorMockup                | 1 ms     |    |
| /Qt4_video-mockup        | Qt4_VideoMockup            | 4 ms     |    |
| /eh1/diff-sampy          | MotorMockup                | 0 ms     |    |
| /eh1/diff-sampx          | MotorMockup                | 0 ms     |    |
| /eh1/diff-phiz           | MotorMockup                | 0 ms     |    |
| /eh1/diff-phiy           | MotorMockup                | 0 ms     |    |
| /eh1/diff-kappaphi       | MotorMockup                | 0 ms     |    |
| /eh1/diff-kappa          | MotorMockup                | 0 ms     |    |
| /mini-diff               | DiffractometerMockup       | 108 ms   |    |
```

### DIFF
--- a/HardwareRepository.py
+++ b/HardwareRepository.py
@@ -218,12 +218,7 @@ class __HardwareRepositoryClient:
                     hwobj_instance.resolveReferences()
 
                     try:
-                        def addChannelsAndCommands(node):
-                            if isinstance(node, BaseHardwareObjects.CommandContainer):
-                                node._addChannelsAndCommands()
-                            for child_node in node:
-                                addChannelsAndCommands(child_node)
-                        addChannelsAndCommands(hwobj_instance)
+                        hwobj_instance._addChannelsAndCommands()
                     except:
                         logging.getLogger('HWR').exception("Error while adding commands and/or channels to Hardware Object %s", hwobj_name)
                         comment = "Failed to add all commands and/or channels"
--- a/HardwareRepository.py
+++ b/HardwareRepository.py
@@ -4,7 +4,7 @@ The Hardware Repository database is a set of XML files describing devices, equip
 and procedures on a beamline. Each XML file represent a Hardware Object.
 The Hardware Repository module provides access to these Hardware Objects, and manages
 connections to the Control Software (Spec or Taco Device Servers).
-""" 
+"""
 
 __author__ = 'Matias Guijarro'
 __version__ = 1.3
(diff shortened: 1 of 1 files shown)


HardwareRepository.py
```diff
@@ -149,115 +150,128 @@ def require(self, mnemonicsList):
...
-               if os.path.exists(file_path):
-                 try:
-                   xmldata = open(file_path, "r").read()
-                 except:
-                   pass
-                 break 
-
-        if True:
-                if len(xmldata) > 0:
+                file_name = hwobj_name[1:] if hwobj_name.startswith(os.path.sep) else hwobj_name
```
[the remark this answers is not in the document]
IvarsKarpics: Well, theoretically yes. I added default value to avoid this error.

HardwareRepository.py
```diff
@@ -149,115 +150,128 @@ def require(self, mnemonicsList):
...
+                        del self.hardwareObjects[name]
+
+                    hwobj_instance.resolveReferences()
+
                     try:
-                        #t0 = time.time()
-                        ho = self.parseXML(xmldata, hoName)
-                        if type(ho) == str:
-                            return self.loadHardwareObject(ho)  
+                        def addChannelsAndCommands(node):
```
marcus-oscarsson: Would it not be better if this method is defined on node ?

IvarsKarpics: What do you mean? Like move this method to HardwareObject class?

HardwareRepository.py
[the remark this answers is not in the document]
marcus-oscarsson: I guess the os.path.exists(file_path) below will catch that, although not if its None ...

HardwareRepository.py
marcus-oscarsson: Yes, that's what I had a in mind, the method only operates on a instance of a CommandConatiner derived object (it seems ...) as argument and does not seem dependent of anything else in this file. So one could probably move it

Although I noticed that its an old solution, that I guess we can continue to live with, if you don't really feel like messing with that, which I understand if you don't

IvarsKarpics: In fact there is no need for recursion here, because anyway each child hwobj will be initialized. Its enough to call hwobj_instance._addChannelsAndCommands()

marcus-oscarsson: Even better, in that case, but the current solution is recursive though

marcus-oscarsson:
Thanks Ivars !

PR #222, opened 2018-10-10
GPHL-specific fixes + tab indentation fix

rhfogh:


beteva:
Hi @rhfogh, you may want to change your comment from EDRF to ESRF :-)
Apart from that, could 'co.gphl.wf.bdg_licence_dir' or 'co.gphl.wf.bin'  in self.software_paths.get() be a configurable and as such put in an xml file?

rhfogh:
Hi @beteva,

Typo fixed.

The funny identifier strings are actually configured in an xml file - gphl-setup.xml to be exact. The reason for using such long and complicated names is that these are the names you need to use to pass the relevant paths to the java workflow (the Java properties names). Peter likes properly namespaced identifiers. So I have defined the identifiers as required by the Java and rename them in the code you see to have a more convenient name for the Python environment variables and for internal MXCuBE use. Alternatively I could use the Java identifiers throughout, or give them with a nicer convention in the xml file and then rename them to the Java names in the MXCuBE code before passing them to the workflow runner.

Do you think it is worth modifying?

marcus-oscarsson:
Hi Rasmus,

I guess you do as you like with those identifiers, it seems to be very specific to your implementation.
I'm merging this now and you can make another PR with any eventual identifier changes ...

Cheers,
Marcus

marcus-oscarsson:
Well, actually before merging ... 

- I noticed that you have done some changes to the SOLEIL PXCollect is that intentional ?
- Another thing: the commits from Ivars or those cherry picks from master ?

Marcus

rhfogh:
Hmm. How do we deal with this?

 Basically my code has evolved over the last two weeks, and I keep pusing the changes regularly. In practice the rhfogh:master is a continuous development branch, so I do not make a separate temporary branch every time I make a pull request. 

In the specific case that means that 1) I have addressed Antonia's comment as part of a subsequent refactoring of my internal workflow directory handling, 2)  the commits from Ivars have been added, which is where the change to SOLEIL PXCollect is coming from. 'Attenuators', 'Small additions to abstract classes' and 'Removed update_lims_with_workflow' are already merged into master (that is where they are coming from), so they do not represent anything new. The changed files are all either GPhL-specific or part of Ivars' commits and so already merged into master. Could you deal with them now on that basis? Or does this need looking at again?

I can see it is a problem that the pull request has, effectively, changed a fair bit since it was first made and reviewed. On the other hand, keeping each pull request isolated in a separate temporary branch when this does not in any way match the actual progress of the work would be a bit of a pain. What would be the way to do this for the future?

marcus-oscarsson:
Hi again,

Well a PR should contain a or a set of well defined features or fixes and should also be done on separate branch. Reality sometimes looks a bit different and we sometimes group unrelated things together which I think is okey as long as it stated in the PR roughly what the PR contains.

Its probably a bad idea for yourself and the people doing the review if the PR is made from a very active branch, like master. So my suggestion would be (and how its frequently done) to have a separate branch for your PR.

I'm otherwise fine with all these changes, I just wanted to make sure that everything was alright. I'm still wondering were Ivars commits comes from though. Could you try to rebase this on origin/master because I imagine that Ivars commits already exists there ?

I completely understand the reasons for the variable namespace and I don't see any reason for changing  it.

Cheers,
Marcus

rhfogh:
What I guess I should do, then, is to fork a dummy branch every time I make a PR and then do the PR from there. Doing the actual development in a series of carefully divergent streams, which need to be merged back once they are accepted, does not sound realistic given that I am working almost 100% on GPhL-specific code.

As for Ivars commits I know where they came from - they were merged in from origin/master after Ivars' PR had been accepted. Which presumably means that if you accept the PR there will be no problems? I tried to do a rebase to orign/master in a test branch, and got some merge conflicts, so I had to actually merge rather than just rebase - that is presumably why we have the extra commits appearing. Indeed I believe the same thing happened originally - I always use rebase to update to origin/master. Not sure I quite understand what you want me to do now - I can remove the PR and make a new one if it helps. If I do that, how do you want me to handle the merge conflicts?

marcus-oscarsson:
Hi again,

Well I guess this is fine as it is, in general I like to keep the git "tree" clean, free from duplicated commits and unnecessary merges. My intention was simply to make sure that everyone was aware of what was being merged because the commits looked(s) a bit strange. 

So ideally what I would have liked to see is a PR with only the relevant commits, without merge commits and Ivars commits.

I will in anyways go a head and merge this, since fixing it will just create extra work for everybody. About creating a separate branch for your PR's, I think you simply have to accept that you cant continue pushing  to a branch that you have made a PR from. Well within reasonable limits, you can of course do smaller fixes and additions.

Cheers,
Marcus

rhfogh:
@marcus-oscarsson

Thanks

"_I think you simply have to accept that you cant continue pushing to a branch that you have made a PR from._"
I agree. Sorry for the extra trouble.

Rasmus

PR #223, opened 2018-10-11
[master] Attenuators

IvarsKarpics:
Changed signal name trslimitsChanged -> limitsChanged

We should agree on a generic naming.
* valueChanged
* stateChanged
* limitsChanged

### DIFF
--- a/HardwareObjects/AbstractAttenuators.py
+++ b/HardwareObjects/AbstractAttenuators.py
@@ -33,7 +33,7 @@ class AbstractAttenuators(Device, object):
         Device.__init__(self, name)
 
         self._value = None
-        self._limits = None
+        self._limits = [0, 100]
         self._state = None
 
     @abc.abstractmethod
@@ -46,7 +46,7 @@ class AbstractAttenuators(Device, object):
 
     def set_limits(self, limits):
         self._limits = limits
-        self.emit('limitsChanged', self._limits) 
+        self.emit('limitsChanged', (self._limits, )) 
 
     def get_limits(self):
         return self._limits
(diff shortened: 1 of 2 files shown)


HardwareObjects/AbstractAttenuators.py
```diff
@@ -46,7 +46,7 @@ def get_value(self):
 
     def set_limits(self, limits):
         self._limits = limits
-        self.emit('limitsChanged', self._limits) 
+        self.emit('limitsChanged', (self._limits, ))
```
rhfogh: It looks like the parameter to the emitted signal would be
([0,100],)
It is hard (for me) to find which function is actually connected to this signal - and some other signals that take limits use the naked list as a parameter. Could you comment on which function takes this, and in general when one should use the list and when one should put it in a tuple when emitting a signal?

IvarsKarpics: If used `self.emit('limitsChanged', self._limits)` There was a problem in the:
https://github.com/mxcube/mxcube/blob/06b0d3a916ba05145ef4bb493940ad86380b10d7/BlissFramework/Bricks/widgets/Qt4_create_task_base.py#L153
set_transmission_limits complained about passed 3 arguments instead of 2, because self._limits is a list and signal arguments need to be passed as a list (correct me if I am wrong).
Regarding the functionality: this signal is used just once here to set the limits of transmission QLineEdit.

rhfogh: OK, It looked weird to me and I could not quickly figure out how it worked. But it sounds like it is all taken care of. Basically all emitted argumetns must be passed as a list, I guess.
AOK.

PR #228, opened 2018-10-12
[2.2] XML files for web version

marcus-oscarsson:
Hi,

Added web version xml file examples and fixed small issue with sample changer mockup (unload)

Marcus


meguiraun:
is it only me? I see `#` character on xml file name, only on these two:
image

Also, which is the differences between qt and web xml files?

marcus-oscarsson:
Hey

Nope its not only you, it was emacs temp files. I've removed them now.

There are some differences between the qt and web xml files for instance the Shapes HardwareObject are not the same, and we added a few attributes to the web version which are not used in the qt version. So its maybe clearer to keep them separate, it also makes it easier to just copy the files as they are.

Marcus

PR #232, opened 2018-10-31
Sardana motor

jordiandreu:
Specific changes for SardanaMotor:

* Add velocity and acceleration in SardanaMotor.py.
* Add alias command names for get_position and get_limits.
* Add move_threshold property (default set to 0).
* Add connnectNotify() calls to inform clients on connection.
* Replace specific SardanaMotor States by AbstractMotor states.


HardwareObjects/SardanaMotor.py
```diff
@@ -74,6 +81,14 @@ def init(self):
         if self.threshold is None:
             self.threshold = self.threshold_default
 
+        try:
+            self.move_threshold = self.getProperty("move_threshold")
```
IvarsKarpics: A pure getProperty call will never throw an exception. If no property with the attribute exists then it will return None. You can define also a default value in case if the attribute do not exist in the xml, for example:
` self.move_threshold = self.getProperty("move_threshold", self.move_threshold_default)`
will assign self.move_threshold_default  if move_threshold do not exist in xml

IvarsKarpics: A small suggestion to improve the code, but in general PR is fine by me.

jordiandreu: Hi @IvarsKarpics,
I have followed your suggestion and committed the changes

jordiandreu: BTW, thanks for the hint. I just copied what i have found in other definitions. This try/except blocks can be found in many different HwObj (for instance, in the GenericDiffractmeter...)

PR #234, opened 2018-10-31
Generic diffractometer revisited

jordiandreu:
Revisiting GenericDifractometer:

* Add delay_state_polling property in GenericDiffractometer.
* Backward compatibility for method calls startCentringMethod and acceptCentring (are called from other modules).
* Allow simultaneous move of several motors by starting move on each of them without waiting for them to stop until they are all running.
* Declare stab for get_osc_max_speed().
* Fix value emitted with signal newAutomaticCentringPoint to emit a list-type value.

### DIFF
--- a/HardwareObjects/GenericDiffractometer.py
+++ b/HardwareObjects/GenericDiffractometer.py
@@ -606,6 +606,8 @@ class GenericDiffractometer(HardwareObject):
                 logging.getLogger("HWR").exception("Diffractometer: problem while centring")
                 self.emit_centring_failed()
 
+    startCentringMethod = start_centring_method   # ensure backward compatibility with queue_entry.py
+
     def cancel_centring_method(self, reject=False):
         """
         """
@@ -670,7 +672,7 @@ class GenericDiffractometer(HardwareObject):
                                                   self.beam_position[0],
                                                   self.beam_position[1],
                                                   msg_cb = self.emit_progress_message,
-                                                  new_point_cb=lambda point: self.emit("newAutomaticCentringPoint", point))
+                                                  new_point_cb=lambda point: self.emit("newAutomaticCentringPoint", (point,)))
               else:
                   self.current_centring_procedure = gevent.spawn(self.automatic_centring)
               self.current_centring_procedure.link(self.centring_done)
@@ -833,8 +835,13 @@ class GenericDiffractometer(HardwareObject):
                             and target values.
         :type motors_dict: dict
         """
+        if not type(motor_positions) is dict:
+            motor_positions = motor_positions.as_dict()
+
+        self.wait_device_ready(timeout)
         for motor in motor_positions.keys():
             position = motor_positions[motor]
+            logging.getLogger("HWR").debug(" Moving %s to %s" % (str(motor), position))
             if type(motor) in (str, unicode):
                 motor_role = motor
                 motor = self.motor_hwobj_dict.get(motor_role)
@@ -842,6 +849,7 @@ class GenericDiffractometer(HardwareObject):
                 if None in (motor, position):
                     continue
                 motor_positions[motor] = position
+            #self.wait_device_ready(timeout)
             motor.move(position)
 
         if self.delay_state_polling is not None and self.delay_state_polling > 0:    
@@ -893,6 +901,8 @@ class GenericDiffractometer(HardwareObject):
             centring_status['motors'] = self.get_positions()
         self.emit('centringAccepted', (True, centring_status))
         self.emit("fsmConditionChanged", "centering_position_accepted", True)
+
+    acceptCentring = accept_centring
 
     def reject_centring(self):
         """
@@ -1138,6 +1148,15 @@ class GenericDiffractometer(HardwareObject):
         Descript. :
         """
         return
+
+    def get_osc_dynamic_limits(self):
+        return (-10000, 10000)
+
+    def get_osc_max_speed(self):
+        """
+        """
+        return None
+        # raise NotImplementedError
 
     def zoom_in(self):
         return
--- a/HardwareObjects/GenericDiffractometer.py
+++ b/HardwareObjects/GenericDiffractometer.py
@@ -853,7 +853,10 @@ class GenericDiffractometer(HardwareObject):
         self.wait_device_ready(timeout)
         for motor in motor_positions.keys():
             position = motor_positions[motor]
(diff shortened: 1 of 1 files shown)


HardwareObjects/GenericDiffractometer.py
```diff
@@ -334,6 +337,11 @@ def init(self):
         except:
            pass # used the default value
 
+	try:
+            self.delay_state_polling = self.getProperty("delay_state_polling")
```
IvarsKarpics: Should this property be in the GenericDiffractometer?

HardwareObjects/GenericDiffractometer.py
```diff
@@ -403,6 +411,11 @@ def is_ready(self):
         return self.current_state == DiffractometerState.tostring(\
                     DiffractometerState.Ready)
 
+    def wait_device_not_ready(self, timeout=5):
```
IvarsKarpics: Hmm, for me it is not clear why one should wait till device is not ready?

HardwareObjects/GenericDiffractometer.py
```diff
@@ -719,15 +734,21 @@ def centring_done(self, centring_procedure):
...
                 logging.exception("Could not move to centred position")
                 self.emit_centring_failed()
             else:
-                #if 3 click centring move -180
-                if not self.in_plate_mode():
-                    self.motor_hwobj_dict['phi'].syncMoveRelative(-180)
+                #if 3 click centring move -180. well. dont, in principle the calculated
+                # centred positions include omega to initial position
+                pass
+                #if not self.in_plate_mode():
```
IvarsKarpics: Why this code is removed?

HardwareObjects/GenericDiffractometer.py
```diff
@@ -825,8 +847,16 @@ def move_motors(self, motor_positions, timeout=15):
...
         :type motors_dict: dict
         """
+        if not type(motor_positions) is dict:
+            motor_positions = motor_positions.as_dict()
+
+        self.wait_device_ready(timeout)
         for motor in motor_positions.keys():
             position = motor_positions[motor]
+            if type(motor) == str:
+                logging.getLogger("HWR").debug(" Moving %s to %s" % (motor, position))
```
IvarsKarpics: I would suggest to remove this log. It will produce to much log lines for each movement.

HardwareObjects/GenericDiffractometer.py
```diff
@@ -837,6 +867,13 @@ def move_motors(self, motor_positions, timeout=15):
             motor.move(position)
         self.wait_device_ready(timeout)
 
+        if self.delay_state_polling is not None and self.delay_state_polling > 0:    
+            # delay polling for state in the
+            # case of controller not reporting MOVING inmediately after cmd
+            gevent.sleep(self.delay_state_polling)
+
+        self.wait_device_ready(timeout)
```
IvarsKarpics: Twice a method self.wait_device_ready is called. The second call should be under the if statement.

HardwareObjects/GenericDiffractometer.py
jordiandreu: We are using the **sample_centring** module and I think it is because the omega axis is returned to the original position also when the motors are moved to the centered position. This later movement moves an extra 180 degrees backward the omega axis.

Let me revisit the centring procedure and I will provide more details on this.

HardwareObjects/GenericDiffractometer.py
jordiandreu: See general comment below.

HardwareObjects/GenericDiffractometer.py
jordiandreu: See general comment below.

HardwareObjects/GenericDiffractometer.py
jordiandreu: See general comment below.

HardwareObjects/GenericDiffractometer.py
jordiandreu: And what do you think about just commenting the lines?

jordiandreu: Hi @IvarsKarpics , after reviewing the centring routines based on the sample_centring module, I have seen that:

* The omega axis (phy role) position is stored in the SAVED_INITIAL_POSITIONS dictionary in the _sample_centring.prepare_ method. This means that the callback defined for the centering Greenlet (centring_done) gets this dictionary updated with the centered positions for the other motors (phy positions is not accounted in the calculations).
* When moving the motors to the centered positions, _phy_ is in the dictionary of centered positions with its initial position. For this, the _else_ statement at the end of the _try/except_ block is not necessary.

Maybe this is not handled in the same way if not using the _sample_centring_ module?

beteva:
Hello @jordiandreu and @IvarsKarpics, seems to me that pre/post delays for moving motors should not be handled by the Diffravtometr hardware object, but by the underlying Motor HO or even the Control system. This is a Generic HO which should not reflect the particularities of a motor. The only generic handling is to see weather you want to move all the motors at the same time (wait for a single motor will be False, and a generic wait_device_ready is used). By the way, when you move motors with a MD2S, there is a command to move all the motors together, so the move_motors function is overwritten.

marcus-oscarsson:
Hello all,

There seems to be some underlying timing related issues here. I completely agree with @beteva that its something that ideally should be handled in the underlying motor or control logic. It could, if you find it more appropriate, or complicated to implement in other ways, be handled in a subclass of GenericDiffractometer.

I'm otherwise fine with this as it is for current master, @IvarsKarpics is there still something you would like to see changed ?

rhfogh:
@beteva 
Can you confirm that when you move 'all motors together', all the motors are actually moved to the desired position? On ID30B I had the experience of the phi and kappa motor movement causing additional movement in the centring motors, so that moving the centringmotors first meant that you would end up with a different centringmotor position from what you passed in. And the motors were moved in random order. I fixed that one by moving the kappa and  kappa_phi motors BEFORE the centringmotors, but any changes made in this area need to ensure that the  all motors arrive at the place they were told to go.

marcus-oscarsson:
Hm, so @IvarsKarpics you have requested for changes but I'm not sure its still the case ?, whats the status of the discussion, are there still some issues to be resolved ?

IvarsKarpics:
Hi @marcus-oscarsson  and others,
Sorry for holding the PR, I will have a look at it asap (latest this evening)

PR #236, opened 2018-10-31
LdapLogin search fields capability

jordiandreu:
Add the possibility to pass search fields in ldap login query to obtain extra information from ldap, for example `fields=["uid", "homeDirectory"]` will include _homeDirectory_ in results for later use.


HardwareObjects/LdapLogin.py
```diff
@@ -35,6 +35,8 @@ def init(self):
         domain   = self.getProperty('ldapdomain')
 	ldapou   = self.getProperty('ldapou')
```
marcus-oscarsson: This did not change now, but please check this indent ?

jordiandreu: Thanks Marcus!
Done!

marcus-oscarsson: Thanks !

PR #237, opened 2018-10-31
Add templates path for ALBA Synchrotron

jordiandreu:
Add PathTemplate model for ALBA synchrotron


HardwareObjects/queue_model_objects_v1.py
```diff
@@ -1312,6 +1312,27 @@ def get_archive_directory(self):
                                              PathTemplate.archive_folder)
             archive_directory = os.path.join(archive_directory,
                                              *folders[4:])
+        elif PathTemplate.synchotron_name == "ALBA":
```
beteva: It is unfortunate that we have to add a path for each institute. I think all these paths (ESRF, ALBA, ...) should go to Session.

HardwareObjects/queue_model_objects_v1.py
```diff
@@ -1327,6 +1327,27 @@ def get_archive_directory(self):
                                              PathTemplate.archive_folder)
             archive_directory = os.path.join(archive_directory,
                                              *folders[4:])
+        elif PathTemplate.synchotron_name == "ALBA": 
+            #archive_directory = os.path.join(PathTemplate.archive_base_directory)
```
IvarsKarpics: I would suggest to remove not used code (#...). I agree with @beteva  that this should go to Session hwobj. We can discuss this during the developers meeting.

HardwareObjects/queue_model_objects_v1.py
jordiandreu: Indeed.

HardwareObjects/queue_model_objects_v1.py
jordiandreu: Done!

IvarsKarpics:
If it is fine for everyone I will accept the PR and if we will find time during the developers meeting we can discuss on how to move this code to Session.py

marcus-oscarsson:
Yes, I think we all agree on this and we will consider it at some stage in the refactoring work

PR #238, opened 2018-10-31
Session extended

jordiandreu:
Group assignment of directory and folder values in specific function
set_base_directories() for better checking of default values and other


HardwareObjects/Session.py
```diff
@@ -80,6 +102,24 @@ def init(self):
 
         queue_model_objects.PathTemplate.set_precision(precision)
 
+    def set_base_data_directories(self, base_directory, base_process_directory, base_archive_directory,
```
IvarsKarpics: Is there a reason for a new method? Is it called also somewhere else

jordiandreu: Not really. It just avoids a larger init method and looks a bit cleaner to me. However, maybe we can adopt a convention for private methods also (i.e. def._set_base_data_directories) if necessary. However, I have no objection to insert it into the init method.

IvarsKarpics: Well, true: it looks cleaner. If no one is against then I will accept the PR by the end of the day

PR #239, opened 2018-10-31
Alba specific changes

jordiandreu:
Massive changes for ALBA-specific HwObjs.


marcus-oscarsson:
Hi,

I do think it makes sense to have an ALBASession, I actually think that is the preferred way of  customizing the paths with our current system. 

It its indeed strange that you need a ALBAISPyBClient2 ?

Regards,
Marcus

marcus-oscarsson:
I will go ahead and merge this PR since this is what you are using, but the question about the ISPyBClient still remains ?

PR #240, opened 2018-10-31
Data analysis

jordiandreu:
This is a merged code with ALBA development.

* Protect from not existent fields.
* Correct bug in extraction of info from edna_input.
* Remove unused code (Threading objects).
* Define STAB method prepare_edna_input.
* Add (again) run_edna method to enable its overwrite for site specific code.


HardwareObjects/DataAnalysis.py
```diff
@@ -221,22 +234,29 @@ def characterise(self, edna_input):
...
             raise RuntimeError("No process directory specified in edna_input")
 
-        msg = "Starting EDNA characterisation using xml file %s" % edna_input_file
+        self.result = self.run_edna(edna_input_file, edna_results_file, path)
+
+        return self.result
+
+    def prepare_edna_input(self, edna_input):
+        """
+        STAB. Allows to manipulate edna_input object before exporting it to file
```
marcus-oscarsson: STAB. ?

vrey01:
Este es sobre lo de “STAB”?

Solo es un comentario, es para indicar que esa funcion debe de ser re-escrita por la clase que derive. STAB = apuñalar. Vamos que se la carguen. 

Puedes quitar el comentario

O es por otra cosa?



Bixente

marcus-oscarsson:
Hola Bixente !

I'm not sure that I understand this correctly. But yes my comment was simply to ask what STAB meant,
So as I understand it basically means that its an abstract method, that needs to be overridden ?

Gracias !
Marcus

vrey01:
Marcus,

Ooops. Sorry. I thought I was answering only to Jordi. 

But yes, you understood perfectly. Your spanish is very good. ;) 

well, it is not abstract in the sense that it does not NEED to be overwritten, it can stay empty. 

Bixente

PR #242, opened 2018-10-31
Video Device and Graphics Manager

jordiandreu:
Merged code with ALBA developments:

* Fix a bug where a core dump is produced randomly when a _qimage_ underlying C object is destroyed before being used.
* Initialize variable to None for subsequent check set_centring_state(True) when start_centring.
* Allow to set width and height in XML to overwrite values (wrong) from Hardware server.
* Allow to set fixed width and height in xml configuration file.


HardwareObjects/GenericVideoDevice.py
```diff
@@ -120,6 +121,15 @@ def init(self):
         except:
             pass
 
+        try:
```
IvarsKarpics: All this could be substituted with `self.scale = self.getProperty("scale", 1.0)` and in xml define float (like 1.0)

HardwareObjects/GenericVideoDevice.py
```diff
@@ -177,7 +189,7 @@ def get_new_image(self):
         """
         raw_buffer, width, height = self.get_image()
 
-        if raw_buffer is not None:
+        if raw_buffer is not None and raw_buffer.any():
```
IvarsKarpics: What any() should return?

HardwareObjects/GenericVideoDevice.py
```diff
@@ -190,9 +202,13 @@ def get_new_image(self):
...
                 qimage = qimage.mirrored(self.cam_mirror[0], self.cam_mirror[1])     
 
+            if self.scale != 1:
+               dims = self.get_image_dimensions()  #  should be already scaled
+               qimage = qimage.scaled(QSize(dims[0], dims[1]))
+
             qpixmap = QPixmap(qimage)
             self.emit("imageReceived", qpixmap)
-            return qimage
+            return qimage.copy()
```
IvarsKarpics: Why one needs to return an copy of a qimage?

HardwareObjects/GenericVideoDevice.py
jordiandreu: Here _any_ returns _True_ if at least one element from the raw buffer (which in our case is a ndarray object) is different than '0'.

HardwareObjects/GenericVideoDevice.py
jordiandreu: We found that this _qimage_ object was deleted (probably by the garbage collector) before being used. This was producing a core dump in our code version (we are using Qt4_TangoLimaVideo.py).

HardwareObjects/GenericVideoDevice.py
jordiandreu: Done!

vrey01:
Totalmente de acuerdo con tus comentarios

Bixente

HardwareObjects/GenericVideoDevice.py
marcus-oscarsson: Hm, i'ts interesting that it helps to make a copy if the problem is that the object is garbage collected. Is it somehow using weak references ?

HardwareObjects/GenericVideoDevice.py
IvarsKarpics: ok

HardwareObjects/GenericVideoDevice.py
marcus-oscarsson: Still curious about the copy ... ?

jordiandreu: Yes, me too. I still have no clue about it.

PR #243, opened 2018-10-31
Fix queue entry

jordiandreu:
* Fix an issue with sample location if value obtained is not a tuple.
* Adapt queue_entry with new naming of centring methods in GenericDiffractometer.    
* Protect mounting coordinates from SC casting value as a tuple.


HardwareObjects/queue_entry.py
```diff
@@ -1196,9 +1196,11 @@ def start_char(self):
         reference_image_collection = char.reference_image_collection
         characterisation_parameters = char.characterisation_parameters
 
-        edna_input = self.data_analysis_hwobj.from_params(\
-             reference_image_collection,
-             characterisation_parameters)
+        if self.data_analysis_hwobj is not None:
+            log.info("  - creating edna input from params")
```
IvarsKarpics: There is no need for this log.

jordiandreu: Done.

HardwareObjects/queue_entry.py
```diff
@@ -1196,9 +1196,10 @@ def start_char(self):
...
 
-        edna_input = self.data_analysis_hwobj.from_params(\
-             reference_image_collection,
-             characterisation_parameters)
+        if self.data_analysis_hwobj is not None:
+            edna_input = self.data_analysis_hwobj.\
+                         from_params(reference_image_collection,
+                                     characterisation_parameters)
 
         self.edna_result = self.data_analysis_hwobj.characterise(edna_input)
```
IvarsKarpics: I assume line 1204 should also go under the if block (line 1199)?
I will test this PR...

IvarsKarpics: I tested with mockup version and it looks fine. See (and accept it if its ok) PR: https://github.com/mxcube/mxcube/pull/267 for updated xml config. If no one is against then I will accept the PR at the end of the day.

PR #248, opened 2018-11-13
Merge of Desy P11 Beamline

jurbschat:
Merge Desy P11 Hardware Repository Files

### DIFF
--- a/HardwareObjects/DESY/DigitalZoomMotor.py
+++ b/HardwareObjects/DESY/DigitalZoomMotor.py
@@ -1,0 +1,170 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


beteva:
Agree that we can have a MotorWPositions HO (we do have Spec, Robodoff and Bliss MotorWPosition), but maybe should discuss the structure - should it inhereit from AbstarctMotor and Device or from a particular motor type, which already inherits from AbstartMotor and has the limits. moving...

HardwareObjects/DESY/DigitalZoomMotor.py
```diff
@@ -0,0 +1,170 @@
...
+__license__ = "GPL"
+
+
+import logging
+from HardwareRepository.BaseHardwareObjects import Device
+from HardwareRepository.HardwareObjects.AbstractMotor import AbstractMotor
+from HardwareRepository.HardwareObjects.AbstractMotor import MotorStates
+
+
+class DigitalZoomMotor(AbstractMotor, Device):
```
IvarsKarpics: AbstractMotor is inherited from the HardwareObject, so there is not need to add Device class.

HardwareObjects/DESY/DigitalZoomMotor.py
```diff
@@ -0,0 +1,170 @@
...
+        self.move(absolute_position)
+
+    def syncMoveRelative(self, relative_position, timeout=None):
+        """
+        Descript. : same as normal moveRelative, for position change is
+                    instantaneous
+        """
+        self.syncMove(self.getPosition() + relative_position, timeout)
+
+    def stop(self):
```
IvarsKarpics: move_relative, syncMove, stop are defined in the AbstractMotor. One should keep the list of arguments as defined in the abstract class (see move method)

PR #251, opened 2018-11-13
Changed configuration handling

rhfogh:
- Replaced getHardwareRepositoryPath with (pre-existing) findInRepository, as the former did not work with multi-directory hardware repositories.

- Added facility to override data directory paths. This allows you to use a beam-line-specific session.xml in mock mode, redirecting teh data to the correct location on the mocking computer. The change has no effect if the relevant configuration file is absent.

- Changed GPhL configuration handling

### DIFF
--- a/HardwareObjects/DataAnalysis.py
+++ b/HardwareObjects/DataAnalysis.py
@@ -43,6 +43,9 @@ class DataAnalysis(AbstractDataAnalysis.AbstractDataAnalysis, HardwareObject):
         self.start_edna_command = self.getProperty("edna_command")
         self.edna_default_file = self.getProperty("edna_default_file")
         fp = HardwareRepository().findInRepository(self.edna_default_file)
+        if fp is None:
+            raise ValueError("File %s not found in repository"
+                             % self.edna_default_file)
         with open(fp, 'r') as f:
             self.edna_default_input = ''.join(f.readlines())
 
--- a/HardwareObjects/DataAnalysis.py
+++ b/HardwareObjects/DataAnalysis.py
@@ -42,9 +42,8 @@ class DataAnalysis(AbstractDataAnalysis.AbstractDataAnalysis, HardwareObject):
         self.collect_obj = self.getObjectByRole("collect")
         self.start_edna_command = self.getProperty("edna_command")
         self.edna_default_file = self.getProperty("edna_default_file")
-        hwr_dir = HardwareRepository().getHardwareRepositoryPath()
-
-        with open(os.path.join(hwr_dir, self.edna_default_file), 'r') as f:
+        fp = HardwareRepository().findInRepository(self.edna_default_file)
+        with open(fp, 'r') as f:
             self.edna_default_input = ''.join(f.readlines())
 
     def get_html_report(self, edna_result):
--- a/HardwareObjects/GenericVideoDevice.py
+++ b/HardwareObjects/GenericVideoDevice.py
@@ -362,7 +362,8 @@ class GenericVideoDevice(Device):
 
     """  Methods to be implemented by the implementing class """
     def get_raw_image_size(self):
-        pass
+        # Must return a two-value list necessary to avoid breaking e.g. ViideoMockup
+        return [None,None]
 
     @abc.abstractmethod
     def get_image(self):
--- a/HardwareObjects/Session.py
+++ b/HardwareObjects/Session.py
@@ -7,9 +7,15 @@ access and manipulate this information.
 import os
 import time
 import logging
+import xml.etree.ElementTree as ET
 
+from HardwareRepository.HardwareRepository import HardwareRepository
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 import queue_model_objects_v1 as queue_model_objects
+
+# Configuration file used to override base_directory paths.
+# Will be ignored if not founc
+parameter_override_file = 'session_parameter_override.xml'
 
 class Session(HardwareObject):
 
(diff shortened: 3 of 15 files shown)


HardwareObjects/Session.py
```diff
@@ -7,10 +7,16 @@
...
 import logging
+import xml.etree.ElementTree as ET
 
+from HardwareRepository.HardwareRepository import HardwareRepository
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 import queue_model_objects_v1 as queue_model_objects
 
+# Configuration file used to override base_directory paths.
+# Will be ignored if not founc
+parameter_override_file = 'session_parameter_override.xml'
```
beteva: Hello @rhfogh, it does not look like a good idea to put hard coded file name. This, if needed, should be handled in the init, as a parameter from the session.xml

beteva:
Hello @rhfogh, the changes in you PR are only applicable for the master branch.
Maybe before accepting them, we should probably see how to merge master and 2.2, as findInRepository is not in the 2.2 branch

HardwareObjects/Session.py
rhfogh: Well, the point of this was that it allowed me to keep the beamline-specific settings and change only the top level directories, WITHOUT having to edit session.xml. I have to mock a number of different beamlines on the same machine, and I would like to use the session.xml as-is to get file name formatting, beamline name etc., and only change the destination of the data. If I do edit session.xml I get into a mess with either having changes that I have to remember not to check in (because they only make sense for myself locally), or to maintain an entire duplicate set-up for each beamline, and keep it up to date by hand. 
If this hack (yes it is a hack) comes to depend on having the right content in session.xml that kind of loses the point, and you might as well go back to editing the directory names directly. 

Alternatively, one might split session.xml in two files - one with absolute directory names, that depend on the directory structure you are runnign in, and anotehr with all the other parameters?

rhfogh:
@beteva 
I thought the point was to make sure all changes were checked into their respective branches before we started merging. Which I am doing quite late, I know, but I am doing it. Doing it your way leaves me with this pending change that I cannot even start doing anythign about.

It might be easier to add findInRepository now, and then sort out any conflicts as part of the merge?
It is a simple and self-contained function (see below)

    def findInRepository(self, relative_path):
        """Finds absolute path of a file or directory matching relativePath
        in one of the hardwareRepository directories

        Will work for any file or directory, but intended for configuration
        files that do NOT match the standard XML file system"""

        if self.server:
            logging.getLogger('HWR').error(
                "Cannot find file in repository - server is in use"
            )
            return
        else:
            if relative_path.startswith(os.path.sep):
                relative_path = relative_path[1:]

            for xml_files_path in self.serverAddress:
                file_path = os.path.join(xml_files_path, relative_path)
                if os.path.exists(file_path):
                    return os.path.abspath(file_path)
            #
            return

HardwareObjects/Session.py
beteva: In fact we have the *.xml files in order NOT to change the HO file itself. Maybe the better approach will be, as you say, to have one Session.py and a GPFLSession.py, which inherits from session and adds the extra parameters/directories you need.

IvarsKarpics: Hmm, I would be against of inheriting core (Session, Queue, queue_*) components. Lets discuss it in two days.

beteva:
The fact I have reacted was that the changes you make are not just adding an extra method, but also change two largely used HO (BeamlineSetup and DataAcquisition). With MXCuBE3 we do not use the server, but the files, so I do not know yet what the impact will be.

HardwareObjects/Session.py
rhfogh: We already have ALBASession(Session), SOLEILSession(Session) and an old MaxLabSession(Session).  Which would be a reason for me to avoid a separate subclass, actually, because I would want to mock the specific local subclass.

hwr_import sounds like a good idea. But do you know of a way to get hold of additional properties that are not inside the object xml?

I.e. if I do something like 

mock_session.xml:

<hwr_import href="/embl_hh_p14/session"/>
<base_directory>/tmp/mxcube_base</base_directory>
<processed_data_base_directory>/tmp/mxcube_procbase</processed_data_base_directory>
<archive_base_directory>/tmp/mxcube_archbase</archive_base_directory>

in a file, is there any way for Session.init() to get hold of the three additional properties?

rhfogh:
You have a point, but near as I can see the behaviour will be correct - at least now I have added error checking to the PR.

- If you have a server the new code will fail, but then the old code would have failed too. The error messages should be better now. 

- If the file exists in the first of the directories on the list, you will get the file, now as before.

- If the file cannot be found at all, you will get an error, now as before; the error messages should just be better. 

- The only difference comes in if 1) you have more than one configuration on your lookup path, and 2) the file you are looking for is not in the first directory but in one of the subsequent ones. Then you would find it with the new code but not with the hold code - which, arguably means that the old code had a bug.

It makes sense if you prefer to look before you leap. But I would stillhope that we could manage to get this PR accepted before we start merging. As MO wrote: "_It will further be very complicated to add features to either 2.2 or 2.3 (master before the merge) while we are in the process of creating 3.0"_.

HardwareObjects/DataAnalysis.py
```diff
@@ -42,9 +42,11 @@ def init(self):
         self.collect_obj = self.getObjectByRole("collect")
         self.start_edna_command = self.getProperty("edna_command")
         self.edna_default_file = self.getProperty("edna_default_file")
-        hwr_dir = HardwareRepository().getHardwareRepositoryPath()
-
-        with open(os.path.join(hwr_dir, self.edna_default_file), 'r') as f:
+        fp = HardwareRepository().findInRepository(self.edna_default_file)
```
marcus-oscarsson: Hm, like this the edna default file has to be located in the getHardwareRepositoryPath which was not the case before. I'm not saying that it's a bad thing, but it might be a potentially breaking change.

HardwareObjects/GenericVideoDevice.py
```diff
@@ -362,7 +362,8 @@ def get_image_dimensions(self):
 
     """  Methods to be implemented by the implementing class """
     def get_raw_image_size(self):
-        pass
+        # Must return a two-value list necessary to avoid breaking e.g. ViideoMockup
```
marcus-oscarsson: Typo in comment Viideo -> Video

Are you sure that you don't want to return numbers i.e. 0, 0 instead ?

marcus-oscarsson:
Hi,

It seems like this PR has some DESSY specific parts from a merge ?

Apart from what Antonia commented I also spotted a few minor things, commented above. I really don't have anything against the findInRepository method. The important thing is that its used in a way that is used in backwards compatible way. I however also think that we should take a closer look to see if we cant find a solution fulfills the same function as findInRepository but in a different way ?

Would it be possible to extract the things that are vital for your workflows to run in a separate PR ?

Marcus

PR #256, opened 2018-11-16
Reformatting code

MartinSavko:
Outcome of executing 
     black --safe ./;
     black --fast ./;
     autopep8 -r -j 0 -i --max-line-length 88 ./

on the entire HardwareRepository. The results seems to me more aesthetically pleasing then if autopep8 run first and black last ...


beteva:
Maybe add -a to autopep8, so it removes the trailing white spaces as well

rhfogh:
Did you check what stumbling blocks caused black --safe to fail? If not, how do we know/ensure that handling of e.g. tabs has not caused changes in the code?

beteva:
Hello @rhfogh, The main reason to run --fast is taht we have print which is not  python 3 compliant.
 I am preparing a new PR, with changes that, with some instruction of how to change the institute specific code problems. It will remove the need to do --fast. Please, be patient

rhfogh:
@beteva No problem then, I can see it is well in hand.

PR #257, opened 2018-11-19
Replaced override_directories hack with CollectEmulator implementation

rhfogh:
This hack should never have been in Session in the first place. It is now moved to CollectEmulator


marcus-oscarsson:
Thanks, well done !

PR #261, opened 2018-11-20
pep8 and black cleanup

beteva:
Applied recipe:
autopep8 -a -r -j 0 -i --max-line-length 88 ./
black --safe ./
Change offending print to print function. Use from __future__ import print_function


marcus-oscarsson:
Great work everyone !, so Ill merge this since Ivars tested it and approved it

PR #258, opened 2018-11-19
Replaced override_directories hack with CollectEmulator implementation

rhfogh:
Removing Session hack, replacing it with code in CollectEmulator.
Bug fix ported from 2.3.0

PR #263, opened 2018-11-22
[master] new mockup dir to store all mockup classes

IvarsKarpics:


beteva:
Hello @IvarsKarpics, tahnks for dealing with the mockups and cleaning the imports.
I will accept the merge request, although Session.py is also part of that PR. Afterwards we will have to deal with PR #258, concerning Session.py as well.

PR #269, opened 2018-11-28
Fixing 'override data directories' - again

rhfogh:
Hi. This should hopefully olve the problem.

NB, note that CollectEmulator is a mock object (I shall move it shortly) and depends on GPhL code.


rhfogh:
Thanks!

beteva:
Hello,
there is still this: parameter_override_file = "session_parameter_override.xml"

PR #270, opened 2018-11-29
[master] New abstract directory

IvarsKarpics:
solves #271


HardwareObjects/GenericDiffractometer.py
```diff
@@ -923,7 +923,6 @@ def move_to_motors_positions(self, motors_positions, wait=False):
         )
         self.move_to_motors_positions_procedure.link(self.move_motors_done)
         if wait:
-            self.wait_device_not_ready()
```
rhfogh: Glad to see this one go - it gave me problems. But, since it is freshly added and I suspect not used elsewhere, should the function definition not be removed as well?

HardwareObjects/mockup/SOLEILDataAnalysisMockup.py
```diff
@@ -2,7 +2,7 @@
 import logging
 import gevent.event
 import subprocess
-import AbstractDataAnalysis
+import abstract.AbstractDataAnalysis
```
rhfogh: I think that will break. It should put 'abstract' (NOT AbstractDataAnalysis) into the top namespace. I think the correct way is:
from abstract import AbstractDataAnalysis

HardwareObjects/mockup/DataAnalysisMockup.py
```diff
@@ -29,9 +29,10 @@
 __license__ = "LGPLv3"
 
 
-class DataAnalysisMockup(AbstractDataAnalysis.AbstractDataAnalysis, HardwareObject):
+class DataAnalysisMockup(AbstractDataAnalysis, HardwareObject):
```
rhfogh: Do we want to have HardwareObject as a superclass of AbstractDataAnalysis?

HardwareObjects/BlissActuator.py
```diff
@@ -9,7 +9,7 @@
 """
 import logging
 
-from AbstractActuator import *
+from abstract.AbstractActuator import *
```
rhfogh: Can we get rid of the 'from x import *' bit next?
Maybe in a subsequent PR?

HardwareObjects/mockup/MultiCollectMockup.py
```diff
@@ -1,5 +1,5 @@
 from HardwareRepository.BaseHardwareObjects import HardwareObject
-from abstract.AbstractMultiCollect import *
+from abstract.AbstractMultiCollect import AbstractMultiCollect
```
rhfogh: This might be a problem. In the version I have the 'from x import *' also imported 'task', which in turn had been imported with 'From TaskUtils import *'. That would have to be ckecked / fixed

HardwareObjects/GenericDiffractometer.py
rhfogh: OK, if it is actually used lower down it makes sense to keep the function definition.

HardwareObjects/mockup/MultiCollectMockup.py
```diff
@@ -1,5 +1,5 @@
 from HardwareRepository.BaseHardwareObjects import HardwareObject
-from AbstractMultiCollect import *
+from abstract.AbstractMultiCollect import *
```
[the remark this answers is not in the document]
rhfogh: I think you also need to import 'task' (from TaskUtils)

HardwareObjects/MicrodiffActuator.py
```diff
@@ -13,7 +13,7 @@
 
 import logging
 import time
-from abstract.AbstractActuator import *
+from abstract.AbstractActuator import AbstractActuator
```
rhfogh: You aslso need to import 'task' (from, TaskUtils)

HardwareObjects/mockup/MultiCollectMockup.py
IvarsKarpics: You are right. Not imported task broke the code (tested). I fixed it

HardwareObjects/mockup/MultiCollectMockup.py
```diff
@@ -1,5 +1,5 @@
 from HardwareRepository.BaseHardwareObjects import HardwareObject
-from abstract.AbstractMultiCollect import AbstractMultiCollect
+from abstract.AbstractMultiCollect import AbstractMultiCollect, task
```
rhfogh: Can't we do 
from TaskUtils import task instead? It is a bit inelegant to get indirectly instead of looking where 'task' is defined.

IvarsKarpics: done

PR #293, opened 2018-11-30
Refactored 'from ... import *' and removed unused imports

rhfogh:


IvarsKarpics:
I thought at some point we agreed that the one who submits a PR should not accept/merge it?

rhfogh:
@IvarsKarpics 
Sorry about that, I was not sure, actually. But I was not sure either what I should do once you had approved it. When I review something I either merge it, or comment explicitly that it sounds OK but someone else needs to look at it. Is that what 'approve' is supposed to mean? I thought it meant 'Ivars says this is OK', which I thought was enough to merge on.

Should I wait for a third opinion next time?

rhfogh:
In fact that was exactly what happened recently with PR #286 : Marcus submitted it, you approved it, and Marcus merged it. I am happy to follow procedures either way, but I think we need to agree about what 'approve' means and how it should be handled.
Thanks to PK for calling my attention to #286.

IvarsKarpics:
I think it  would be good if we had at least 3 different persons involved in PR create/accept/merge process (like a pair-reviewing). Maybe if the PR is accepted and no one merges it for two weeks then the author may accept it.

rhfogh:
We could do that - though it is a new rule. The risk is, obviously, that PR's will languish for those two weeks, which will be messy now we are moving fast. There has been a problem, I seem to remember, with your PRs not being dealt with, and I have been the only person who did anything serious about some of them. If we now need a third person, who will do the looking? 

Anyway, what do other people think?

beteva:
I guess if you need more than one person to review, one could ask for it via the Reviewers (top right on this page). Otherwise, the one who reviews is the one who accepts the PR.

marcus-oscarsson:
Sounds like a good idea to me as well

PR #297, opened 2018-12-04
Refactored some 'from ... import *' and cleaned up imports

rhfogh:
See what you think

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -1,11 +1,10 @@
 import sys
 import logging
-import types
-import dispatcher
-from dispatcher import *
+from dispatcher import dispatcher
 from CommandContainer import CommandContainer
 
 if sys.version_info > (3, 0):
+    # This is bad practice, but presumably there is a good reason?
     from HardwareRepository import *
 else:
     import HardwareRepository
--- a/CommandContainer.py
+++ b/CommandContainer.py
@@ -7,10 +7,9 @@ for command launchers and channels (see Command package).
 - C*Object, command launcher & channel base class
 """
 
-import types
 import weakref
 import logging
-from dispatcher import *
+from dispatcher import dispatcher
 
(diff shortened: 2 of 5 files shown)


BaseHardwareObjects.py
```diff
@@ -1,11 +1,10 @@
 import sys
 import logging
-import types
-import dispatcher
-from dispatcher import *
+from dispatcher import dispatcher
 from CommandContainer import CommandContainer
 
 if sys.version_info > (3, 0):
+    # This is bad practice, but presumably there is a good reason?
```
marcus-oscarsson: Hm, its a good question, how would this even work in Python 3. Because HardwareRepository.HardwareRepository() would probably not work properly right ?

IvarsKarpics: I think it is worth consider compatibility with Python3 as a next milestone (after 3.0.0)

marcus-oscarsson: I agree, but this seems to be a fix for Python 3, or am I mistaken ?. But looking at the import in the statement is also looks like that would break or ?

beteva: agree with @marcus-oscarsson, this  concerns Python3. I believe this must have been put there by @douglasbeniz some time ago. It looks like there should be no reason to use this.

marcus-oscarsson: You can remove the condition in my opinion

vrey01:
If kept, that line will break code every time there is a 
   HardwareRepository.HardwareRepository() 
call later in that module.

It should be removed

Bixente
______________________

Bixente Rey Bakaikoa
vrey01@gmail.com
+34 696.613.500

BaseHardwareObjects.py
rhfogh: OK, I shall merge this for now (it is approved, after all), and remove the conditional 'import *' in my next PR.

Issue #291, opened 2018-11-30
Move example xmls from mxcube/ExampleFiles to HardwareRepository/examples?


IvarsKarpics:
Should we move site specific configuration xmls from mxcube/ExampleFiles to a new directory HardwareRepository/examples?

rhfogh:
Sounds like a good idea - these parameters are not really UI-specific. But we would have to see that the MXCuBE3 people feel about it.

marcus-oscarsson:
The xml example files are already in the test directory of HardwareRepository, alltough we can move those files to another place if you think its better. I have nothing against including the site specific files as well, we should perhaps be careful of what we put there, its after all a example directory. 

The other files that are not configuring hardware objects should in my opinion not at all be in the HardwareRepository repository.

Marcus

PR #301, opened 2018-12-06
Updated Qt xmls

IvarsKarpics:
Following xmls will be removed from mxcube/ExampleFiles/HardwareObjects.xml
and site specific examples will be moved to HardwareRepository/examples directory (PR will follow soon)


IvarsKarpics:
Refers to #291

PR #302, opened 2018-12-06
Moved example xmls from mxcube/ExampleFiles to examples

IvarsKarpics:
Moved all site specific example xmls files from mxcube/ExampleFiles to HardwareRepository/examples


IvarsKarpics:
Refers to #291

rhfogh:
Is 'examples' really the best directory name? Would 'configuration' not be better - after all this is not just for info and education, it is supposed to be what people are actually using, or no?

IvarsKarpics:
'configuration' sounds good to me.

PR #304, opened 2018-12-07
Added pytest to check beamline setup attributes

IvarsKarpics:


IvarsKarpics:
One can accept this PR when all travis tests pass

__init__.py
```diff
@@ -2,7 +2,13 @@
 from logging.handlers import RotatingFileHandler
 import os
 import sys
-import HardwareRepository
+
+if sys.version_info > (3, 0):
+    # This is bad practice, but presumably there is a good reason?
+    # replaced back, because it breaks for python3
```
rhfogh: Can someone explain what is happening here? Is mxcube3 using  a different Pythonpath from mxcube2? Or does the import statement work differently? This is really opaque.

IvarsKarpics: Its nothing to do with mxcube3. The import statement do not work in python 3, so this is a workaround. In general I would like to trigger a discussion about naming, because, for example in BaseHardwareObject we see something like:

import HardwareRepository
hw_object = HardwareRepository.HardwareRepository().getHardwareObject(reference)

It is difficult to understand what HardwareRepository here means: is it a module, class or method.  A quick solution would be to substitute HardwareRepository() with getHardwareRepository(). A proper solution would be to rename module or (easier) class.

rhfogh: I completely agree about the namingconfusion  and the need for renaming.

However, I think there is another problem: circular imports, and that this is why otherwise sensible import statements will not work in Python 2. The problem is that 

- HardwareRepository.__init__ imports HardwareRepository.HardwareRepository
- HardwareRepository.HardwareRepository imports BaseHardwareObjects
- BaseHardwareObjects imports HardwareRepository.HardwareRepository

See my PR #306, that breaks the circular imports and proposes a form of import statement that might work but needs testing.

rhfogh: Not sure I should do this myself now (and I certainly have no opinion on the testing code), but I think this one should be merged, unless there are other problems with it than the one Ivars and I have been discussing. Those can then be dealt with later, as part of my #306

PR #306, opened 2018-12-10
Resolved circular import and python 2v3 problem (?)

rhfogh:
Two parts:
1) removed circular import from BaseHardwareObjects, and proposed an import statement that ought to work for both Python 2 and Python3. This is tested only in Python 2; could comeone test it in Python 3?

2) cathc-up removsal of queue_model_objects_v1 and otehr unused import statements

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -3,12 +3,20 @@ import logging
 from dispatcher import dispatcher
 from CommandContainer import CommandContainer
 
-if sys.version_info > (3, 0):
-    # This is bad practice, but presumably there is a good reason?
-    from HardwareRepository import *
-else:
-    import HardwareRepository
+# if sys.version_info > (3, 0):
+#    # This is bad practice, but presumably there is a good reason?
+#    # replaced back, because it breaks for python3
+#    from HardwareRepository import HardwareRepository
+# else:
+#     import HardwareRepository
 
+def getHardwareRepository():
+    """Get a HardwareRepository instance.
+
+    NB Necessarty because putting the import statement at the top level
+    leads to circular imports"""
+    from .HardwareRepository import HardwareRepository
+    return HardwareRepository()
 
 class PropertySet(dict):
     def __init__(self):
@@ -152,7 +160,7 @@ class HardwareObjectNode:
                 self.__references.pop()
             )
 
-            hw_object = HardwareRepository.HardwareRepository().getHardwareObject(
+            hw_object = getHardwareRepository().getHardwareObject(
                 reference
             )
 
@@ -305,7 +313,7 @@ class HardwareObject(HardwareObjectNode, CommandContainer):
         return self.name()
 
     def __setstate__(self, name):
-        o = HardwareRepository.HardwareRepository().getHardwareObject(name)
+        o = getHardwareRepository().getHardwareObject(name)
         self.__dict__.update(o.__dict__)
 
     def __bool__(self):
@@ -397,7 +405,7 @@ class HardwareObject(HardwareObjectNode, CommandContainer):
 
             if isinstance(node, HardwareObject):
                 if len(updates) > 0:
-                    HardwareRepository.HardwareRepository().update(node.name(), updates)
+                    getHardwareRepository().update(node.name(), updates)
                 return []
             else:
                 return updates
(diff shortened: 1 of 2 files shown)


BaseHardwareObjects.py
```diff
@@ -3,12 +3,20 @@
...
-else:
-    import HardwareRepository
+# if sys.version_info > (3, 0):
+#    # This is bad practice, but presumably there is a good reason?
+#    # replaced back, because it breaks for python3
+#    from HardwareRepository import HardwareRepository
+# else:
+#     import HardwareRepository
 
+def getHardwareRepository():
```
IvarsKarpics: I would suggest to move the method to HardwareRepository.py and rename it to getHardwareRepositoryInstance.

IvarsKarpics: Sounds good to me. Lets see what others think...

rhfogh: This would amount to 
- renanimg the HardwareRepository.HardwareRepository() *function* (sic) to 
HardwareRepository.getHardwareRepositoryInstance() (clearly an improvement), 
- fixing all the 42 places in the code that calls it, and 
- moving the imports in BaseHardwareObjects down inside the respective function bodies. 
Can do.

As you say, let us see what the others think before starting to make changes.

marcus-oscarsson: Its definitely an improvement how about accessing it through a global variable called HR or similar. I don't really mind and I think its clear improvement but writing getHardwareRepositoryInstance() every time might get a bit tiring :) 

You could then do `from HardwareRepository import HR`, we would of-course need to make sure HR is properly
initialized. We could also add a member to a HardwareObject which points to its instance of HardwareRepository, 
which would perhaps in many cases avoid the circular reference problem.

Cheers,
Marcus

rhfogh: @marcus-oscarsson 
That looks hard to set up.

The RepositoryServer setup is quite complex, with a _hwrserver instance and a separate  __HardwarerepositoryClient instance. There is provision for either a SpecConnectionsManager or simply a series of directories to be the source of HardwareObjects The actual location of the xml files (when you are dealing with xml files) are set from an environmental variable for testing (in scores of different places, from the same environment variable). For running the program it is set at the start from startGui. 

So, having a single instance, reliably initialised to one value, does not look like an easy thing to achieve while keeping all the existing functionalty. 

I could do a fair degree of simplification without breaking anything, but I would have to do it through a function (not an accessible instance) on HardwareRepository to keep functionality. Alternatively, if you want to refactor the whole thing, you would almost have to get someone to do it who understands the reasons behind the current complexity. Which I do not.

marcus-oscarsson: I see, well I think your current suggestion is clear improvement so lets go for that for the moment, since it seems like implementing it will be quite fast. I can try to get some time after Christmas to look at the possibilities to simplify the HardwareRepository. What do you think about that ?

Marcus

rhfogh: That sounds excellent. While I am at it I think I shall put the defaulting of the HardwareDirectiory to the one set by the envieonment variable into the gette function - that saves haveing the same five lines of code in every single test function.
Anyaway, as soon as I can get to it.

rhfogh:
I'll withdraw this one and make a replacement instead

beteva:
Hello everybody, There is something that does not go in the line we have discussed.  There are PR changing to  getHardwareRepository, which uses CamelCase we want to remove. Could we, please, first solve that and than change all the files, just not to do the work again. @marcus-oscarsson had a proposal which we did not even discuss!!!

rhfogh:
Well, yes, we do need to change over to snake_case. But I am still a little confused what you are saying exactly:

- What proposal of Marcus' do you mean exactly? 

- For a change that ranges so widely across both repositories, like this one, I think it might sometimes be better to do the substantial change separate from the format change. If it is just a format change you can read a lot of files and nod through the changes. If you need to actually understand them, because some of them mean something, it takes much longer. 

- Anyway, maybe I should have made the formatting change at the same time. But at the point we are at now, where the PR is accepted, and the corresponding PR for the other repository is still waiting, do you want us to reopen the PR and redo all the work? Or could we get this merged and then start doing case changes as a separate operation? That is certainly what I would prefer if I had to do the work.

PR #307, opened 2018-12-10
Fixed lims xml

IvarsKarpics:


rhfogh:
I find the string 'ISPyBClient2Mockup' in three places:

- configuration/desy_p11/lims-client-mockup.xml
- test/xml-qt/lims-client-mockup.xml
- test/xml-web/lims.xml

Should not all three be fixed?

IvarsKarpics:
Yes, I will fix it

PR #309, opened 2018-12-11
EMBL hwobj update

IvarsKarpics:


IvarsKarpics:
Sorry, I had to pull all changes from all beamlines and in the end it was difficult to merge in the current version, so i rolled back to version without black formatting. Somehow I can not install black on my ubuntu machine. It looks like it needs python3.6.
Maybe wait a little bit before I set up black and submit the changes.

rhfogh:
Sounds like a perfectly reasonable explanation.

HardwareObjects/EMBL/EMBLAutoProcessing.py
```diff
@@ -48,11 +48,13 @@ def __init__(self, name):
         self.current_autoproc_procedure = None
 
     def init(self):
-        self.autoproc_programs = self["programs"]
+        try:
+            self.autoproc_programs = self["programs"]
+        except BaseException:
```
rhfogh: As  I read it, you whoudl never put BaseExcption in an except clause.
AT most you can put excdeption, which at least does nto catch e.g. keyboard interrupts.
What you ought to put here is the expected exception, presumably KeyError

beteva: Actually BaseException is added by black if it finds except without any exception specified. I agree that in this case KeyError is more appropriate. And indeed black needs >=python3.6

IvarsKarpics: Fixed

HardwareObjects/EMBL/EMBLMiniDiff.py
```diff
@@ -34,7 +33,7 @@
         )
 
 from GenericDiffractometer import GenericDiffractometer
-from HardwareRepository.TaskUtils import task
+from HardwareRepository.TaskUtils import *
```
rhfogh: This is reverting a recent fix, Could you not avoid it?

IvarsKarpics: sorry, I will fix it

HardwareObjects/EMBL/EMBLMiniDiff.py
```diff
@@ -300,11 +305,17 @@ def refresh_omega_reference_position(self):
             reference_pos = self.omega_reference_motor.get_position()
             self.omega_reference_motor_moved(reference_pos)
 
+    def set_static_pixels_per_mm(self, pixels_per_mm):
+        self.static_pixels_per_mm = pixels_per_mm
```
rhfogh: If possible it would be better ot avoid storing the same value in two different ways

rhfogh: OK, so pixels_per_mm_x is not in general the same as static_pixels_per_mm[0], even if they are set to the same value in this code? A bit unobvious from the variable names, the ignorant could easily mix them up, but OK, this is an EMBL object, so as long as you are on top of it ...

PR #311, opened 2018-12-11
Fixed Session hwobj

IvarsKarpics:
* Fixed archive path.
* Commented two not used variables
* new EMBLSession.py

### DIFF
--- a/HardwareObjects/Session.py
+++ b/HardwareObjects/Session.py
@@ -164,36 +164,24 @@ class Session(HardwareObject):
         else:
             start_time = time.strftime("%Y%m%d")
 
-        if self.synchrotron_name == "EMBL-HH":
-            if os.getenv("SUDO_USER"):
-                user = os.getenv("SUDO_USER")
-            else:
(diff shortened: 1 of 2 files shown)


HardwareObjects/Session.py
```diff
@@ -135,15 +135,16 @@ def set_base_data_directories(
...
+        #self.archive_data_folder_name = archive_folder
 
         if self.base_directory is not None:
             queue_model_objects.PathTemplate.set_data_base_path(self.base_directory)
 
         if self.base_archive_directory is not None:
             queue_model_objects.PathTemplate.set_archive_path(
-                self.base_archive_directory, self.archive_folder_name
+                self.base_archive_directory, archive_folder
             )
```
rhfogh: Just a suggestion - should 
        raw_folder="RAW_DATA",
        process_folder="PROCESS_DATA",
        archive_folder="ARCHIVE",
be put a class-level constants, so code can access them from other places, rather than having them only here as defalut values?
This should not hold up the commit, it is just a suggestion.

PR #313, opened 2018-12-12
Solves issue #283

IvarsKarpics:


rhfogh:
Would this not be a good opportunity to get rid of 'from ...SampleChanger import *'?
There are only three objects defined in the file: SampleChanger, SampleChangerMode, and SampleChangerState

rhfogh:
Also - I am happy either way - but were we not going to keep the SampleChanger module in the sample_changer directory, and only move the concrete implementations? If e.g. Component and Container are staying, should AbstractSampleChanger then not stay too?

HardwareObjects/Cats90.py
```diff
@@ -24,7 +24,7 @@
 import PyTango
 import logging
 
-from GenericSampleChanger import *
+from abstract.AbstractSampleChanger import *
```
marcus-oscarsson: We should perhaps also do something about these * imports, but maybe in another PR. Ill create an issue for all * imports. 

By the way I'm actually already on early Christmas vacation, so I'm not really working ;)

Cheers,
Marcus

marcus-oscarsson:
Thanks for doing this @IvarsKarpics , I wanted to do it my self but I did not find the time before I left for vacation. There is a conflict here that needs to be resolved

marcus-oscarsson:
@rhfogh I also commented on the * imports, and I noticed your comment on it after. I created an issue for removing any remaining * import (#320)

marcus-oscarsson:
@rhfogh You are also right somehow for the AbstractSampleChanger. We said however that all abstracts should go to /abstract. Thus I think its maybe more consequent to move it to /abstract, another solution would be to move all of sample_changer to abstract. Since what one often really wants to use in sample_changer is the AbstractSampleChanger

rhfogh:
I think we should move all of sample_changer fo abstract. The pieces belong together, and they collectively make up a module describing and abstract sample changer, right?

marcus-oscarsson:
That's what I thought as well, @IvarsKarpics what do you think ?

beteva:
+1 following @marcus-oscarsson and @rhfogh

Issue #314, opened 2018-12-12
Clean up hardware objects


IvarsKarpics:
Move site specific hardware objects to site specific directories (based on the[ table of used hardware objects](https://docs.google.com/document/d/1JgEI23YaVbV9FZewOeHRFJNyh7lXF5P5v-pa9SHKM_s/edit#heading=h.l75ak8uw4342)).

IvarsKarpics:
Hi @beteva , @marcus-oscarsson ,

I see duplicated BlissActuator.py in HardwareObjects and in HardwareObjects/ESRF. Which file should be kept?

beteva:
On the other hand @IvarsKarpics is right - everything labeled ESRF... (like the ESRFCryoMon.py) should be moved to the ESRF directory.

beteva:
The HardwareObjects/ESRF/BlissActuator.py can be removed. Thanks @IvarsKarpics

PR #315, opened 2018-12-12
Issue #314

IvarsKarpics:


marcus-oscarsson:
Hi @IvarsKarpics 

I'm not so sure that moving Bliss related hardware objects to ESRF is the best solution, since they are intended to be general for the Bliss control system. Its true that its currently only being used at ESRF but it could possible be used at other sites and does not have to be ESRF specific. 

What does @beteva say ?

beteva:
Agree with @marcus-oscarsson. It is a control system like Tango, Tine, Epics, spec ..., its place is in the HardwareObjects and not in ESRF only.

PR #316, opened 2018-12-12
Small fixes in config for the web version

IvarsKarpics:
* Added missing reference to camera
* Added missing energy_motor and cleaned up other related xml
* Removed PyTango imports from mockup

Otherwise mxcube3 runs(at least starts) without problem.


marcus-oscarsson:
Thanks @IvarsKarpics

marcus-oscarsson:
@meguiraun, @IvarsKarpics , sorry noticed that you Mikel was also marked as reviewer. I hope you are fine with this :)

PR #317, opened 2018-12-12
Cleared circular imports and changed to getHardwareRepository()

rhfogh:
Goes with similare PR for mxcube


rhfogh:
Had to change the test code as well, but the test now passes

PR #318, opened 2018-12-12
Add gphl HardwareObjects to setup

rhfogh:
This one may be controversial, but I would like an opinion. It would simplify my life if I could use the standard xml-qt beamline-setup also for my own mock runs, without having to rely on a different local copy one way or the other. That would require adding links to two hardware object files. If the gphl-setup file is not found (as it normally will not be) all that happens is that you get a line in the log saying that the GPhL code is not installed.

Let me know what you think


rhfogh:
OK. I did not get thet error, because it came from a missing dependency of GphlWorkflow - which I of course had installed. I have now moved also gphl-workflow.xml away from the main config directory and into gphl_mock. That should remove the error message, and people will always have gphl-workflow and gphl-setup configured if they want to use it.

Can you see how it looks now?

PR #327, opened 2018-12-17
Added Info on how to test a PR

IvarsKarpics:


CONTRIBUTING.md
```diff
@@ -60,8 +60,15 @@ Pull request (PR) is the most convinient way of submitting a new code to the rep
 
 ### Accepting a pull request
 
-* The author of a pull request may request a PR review from a certain amount of developers.
+* The author of a PR may request a PR review from a certain amount of developers.
 * A reviewer can Comment, Approve or Request changes.
+* Before accepting the PR reviewer has to test the proposed code changes. To test the PR pull the proposed PR and run mxcube:
```
rhfogh: That is reasonable enough - but it does mean that I, for instance, will be unable to approve a lot of PRs, which contain code I cannot test, and will have a much higher activation energy for most of the rest. The recent 'moved xml test files' is a case in point. It seems OK to merge it after a quick look, given what it does, but it would take me half an hour or more to test it. Inevitably I will approve feewer PRs, and slower.

IvarsKarpics: Hi @rhfogh ,

I agree. Maybe we have to reformulate that sentence. In general those are contributing guidelines and do not cover everything.

marcus-oscarsson: I agree, and this is also why we have CI

rhfogh: Could we put in that before you *make* a PR you must try out the proposed code changes? Today I have, not for the first time, spent time on trying to find out why my mock installation was broken after fetching down the latest master. Only to find that changes that were merged to HardwareRepository a few days ago left MXCuBE unstartable, and the changes to mxcube needed to make it work again were only put up today? 

I know we are working very fast these days, and I appreciate how much work is being done. But could people commit a consistent and tested set of changes to both repositories, so that they at least, know that the program runs? Or, failing that, mark your PRs as WIP or 'incomplete', so that the rest of us do not fetch and merge them till they work for the originator?

IvarsKarpics: I do my best to keep both (mxcubeqt and HardwareRepository) repositories in sync. If there is a unstable, not running code please submit an issue or contact directly the one who submitted broken code. Indeed there are lots of structural changes lately and there will be even more, because next year I will start to cleanup (fix imports, rename directories, autopep and etc.) the mxcube qt code, so I suggest to be patient and follow to all changes.
Anyway, what should I do with this PR? Remove the sentence about testing?

marcus-oscarsson: I think that its great if each application that uses HardwareRepository is tested with it to ensure integrity. Its however up to the developers of that application, in our case Qt and Web front ends to make sure that the right version of HardwareRepository is used, and that the tests for that application runs correctly.

We should maybe add something about also adding a test case for new code so that the CI actually works as intended. For instance we could say that, at least, each abstract class should have a test case that tests its interface.

I'ts of course great if one tries to run MXCuBE as well, we could actually add a kind of integration test that initializes HardwareRepositry the same way as running MXCuBE.

In addition to this I always run manual tests on the beamlines in anyways, since running automatic tests with the hardware we have on the beamlines might get interesting ...

So finally, I would propose that we try to do our best to add test cases so that we can rely on the CI as much as possible.

IvarsKarpics: I agree that we have to use travis CI as much as we can. I am now trying to set up a test for qt gui with QtTest and the idea would be to simulate actual user interaction, so it tests HardwareRepository and gui at the same time.

rhfogh: @IvarsKarpics Thanks! I shall pull that down ASAP

rhfogh: @IvarsKarpics @marcus-oscarsson 
I was not (just) moaning - some of us have done an amazing amount of work recently, and 'when you chop wood the splinters fly', as the Russians say.

But I was thinking about having a requirement that before you make a PR you must open the program in mock mode and check if there are new errors.  That should be a fairly quick test, even if you have to do it manually, and it would force people to get a consistent state and avoid a few obvious errors. It is probably easier for the person making a PR to make that kind of test than it would be for anyone else.

marcus-oscarsson: @IvarsKarpics @rhfogh , thanks for the fix Ivars and I don't think we perceived it as you were moaning Rasmus :). In anyways, I think its good practice to ensure that the changes one just have made are well reflected in the application and that the application still works, so adding (or keeping) the sentence about running MXCuBE in mockup mode is fine in my opinion. I'm more for automating this but manual checks are always a good complement and sometimes even necessary. We have not said if we should run both MXCuBEs or if one is sufficient ?

PR #328, opened 2018-12-17
Added new collection items to handle ssx and imaging dc

IvarsKarpics:
* Small fixes for XMLRPC : by defaults set all interfaces to false:
* Cleaned up GenericParallelProcessing. Soon will be renamed to AbstractParallelProcessing
* Added two new experiment types: still (for serial mx) and imaging (xray tomography).
* Added missing methods

### DIFF
--- a/HardwareObjects/GenericParallelProcessing.py
+++ b/HardwareObjects/GenericParallelProcessing.py
@@ -21,14 +21,15 @@ import os
 import time
 import logging
 import gevent
-import numpy
 import subprocess
+import numpy as np
 
 from copy import copy
 from scipy import ndimage
 import matplotlib.pyplot as plt
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
+from QtImport import *
 import SimpleHTML
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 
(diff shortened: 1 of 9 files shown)


HardwareObjects/XMLRPCServer.py
```diff
@@ -155,14 +155,11 @@ def init(self):
         Method inherited from HardwareObject, called by framework-2.
         """
 
-        self.all_interfaces = self.getProperty("all_interfaces")
+        self.all_interfaces = self.getProperty('all_interfaces', "False")
```
rhfogh: Are you sure this will work? It looks to me like self.all_interfaces will default to the string "False", which is *True* by definition in python (as a no-empty string).

Did you test this?

HardwareObjects/GenericParallelProcessing.py
```diff
@@ -21,14 +21,15 @@
...
-import numpy
 import subprocess
+import numpy as np
 
 from copy import copy
 from scipy import ndimage
 import matplotlib.pyplot as plt
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
+from QtImport import *
```
rhfogh: Qt import in a HardwareObject (and with 'import *' too).
Presumably this is unavoidable?

HardwareObjects/XMLRPCServer.py
IvarsKarpics: Thanks @rhfogh  for the review. Indeed in my case all_interface was True, so It did not failed. I will fix it (also the QtImprt) in a minute.

HardwareObjects/GenericParallelProcessing.py
IvarsKarpics: This import was not necessary

IvarsKarpics:
Hi @rhfogh ,

I edited code according to you suggestions. Is it ok to accept the PR?

PR #329, opened 2018-12-17
Added proposed licensing notices.

rhfogh:
To discuss and later expand once we agree.


rhfogh:
This is a proposal, with license info added only to GPhL files. Once we agree they shold spread everywhere.

This proposal is based on the information in https://www.gnu.org/licenses/gpl-howto.html

The main points are:

- Adding COPYING.LESSER as well as COPYING

- Putting a top-level statement in README (My idea, but surely it makes sense)

- Copyright holder and years

- License statement in each file (to make sure the license remains if files are taken out)

- Optionally one could put the LGPL logo on the spalsg screen, but I am not sure I think it is worth it.

beteva:
Hello @rhfogh . Thank you very much for reviewing all this. I have one question and one comment. Does the phrase¨ without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.¨ implies that you cans sell a product that contains part or whole MxCUBE? I know it was there before, but as we review it, is is still correct.
The comment is that __copyright__ = and __license__ = should come after the imports (otherwise black reports an error, especially if you use from warnings import warn), so maybe we can put the whole phrase ¨MXCuBE is free software...¨ at the beginning of the file and  the __license__ = """This file is part of MXCuBE. """ after the imports.

rhfogh:
@beteva 

Thanks. 

- The proposal I saw was to put the license at the start, but I was a bit unsure - and I think black has the clinching argument. I am not quite sure what bits of text you want to put where, though, can you elaborate? One possibility would be to put the license in the header comment, and the copyright after the imports. Another would be to put both after the imports. 

- The points about MERCHANTABILITY and FITNESS FOR PURPOSE are directly as recommended by GNU. As I understand it you can actually sell a product that includes MXCuBE in whole or in part, that is part of LGPL, but that is not what this is about. Rather I think it says that we are making no guarantees whatsoever, so if you use MXCuBE you have the entire responsibility for any consequences. Even if a bug in MXCuBE crashes your synchrotron or causes someone to be irradiated, you still cannot sue us for damages.

A technical point:
I would propose that we finish the discussion first, and only then do I withdraw this proposal and replace it with what we agree we should do.

beteva:
@rhfogh I was thinking of putting the license in the header comment, and the copyright after the imports. Thus the code will be easier to read.
Thanks for explaining the terms and sorry for being so ignorant as soon as it comes to licensing.
OK for COPYING.LESSER. Maybe instead of README, we should put all that concerns legal aspects to LICENSING or something similar. OK for the copyright as well, but I think it will be nice to have \__author\__ as well. I would avoid the LGPL logo on the splash screen.

README.md
```diff
@@ -4,3 +4,17 @@
 HardwareRepository is the back-end level of [MXCuBE Qt](https://github.com/mxcube/mxcube/) and [MXCuBE Web](https://github.com/mxcube/mxcube3/)  allowing to access the beamline control system and instrumentation. It acts as a container of single python objects (called hardware objects).
 
 Please read the [contributing guidlines](https://github.com/mxcube/HardwareRepository/blob/master/CONTRIBUTING.md/) before submiting your code to the repository.
+
+
+	MXCuBE is free software: you can redistribute it and/or modify
```
IvarsKarpics: Maybe remove the spaces before the text?

HardwareObjects/GphlMessages.py
```diff
@@ -3,6 +3,25 @@
 """ Abstract beamline interface message classes
 """
 
+__copyright__ = """ Copyright © 2016 - 2018 by Global Phasing Ltd. """
+
+__license__ = """This file is part of MXCuBE.
```
IvarsKarpics: I would suggest to use __license__ = "LGPLv3+" and move the statements about the licence to the top of the file, like in:
https://github.com/mxcube/HardwareRepository/blob/b15650e98f5c9aaa9630b09c72f8aebe6e9cfd71/HardwareObjects/GenericParallelProcessing.py#L1-L18

rhfogh:
I am withdrawing this PR and replacing it with one that incorporates your comments.

PR #331, opened 2018-12-19
Proposed license info

rhfogh:
New proposal for adding license information, incorporating input from @beteva, @IvarsKarpics, @marcus-oscarsson 

This is intended as a  template for everybody to use, so once we are sure we agree it should start to spread.


HardwareObjects/GphlWorkflowConnection.py
```diff
@@ -43,13 +63,9 @@
...
         robustapply.robust_apply = robustapply.robustApply
 
-
-__copyright__ = """
-  * Copyright © 2016 - ${YEAR} by Global Phasing Ltd. All rights reserved
-"""
-
-__author__ = "rhfogh"
-__date__ = "04/11/16"
+__copyright__ = """ Copyright © 2016 - 2018 by Global Phasing Ltd. """
```
IvarsKarpics: In general I am fine with the PR. I just have one question: should we then add our own copyright to each file? What about core and abstract classes? Should we use just MxCuBE colaboration for all files?

rhfogh: I was wondering about that, actually. There would certainly be a case for having 'Copyright MXCuBE collaboration' on all files. The point where it matters is, I guess, that it is only the copyright holder who can relicense under another license. 

Maybe we need to pass this one on to the steering committee?

HardwareObjects/GphlMessages.py
```diff
@@ -1,26 +1,24 @@
 #! /usr/bin/env python
 # encoding: utf-8
 """ Abstract beamline interface message classes
```
beteva: Suggest this lines goes after the Licence chapter and not included in it.

HardwareObjects/GphlMessages.py
```diff
@@ -1,26 +1,24 @@
...
-"""
-
-__copyright__ = """ Copyright © 2016 - 2018 by Global Phasing Ltd. """
 
-__license__ = """This file is part of MXCuBE.
+License:
```
beteva: Suggest to remove this empty line

HardwareObjects/GphlMessages.py
```diff
@@ -1,26 +1,24 @@
...
 
-	MXCuBE is free software: you can redistribute it and/or modify
-	it under the terms of the GNU Lesser General Public License as published by
-	the Free Software Foundation, either version 3 of the License, or
-	(at your option) any later version.
+This file is part of MXCuBE.
```
beteva: Suggest to remove this empty line

HardwareObjects/GphlMessages.py
```diff
@@ -1,26 +1,24 @@
...
-	along with MXCuBE.  If not, see <https://www.gnu.org/licenses/>.
-	"""
+MXCuBE is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
 
+You should have received a copy of the GNU Lesser General Public License
+along with MXCuBE.  If not, see <https://www.gnu.org/licenses/>.
+"""
```
beteva: Put here `""" Abstract beamline interface message classes  """`
Thus we can insert all the specific documentation.

HardwareObjects/GphlMessages.py
rhfogh: Hmmm, I am not sure. The header comment is where it should say what the file is for. Is it not better to put that at the top, where people see it properly, rather than at the end of the license statement, where the eye most likely is already glazed over? What do other people think?

Anyway, will black not complain if we put '__license__ =' ahead of the import statements?

HardwareObjects/GphlMessages.py
beteva: In this file there is only one line, but in many files you have more than this. The information will be somehow diluted. Or maybe we should do as before - the licence chapter has `#` for each line, while the rest is within docstrings.

HardwareObjects/GphlWorkflowConnection.py
marcus-oscarsson: Hm, yes I think its something that should be discussed in the steering committee. It would be natural to have 'Copyright MXCuBE collaboration' on all files. But I understand that it would be good as a precaution to use something else in some cases.

marcus-oscarsson:
@rhfogh thanks for this, we need to discuss it internally here so there might be some additions later on. However I will approve the PR as such ...

PR #332, opened 2018-12-19
Fixed bug introduced by PR #328

IvarsKarpics:


IvarsKarpics:
strange that it was not noticed by travis and pytest...

HardwareObjects/GenericParallelProcessing.py
```diff
@@ -112,7 +112,7 @@ def init(self):
         )
         self.start_command = str(self.getProperty("processing_command"))
         self.kill_command = str(self.getProperty("kill_command"))
-        self.grid_properties = eval(self.getProperty("grid_properties", []))
+        self.grid_properties = eval(self.getProperty("grid_properties", "[]"))
```
rhfogh: Are you sure about this? Looking at the PYthon documentation (both 2.7 and 3.7) I get:
 eval(expression, globals=None, locals=None)
    The arguments are a string and optional globals and locals. If provided, globals must be a dictionary. If provided, locals can be any mapping object.

So the list is wrong, but the string would also be wsrong, or no?

IvarsKarpics: why string is wrong? getProperty will return str if grid_properties is defined, otherwise eval will take default value "[]"

rhfogh: My apologies - I cannot read. I thought "[]" was a parameter to eval, not to getProperty.
Objection withdrawn.

PR #333, opened 2018-12-19
Improved pytests for travis

IvarsKarpics:
I realized that the existing pytest is not accurate. It loads beamline setup, but actually do not test if all necessary hwobj are initialized. So I added a more detailed check for attributes.
* also there was no test against web xmls: added them as well.


requirements_2.txt
```diff
@@ -7,3 +7,4 @@ PyDispatcher
 numpy
 scipy
 matplotlib
+Pillow
```
marcus-oscarsson: Why is this file called requirements_2 ?

marcus-oscarsson: Actually I noticed that there are now requirements, requirements_2 and requirements_3. I guess these were added to handle the requirements differences between web and qt front ends ?.

I think the specific details of each front end should be tested in respective repository, so these tests, should exclusively test HardwareRepository. The same goes for the requirements, I can understand that there might be some exceptions and that one need a different library in some cases. Those should however be quite few ?

If it is the case could we give the requirement files a more intuitive name like with requirements_qt and requirements_web, that contain the additional requirements. I think however it would be great if we could stick
to a single requirements file for all of HardwareRepository.

marcus-oscarsson:
Thanks @IvarsKarpics it was exactly this kind of test that I mentioned in #327. I agree that all Qt related tests should be done in the Qt front end repo. Do you still rely on Qt in HardwareRepository ?

requirements_2.txt
IvarsKarpics: Hi @marcus-oscarsson , no requirements_2 is for python2 and ..3 for python3. I dont know how other people deal with various requirements for python 2 and 3, but this was a quick solution (see if part in .yaml file). I guess one can come up with a single file for both versions.

IvarsKarpics:
Yes, we need Qt in HardwereRepository. All graphics are written in Qt (Qt4_GraphicsLib and Qt4_GraphicsManager)

test/pytest/conftest.py
```diff
@@ -4,26 +4,15 @@
...
-    hwr.connect()
-    return hwr
+def hwr_qt():
+    return os.path.join(HWR, "configuration/xml-qt")
 
 @pytest.fixture(scope="session")
-def blsetup(hwr):
-    return hwr.getHardwareObject("beamline-setup")
+def hwr_web():
+    return os.path.join(HWR, "configuration/xml-qt")
```
rhfogh: Surely that should be 'configuration/xml-web'?

marcus-oscarsson: I also created an issue for harmonizing the XML files #335

marcus-oscarsson:
I have not looked at Qt4_GraphicsLib and Qt4_GraphicsManager in quite some time. But I guess Qt4_GraphicsLib could be fairly easily moved to the front-end repo ?. I realize that Qt4_GraphicsManager could be quite complicated to move though, what do you think ?. 

It would finally make it easier to manage the tests, dependencies and make the code base a bit cleaner, if we could move the front-end specific code to their respective repositories.

Let me know what you think ?

test/pytest/conftest.py
IvarsKarpics: ups! Sorry for typo. I fixed it

marcus-oscarsson: :), great thanks !

IvarsKarpics:
Hi @marcus-oscarsson ,

Moving Qt4_GraphicsManager to the front level is durable, but not a straight forward, because lots of hardware objects really on the graphics hardware object.
For time being I excluded the check for the graphics hardware object.

marcus-oscarsson:
I see, that's what I thought as well. I guess that's okey as long as there is now "strict" dependency on Qt so that people that don't need it version will have to deal with that. Its not currently a problem so I think we can leave it like this if nobody else disagrees, the Qt dependency can be seen as MXCuBE2 specific and are installed with MXCuBE2 in anyways.

marcus-oscarsson:
@rhfogh is it fine like this, or can we merge ?

rhfogh:
@marcus-oscarsson 
Fine with me. I am a bit hung up at the moment, but no objections.

Issue #276, opened 2018-11-29
AbstractDiffractometer


IvarsKarpics:

IvarsKarpics:
Will be close when PR #334

marcus-oscarsson:
Replaced by #672

PR #334, opened 2018-12-21
[WIP] AbstractDiffractometer issue #276

beteva:
AbstractDiffractometer initial version.
AbstractCentring initial version
Microdiff changed to follow AbstractDiffractometer.

### DIFF
--- a/HardwareObjects/Microdiff.py
+++ b/HardwareObjects/Microdiff.py
@@ -1,322 +1,207 @@
-from HardwareRepository.BaseHardwareObjects import Equipment
-import tempfile
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Example xml file
+<object class = "Microdiff"
+  <username>MD2S</username>
+  <exporter_address>wid30bmd2s:9001</exporter_address>
+  <exporter_commands>
+    <move_multiple_motors>SyncMoveMotors</move_multiple_motors>
+    <move_to_phase>startSetPhase</move_to_phase>
+    <scan_limits>getOmegaMotorDynamicScanLimits</scan_limits>
+    <abort>abort</abort>
+    <move_motors_sync>startSimultaneousMoveMotors</move_motors_sync>
+    <osc_scan>startScanEx</osc_scan>
+    <helical_scan>startScan4DEx</helical_scan>
+    <mesh_scan>startRasterScan</mesh_scan>
+    <still_scan>startStillScan</still_scan>
+    <set_plate_vertical>setPlateVertical</set_plate_vertical>
+  </exporter_commands>
+  <exporter_channels>
+    <scale_x>CoaxCamScaleX</scale_x>
+    <scale_y>CoaxCamScaleY</scale_y>
+    <head_type>HeadType</head_type>
+    <kappa_enable>KappaIsEnabled</kappa_enable>
+    <current_phase>CurrentPhase</current_phase>
+    <hwstate>HardwareState</hwstate>
+    <swstate>State</swstate>
+    <scan_range>ScanRange</scan_range>
+    <scan_start_angle>ScanStartAngle</scan_start_angle>
+    <scan_nb_frames>ScanNumberOfFrames</scan_nb_frames>
+    <scan_exposure_time>ScanExposureTime</scan_exposure_time>
+    <scan_detector_gate_pulse_enabled>DetectorGatePulseEnabled</scan_detector_gate_pulse_enabled>
+    <detector_gate_pulse_readout_time>DetectorGatePulseReadoutTime</detector_gate_pulse_readout_time>
+  </exporter_channels>
+</object>
+"""
+
 import logging
-import math
-import os
-import time
-from HardwareRepository import HardwareRepository
-import MiniDiff
-from HardwareRepository import EnhancedPopen
-import copy
 import gevent
+import math
+import gevent
+import time
+
+# import AbstractDiffractometer
+from AbstractDiffractometer import AbstractDiffractometer, DiffrHead, DiffrPhase
+
 import sample_centring
 
 MICRODIFF = None
 
 
-class Microdiff(MiniDiff.MiniDiff):
+class Microdiff(AbstractDiffractometer):
     def init(self):
-        global MICRODIFF
-        MICRODIFF = self
-        self.timeout = 3
-        self.phiMotor = self.getDeviceByRole("phi")
-        self.exporter_addr = self.phiMotor.exporter_address
-        self.x_calib = self.addChannel(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "x_calib",
-            },
-            "CoaxCamScaleX",
-        )
-        self.y_calib = self.addChannel(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "y_calib",
-            },
-            "CoaxCamScaleY",
-        )
-        self.moveMultipleMotors = self.addCommand(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "move_multiple_motors",
-            },
-            "SyncMoveMotors",
-        )
-        self.head_type = self.addChannel(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "head_type",
-            },
-            "HeadType",
-        )
-        self.kappa = self.addChannel(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "kappa_enable",
-            },
-            "KappaIsEnabled",
-        )
-        self.phases = {
-            "Centring": 1,
-            "BeamLocation": 2,
-            "DataCollection": 3,
-            "Transfer": 4,
-        }
-        self.movePhase = self.addCommand(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "move_to_phase",
-            },
-            "startSetPhase",
-        )
-        self.readPhase = self.addChannel(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "read_phase",
-            },
-            "CurrentPhase",
-        )
-        self.scanLimits = self.addCommand(
-            {
-                "type": "exporter",
-                "exporter_address": self.exporter_addr,
-                "name": "scan_limits",
-            },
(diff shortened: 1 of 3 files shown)


HardwareObjects/Microdiff.py
```diff
@@ -1,322 +1,207 @@
...
-            if self.swstate_attr.getValue() == "Ready":
-                return True
-        return False
-
-    def _wait_ready(self, timeout=None):
-        # None means infinite timeout
-        # <=0 means default timeout
-        if timeout is not None and timeout <= 0:
+    def wait_ready(self, timeout=None):
+        """ Waits when diffractometer status is ready
```
marcus-oscarsson: We should agree on the style used in the doc strings. This uses google style doc strings, while we so far have been using the original sphinx python domain style. I personally like the google version but I think its good if we could agree on one style, see #337

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+        return motors_positions_dict
+
+    def getPositions(self, motors_positions_list=[]):
+        warn(
+            "getPositions is deprecated. Use get_motor_positions instead",
+            DeprecationWarning,
+        )
+        return self.get_motor_positions(motors_positions_list)
+
+    def motor_positions_to_screen(self, centred_positions_dict):
```
marcus-oscarsson: We should perhaps use the common centering math object for doing  this

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+    def get_phase(self):
+        """ Get the current phase
+        Returns:
+            phase (enum): DiffrPhase enumeration element
+        """
+        return self.current_phase
+
+    """ snapshots """
+
+    def take_snapshots(self, nb_snapshots=0, rot_offset=90., wait=True):
```
marcus-oscarsson: I guess we can refine this along the road, but take_snapshots does not necessarily have to belong to the diffractometer

HardwareObjects/Microdiff.py
```diff
@@ -1,322 +1,207 @@
...
+            the availble motors
+        Args:
+            motors_positions_list (list): List of motor names or hwobj
+        Returns:
+            motors_positions_dict (dict): role:position dictionary
+        """
+        motors_positions_dict = AbstractDiffractometer.get_motor_positions(
+            self, motors_positions_list
+        )
+        if not self.in_kappa_mode():
```
marcus-oscarsson: I'm not sure about the in_kappa_mode. Id prefer that we always get all motors if nothing else is passed in the call. One could imagine an additional argument for instance include_kappa if you think its necessary ?

HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This has been put here only to trigger a conversation. I agree with @marcus-oscarsson that this method's place is rather in centring maths.

HardwareObjects/abstract/AbstractDiffractometer.py
beteva: I was also hesitating whether this method belongs to the diffractometer HO, but as this is the class that combines motors and camera, I've put it here.

marcus-oscarsson:
Maybe short notice, but we would appreciate some more feedback on this subject before tomorrows meeting. 

Cheers 
Marcus and Antonia !

marcus-oscarsson:
Thanks for your feedback @IvarsKarpics, I think @beteva 's intention with this PR was to provide a basis for the general concepts so that we have something to discuss around. The code itself should perhaps be interpreted more as pseudocode. 

The AbstractDiffractometer and the AbstractCollect are in my opinion one of the more important parts of the refactoring so I think it would be good if we could continue the discussion.

I believe we would be able to sort out the general concepts here in this PR for instance and then create a mockup implementation. It might as you say become necessary with more "intensive" interaction at some point, we can maybe start by having a video conference just around this topic. 

Cheers,
Marcus

beteva:
Thank you @IvarsKarpics for the numerous comments. As @marcus-oscarsson has mentioned, my intentions were to provoke a discussion, rather than imposing a finished solution. I have created the two abstract classes and changed Micriodiff to inherit and use the methods of Abstaract*. What I would really like is to agree on the needs and the methods (especially for the centring).

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+    Waiting = 18
+    Positioned = 19
+    Starting = 20
+    Loading = 21
+    Unknown = 22
+    Alarm = 23
+    Fault = 24
+    Invalid = 25
+    Offline = 26
+
```
jordiandreu: Maybe we could review this enum to remove some state values for simplicity.

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+
+@enum.unique
+class DiffrPhase(enum.Enum):
+    CENTRE = "Centring"
+    COLLECT = "DataCollection"
+    SEE_BEAM = "BeamLocation"
+    TRANSFER = "Transfer"
+    SEE_SAMPLE = "LightSample"
+
```
jordiandreu: We have phase concept defined at ALBA in a Tango DS. Mapping these definitions to our, the one missing from our side would be CENTRE which for us is equivalent to SEE_SAMPLE. What would be the difference between these two phases?

jordiandreu: Maybe we also need to add the UNKNOWN phase.

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+        # motors and direction initialisation
+        self.motors_hwobj = {}
+        self.directions = {}
+        try:
+            for role in self["motors"].getRoles():
+                motor = self["motors"].getObjectByRole(role)
+                self.motors_hwobj[role] = motor
+                self.directions[role] = 1
+                self.connect(motor, "stateChanged", self.emit_motor_state_changed(role))
```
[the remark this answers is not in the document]
jordiandreu: We are adding the role to the signal name, is this necessary?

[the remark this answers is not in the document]
jordiandreu: > I think this will not work or am I wrong? One can not add parameters to the slot.

No, but maybe defining a kind of slot factory method, which will return a function according to the role value.

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+        except IndexError:
+            pass
+
+        # actuators
+        self.actuators_hwobj = {}
+        try:
+            for role in self["actuators"].getRoles():
+                self.actuators_hwobj[role] = self["actuators"].getObjectByRole(role)
+        except IndexError:
+            print("No actuators configured")
```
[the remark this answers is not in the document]
jordiandreu: It would be worth to define a logger for each HardwareObject, inheriting from the HWR logger? In this way it would be easier to track from which module the log message comes from?

inside  __init__ method
> self.logger = logging.getLogger("HWR.AbstractDiffractometer")

HardwareObjects/abstract/AbstractDiffractometer.py
jordiandreu: I do agree

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+from __future__ import print_function
+import gevent
+import logging
+import enum
+from warnings import warn
+
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+from HardwareRepository.TaskUtils import cleanup, error_cleanup
+
+__credits__ = ["MXCuBE colaboration"]
```
jordiandreu: collaboration

HardwareObjects/abstract/AbstractDiffractometer.py
jordiandreu: What do you think about these naming:

SEE_SAMPLE -> SAMPLE_VIEW
SEE_BEAM -> BEAM_VIEW

HardwareObjects/Microdiff.py
```diff
@@ -326,221 +211,114 @@ def oscilScan4d(self, start, end, exptime, motors_pos, wait=False):
...
+        Returns:
+            phase (enum): DiffrPhase enumeration element
+        """
+        phase = self.channels["current_phase"].get_value()
+        for i in DiffrPhase:
+                if phase == i.value:
+                    self.current_phase = i
+        return self.current_phase
+
+    def get_head_type(self):
```
jordiandreu: This method is already defined in line 120.

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+            x = (phiy * self.pixels_per_mm_x) + self.beam_position[0]
+            y = dy + (phiz * self.pixels_per_mm_y) + self.beam_position[1]
+
+            return x, y
+        else:
+            raise NotImplementedError
+
+    """ Head type and modes """
+
+    def get_head_type(self):
```
jordiandreu: I think this method is expected to be implemented for each Diffractometer class inheriting from this AbstractDiffractometer and match the enum class defined for the head types.

```suggestion
    def get_head_type(self):
        raise NotImplementedError
```
And then also remove the **try/except** block as @IvarsKarpics commented in line 199.

jordiandreu: I think this method is expected to be implemented for each Diffractometer class inheriting from this AbstractDiffractometer and match the enum class defined for the head types.

```suggestion
    def get_head_type(self):
        raise NotImplementedError
```
And then also remove the **try/except** block as @IvarsKarpics commented in line 199.

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+
+    """ phases """
+
+    def set_phase(self, phase=None, wait=False):
+        """Sets diffractometer to selected phase
+        Args:
+            phase (str or enum): desired phase
+            waith (bool): wait (True) or not (False) until the phase is set
+        """
+        if isinstance(phase, str):
```
jordiandreu: Maybe:
`raise NotImplementedError`
since this needs to be implemented for each specific diffractometer?

HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,566 @@
...
+        Args:
+            timeout (int): timeout [s]
+        :type timeout: int
+        """
+        pass
+
+    """ motors handling """
+
+    def move_motors(self, motors_positions_dict, wait=False, timeout=20):
+        """ Move simultaneously specified motors to the requested positions
```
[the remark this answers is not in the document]
beteva: Agree, but the only way to make them move really simultaneously is that the underlying control can do it. Here I've only put a way to move quasi simultaneously, and indeed it also relies on the fact that a motor can start moving, without waiting the move to finish.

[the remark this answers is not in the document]
beteva: In fact I will just change the documentation. The idea of the method is that you can choose to move the motors simultaneously or not.

IvarsKarpics:
@beteva  and others: I would suggest to accept this PR and I will prepare an another PR with changes done during the dev meeting.

jordiandreu:
> @beteva and others: I would suggest to accept this PR and I will prepare an another PR with changes done during the dev meeting.

Better CLOSE it an open a new one? Go ahead.

marcus-oscarsson:
I'm also fine with merging it, there seems to be a merge conflict though. We can merge when that is solved ...

marcus-oscarsson:
Closing this as discussed in the January developers meeting, to be reopened in a different from at a later time

PR #338, opened 2019-01-08
Initial values of beam info

marcus-oscarsson:
Hi,

Making sure that initial values are set Initializing beam_size_aperture and beam_size_slits members with their corresponding "hardware" values

Marcus


marcus-oscarsson:
@IvarsKarpics I guess we can marge this then as you approved it, or do you want someone else to take a look ?

PR #339, opened 2019-01-09
Mockup channel and command

IvarsKarpics:
Added mockup channel and command to test the channel and command interface

### DIFF
--- a/CommandContainer.py
+++ b/CommandContainer.py
@@ -280,7 +280,7 @@ class CommandContainer:
 
                 newChannel = TineChannel(channelName, channel, **attributesDict)
             except BaseException:
-                logging.getLogger("GUI").exception(
+                logging.getLogger("HWR").exception(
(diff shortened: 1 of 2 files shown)


CommandContainer.py
```diff
@@ -315,6 +315,23 @@ def add_channel(self, attributesDict, channel, addNow=True):
...
+                try:
+                    attributesDict["default_value"] = float(self.default_value)
+                except AttributeError:
+                    pass
+
+            try:
+                from Command.Mockup import MockupChannel
+                newChannel = MockupChannel(channelName, channel, **attributesDict)
+            except BaseException:
+                logging.getLogger("GUI").exception(
```
marcus-oscarsson: Why is this logging to the "GUI" logger ?

IvarsKarpics: ups... I will fix it

PR #343, opened 2019-01-17
Added python3 support

IvarsKarpics:
Tested against mxcubeQt


HardwareObjects/GenericDiffractometer.py
```diff
@@ -944,6 +944,8 @@ def move_motors(self, motor_positions, timeout=15):
             motor_positions = motor_positions.as_dict()
 
         self.wait_device_ready(timeout)
+        result_dict = {}
```
rhfogh: Something must be missing. result_dict is created and filled ut not used for anything later.

HardwareObjects/abstract/sample_changer/Sample.py
```diff
@@ -1,5 +1,5 @@
-from Component import *
 import sys
+from .Component import *
```
rhfogh: I think the following shoudl work, and it is nicer:
from .Component import Component

HardwareObjects/abstract/sample_changer/Container.py
```diff
@@ -1,4 +1,4 @@
-from Sample import *
+from .Sample import *
```
rhfogh: I think the following should work, and it is nicer:

from .Component import Component
from .Sample import Sample

HardwareObjects/mockup/MicrodiffZoomMockup.py
```diff
@@ -39,11 +39,12 @@ def isReady(self):
 
     def sortPredefinedPositionsList(self):
         self.predefinedPositionsNamesList = self.predefinedPositions.keys()
-        self.predefinedPositionsNamesList.sort(
-            lambda x, y: int(
-                round(self.predefinedPositions[x] - self.predefinedPositions[y])
+        if hasattr(self.predefinedPositionsNamesList, "sort"):
```
rhfogh: An attribute called predefinedPositionsNamesList must be a list, and therefore must be sortable.
If this is a problem, the solution would be to make sure it always is a list.
Is it because the .keys() function returne an iterator, in newer pythons? Then you should cast it to a list explicitly.

HardwareRepository.py
```diff
@@ -25,9 +25,13 @@
 except ImportError:
     pass
 
-import HardwareObjectFileParser
-import BaseHardwareObjects
-from dispatcher import dispatcher
+from . import BaseHardwareObjects
+from . import HardwareObjectFileParser
+
+if sys.version_info[0] == 3:
```
rhfogh: Too bad this kind of thing cannot be avoided, but presumably you tried and it is not possible.

IvarsKarpics: Yes, I tried different options, but none of them worked.

marcus-oscarsson:
Does the HWR start with these changes using Python3 or is it just changes in progress ?

HardwareRepository.py
IvarsKarpics: Thanks @marcus-oscarsson  for the suggestion. It works.

marcus-oscarsson: :) Great !

IvarsKarpics:
@marcus-oscarsson : yes the HWR and Qt gui starts now without errors with python3 and python2

marcus-oscarsson:
That's great, so should soon work with MXCuBE3 as well then :)

HardwareObjects/queue_model_objects.py
```diff
@@ -1642,8 +1642,8 @@ def __init__(self, motor_dict=None):
             setattr(self, motor_name, None)
 
         if motor_dict is not None:
-            for motor_name, position in motor_dict.iteritems():
-                setattr(self, motor_name, position)
+            for motor_name in motor_dict.keys():
```
rhfogh: Would it not be better with 
for motor_name, position in motor_dict.items():

HardwareObjects/mockup/MicrodiffZoomMockup.py
IvarsKarpics: I removed all that code for time being because it works also without sorting. If it fails then I suggest to open an another issue.

HardwareObjects/queue_model_objects.py
IvarsKarpics: done

marcus-oscarsson:
Hi @IvarsKarpics 

Actually running this I get quite a few issues with relative imports, for instance in queue_entry were there is from . import. Are you getting the same ?

Marcus

beteva:
Maybe this can sole the import problem
from __future__ import absolute_import

marcus-oscarsson:
Hm, I don't think so since the tests are with python3

marcus-oscarsson:
Hi @IvarsKarpics, I'm using python 3.7 and I get some interesting errors with the following message,

    ImportError: attempted relative import with no known parent package

Its on all the "from . import" statements. I thought that you've maybe seen the error and made a fix for it somewhere. 

I believe its due to another relative import somewhere so that the package name gets lost, I'm looking into it ...

HardwareObjects/mockup/MicrodiffZoomMockup.py
```diff
@@ -27,73 +24,65 @@ def init(self):
...
-    def sortPredefinedPositionsList(self):
-        self.predefinedPositionsNamesList = self.predefinedPositions.keys()
-        self.predefinedPositionsNamesList.sort(
-            lambda x, y: int(
-                round(self.predefinedPositions[x] - self.predefinedPositions[y])
-            )
-        )
+    def sort_positions_list(self):
+        self.positions_names_list = list(self.predefined_positions.keys())
+        #self.positions_names_list.sort()
```
rhfogh: Why not sort it? For a newly made list it cannot hurt

HardwareObjects/GenericDiffractometer.py
```diff
@@ -957,10 +958,10 @@ def move_motors(self, motor_positions, timeout=15):
             if type(motor) in (str, unicode):
                 motor_role = motor
                 motor = self.motor_hwobj_dict.get(motor_role)
-                del motor_positions[motor_role]
+                #del motor_positions[motor_role]
```
rhfogh: I think you need to be VERY careful when checking this one.
motor)positions is a mutable input value. We ahve no idea if code elsewhere is relying on the
motor_positions dictionary being mlodified in this call.
Of course this should not happen - the dictoinary should remain unmodified, and a modified copy
should be returned. But I would not be too sure what is actually going on.

marcus-oscarsson: I think you are absolutely right. So @IvarsKarpics have for the moment commented this line, which I think is good as well. There is a PR for the new AbstractDiffractometer #334 were this kind of discussion will be very important. We already mentioned on the meeting that we would like to use more mutable data structures so I think Motors should naturally evolve in that direction as well as other things.

rhfogh: @marcus-oscarsson I do not remember the discussion on mutable data structures. It will need a lot of thought. A mutable singleton, essentially an object, representing the current state of  motors would make some sense. But mutable input parameters to functions, or sharing mutble attributesof objects, means that it becomes extremely difficult to konw what you are modifying and what teh effecsts wil be if you modity that data structure.

marcus-oscarsson: Sorry @rhfogh I ofcourse meant immutable, I hope it makes more sense then.

The discussion around the *immutable* data structures should be in the meeting minutes of the November face to face meeting for those who wants to go back to read it :).

PR #349, opened 2019-01-30
Added pylint to travisci

IvarsKarpics:


IvarsKarpics:
Hi @marcus-oscarsson ,

I was trying to enable pylint for various directories, but all of them, as expected, failed. So, I disabled it for time being. I need we need a face to face meeting to do a cleaning according to the pylint result. For example, are you using the TacoDevice_MTSafe.py file. Pylint was complaining a lot about this file. Maybe we can create a directory called archive that will not be tested with pylint?

beteva:
Hi  @IvarsKarpics,  Archive director is good idea. All we have to agree on is what to go there.

marcus-oscarsson:
@IvarsKarpics, @beteva 

Yes there are probably a few files that we can store in an archive directory, and as @beteva says we just need to agree on which ones go there.

I think we need to come up with reasonable configuration for pylint in order to have real benefit of it. I think that @rhfogh is looking into that

Cheers,
Marcus

IvarsKarpics:
Good. Then maybe lets  talk about this during our next web meeting.

IvarsKarpics:
Hi @marcus-oscarsson . What is the status of this PR? I removed the pylint test, but fixed few bugs indicated by pylint.

marcus-oscarsson:
Hi, 

I wanted to wait a bit to  get some input from @rhfogh regarding the PyLint config.

Thanks,
Marcus

PR #353, opened 2019-02-07
Added more badges to the README

IvarsKarpics:


README.md
```diff
@@ -1,4 +1,6 @@
 [![Build Status](https://travis-ci.org/mxcube/HardwareRepository.svg?branch=master)](https://travis-ci.org/mxcube/HardwareRepository)
+[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/ambv/black)
```
marcus-oscarsson: Not sure about this one, I agree that its nice with something the says that we are using a code style. However we are not really applying black on a regular basis

IvarsKarpics: Hmm, we did apply autopep and black to all HardwareRep. during the developers meeting. Why not to inform people that we use black coding guidelines?

marcus-oscarsson: We did, but then we should also have something that ensures that we are following them or at-least agree that we run black on a regular basis

IvarsKarpics: Ok, I removed the badge. During the next face to face meeting we can discuss how we can follow the black guidelines.

Issue #335, opened 2019-01-04
Configuration files mxcube3 and mxcube2


marcus-oscarsson:
Unify the XML configuration files so that the same set of files can be used for both MXCuBE3 and MXCuBE2

IvarsKarpics:
To be done in two weeks

marcus-oscarsson:
This will become much easier when the abstract classes are in place and settled, we are in effect already doing this :)

marcus-oscarsson:
A start could be to create a folder for configuration files that already are or can be made common

PR #355, opened 2019-02-15
Added methods to test energy hwobj

IvarsKarpics:


test/pytest/test_hwr_qt.py
```diff
@@ -13,3 +14,22 @@ def test_hwr_qt(hwr_qt):
                  "collect", "parallel_processing", "xrf_spectrum", "detector")
     for role in role_list:
         assert hasattr(blsetup_hwobj, "%s_hwobj" % role)
+    """
+    energy_hwobj = hwr.getHardwareObject("energy-mockup")
```
marcus-oscarsson: Why is this specific to the qt version, because of the XML files ?

IvarsKarpics: Yes. If we join the config xmls (#335) then we would not need those two test methods.

marcus-oscarsson: I see, I think that should definitely be a priority if it prevents us from for instance sharing tests. We one could create a fixture which simply returns the correct object depending on the version used. 

I also think that this test should be moved to a separate routine that tests only the energy hardware object.

IvarsKarpics: Ok, I will have a look how we can join the config files.

IvarsKarpics: I split the test method in several methods and now there is a method that checks just the energy hwobj. In a similar method one should check all relevant hardware objects.

test/pytest/test_hwr_qt.py
```diff
@@ -1,15 +1,46 @@
...
-                 "session", "lims_client", "data_analysis", "energyscan",
-                 "collect", "parallel_processing", "xrf_spectrum", "detector")
-    for role in role_list:
-        assert hasattr(blsetup_hwobj, "%s_hwobj" % role)
+    global HWR
+    HWR = HardwareRepository.getHardwareRepository(hwr_qt)
+    HWR.connect()
+
+def test_energy():
+    energy_hwobj = HWR.getHardwareObject("energy-mockup")
```
marcus-oscarsson: its, perhaps better if the method takes the hardware object to test as an argument so we can use or in some other more modular way. So that we can reuse if for the web version ?

IvarsKarpics: Ok, I see. I created a bit different structure and created a file test_energy.py that will test energy hwobj. As we have two configuration I am passing a list of two energy hwobj. In the end when we have a single mockup config this will take a single energy hwobj.

marcus-oscarsson: Awesome :+1:

PR #356, opened 2019-02-18
Update of EMBL hardware objects

IvarsKarpics:
* Updated EMBL hardware objects

### DIFF
--- a/HardwareObjects/EMBL/EMBLCRL.py
+++ b/HardwareObjects/EMBL/EMBLCRL.py
@@ -82,7 +82,7 @@ class EMBLCRL(HardwareObject):
     def convert_value(self, value):
         """Converts int to list of bits"""
         if value:
-            if type(value) in (list, tuple):
+            if isinstance(value, (list, tuple)):
                 lens_combination = 0
                 for index in range(self.lens_count):
                     lens_combination = lens_combination + value[index] * pow(2, index)
--- a/HardwareObjects/EMBL/EMBLCRL.py
+++ b/HardwareObjects/EMBL/EMBLCRL.py
@@ -81,16 +81,17 @@ class EMBLCRL(HardwareObject):
 
     def convert_value(self, value):
         """Converts int to list of bits"""
-        if type(value) in (list, tuple):
-            lens_combination = 0
-            for index in range(self.lens_count):
-                lens_combination = lens_combination + value[index] * pow(2, index)
-        else:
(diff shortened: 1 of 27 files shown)


IvarsKarpics:
After some cleanup this PR is ready to be reviewed, merged. It also contains pylint check for EMBL hardware objects. At some point we should check also abstract, mockup and all other hardware objects with pylint.

HardwareObjects/EMBL/EMBLCRL.py
```diff
@@ -81,16 +81,17 @@ def init(self):
...
-            for index in range(self.lens_count):
-                lens_combination = lens_combination + value[index] * pow(2, index)
-        else:
-            lens_combination = [0, 0, 0, 0, 0, 0]
-            for index in range(self.lens_count):
-                lens_combination[index] = (value & pow(2, index)) / pow(2, index)
-
-        return lens_combination
+        if value:
+            if type(value) in (list, tuple):
```
rhfogh: Better:
if isinstance(value, (list, tuple)):

HardwareObjects/EMBL/EMBLPPUControl.py
```diff
@@ -69,13 +69,14 @@ def all_restart_changed(self, status):
         self.restart_running = status
 
     def file_info_changed(self, value):
+        value = (value)
```
rhfogh: Something rotten here:
1) 'value = (value)' is a no-op
2) if you meant 'value = (value,)' the length is always 1
3) if you are exrreacting a single value from a list or tuple, the list of tuple should be named 'values' (plural)

rhfogh:
Has it been decided that all imports must be from the top level i.e. that HardwareRepository should be removed from the PythonPath)? I highly approve, I just want to know the exact state of play

IvarsKarpics:
Hi @rhfogh , thank you for the useful suggestions! I will update my code.

marcus-oscarsson:
@rhfogh Nothing have been decided yet but I think we basically all agree that its a good idea. It will most likely have some practical implications on how things are referenced from XML files and how that's going to be handled needs to be discussed.

I see no real reason for not using absolute imports, without appending additional paths to the python path, from python code we write.

IvarsKarpics:
Pylint via travis indicated a typo on my code. Nice! we should definitely run pylint on all HardwareRepository.

marcus-oscarsson:
Yes I agree, I was hoping that we eventually could run it without the -E option

PR #357, opened 2019-02-19
Adaptations needed after API changes

marcus-oscarsson:
Dear all,

Some adaptations needed after the API changes we have introduced recently, along with some code cleanup.

For Qt developers pay attention to:

                    snapshot = self.shape_history.get_scene_snapshot()
to

                    snapshot = self.shape_history.get_snapshot()

Marcus

### DIFF
--- a/HardwareObjects/GenericDiffractometer.py
+++ b/HardwareObjects/GenericDiffractometer.py
@@ -196,7 +196,7 @@ class GenericDiffractometer(HardwareObject):
         self.image_height = None
 
         self.current_state = None
-        self.current_phase = None
+        self.current_phase = GenericDiffractometer.PHASE_CENTRING
         self.transfer_mode = None
         self.current_centring_procedure = None
         self.current_centring_method = None
--- a/HardwareObjects/GenericDiffractometer.py
+++ b/HardwareObjects/GenericDiffractometer.py
@@ -31,11 +31,10 @@ import numpy
 import queue_model_objects
 
 try:
-   unicode
+    unicode
 except:
-   # A quick fix for python3
-   unicode = str
- 
+    # A quick fix for python3
+    unicode = str
 
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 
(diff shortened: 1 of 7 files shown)


HardwareObjects/GenericDiffractometer.py
```diff
@@ -196,7 +195,7 @@ def __init__(self, name):
         self.image_height = None
 
         self.current_state = None
-        self.current_phase = None
+        self.current_phase = GenericDiffractometer.PHASE_CENTRING
```
IvarsKarpics: I would keep value None as not initialized value (all variables in __init__ should be better None) and then initiate it in the init method of current diffractometer (Mockup, etc).

HardwareObjects/queue_entry.py
```diff
@@ -1091,7 +1095,7 @@ def collect_dc(self, dc, list_item):
                 else:
                     pos_dict = self.diffractometer_hwobj.getPositions()
                     cpos = queue_model_objects.CentredPosition(pos_dict)
-                    snapshot = self.shape_history.get_scene_snapshot()
+                    snapshot = self.shape_history.get_snapshot()
```
IvarsKarpics: is this a snapshot with or without overlays? How we can distinguish them? I guess if you call get_snapshot of shape_history you will get a snapshot with overlays and get_snapshot from camera will return a raw image.

HardwareObjects/GenericDiffractometer.py
marcus-oscarsson: Yes that's probably nicer, I made a fix for it !

HardwareObjects/queue_entry.py
marcus-oscarsson: Our get_snapshot method on ShapeHistory is for the moment empty, but we do have a get_snapshot that takes a overlays keyword argument.

That's another reason for moving the entire snapshot functionality along with a bunch of other things to a more beamline wide entity.

marcus-oscarsson:
@IvarsKarpics So I can go ahead and merge this then ?

IvarsKarpics:
Sorry, I forgot to merge it. Now its merged

PR #358, opened 2019-03-26
Reworked AbstractShutter

marcus-oscarsson:
Hi

Moved AbstractShutter from AbstractBase and created AbstractNState that is used to define an AbstractShutter. This could eventually be reused for AbstractActuators as well.

Cheers,
Marcus


HardwareObjects/abstract/AbstractNState.py
```diff
@@ -0,0 +1,90 @@
...
+    @abc.abstractmethod
+    def state(self):
+        """
+        Returns:
+           str: The current state name 
+        """
+        return
+
+    @abc.abstractmethod
+    def is_ok(self):
```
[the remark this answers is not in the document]
marcus-oscarsson: Sure, I actually used what was there before but we can call it is_ready. Actually is_valid would be more accurate since that's what iis checking. It might not be ready but valid :)

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class AbstractShutter(HardwareObject, AbstractNState):
+    STATE = ShutterState
+
+    def __init__(self, name):
+        HardwareObject.__init__(self, name)
+
+    @abc.abstractmethod
+    def open(self):
```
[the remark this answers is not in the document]
marcus-oscarsson: I was thinking that as well, but then I noticed that it was never used for shutters so I though it was unnecessary. But Ill add them for completeness in case someone wants to do it open or close synchronously ....

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -0,0 +1,90 @@
...
+    @abc.abstractmethod
+    def is_ok(self):
+        """ Checks if the shutter is in one of its predefined states """
+        return
+
+
+@unique
+class ShutterState(Enum):
+    """
+    Defines the valid Shutter states
```
beteva: Using Enum is a very good idea. Could we make it IntEnum instead of Enum. Thus we can also combine numerical values.

HardwareObjects/abstract/AbstractNState.py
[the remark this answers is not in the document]
beteva: Agree with @IvarsKarpics, we should start adding timeout and wait args whenever we have a potential blocking action or as @marcus-oscarsson pointed out - a potential synchronous action on several objects.

HardwareObjects/abstract/AbstractNState.py
marcus-oscarsson: Yes, I guess that wont be a problem

IvarsKarpics:
Hi @marcus-oscarsson . Could you please have a look at and potentially fix the issues indicated by the codacy?

marcus-oscarsson:
Hi, tried to fix most of the Codacy linting errors. I skipped the test directory which we probably should exclude from checks, because of asserts and other linter unfriendly statements

Marcus

IvarsKarpics:
Hi @marcus-oscarsson . Thank you for the cleanup. I added test directory to the ignored list. From now it will not be checked.
From my side I am fine with the PR.

marcus-oscarsson:
So I think you can merge it then since the other things have been fixed...

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class AbstractShutter(HardwareObject, AbstractNState):
+    """
+    Defines the common interface for shutters
+    """
+    STATE = ShutterState
+
+    def __init__(self, name):
+        HardwareObject.__init__(self, name)
```
IvarsKarpics: Should the AbstractNState.__init__(self, name) also needs to be called?

HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -1,47 +1,64 @@
...
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+import time
+import random
+from abstract.AbstractNState import AbstractShutter
```
IvarsKarpics: If we use abstract import, then this should be HardwareRepository.HardwareObjects.abstract.AbstractNState

HardwareObjects/abstract/AbstractNState.py
marcus-oscarsson: I dont think that would be necessary since AbstractNState does not inherit HardwareObject ...

HardwareObjects/mockup/ShutterMockup.py
marcus-oscarsson: You mean if we use absolute imports ?

marcus-oscarsson: Then surely :)

HardwareObjects/abstract/AbstractNState.py
IvarsKarpics: okay

HardwareObjects/mockup/ShutterMockup.py
rhfogh: You get senaky bugs if you import the same file via two different paths (essentially they behave as two different modules, that happen to have the same name). So to avoid that we need to luse the same path always, which has to be the long one

marcus-oscarsson: @rhfogh I agree, I had the issue myself, its been updated a little while ago

marcus-oscarsson:
@IvarsKarpics you can merge @beteva is fine with the PR

marcus-oscarsson:
Thanks :+1: :)

PR #359, opened 2019-03-27
Removed QtImport file and cleanup up Qt* classes

IvarsKarpics:
* Removed QtImport file from the HardwarRepository
* Other files obtain Qt classes via gui.utils.QtImport.

At some point all classes requiring Qt could go to the gui part.


HardwareObjects/QtGraphicsLib.py
```diff
@@ -47,9 +47,9 @@
 import logging
 from datetime import datetime
 
-import QtImport
+from gui.utils import QtImport
 
-import queue_model_objects
+from . import queue_model_objects
```
IvarsKarpics: I will fix this to import from HardwareRepository

IvarsKarpics:
Thank you @rhfogh !

PR #362, opened 2019-03-28
GPhL and cleanup multi-function PR

rhfogh:
This PR mixes GPhL changes (ported from 2.3.0) and cleanup (which was in part needed to fix bugs).

There are the following parts:

- Upgrade of GPhL-specific files and classes (including CollectEmulator  and queue models and entries)
- Changing all imports to start at HardwareRepository or gui, where relevant
- Fixed spelling of 'collaboration'
- Run black on gphl and some other files
- Added rough draft of gphl config files to P14
- Added 'detdistance' to collection parameters as parallel alternative to 'resolution'
- minor mockup improvements
- Refactored Session folder handling
- Added AbstractCollect debug output
- Minor fixes and style improvements, including removal of unneeded imports.

Apologies for the large size.


HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -259,8 +259,9 @@ def do_collect(self, owner):
                 )
                 self.set_energy(self.current_dc_parameters["energy"])
 
-            if "resolution" in self.current_dc_parameters:
-                resolution = self.current_dc_parameters["resolution"]["upper"]
+            dd = self.current_dc_parameters.get("resolution")
```
IvarsKarpics: I know that dd is a nameless variable, but pylint will complain. Could it be a `collect_resolution`?

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -277,6 +278,16 @@ def do_collect(self, owner):
             # In order to call the hook with original parameters
             # before update_data_collection_in_lims changes them
             # TODO check why this happens
+
+            # Motor positions debug output
+            ll = sorted((tt[0] if isinstance(tt[0], (str, unicode)) else tt[0].name(), tt[1])
+                        for tt in self.current_dc_parameters['motors'].items())
+            logging.getLogger('HWR').debug("MOTORS ACQ target: "
```
IvarsKarpics: Sorry, but I do not think we need these debug messages in the log file. Maybe it could be a new logger, that by default is not printed?

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -277,6 +278,16 @@ def do_collect(self, owner):
...
+                                           + ', '.join('%s=%s' % (tt) for tt in ll))
+
+            ll = sorted(self.diffractometer_hwobj.get_positions().items())
+            logging.getLogger('HWR').debug("MOTORS ACQ result: "
```
IvarsKarpics: same here

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -770,6 +781,18 @@ def move_to_centered_position(self):
         """
         Descript. :
         """
+        dd = self.diffractometer_hwobj.get_positions()
+        logging.getLogger('HWR').debug(
```
IvarsKarpics: and here

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -770,6 +781,18 @@ def move_to_centered_position(self):
...
+            'MOTORS-premove ' +
+            ', '.join('%s:%s' % tt for tt in sorted(dd.items()))
+
+        )
+        dd = self.current_dc_parameters['motors']
+        logging.getLogger('HWR').debug(
```
IvarsKarpics: and also here

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -779,6 +802,11 @@ def move_to_centered_position(self):
                     positions_str += " %s=%f" % (motor.getMotorMnemonic(), position)
         self.current_dc_parameters["actualCenteringPosition"] = positions_str
         self.move_motors(self.current_dc_parameters["motors"])
+        dd = self.diffractometer_hwobj.get_positions()
```
IvarsKarpics: finally also here

HardwareObjects/mockup/CollectEmulator.py
```diff
@@ -244,23 +242,23 @@ def data_collection_hook(self):
...
-
-        logging.getLogger("HWR").debug(
-            "Emulator:  actual position "
-            + ", ".join(
-                "%s=%s" % tt
-                for tt in sorted(self.diffractometer_hwobj.get_positions().items())
-                if tt[1] is not None
-            )
-        )
+        # logging.getLogger("HWR").debug(
```
IvarsKarpics: Maybe remove this code?

HardwareObjects/queue_model_objects.py
```diff
@@ -1506,6 +1502,8 @@ def __init__(self):
         self.energy = float()
         self.centred_position = CentredPosition()
         self.resolution = float()
+        # detector_distance ('detdistance') used if resolution is 0 or None
+        self.detdistance = float()
```
IvarsKarpics: `self.detector_distance` please

HardwareObjects/queue_model_objects.py
```diff
@@ -1671,6 +1670,8 @@ def as_dict(self):
                 ],
             )
         )
+        #
```
IvarsKarpics: ?

HardwareObjects/abstract/AbstractCollect.py
rhfogh: I have noticed this one, and using this kind of variable name is one of my general practices. If people do not want to relax the rule in pylint, I was thinking of changing the variable name to 'ddd', or maybe 'adict'. The meaning is 'A temporary value containing a dictionary' Similarly 'll' for lists, 'tt' for tuples, 'ss' for strings, 'ii' for ints, and 'xx' for floats and others.

The argument is that a name like 'collect_resolution' suggests something with a clear meaning. If you want to modify it, you would need to search the entire function to see if 'collect_resolution' is used again lower down, and if so whether it is supposed to be the same value, or just another temporary value with similar contents.  Of course in this particular case it is not even a float (as 'collect_resolution'' would suggest), but rather a dictionary of the form '{"upper":some_float}'. I maintain that it makes more sense to mark variables clearly as temporary and of local scope.

For now, can I propose that we accept these variables as they are, and agree how this kind of variables can or should be named? If (when?) I am voted down I will then make the change across my code, not merely in the places that are pointed out.

HardwareObjects/abstract/AbstractCollect.py
rhfogh: See above

HardwareObjects/abstract/AbstractCollect.py
rhfogh: I guess this kind of thing is really debug output, and OK, I shall take it out. Do you mind if I just comment out some of them for now, since the problem of which motor values are actually set when is, AFAIAC, a current problem acrioss multiple beamlines?

HardwareObjects/abstract/AbstractCollect.py
rhfogh: See above

HardwareObjects/abstract/AbstractCollect.py
rhfogh: See above

HardwareObjects/abstract/AbstractCollect.py
rhfogh: See above

HardwareObjects/mockup/CollectEmulator.py
rhfogh: OK

HardwareObjects/queue_model_objects.py
rhfogh: OK

HardwareObjects/queue_model_objects.py
rhfogh: OK

HardwareObjects/abstract/AbstractCollect.py
IvarsKarpics: I see your point. It makes sense to use temporary variables like dd. Then I would suggest to keep it as it is and during the next meeting we can come up with a list of names. And maybe names that are ok with pylint. For example not dd, but temp_dict

HardwareObjects/abstract/AbstractCollect.py
IvarsKarpics: Well, I am fine with the log message. For testing and debugging they are very useful, but in the production mode they should not appear (I imagine scrolling long log files on Sunday afternoon via ssh). Maybe we can add command line argument --test, which by default is false, if it is true then "TEST" log will be printed.

rhfogh: I have commented it out for now. We can discuss whether to add a test flag - or simply set the loggers to skip debug level messages.

IvarsKarpics: Ok

HardwareObjects/abstract/AbstractCollect.py
marcus-oscarsson: I agree, its often useful to have short names for temporary variables for clarity and convince. There are also some instance when one simply wants to use a one or two letter variable name. So I would vote for relaxing the pylint rule for variable lengths (actually I think it's the snake case rule ...)

HardwareObjects/abstract/AbstractCollect.py
marcus-oscarsson: Would it no just be enough to use setLevel on the logger, or create a separate logger with even lower importance called test or something, or did I misunderstand something ?

Command/Epics.py
```diff
@@ -10,7 +10,7 @@
 from Poller import *
 
 # from .CommandContainer import CommandObject, ChannelObject
-from CommandContainer import CommandObject, ChannelObject
+from HardwareRepository.CommandContainer import CommandObject, ChannelObject
```
marcus-oscarsson: Thanks, great :)

HardwareObjects/ESRF/BM14MultiCollect.py
```diff
@@ -4,8 +4,8 @@
 import logging
 
 from HardwareRepository.TaskUtils import task
-from ESRFMultiCollect import CcdDetector, ESRFMultiCollect, TunableEnergy
-from detectors.TacoMarDetector import Mar225
+from .ESRFMultiCollect import CcdDetector, ESRFMultiCollect, TunableEnergy
+from HardwareRepository.HardwareObjects.TacoMarDetector import Mar225
```
marcus-oscarsson: Thanks again :)

HardwareObjects/ESRF/ID30BMultiCollect.py
```diff
@@ -4,8 +4,8 @@
 import os
 
 from HardwareRepository.TaskUtils import task
-from ESRF.ESRFMultiCollect import ESRFMultiCollect, PixelDetector, TunableEnergy
-from detectors.LimaPilatusDetector import Pilatus
+from .ESRFMultiCollect import ESRFMultiCollect, PixelDetector, TunableEnergy
+from HardwareRepository.HardwareObjects.LimaPilatusDetector import Pilatus
```
marcus-oscarsson: Did you use a tool for this or did you do it by hand ?

HardwareObjects/GphlMessages.py
```diff
@@ -974,6 +998,24 @@ def allowedWidths(self):
...
+        """Get sweeps in acquisition order.
+
+        WARNING we do not have the necessary information.
+        So we sort in increasing order by angles, in alphabetical name order,
+        (in pracite, 'kappa', 'kappa_phi', 'phi')
+        HORRIBLE HACK, but as it happens this will give
+        at least the first one correct mostly
+        and anyway is the best approximation we have
+
+        TODO get this right, once the workflow allows"""
```
marcus-oscarsson: Whats missing to make it right ?

marcus-oscarsson:
Thanks for all this work

HardwareObjects/GphlMessages.py
rhfogh: The problem is in the Java workflow code from GPhL. At this point in the code the acquisition order of the sweeps (which I need) is not made available to the MXCuBE code. It is on Peter Kellers TODO list, but changes in the Java has its own times and there are otehr priorities as well there.

HardwareObjects/ESRF/ID30BMultiCollect.py
rhfogh: By hand, from inside Pycharm. I would not dare do this kind of thing without at least manual checking. It was some combination of Pycharms error highlighting, and making global searches for 'import .', 'import ESRF', or 'import detectors', once I knew what to look for.

HardwareObjects/abstract/AbstractCollect.py
beteva: @marcus-oscarsson is right, it is not the length but the snake case. d_d would have worked. Agree on setting up a rule, just to make pylint happy.

HardwareObjects/abstract/AbstractCollect.py
beteva: I would not change the logger level, as the logger might also by used by some side procedures, like in BLISS for us. I thought we had different loggers (HWR, user_level, ...), which probably needs some clarification/cleanup. A topic for one of our internet meetings?

HardwareObjects/abstract/AbstractCollect.py
rhfogh: According to the Codacy issues, the problem is:
"Variable name "gg" doesn't conform to u'[a-z_][a-z0-9_]{2,30}$' pattern"
As I read it, the problem is that the length must be at least three - which is why 'd_d' would work.

I would rather like to keep my current style (and 'temp_dict', while descriptive, is too long for my taste). At worst I could add a numeral, and change over to 'dd0', 'dd1', etc. You sometimes need several different variable after all.

Let us see what we can come up with?

HardwareObjects/abstract/AbstractCollect.py
marcus-oscarsson: Changing the log level of the HWR logger only effects that logger and not the ones used in Bliss

PR #365, opened 2019-04-01
Added tests for detector and energy

IvarsKarpics:


test/pytest/test_detector.py
```diff
@@ -0,0 +1,20 @@
+def test_detector_atributes(detector_list):
+    for detector in detector_list:
+        assert not detector is None, "Detector hardware objects is None (not initialized)"
+        current_distance = detector.get_distance()
+        distance_limits = detector.get_distance_limits()
+        exp_time_limits = detector.get_exposure_time_limits()
+        has_shutterless = detector.has_shutterless()
+
+        assert isinstance(current_distance, (int, float)), "Distance value has to be int or float. Now %s" % str(current_distance)
```
marcus-oscarsson: Hm, would it not be better to say that it should be float ... always ...

test/pytest/test_detector.py
```diff
@@ -0,0 +1,20 @@
...
+        assert isinstance(distance_limits, (list, tuple)), "Distance limits has to be defined as a tuple or list"
```
marcus-oscarsson: Same here actually, even if its of maybe less importance

test/pytest/test_detector.py
```diff
@@ -0,0 +1,20 @@
...
+        assert not None in distance_limits, "One or several distance limits is None"
+        assert distance_limits[0] < distance_limits[1], "First value of distance limits has to be the low limit"
+
+        assert hasattr(detector, "set_distance")
```
marcus-oscarsson: You don't really have to do this of the method is marked as Abstract in the definition which assume should be the case
since there is a test case like this ?

test/pytest/test_energy.py
```diff
@@ -16,3 +16,9 @@ def test_energy(energy_list):
         assert hasattr(energy, "can_move_energy")
```
marcus-oscarsson: Same as above

test/pytest/test_detector.py
IvarsKarpics: In our case distance is in mm (int) and the test would fail. Well, as this one tests mockup it could be also just float

test/pytest/test_detector.py
IvarsKarpics: Ok, it makes sense to remove them because later there is actual set_distance test method

IvarsKarpics:
Danke!

marcus-oscarsson:
Bitte !

PR #370, opened 2019-04-15
Pylint-suggested fixes in GPhL code - and a few obvious errors

rhfogh:


HardwareObjects/mockup/CollectEmulator.py
```diff
@@ -26,12 +26,12 @@
 import subprocess
 import logging
 import re
+from collections import OrderedDict
 import f90nml
+import api
```
marcus-oscarsson: So is this the Beamline object we discussed or the API ?

rhfogh: Dear MArcus,

No, this is Ivars api object. It should be changed over to the one we discussed, assuming that we do agree on that, of course.

marcus-oscarsson: Okey, so you will do that in another PR then I imagine ?

rhfogh: Yes

HardwareObjects/BlissRontecMCA.py
```diff
@@ -1,5 +1,6 @@
 from HardwareRepository.BaseHardwareObjects import HardwareObject
-from AbstractMCA import *
+from abstract.AbstractMCA import AbstractMCA
```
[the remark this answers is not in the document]
rhfogh: Indeed. Fixed now.

PR #372, opened 2019-04-16
Add missing decode method (issue 371)

jordiandreu:
Add the decode method necessary to convert
the data object to be uploaded to ISPyB.

This method is used in the create_session method.

### DIFF
--- a/HardwareObjects/ISPyBClient.py
+++ b/HardwareObjects/ISPyBClient.py
@@ -109,16 +109,9 @@ def utf_encode(res_d):
         if isinstance(value, dict):
             utf_encode(value)
 
-        if type(res_d[key]) in (int, float, bool, str):
-            # Ignore primitive types
-            pass
-        elif isinstance(res_d[key], suds.sax.text.Text):
-            # utf-8 encode Text data
-            try:
-                res_d[key] = res_d[key].encode("utf8", "ignore")
-            except BaseException:
-                pass
-        else:
+        try:
+            res_d[key] = value.encode("utf8", "ignore")
+        except BaseException:
             # If not primitive or Text data, complext type, try to convert to
             # dict or str if the first fails
             try:
--- a/HardwareObjects/ISPyBClient.py
+++ b/HardwareObjects/ISPyBClient.py
@@ -105,9 +105,9 @@ def in_greenlet(fun):
 
 
 def utf_encode(res_d):
-    for key in res_d.iterkeys():
-        if isinstance(res_d[key], dict):
-            utf_encode(res_d)
+    for key, value in res_d.items():
+        if isinstance(value, dict):
+            utf_encode(value)
 
         if type(res_d[key]) in (int, float, bool, str):
             # Ignore primitive types
@@ -122,10 +122,10 @@ def utf_encode(res_d):
             # If not primitive or Text data, complext type, try to convert to
             # dict or str if the first fails
             try:
-                res_d[key] = utf_encode(asdict(res_d[key]))
+                res_d[key] = utf_encode(asdict(value))
             except BaseException:
                 try:
-                    res_d[key] = str(res_d[key])
+                    res_d[key] = str(value)
                 except BaseException:
                     res_d[key] = "ISPyBClient: could not encode value"
 
@@ -135,10 +135,9 @@ def utf_encode(res_d):
 def utf_decode(res_d):
     for key, value in res_d.items():
         if isinstance(value, dict):
-            utf_decode(res_d)
-
+            utf_decode(value)
         try:
-            res_d[key] = res_d[key].decode('utf8', 'ignore')
+            res_d[key] = value.decode('utf8', 'ignore')
         except BaseException:
             pass
 
--- a/HardwareObjects/ISPyBClient.py
+++ b/HardwareObjects/ISPyBClient.py
@@ -133,8 +133,8 @@ def utf_encode(res_d):
 
 
 def utf_decode(res_d):
-    for key in res_d.iterkeys():
-        if isinstance(res_d[key], dict):
+    for key, value in res_d.items():
+        if isinstance(value, dict):
             utf_decode(res_d)
 
         try:
--- a/HardwareObjects/ISPyBClient.py
+++ b/HardwareObjects/ISPyBClient.py
@@ -132,6 +132,19 @@ def utf_encode(res_d):
     return res_d
 
 
+def utf_decode(res_d):
+    for key in res_d.iterkeys():
+        if isinstance(res_d[key], dict):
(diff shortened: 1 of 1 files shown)


HardwareObjects/ISPyBClient.py
```diff
@@ -132,6 +132,19 @@ def utf_encode(res_d):
     return res_d
 
 
+def utf_decode(res_d):
+    for key in res_d.iterkeys():
```
IvarsKarpics: there is no iterkeys in python3 (we should enable pylint for all HardwareRepository asap). Please modify code to be compatible with python3. For example:
for example:
```
for key, item in res_d.items():
     if isinstance(item, dict):
        ....
```

jordiandreu: Sorry, I just copied from my current old python2 branch...fixed.

HardwareObjects/ISPyBClient.py
```diff
@@ -132,6 +132,19 @@ def utf_encode(res_d):
     return res_d
 
 
+def utf_decode(res_d):
+    for key, value in res_d.items():
+        if isinstance(value, dict):
+            utf_decode(res_d)
```
rhfogh: Surely taht should be utf_decode(value) ?

HardwareObjects/ISPyBClient.py
```diff
@@ -132,6 +132,19 @@ def utf_encode(res_d):
...
+
+        try:
+            res_d[key] = res_d[key].decode('utf8', 'ignore')
```
rhfogh: I am puzzled:
First, why is only the value converted, not the key?
Second, what about all the non-string values?
For the second point, would it not be better to do
if isinstance(value,str): instead of the try:? Or is that too complicated when supporting both PYthon 2 and 3?

HardwareObjects/ISPyBClient.py
jordiandreu: Sure, you are right. I will also apply this change to the utf_encode method, which also contains the same bug.

HardwareObjects/ISPyBClient.py
jordiandreu: > First, why is only the value converted, not the key?

Usually, the keys are assigned by the developer and they are usually alfa-numeric characters (with few special characters) that need no conversion in western codecs.

> Second, what about all the non-string values?
For the second point, would it not be better to do
if isinstance(value,str): instead of the try:? Or is that too complicated when supporting both Python 2 and 3?

Using the try/except you only convert objects which have the encode/decode method i.e. you do not have to have a list of types supported. For instance, the value could be a _str_ or _suds.sax.text.Tex_t type, both having the encode method.

> Third, what is being converted which way - I am always confused by the en/decoding.

I think it in the following way...(maybe wrong?)
Given a byte string (whatever the encoding is) the method _encode("utf8")_ will try to encode this content to utf-8 string object. Later on, you can decode the utf-8 string again to the original encoding by using the _decode("utf8")_.

Of course, it is possible that some character represented in the byte string has no representation in the coded requested and then the encode method fails. For instance:

I think this agrees with Duncan's answer in the post entry you sent.


I post also a test script to clarify this issue. I will review the code and push some new commit to the pull request and keep with the discussion.

```
#!/usr/bin/python
# -*- coding: iso8859-15 -*-

def utf_encode(res_d):
    for key, value in res_d.items():
        if isinstance(res_d[key], dict):
            utf_encode(value)

        if True:#isinstance(res_d[key], suds.sax.text.Text):
            try:
                res_d[key] = res_d[key].encode('utf8', 'ignore')
            except:
                pass

    return res_d


def utf_decode(res_d):
    for key, value in res_d.items():
        if isinstance(value, dict):
            utf_decode(value)

        try:
            res_d[key] = res_d[key].decode('utf8', 'ignore')
            print("decoding {} to {}".format(value, res_d[key]))
        except BaseException as e:
            print(e)

    return res_d


if __name__ == "__main__":
    d = {'a':1, 'b':True, 'c':1.01, 'd':'Bárbara Luísa', 'e': {'aa':1, 'bb':False, 'c':'Ramón'}, 'f':'John Smith'}
    print('original: {0}'.format(d))
    d_encoded = utf_encode(d)
    print('encoded: {0}'.format(d_encoded))
    print('decoded (original again: {0}'.format(utf_decode(d_encoded)))
    utf_decode(d_encoded)
```

rhfogh: First:
You are right.

Second:
You are right. 

Third:
Not so sure. 

It is tricky that in python 2 both string and unicode have both encode and decode methods.
But if you look in the python 3 documentation, you find that bytes and bytearray have a decode method but *not* an encode method. Whereas str (which in Python 3 is the same as unicode) has an encode but *not* a decode method. That would mean that you have (as it were)

def encode(value:str) -> bytes:

def decode(value:bytes) -> str:

Which I think is the opposite of what you are saying. The functions may work as they stand, but this is so confusing that I think we have to be quite sure which way it goes. 

What do you think?

jordiandreu: Yes, you are right. In my third point I was telling just the other way around it really is.

I post here a talk quite interesting about the topic I red some time ago:

https://nedbatchelder.com/text/unipain.html

and a short summary (also to clarify myself ) as far as I understood:

The key is that python2 do some automatic decoding from bytes (str string) to code points (unicode string) using the default codec on the system ('ascii'). This fails when non ascii characters need to be converted to unicode string with the default codec (which is ascii).

String literals are stored in bytes (str) by default and converted to unicode automatically (if possible) when needed:

Then "a" == u"a" is True, cause it is automatically converted to unicode using the ascii codec when being compared, which is equivalent to the explicit expression b"a" == u"a".
However "á" == u"á" is False cause no automatic conversion (using the ascii codec) can be performed since it will raise an exception. However, you can explicitly decode to unicode using the utf-8 codec:
"á".decode('utf-8') == u"á" is True

In python3 things are much clear cause no automatic conversion from bytes to unicode is done since string literals are stored as unicode [code points](https://en.wikipedia.org/wiki/Code_point) by default (and unicode type has been replace by str). If you want to use bytes, then you need to do it explicitly.

python2:
type("a") = type(b"a") = <type 'str'> and type(u"a") = <type 'unicode'>

python3:
type("a") = type(u"a") = <type 'str'> and type(b"a") = <type 'bytes'>

Then you can think the encode/decode methods for both python2 and python3 as:

encode(code point) --> bytes
decode(bytes) --> code points

rhfogh: AOK now. I read the presentation - and the one that one pointed to - and I understand it slightly better.
I hate how sanctimoniuos those guys sond, though

Issue #363, opened 2019-03-28
Bug in new AbstractShuttter / ShutterMockup


rhfogh:
The new ShutterMockup causes an error on program start (see below) - which means I cannot run the mock version.

I also have some style objections (apologies for not looking before the PR was merged:
- It is not clear from parameter names and docs when you are dealing with a state object, its name, or its value
- If there is a set_state there should be a get_state() to match
- IMHO function names should be verbs. It is not clear from the name that state() is a function or what it does.
- DuoStateBrick seems to be using a number of functions that are not in the ShutterMockup:
openShutter, closeShutter, moveToPosition, getState, getShutterState, ...

Error follows:

('@~@~', <DuoStateBrick.WrapperHO object at 0x7fc3d34fb770>, <ShutterMockup.ShutterMockup object at 0x7fc3d439dc10>)
2019-03-28 17:39:36,004 |root |ERROR  | Error while setting property mnemonic for shutter_brick (details in log file).
Traceback (most recent call last):
  File "/MXCuBE/mxcube/gui/BaseComponents.py", line 1022, in _property_changed
    self.property_changed(property_name, old_value, new_value)
  File "/MXCuBE/mxcube/gui/bricks/DuoStateBrick.py", line 243, in property_changed
    self.wrapper_hwobj.getState()
  File "/MXCuBE/mxcube/gui/bricks/DuoStateBrick.py", line 481, in getStateShutter
    state = self.dev.getShutterState()
  File "/MXCuBE/mxcube/HardwareRepository/BaseHardwareObjects.py", line 327, in __getattr__
    raise AttributeError(attr)
AttributeError: getShutterState

PR #373, opened 2019-04-16
Enabled pylint check for mockup objects

IvarsKarpics:
* Added missing imports.
* Fixed import to use absolute imports.
* Removed duplicated methods (for example in the ISPyBClientMockup)
* Fixed PlateManipulatorMockup
* Enable pylint check for mockup directory
* Rewrote MachineInfoMockup to use OrderedDict
* Solved #363 

I could not sort out Oxford700Mockup.py and LimaDetectorMockup.py (unable to import PyTango). So for time beeing I disabled pylint check for these two files.

### DIFF
--- a/HardwareObjects/mockup/MachineInfoMockup.py
+++ b/HardwareObjects/mockup/MachineInfoMockup.py
@@ -19,8 +19,10 @@
 
 import os
 import time
-from gevent import spawn
 from random import uniform
+from collections import OrderedDict
+
+from gevent import spawn
 
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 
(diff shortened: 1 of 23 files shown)


HardwareObjects/mockup/MachineInfoMockup.py
```diff
@@ -41,50 +43,50 @@ def __init__(self, name):
...
-        temp_dict["title"] = "Flux"
-        self.values_list.append(temp_dict)
-
-        temp_dict = {}
-        temp_dict["value"] = "-"
-        temp_dict["in_range"] = True
-        temp_dict["title"] = "Disk space"
-        temp_dict["align"] = "left"
-        self.values_list.append(temp_dict)
+        self.values_ordered_dict = OrderedDict()
```
rhfogh: This is fine - not sure why it needs to be ordered, though. Do we depend on looping through these records in order somewhere? Anyway, just fine with doing it this way.

HardwareObjects/mockup/MachineInfoMockup.py
```diff
@@ -176,7 +179,7 @@ def sizeof_fmt(self, num):
     def sizeof_num(self, num):
         """Returns disk space formated in exp value"""
 
-        for x in ["m", unichr(181), "n"]:
+        for x in ["m", chr(181), "n"]:
```
rhfogh: Will this work in Python2? Trying (in Python 2.7) I get:
>>> chr(181)
'\xb5'
>>> print(chr(181))
�
>>> print(unichr(181))
µ

HardwareObjects/mockup/MachineInfoMockup.py
IvarsKarpics: We would like to display the parameters in the gui in the ordered way. On the top machine current, status, etc

HardwareObjects/mockup/MachineInfoMockup.py
IvarsKarpics: If you do that in terminal, yes you will get that. But in the Qt widget it is fine.

rhfogh: Cool

HardwareObjects/mockup/MachineInfoMockup.py
rhfogh: Makes sense

IvarsKarpics:
Ok, after merging the PR #370 I will update this PR and most likely resolve conflicts.

IvarsKarpics:
Hi @rhfogh . I updated pr with the changes introduced with PR #370 . Now it is ready to be merged

IvarsKarpics:
Thanks!

PR #375, opened 2019-04-24
Added Python 2/3-compatible string type checking

rhfogh:
Also some minor fixes, and fixes to global phasing code


IvarsKarpics:
Well, I thought we agreed that the one who makes PR is not allowed to merge it. I accepted it, but it does not mean it can be merged (maybe someone else has an opinion). Generally I would merge it after few days if nobody raise a voice.

rhfogh:
Sure, we can do it that way too.

PR #376, opened 2019-04-25
Bugfix to handle compression tag correctly

IvarsKarpics:
Added a static method to set compression state for diffraction plan. A corresponding PR to mxcube will be made after this PR has been accepted.

### DIFF
--- a/HardwareObjects/queue_model_objects.py
+++ b/HardwareObjects/queue_model_objects.py
@@ -756,10 +756,14 @@ class Characterisation(TaskNode):
         self.characterisation_software = None
         self.wait_result = None
         self.run_diffraction_plan = None
-        self.diff_plan_compression = True
 
         self.auto_add_diff_plan = True
         self.diffraction_plan = []
+
+    @staticmethod
+    def set_char_compression(state):
+        print("setttt ", state)
+        Characterisation.diff_plan_compression = state
 
     def get_name(self):
         return "%s_%i" % (self._name, self._number)
--- a/HardwareObjects/queue_model_objects.py
+++ b/HardwareObjects/queue_model_objects.py
@@ -762,7 +762,6 @@ class Characterisation(TaskNode):
 
     @staticmethod
     def set_char_compression(state):
-        print("setttt ", state)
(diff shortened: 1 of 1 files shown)


HardwareObjects/queue_model_objects.py
```diff
@@ -756,11 +756,14 @@ def __init__(
         self.characterisation_software = None
         self.wait_result = None
         self.run_diffraction_plan = None
-        self.diff_plan_compression = True
 
         self.auto_add_diff_plan = True
         self.diffraction_plan = []
 
+    @staticmethod
```
rhfogh: Should Characterisation.diff_plan_compression not be initialilsed to a default value(with a documenting comment) in the class definition?

Just out of curiosity (no objection), wa the static attribute?

IvarsKarpics: Well, it is a long story :) Let me explain the situation. There is a checkbox called Compression in the  standard collection widget to enable file compression, but there is no checkbox in the characterization widget. So if the characterization succeeds then it has to have the compression flag as it is for standard collections.
To solve this puzzle I introduced diff_plan_compression variable, that will indicate if the diffraction plan has compression. It is set/unset when user clicks on the compression checkbox in data_path_widget. If I put the variable in the init then when the Characterisation item is created the diff_plan_compression will be set as defined in init and the diffraction plan may have wrong compression state. Static method will set the compression state at the gui start (via task_toolbox_brick) and it will be changed when the user clicks the checkbox. And not overwritten  in the __init__
Maybe there is a better solution to solve this.

rhfogh: Makes sense. Thanks for telling - I was just curious. Still think that the value should be initialised in the code, though, even if that is awways overwritten before use.

IvarsKarpics: We have  similar situation in the PathTemplate where several variables (base_directory, template, precision, etc) are set via static methods. Maybe this could be one topic to be discussed during the next developers meeting.

IvarsKarpics:
I found a solution to define value in __init__ and set it correctly via gui.

PR #377, opened 2019-04-25
WIP: First for-discussion draft of a Beamline object

rhfogh:
This just to start discussion. Some points:

- Defining the roles as properties makes them visible to linters, and guarantees that roles have the right spelling. 

- The hierarchy is an alternative to putting everything into a single object (simple, but crowded). It does require that people can find things easily, and the right place to put things will surely change under discussion. I propose to have a 'beam' object; alternatively its contents could be moved to the top level, but that would be rather more confused. 

- All hardwareObjects are configured in only *one* place, either at the top level (here) or within one of the objects given here. 

- All properties in this file are mandatory, lower in the hierarchy they can be optional. 

- A related discussion is to what extent we want people to access objects lower in the hierarchy, as opposed to putting wrapper functions in the top level object. I.e., should it be 'detector.set_distance(value)' or detector.distance_motor.set_position(value)'?

### DIFF
--- a/Beamline.py
+++ b/Beamline.py
@@ -1,0 +1,227 @@
+#! /usr/bin/env python
+# encoding: utf-8
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Beamline class serving as singleton container for links to top-level HardwareObjects
+
+Starting point for discussion. Some points:
+
+- We  could move the contents of 'beam' (below) to the top level.
+  do we want that, with the attendant confusion?
+
+- Do we want people to access teh contained hardware objecs (flux, motors, ...)
+  or to we want wrapper functions in beam, diffractometer, etc.
+
+- Are things in the right place? E.g. 'resolution' could certainly be elsewhere.
+
+
+
+The proposed structure would be somethign like the following,
+where the topmost level is contained in this object.
+
+- beam: AbstractBeam # NB still TBD. Replaces and extends BeamInfo
+
+    # Xray beam upstream of diffractometer, and related matters
+
+    flux: AbstractFlux
+    transmission: float
+    energy: AbstractEnergy
+    fastshutter: AbstractNState
+    safetyshutter: AbstractNState
+    door_interlock: AbstractNState
+    # attenuators: AbstractAttenuator
+    # slits: AbstractSlits
+    # apertures: AbstractAperture
+
+- diffractometer: AbstractDiffractometer
+
+    # The diffractometer and related motors and actuators
+
+    omega: AbstractMotor
+    kappa: AbstractMotor
+    phi: AbstractMotor
+    chi: AbstractMotor
+    get_rotation_motors()
+
+    alignmentx: AbstractMotor
+    alignmenty: AbstractMotor
+    alignmentz: AbstractMotor
+    centringx: AbstractMotor
+    centringy: AbstractMotor
+    get_alignment_motors()
+
+    frontlight: AbstractNState
+    backlight: AbstractNState
+    scintillator: AbstractNState
+    beamstop: AbstractNState
+    capillary: AbstractNState
+
+- graphics
+
+    # The OAV, camera, and related.
+
+    camera: AbstractVideoDevice
+    zoom: AbstractNState
+    # focus??  Generallyidentical to one of the alignment motors.
+    # Shape history stored here
+
+- detector: AbstractDetector
+
+    # The detector and attached motors and actuators
+
+    distance: AbstractMotor
+    resolution: AbstractMotor
+    two_theta: AbstractMotor
+    detector_cover: AbstractNState
+
+- sample_changer: AbstractSampleChanger
+
+- queue: QueueManager
+
+- queue_model: QueueModel
+
+- session: Session
+
+- lims_client: AbstractLimsClient # NBNB still TBD.
+
+Experimental procedures
+
+    # Top level objects? Dictionary?
+
+    energy_scan
+    mxcollect
+    xrf_spectrum
+    xray_imaging
+    gphl_workflow
+    edna
+    centring ?
+
+Server links
+
+    # Top level objects? Dictionary?
+
+    rpc_server
+    redis_client
+
+Software procedures
+
+    # Top level objects? Dictionary?
+
+    auto_processing
+    data_analysis
+    parallel_processing
+
+"""
+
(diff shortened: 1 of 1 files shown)


rhfogh:
If people like this, I can make a draft of the changes this would suggest for e.g. AbstractDetector or AbstractDiffractometer.

Beamline.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+The proposed structure would be somethign like the following,
+where the topmost level is contained in this object.
+
+- beam: AbstractBeam # NB still TBD. Replaces and extends BeamInfo
+
+    # Xray beam upstream of diffractometer, and related matters
+
+    flux: AbstractFlux
```
IvarsKarpics: Thank you  @rhfogh for preparing the first version of beamline object. I have few questions and do not know if it is better to make comments here or discuss them during the face to face meeting. For example: Do you propose structure like beamline.beam.flux or beamline.flux?

Beamline.py
```diff
@@ -0,0 +1,227 @@
...
+    transmission: float
+    energy: AbstractEnergy
+    fastshutter: AbstractNState
+    safetyshutter: AbstractNState
+    door_interlock: AbstractNState
+    # attenuators: AbstractAttenuator
+    # slits: AbstractSlits
+    # apertures: AbstractAperture
```
IvarsKarpics: Apertures or aperture?

Beamline.py
```diff
@@ -0,0 +1,227 @@
...
+
+- sample_changer: AbstractSampleChanger
+
+- queue: QueueManager
+
+- queue_model: QueueModel
+
+- session: Session
+
+- lims_client: AbstractLimsClient # NBNB still TBD.
```
IvarsKarpics: Maybe just `lims`

Beamline.py
rhfogh: That is up for discussion, but I am proposing beamline.beam.flux.

Originally I thought to put all hardware objects inside beamline, including individual motors. That is simpler,  but Marcus and Antonia convinced me it would be too messy with everything in one place. I think that once we have a hierarchy it is better to have all the topmost objects as mandatory singletons, instead of putting varying numbers of actuators, slits, and attenuators directly inside beamline in parallel with fixed objects like the diffractometer.

Beamline.py
rhfogh: This is a TBD comment rather than a precise proposal. As I understand it, different beamlines will have different numbers of slits, attenuators etc. with different names. I do not know enough to propose whether there should be something like an attenuators dictionary, whether there are fixed attenuator names that should be used (like for diffractometer motors), or whether each beamline can use its own local role names. 

Input welcome.

Beamline.py
rhfogh: Cool.

Beamline.py
rhfogh: @IvarsKarpics I think it could be useful to have a discussion here and now. Then I can produce a better and more detailed draft before the f2f meeting, and we will have less issues to deal with there.

marcus-oscarsson:
Thank you for all this work, Its a good start

Beamline.py
marcus-oscarsson: So I guess beam-shape and size also goes with beam then ?

Beamline.py
marcus-oscarsson: One why would be to say that we set beam size and shape from the beam interface and what it does under the hood is beamline specific. I imagine though that one might need more fine grained control over for instance the slits, so we might also need a more well defined interface for the standard beam defining equipment.

Beamline.py
```diff
@@ -0,0 +1,227 @@
...
+    beamstop: AbstractNState
+    capillary: AbstractNState
+
+- graphics
+
+    # The OAV, camera, and related.
+
+    camera: AbstractVideoDevice
+    zoom: AbstractNState
+    # focus??  Generallyidentical to one of the alignment motors.
```
marcus-oscarsson: Indeed but the semantics are different so I think there should be a focus motor defined, the user has to live with that the sample moves, if it as it often is identical to sampx

Beamline.py
```diff
@@ -0,0 +1,227 @@
...
+
+- queue_model: QueueModel
+
+- session: Session
+
+- lims_client: AbstractLimsClient # NBNB still TBD.
+
+Experimental procedures
+
+    # Top level objects? Dictionary?
```
marcus-oscarsson: Data class ?

marcus-oscarsson: Ohm, sorry we cant use those since we also want to be Python 2.7 compatible. How about just a static class ?

Beamline.py
rhfogh: @marcus-oscarsson That would be my idea.

IvarsKarpics:
I prepared a first implementation of beamline object, see PR #400 . Should we close this PR and continue working on #400 ?

jordiandreu:
> I prepared a first implementation of beamline object, see PR #400 . Should we close this PR and continue working on #400 ?

I do agree.

rhfogh:
I have some reservations on using __getattrribute__ instead of explicit properties. Sure it saves on typing all the attributes excplicitly, it means that your development environments do not know whether a given attribute exists, and that it looks acepted to put any property you want in without documenting it in a visiible place. Can we keep both PR alive till I have time to come up with a more detailed reply?

IvarsKarpics:
My first version of beamline object had hard coded attributes: https://github.com/mxcube/HardwareRepository/pull/400/commits/6a87dbecb135784882e41df4833fd5c9a6f8a871
No problem to return this implementation. However I would like to add also dynamic setattribute part, because it would allow to add site specific hardware objects.

rhfogh:
Would it not be better to subclass the Beamline object, and put the site specific objects and th eir roles into the subclass? Then everybody could see qhich objects were there, adn what theyr are called.

IvarsKarpics:
Yes, that would help a lot!

rhfogh:
Maybe we could have e.g. 

class EMBLBeamlineObject(BeamlineObject):

The subclass would have additional properties (but still keep those from the superclass), and the subclass __init_ (and init() if any) could call those from the superclass.

The local config file would anyway have to have values for all attributes, general or site-specific as they might be, and we would need to point the aplication to the right place to pick up this file. We would only have to give the class of the BeamlineObject in the config file as well.

marcus-oscarsson:
I also though that dynamic attributes would be a good idea in the beginning, but its actually just convenient for the person writing the code the first time. The result will not be as clear and it will be difficult to know without double checking configuration or additional documentation (which often does not exist) what the Beamline object actually provides. I also think that it somehow short circuits good OOP practice , as @rhfogh mentioned it would be better to simply subclass Beamline and add the site specific bits. Although even that can be argued as something we should avoid as for as possible within our community, since we would like to share as much logic as we can.

So I think I would also prefer a at least a minimal set of explicitly defined attributes, like the ones we discussed on the meeting. Its of course difficult to avoid creating sub-classes of Beamline for site specific logic but it would be good if it was done with care :)

IvarsKarpics:
Yes, I see you point @rhfogh , but if we agreed that beamline object will not be a hardware object and the name of site specific Beamline object is in then config file, then how technically you will import this site specific class? Should we make our life easier and rename BeamlineSetup hwobj to Beamline and be happy.

rhfogh:
The reason I think it should not be a HardwareObjeect is that it should not have all the extra functionality that HOs have. Especially it should *not* have getObjectByRole - we do not want people to fall into temptation. 

But we did agree that there should be a configuration file for the Beamline object - in yaml - so the application will (must) be able to find that file. And it will not be hard to put in that file a string that gives that name of the class (and module) to import. We can then import that class (be it Beamline EMBLBeamline or whatever) aand instantiate it. Effectivevely e will be doing whtat we are doing for HardwareObjects, just a a single special case. We do not need to have all the extera baggage from the HArdwareObjects in order to do that.

rhfogh:
Superseded by PR #402

PR #380, opened 2019-05-06
Changes to fit to P14, including post-test fixes

rhfogh:
Changed from calling collect_hwobj to calling underlying hwobj
dditional GPhL bug fix

### DIFF
--- a/ConvertUtils.py
+++ b/ConvertUtils.py
@@ -25,9 +25,8 @@ __date__ = "19/06/17"
 
 
 # 'Borrowed' from six, pending installation as a dependency
-PY2 = sys.version_info[0] == 2
-PY3 = sys.version_info[0] == 3
-if PY3:
+PYVERSION = sys.version_info[0]
+if PYVERSION > 2:
     string_types = str,
     text_type = str
     binary_type = bytes
@@ -133,8 +132,8 @@ def ensure_text(s, encoding='utf-8', errors='strict'):
 #     """
 #     if not isinstance(s, (text_type, binary_type)):
 #         raise TypeError("not expecting type '%s'" % type(s))
-#     if PY2 and isinstance(s, text_type):
+#     if PYVERSION < 3 and isinstance(s, text_type):
 #         s = s.encode(encoding, errors)
-#     elif PY3 and isinstance(s, binary_type):
+#     elif PYVERSION > 2 and isinstance(s, binary_type):
 #         s = s.decode(encoding, errors)
 #     return s
--- a/ConvertUtils.py
+++ b/ConvertUtils.py
@@ -27,13 +27,13 @@ __date__ = "19/06/17"
 # 'Borrowed' from six, pending installation as a dependency
 PYVERSION = sys.version_info[0]
 if PYVERSION > 2:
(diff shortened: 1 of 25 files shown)


IvarsKarpics:
Hi @rhfogh , could you please resolve the conflict. I would suggest to make a copy of GphlWorkflowQueueEntry, checkout out the queue_entry from master and paste GphlWorkflowQueueEntry into HardwareObjects/Gphl/GphlQueueEntry.py

rhfogh:
@IvarsKarpics 
In progress.

rhfogh:
All post-test bug fixes added. No longet WIP, but ready to review

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -504,6 +504,9 @@ def get_beam_centre(self):
         else:
             return None, None
 
+    def get_beam_centre_pix(self):
+        return self.detector_hwobj.get_beam_centre_pix()
```
IvarsKarpics: Maybe you can add get_beam_centre_pix to AbstractDetector?

ConvertUtils.py
```diff
@@ -16,24 +16,30 @@
...
 
 # Constants
 
-# Conversion from kEv to A, wavelength = H_OVER_E/energy
-H_OVER_E = 12.3984
-
 
-# Enumerations:
+# 'Borrowed' from six, pending installation as a dependency
+PY2 = sys.version_info[0] == 2
```
IvarsKarpics: Looks like PY2 is not used and PY3 is used just once. Maybe remove them?

HardwareObjects/abstract/AbstractCollect.py
rhfogh: Not qure how to do this. Some beamlines definitely use get_beam_centre to return values in pixels, see e.g. this from ALBACollect:
        beamx, beamy = self.detector_hwobj.get_beam_centre()
        self.image_headers["Beam_xy"] = "(%.2f, %.2f) pixels" % (beamx, beamy)

Or this, from BIOMAXCollect:
        beam_centre_x, beam_centre_y = (
            self.get_beam_centre()
        )  # self.get_beam_centre_pixel() # returns pixel
        config["BeamCenterX"] = beam_centre_x  # unit, should be pixel for master file
        config["BeamCenterY"] = beam_centre_y

Or this, from DetectorMockup:
    def get_beam_centre(self):
        """Get approx detector centre (default to Pilatus values)"""
        xval = self.getProperty('width', 2463)/2. + 0.4
        yval = self.getProperty('height', 2527)/2. + 0.4
        return  xval, yval

Should we not either leave things unchanged, or make a PR where we fix everything in a consistent manner, which people can then discuss?

ConvertUtils.py
rhfogh: Checked in proposed fix (The new PYVERSION might be useful to have)

HardwareObjects/abstract/AbstractCollect.py
IvarsKarpics: Ok, lets leave it as it is and agree on it later.

PR #382, opened 2019-05-15
EnergyMockup is_ready

marcus-oscarsson:
Hi

 - Small improvement of EnergyMockup is_ready

Marcus

### DIFF
--- a/HardwareObjects/abstract/AbstractEnergy.py
+++ b/HardwareObjects/abstract/AbstractEnergy.py
@@ -42,7 +42,7 @@ class AbstractEnergy(HardwareObject):
         self.getCurrentWavelength = self.get_current_wavelength
 
     def is_ready(self):
-        return True
+        return not self._moving
 
     def abort(self):
         self._aborted = True


HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, name):
         self.getCurrentWavelength = self.get_current_wavelength
 
     def is_ready(self):
-        return True
+        return self._moving
```
rhfogh: Should this not be:
    return not self._moving

rhfogh:
Not sure I agree with this:

Strategically: Why do we want text and message functions in the HardwareObject? Is this not a UI issue - so that you should get the state from the HO, and then set the text (or icon, or color change, ...) you want for that state at the UI level? Could that text ever be used for anything other than displaying? Or, alternatively, is there not a state, Open/Closed, where you could use the state name for this purpose?

Technically: If we do decide to do this, do we really need separate functions to return the value of a static class attribute? Would it not be better to just use self.OPEN_TEXT and self.CLOSE_TEXT ? Or a readOnly property, if you want to guard against overwriting?

marcus-oscarsson:
@rhfogh Yes, I agree, I maybe submitted this a bit to fast, there are probably a better way of doing this.

HardwareObjects/abstract/AbstractEnergy.py
marcus-oscarsson: Fixed :)

marcus-oscarsson:
Thanks for the comments, I fixed the is_moving and removed the commits related to the action labels

PR #383, opened 2019-05-17
[WIP] Abstract Attenuators

IvarsKarpics:
* Defines AbstractAttenuators with 3 properties
* Adjusted mockup and Attenuators class.
* Added pytest

### DIFF
--- a/HardwareObjects/abstract/AbstractAttenuators.py
+++ b/HardwareObjects/abstract/AbstractAttenuators.py
@@ -1,64 +1,101 @@
 #
 #  Project: MXCuBE
-#  https://github.com/mxcube.
+#  https://github.com/mxcube
 #
 #  This file is part of MXCuBE software.
 #
 #  MXCuBE is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
+#  it under the terms of the GNU Lesser General Public License as published by
 #  the Free Software Foundation, either version 3 of the License, or
 #  (at your option) any later version.
 #
 #  MXCuBE is distributed in the hope that it will be useful,
 #  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU General Public License for more details.
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+#  GNU Lesser General Public License for more details.
 #
-#  You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


IvarsKarpics:
A corresponding PR to mxcube/mxcube will follow.

IvarsKarpics:
After agreeing on the variable and method names I will change all the classes that use the transmission hwobj.

HardwareObjects/abstract/AbstractAttenuators.py
```diff
@@ -1,64 +1,112 @@
...
+        """
+        Returns current transmission in %
+        :return: float (0 - 100)
+        """
+        return self.__transmission_value
 
-    def set_limits(self, limits):
-        self._limits = limits
-        self.emit("limitsChanged", (self._limits,))
+    def set_transmission(self, value, timeout=None):
```
IvarsKarpics: Should all setters (like set_transmission, set_energy, set_position) have arguments: `value, wait=False,timeout=None` ? Should all hardware objects contain `wait_ready(self, timeout)` method?

rhfogh: That is would require hardware knowledge which I do not have. Ideally one would prefer to have only one way to handle waiting - either via the setters, or via a wait_ready function - not both. If you know which functions are likely to cause waits and can come up with sensible time-outs then it would seem  more sensible to do *without* the wait_ready(). If it is unpredictable what a geven HO might be waiting for and how long, then maybe you should have *only* wauit_ready, and do without timeout and wait. I would prefer the former, but it all depends.

Doies anyone know?

IvarsKarpics: Well, I would prefer to have 
```
if wait:
    self.wait_ready(timeout)
```
within the setters, because I would like to have a single method to achieve synchronous and asynchronous behavior. There might be a case, when you start several commands (like start moving several detector stage motors at the same time) and wait till all of them are ready. But that could be a specific case and still you can use wait_ready.

IvarsKarpics:
No, no, I will not just merge the PR. Maybe we can start discussion now and do the actual merging during the dev meeting.

rhfogh:
@IvarsKarpics 
I am sure that you would not do that - but a third party, like Marcus, might legitimately decide to do so. Hence the warning.

marcus-oscarsson:
So this is WIP and intended to be discussed on the meeting then ?

IvarsKarpics:
Yes @marcus-oscarsson . I changed the name. To be finalized during the face to face meeting.

IvarsKarpics:
Closing the PR as we agreed during the dev meeting

PR #385, opened 2019-05-17
WIP Proposal for new abstract Transmission class

rhfogh:
Alternative proposal for AbstractTransmission (was AbstractAttenuators), for discussion.
Contains an abstract class, a mock, and a sample implementation.

Some comments:

- Transmission is a better name, since it is singular (unliker 'Attenuators') and since the associated value is a transmission. How it is implemented underneath is a matter for lower layers.

- We should not have both public functions and properties, that clutters. And since we need some parameters, like 'timeout', functions it is.

- It is better to have only one parameter, 'timeout', as proposed by Ivars, and avoid the 'wait' parameter. The syntax here has no way to say 'wait forever', if you want that you must give a long timeout, e.g. 999999s

- I think it is better to have the parameters of getters and setters named uniformly 'value'. The signals and internal attributes have more descriptive names. Also my names are a bit shorter than Ivars'

- Any implementation casn be made by implementing the six abstract methods in the AbstractTransmission. As much work as possible has been put in the acutal abstract class.

- The implementation must take care of sending all signals, I can see no way of doing that from the top layer reliably. All state should be kept in the subclasses.

I shall shortly add another file, with an implementation matching Ivars Attenuators class

### DIFF
--- a/HardwareObjects/abstract/AbstractTransmission.py
+++ b/HardwareObjects/abstract/AbstractTransmission.py
@@ -1,0 +1,169 @@
+#! /usr/bin/env python
+# encoding: utf-8
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Defines abstract Transmission
+"""
+
+from __future__ import division, absolute_import
+from __future__ import print_function, unicode_literals
+
+from abc import ABCMeta, abstractmethod
+from enum import Enum, unique
+import gevent
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+
+__credits__ = [" Copyright © 2016 - 2019 by Global Phasing Ltd. All rights reserved"]
+__license__ = "LGPLv3+"
+__category__ = "General"
+__author__ = "rhfogh"
+__date__ = "20190517"
+
+
+# NB we need a set of general motor states, with a minimum number of possibilities
+# which can be shared among many abstract classes
+# This is just a placeholder, based on ShutterStateprobably we need different values
+@unique
+class MotorState(Enum):
+    """
+    Defines the valid General motor states
+
+    NB We want Enum, NOT IntEnum, because we want MotorState.READY == 2 to be False
+    """
+
+    UNKOWN = 0
+    READY = 1
+    MOVING = 2
+    AUTOMATIC = 3
+    DISABLED = 4
+    FAULT = -1
+    ERROR = -2
+
+
+class AbstractTransmission(HardwareObject):
+    """Abstract transmission"""
+
+    __metaclass__ = ABCMeta
+
+    STATE = MotorState
+
+    def __init__(self, name):
+        super(AbstractTransmission, self).__init__(name)
+
+    def set_transmission(self, value, timeout=None):
+        """
+        Sets transmission. NB actual value set may differ from input value
+        :param value: float (0 - 100)
+        :return:
+        """
+        limits = self.get_limits()
+        if None not in limits and (value >= limits[0]) != (value <= limits[1]):
+            raise ValueError("transmission value %s outside imits %s" % (value, limits))
+        if timeout is None:
+            self._set_transmission(value)
+        else:
+            gevent.with_timeout(timeout, self._set_transmission, value)
+
+    def set_limits(self, value):
+        """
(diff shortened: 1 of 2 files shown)


rhfogh:
Have to leave now, so no more on this.

Also, I realise I do not properly understand Ivars PR. How is the information transferred from the Python HO to the underlying channels? The values that are returned are in these __transmission_value etc. attributes, but how do they get to the HO? and how do they get back?

Given my lack of understanding Ivars can surely improve on this proposal. But hopefully the discussion can help.

HardwareObjects/mockup/TransmissionMockup.py
```diff
@@ -1,46 +1,108 @@
...
+class TransmissionMockup(AbstractTransmission.AbstractTransmission):
+    """MockTransmission implementation using new AbstractTransmission
+    """
 
-    def getAttFactor(self):
-        return self.value
+    def __init__(self, name):
+        super(TransmissionMockup, self).__init__(name)
+
+        self._transmission = None
```
IvarsKarpics: Did not we agree that we will use properties?

beteva: My (few :-) comments:
1. Why do we have MotorState in AbstractTransmission. On top of it I personally do not like negative values. In the same line it is weird to have the __matt_state_to_state method.
2. When we talk about states, would it be better to have a generic states enum, good for all the Hardware Objects or we want separate states for each abstract type? I would vote for a generic one, with only few values and should be IntEnum.
3.  I also thought that we want properties rather than functions for get/set.
4. To calculate the transmission, we need to know the energy. Although it is easier to leave this for the underlying HO/control system, just in case we should as well foresee a method for getting the current energy.
5. I am strongly opposed to define the methods as abstract (with the abc module). We should avoid as much as possible, except for very few methods, as the control systems differ from one site to another and there might not be a way to define a concrete method to overload the abstract one. This does not mean that the methods are not defined in the Abstract class.
6. The methods doc strings do not seem to be google style (as we decided)
7. Last but not least: the credits (especially in the abstract classes) should go to the MxCUBE collaboration and not to a particular member of the collaboration and even less to a particular person.

HardwareObjects/abstract/AbstractTransmission.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Wait up to timeout s till device is ready
+        :param timeout: period to wait (in s) before returning, default to config value
+        :return:
+        """
+        if timeout is None:
+            timeout = self.getProperty("default_timeout", 20)
+        with gevent.Timeout(
+            timeout,
+            Exception("%s: Timeout waiting for ready" % self.__class__.__name__),
```
beteva: Exception is too generic. Why not RuntimeError or TimeoutError.

HardwareObjects/abstract/AbstractTransmission.py
```diff
@@ -0,0 +1,169 @@
...
+        ):
+            while self.get_state() is not self.STATE.READY:
+                gevent.sleep(0.1)
+
+    @abstractmethod
```
beteva: Why imposing?

HardwareObjects/mockup/TransmissionMockup.py
rhfogh: @IvarsKarpics @beteva 
I think it is a bad idea to have both properties and also functions to interact with the same attribute. Going via the property would then be an alternative setter that bypasses silently the normal timeout handling - which sounds like a source of confusion to me. And we cannot have only properties, since the set/move functions takes extra parameters. We could maybe have a read-only property instead of the getter function, and use a move_ function to set things. To be discussed

@beteva 

- I suggest we should have a generic states enum shared widely. I just picked MotorStates because 1) it was around, and 2) I thought transmission counted as a sort of motor. The problem is that if the states come as integers from some external channel one would need to convert them to the standard enum. Hence the __matt_state_to_state dictionary. But I'll happily hear from people who understand this better.

- The get_energy function should be on the energy object or beamline object, IMHO, not on the transmission object.

- The doc strings are not yet set properly, I know. It is on my TODO list, meanwhile this was a discussion starter.

- This is my standard credits setting - again it was a quick discussion starter. When doing it for real, I would only put this kind of string on gphl-specific code. Would that be more accerptable?

- Last but not least, The point of the abstractmethod is to make sure that there are functions that one can rely on and code to. If you can code a function so that it only calls other functions, you can put it into the abstract class. But if it needs to be implemented lower down, it is no good putting a dummy function (that gives the wrong result) in the abstract class. If yours is the only beamline that for some reason is unable to implement, say, get_transmission in a meaningful way there is of course no help for it - any code that relies on get_transmission will break, and we could deal with that by having the abstract class function raise NotImplementedError. But there is an alternative case: You might be perfectly able to implement, say, get_transmission, but if you are in a hurry and do not need it yourself, you might choose not to bother, or to give it a different name or signature that fits better in your local context. This is where the abstractmethod property serves to keep people out of temptation.

HardwareObjects/abstract/AbstractTransmission.py
rhfogh: By all means - I had not thought abotu the possibility

HardwareObjects/abstract/AbstractTransmission.py
beteva: Even if I am strongly, I am not completely opposed to the abstract decorator. The only message is that we should really use it for few methods - e.g. only get_/set_transmission in this case and not init, limits ...

beteva:
@rhfogh some of the subjects (like the state enum) I've only put here in order to start a discussion. We should see this at our face to face meeting.

Of course you can put your credits on the gphl-specific code ;-)

rhfogh:
Superseded by new WIP

PR #400, opened 2019-06-17
[WIP] Beamline object

IvarsKarpics:
* Reads configuration file and sets Beamline attributes based on hard coded roles (see table).
* Beamline object is initialized in api.


HardwareRepository.py
```diff
@@ -16,26 +16,33 @@
...
 __author__ = "Matias Guijarro"
 __version__ = 1.3
 
-_instance = None
-_hwrserver = None
+_hwr_instance = None
+_hwr_path = None
 _timers = []
 
+beamline = None
```
marcus-oscarsson: This looks nice, good idea :)

Beamline.py
```diff
@@ -0,0 +1,108 @@
...
+
+from __future__ import division, absolute_import
+from __future__ import print_function, unicode_literals
+
+import os
+
+__license__ = "LGPLv3+"
+
+
+class Beamline(object):
```
marcus-oscarsson: I thought we said that @rhfogh would edit his original proposal or am I mistaken. In anyways, I think we discussed the dynamic mapping of the configuration file to python attributes and it seemed like most people were against it. I was the one that suggested it so I'm not against the idea. I think we then decided that we were going to have a more explicit interface  instead. Or is that just me that hit the "door" to hard (sorry, meeting participants joke ... :) )

rhfogh: @marcus-oscarsson @IvarsKarpics 
We did indeed agree on having explicit mappings instead. And I think we should stick to that. My proposed version would be a top level Beamline with a s many properties as we can agree are general, and then subclasses with just those properties that are better seen as beamline specific. 

We also agreed that I would be editing my proposal and we should go on from there, but as it happens I do not have time for another week - we got our consortium meeting coming up - and anyway we said that there should be a two week period for people to comment. I shall certainly be looking at whatever proposals are current as soon as I have time, but if Ivars meanwhile has time to come up with an alternative starting point we can also look at, I can only be happy with that.

PR #402, opened 2019-07-03
WIP: Yaml-configured BeamlineObject

rhfogh:
This is a new proposal for the beamline object, with yaml configuration. 

There is still a fair bit to do, but before I add functionality, try to extend this to e.g. the diffractometer, or modify the code to fit the new system, we should agree if this is indeed the way to go. 

Comments
=========

  -  The framework will support configuration of any contained object, so I propose it could eventually become a replacement for HardwareObjectNode. I have therefore put it in HardwareRepository and BaseHardwareObjects. For a start (transition phase?) it should support both yaml-configured objects and xml-configured HardwareObjects as contained objects. 

  - All contained objects must be defined as properties, with a single role name per object. They are optional, so you can simply omit them from the configuration file. This is not limited to HardwareObjects, you could configure e.g. AcquisitionParameter objects.

  - The BeamlineObject (and any other object that uses the framework) can be subclassed - you just need to add any additional properties and update a single short function in the subclass

  - Each object belongs to one (and only one) category, and you can get objects by category. That allows you to define and get e.g. all procedures or all centring functions. 

  - You can put lists, maps, or scalars in as non-object properties. These do not have to be defined as properties, but for safety the corresponding attribute has to be set in the class.\_\_init\_\_

TODO
====

  - DONE Check the ruamel.yaml module, which looks like a better replacement for standard yaml (which has a lot of gotchas) (ACTION @rhfogh )

  - DONE Check the yaml formatting we want and the dump options this corresponds to, and rewrite the example to match. (ACTION @rhfogh)

  - Add  output for loaded objects (with load time), failed loads, and non-configured objects. (ACTION @rhfogh)

### DIFF
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -32,7 +32,7 @@ __license__ = "LGPLv3+"
 __author__ = "Rasmus H Fogh"
 
 from collections import OrderedDict
-from BaseHardwareObjects import ConfiguredObject
+from HardwareRepository.BaseHardwareObjects import ConfiguredObject
 
 
 class Beamline(ConfiguredObject):
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -27,7 +27,7 @@ All HardwareObjects
 from __future__ import division, absolute_import
 from __future__ import print_function, unicode_literals
 
-__copyright__ = """ Copyright © 2019 by Global Phasing Ltd. """
+__copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
 __license__ = "LGPLv3+"
 __author__ = "Rasmus H Fogh"
 
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -1,0 +1,467 @@
+#! /usr/bin/env python
+# encoding: utf-8
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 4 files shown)


rhfogh:
The yaml configuration file now has the formatting I would propose as standard. Comment welcome.

Also the code now uses ruamel.yaml, which 1) uses yaml 1.2 with fewer gotchas, 2) has a round-trip mode that lets you load and write out while preserving comments and fomatting, 3) is clearly in current, active development.

marcus-oscarsson:
Thanks for doing all this, I think the idea of using ruamel.yaml is a good idea :). Ill take a closer look at the rest as soon as I can  ! :+1:

HardwareObjects/Beamline.py
```diff
@@ -0,0 +1,467 @@
...
+
+"""Beamline class serving as singleton container for links to top-level HardwareObjects
+
+All HardwareObjects
+"""
+
+from __future__ import division, absolute_import
+from __future__ import print_function, unicode_literals
+
+__copyright__ = """ Copyright © 2019 by Global Phasing Ltd. """
```
marcus-oscarsson: We should probably be careful with the copyright ... right :) ?

HardwareObjects/Beamline.py
```diff
@@ -0,0 +1,467 @@
...
+        return self._objects.get("gphl_connection")
+
+    __object_role_categories["gphl_connection"] = "hardware"
+
+    # centring
+
+    # NB Could centring be treated as procedures instesad?
+
+    @property
+    def n_click_centring(self):
```
marcus-oscarsson: The centering methods can maybe be retrieved through one method with options ?

HardwareObjects/Beamline.py
rhfogh: Fixed. I shall change my default settings. I left them, because it is easier to give away copyright than to take it back, in case you have made a mistake, but we cannot have this occurring all the time.

HardwareObjects/Beamline.py
rhfogh: Indeed. 
In fact it may well be that the system with procedures, one category per object etc. could be improved on, but I thought it was better to have a clear starting point for discussion.

rhfogh:
I have now added a 'name' attribute to ConfiguredObject. This makes littel difference to the BeamlineObject, but serves as preparation for using the ConfiguredObject also for normal HardwareObjects. See separate Issue for discussion.

HardwareRepository.py
```diff
@@ -29,13 +29,111 @@
 from HardwareRepository.dispatcher import dispatcher
 from HardwareRepository.ConvertUtils import string_types
 
+from ruamel.yaml import YAML
```
IvarsKarpics: Is there a simple way to install it for python2. By default it is for python3.

rhfogh:
Obsoleted by PR #409

PR #407, opened 2019-07-16
Python 3 compatibility and a few updates for the new HWR API

marcus-oscarsson:
Dear all,

This PR mostly contains some missing Python 3 compatibility fixes but also a few adaptations to the new HWR API. The most important Python 3 fixes are for the Exporter and ISPyB clients, there are also a few missing list conversions,

I've also updated the Energy hardware object to correspond more to AbstractEnergy and I updated the
AbstractAperture.

It's maybe not custom but since I took the time to regexp replace a few things I've also included a commit with site specific changes. There is nothing controversial there so I don't think there should be an issue. Just let me know in case you see something strange, and Ill simply remove that commit.

Cheers,
Marcus


HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -38,8 +38,8 @@ def __init__(self, name):
         self.getEnergyLimits = self.get_energy_limits
         self.getWavelengthLimits = self.get_wavelength_limits
         self.isReady = self.is_ready
-        self.getCurrentEnergy = self.get_current_energy
-        self.getCurrentWavelength = self.get_current_wavelength
+        self.get_current_energy = self.get_current_energy
```
IvarsKarpics: This looks like a result of automatic find and replace :)

marcus-oscarsson: Yes, I should of course not have changed those :)

marcus-oscarsson: So, I've fixed those :)

HardwareObjects/abstract/AbstractAperture.py
```diff
@@ -94,16 +91,14 @@ def set_diameter_index(self, diameter_index):
                 "Aperture: Diameter index %d is not valid" % diameter_index
             )
 
-    def get_diameter_size(self):
-        # TODO rename to get_diameter_size_mm
+    def get_diameter_size_mm(self):
```
jordiandreu: Shall we agree on a set of base units and do not add them into the methods name?

marcus-oscarsson:
I somehow agree with you that it seems unnecessary to have _mm in the name, I had the impression that it was decided this way. I'm personally perfectly fine with simply get_diameter_size as long as we agree on the return unit. I would also in that case vote for microns as unit (as we are already not using SI units)

marcus-oscarsson:
What does @IvarsKarpics, @beteva, @MartinSavko and @rhfogh think ?

IvarsKarpics:
Hi,

I am also in favor not to use the units in the methods. I guess we should add a list of units in the contributing guidelines.

marcus-oscarsson:
Okey, So Ill remove the units from the method names, since it seems like there have been some misunderstanding of what we actually wanted.

beteva:
Agree that we do not need to put the units in the method name. We probably would not need to agree on units, as this is abstract method which only gives a value. The units would depend on the local implementation.

beteva:
This PR is probably not the right place to discuss, but the are few things in the AbstractAperture which bother me. I will open an issue.

rhfogh:
@beteva 
I would *not* agree that choice of units should be left to local implementations. Some people (like me) might want to write code that would work on more than one beamline, in which case the units would pretty much have to be the same.

marcus-oscarsson:
I guess that @MartinSavko is perhaps on vacation, we can maybe wait a few more days for his review and then merge ...

marcus-oscarsson:
@IvarsKarpics, All, I guess this can be merged now, what do you think ?

rhfogh:
On closer examination the removal of the findInRepository call from DataAnalysis was correct (fits with using absolute file names) and so I shall make no changes here.

Rasmus

PR #411, opened 2019-09-04
ApertureMockup fix

MartinSavko:
This adds two methods to ApertureMockup.py: get_diameter_list() and get_position_list(). Both methods are expected by ApertureBrick.py. This fixes error when starting example gui with aperture mockup hardware object.


IvarsKarpics:
I created a new PR: https://github.com/mxcube/mxcube/pull/358

MartinSavko:
Thanks Ivars, I agree. I am closing the PR.

PR #412, opened 2019-09-12
Python 2 and 3  compatability

marcus-oscarsson:
Hi 

Yet another PR with Python 2 and 3 compatibility issues mostly:

- iteritems to items
- itervalues to values
- and iterkeys to "for key in"

and a few others.

I've also updated some camel cases to snake case, mostly related to diffractometer. The problem have somehow been circumvented by the GenericDiffractometer that makes a mapping between the two casings. I've simply done a multifile sed replace, so this mapping can eventually be removed.

As last time since I did the sed replace Ive also included commits with changes to other sites HardwareObjects, please have a look they are tagged with [SITE]. Let me know if there is something you dont want to change and Ill remove the commit and you have to make the update
on your own.

Macus


HardwareObjects/Session.py
```diff
@@ -9,7 +9,7 @@
 import socket
 
 from HardwareRepository.BaseHardwareObjects import HardwareObject
-from HardwareRepository.HardwareObjects import queue_model_objects
+from queue_model_objects import PathTemplate
```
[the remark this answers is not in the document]
IvarsKarpics: Hmm, I also thought that if you use absolute import over the code it would return the same object. I would suggest to add this topic to our next web meeting. Maybe in the future we could avoid using the static methods, for me there are not so transparent and some time ago I had some issues with that.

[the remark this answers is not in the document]
marcus-oscarsson: I guess static methods are fine but one should probably careful static variables.

In anyways to proceed with this right now, what are you thoughts on a possible solution. Do we keep it as it is for now, move `to_collect_dict` or maybe something else ?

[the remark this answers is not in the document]
rhfogh: This is a subtle bug indeed. But I think there is a solution:
Remove HardwareObjects  from the PythonPath, and do all imports everywhere starting from HardwareRepository.
I have not tested it, but if an object could be imported both via HardwareRepository.HardwareObjects, and just via HardwareObjects, it bevcomes hard or impossible to decide whether code in the file itself should be slotted under one or the other. SUrely if there is only one possible path, Python would be able to do the bookkeeping?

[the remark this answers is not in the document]
marcus-oscarsson: Hm, yes the only problem is that we are relying on HardwareObjects to on the python path, maybe moving it to the end would work, at least temporarily until we can remove the HardwareObjects path

marcus-oscarsson:
Hi all,

So I managed to find a solution to the import problem. I had forgotten one relative import and it seems that a previous relative import also changes how the objects in the module itself is loaded. So instead of loading it from the full path it loads it from the relative part.

While doing this I found a bug in the wrapper addCommand which did not return the added command. I then consequently did rename campaign to add_command.

I took as previously the liberty to commit the rename changes to other sites, just let me know if you would like me to remove those.

Cheers,
Marcus

HardwareObjects/FlexHCD.py.orig
```diff
@@ -0,0 +1,691 @@
+import base64
```
rhfogh: Do we normally check in this kind of file copy?

HardwareObjects/Microdiff.py
```diff
@@ -499,15 +501,15 @@ def start3ClickCentring(self, sample_info=None):
                 chi_angle=self.chiAngle,
             )
 
-        self.currentCentringProcedure.link(self.manualCentringDone)
+        self.current_centring_procedure.link(self.manualCentringDone)
 
-    def interruptAndAcceptCentring(self):
+    def interruptAndaccept_centring(self):
```
rhfogh: Why partial CamelCase instead of full snake_case?

HardwareObjects/SOLEIL/PX1/PX1Energy.py.orig
```diff
@@ -0,0 +1,298 @@
+# from qt import *
```
rhfogh: Again, do we normally sheck in this kind of file?

HardwareObjects/FlexHCD.py.orig
marcus-oscarsson: Ohh, no sorry thats a misstake on my part :)

HardwareObjects/Microdiff.py
marcus-oscarsson: Just a accidental rename, ill make all of it camel case :)

marcus-oscarsson:
Hello,

I've taken care of the issues you've mentioned above

Marcus

CommandContainer.py
```diff
@@ -384,8 +384,8 @@ def getCommandNamesList(self):
         return list(self.__commands.keys())
 
     def addCommand(self, arg1, arg2=None, addNow=True):
-        warn("addCommand is deprecated. Use add_command instead", DeprecationWarning)
-        self.add_command(arg1, arg2, addNow)
+        warn("add_command is deprecated. Use add_command instead", DeprecationWarning)
```
MartinSavko: just a typo -- warning is about addCommand ...

marcus-oscarsson: oops :)

rhfogh: Not sure I understand hwo you solved the import problem. Session.py apparently now says "from queue_model_objects import PathTemplate"
So, diid you simply get rid of relative imports alt9ogether?
Why do we have a mixture (in teh code, also elsewhere, of "import queue_model_objects and "from HardwareRepository.HardwareObjects import queue_model_objects" - would it not be better, nay necessary, to use the second form everywhere?

marcus-oscarsson: @rhfogh Could you make the comment on the appropriate line iof the code so I see were it refers to I think I'm using absolute imports everywhere even in Session.py

I guess we had a mixture because we recently moved to absolute imports and not all of them were changed ?

rhfogh: It is HardwareObjects.Session.py, line 12. It is marked as 'Outdated'; and I do not seem to be able ton add a comment in situ.

Anyway, he question is not bettween absolute and relative imports - I believe that you have removed all relative imports.

It is between 
"import queue_model_objects" (and variants with 'from' etc.)
and
"from HardwareRepository.HardwareObjects import queue_model_objects" (and variants with 'from' etc.)

As I understand it, one imports relative to the top level, and the other imports relative to HardwareRepository.HardwareObjects, and I believe that this will give you two diffferent imports resulting in two different objects.

marcus-oscarsson: Could you point to a place in the code were there are these differences :) ?

Import queue_model_objects would not work if the HardwareObject directory is not in the Python path.

So we should to my understanding avoid the usage of `import queue_model_objects` and use the import from the package root which is `HardwareRepository` so `from HardwareRepository.HardwareObjects import queue_model_objects` would result in an absolute import.

So would also  `import queue_model_objects` since HardwareObjects are on the python Path, but we always want to make sure that we use `from HardwareRepository.HardwareObjects import queue_model_objects` so we get the same object.

Are there still any `import queue_model_objects` lingering ?

rhfogh: I made a mistake - my apologies. Comment withdrawn.

I did find a lot of "import queue_model_objects" when searching throught the code base with pycharm - but on closer inspection it turned out that they are all in gui/qt3, Which means they are obsolete and irrelevant.

I completely agree with your explanation.

marcus-oscarsson: Ahh, allright dont worry,  I thought that something like that might have been case :)

CommandContainer.py
```diff
@@ -730,7 +730,7 @@ def add_command(self, arg1, arg2=None, addNow=True):
 
     def _addChannelsAndCommands(self):
         [self.addChannel(*args) for args in self.__channelsToAdd]
```
IvarsKarpics: If we do a major refactoring,then also we have to change addChannel to add_channel.

HardwareObjects/BeamlineSetup.py
```diff
@@ -422,29 +422,26 @@ def get_default_path_template(self):
     def _get_energy(self):
         try:
             energy = self.energy_hwobj.get_current_energy()
-        except AttributeError:
-            energy = 0
-        except TypeError:
+        except Exception as ex:
+            logging.getLogger("HWR").exception("Could not read energy")
             energy = 0
```
IvarsKarpics: Should we return 0 or None? Maybe better None, because 0 is a valid number (for example transmission can be 0)

HardwareObjects/BeamlineSetup.py
```diff
@@ -422,29 +422,26 @@ def get_default_path_template(self):
...
-        except TypeError:
+        except Exception as ex:
+            logging.getLogger("HWR").exception("Could not read transmission")
             transmission = 0
 
         return transmission
 
     def _get_resolution(self):
         try:
             resolution = self.resolution_hwobj.getPosition()
```
IvarsKarpics: getPosition -> get_position

CommandContainer.py
marcus-oscarsson: I think that's also what we said on yesterdays meeting, that all commands that today have deprecation warning should be updated

HardwareObjects/BeamlineSetup.py
marcus-oscarsson: Hm this is maybe not a part of this PR but an interesting question. I guess that its 0 because its a valid value and we in that way don't need to handle the display of a None energy. We usually check the state of the HO to verify if its "online" or not. So this normally not a problem for the UI since the control in question is marked as "offline", but it could maybe have some relevance for other parts of the code. 

Its at the same time been like this for sometime so I guess we can keep it like this if its not causing anyone sees any problem with it ?

HardwareObjects/BeamlineSetup.py
marcus-oscarsson: Yes, as we said yesterday. We will merge @rhfogh's PR and then go through the most common camel case to snake case conversions in a sperate PR.

The fix in this PR was only for the broken addCommand

marcus-oscarsson:
Merged by #416

marcus-oscarsson:
Merged by #416

PR #413, opened 2019-09-20
AbstractProcedure

IvarsKarpics:
Comments are welcomed!

### DIFF
--- a/HardwareObjects/abstract/AbstractProcedure.py
+++ b/HardwareObjects/abstract/AbstractProcedure.py
@@ -58,8 +58,13 @@ class AbstractProcedure(HardwareObject):
             self.procedure_failed = True
             msg = "Procedure execution failed (%s)" % str(ex)
             logging.getLogger("HWR").error(msg)
-        finally:
+
+        try: 
             self.post_execute(data_model)
+        except Exception as ex:
+            msg = "Procedure post execution failed (%s)" % str(ex)
+            logging.getLogger("HWR").error(msg)
+        finally:
             self.ready_event.set()
 
     def pre_execute(self, data_model):
--- a/HardwareObjects/abstract/AbstractProcedure.py
+++ b/HardwareObjects/abstract/AbstractProcedure.py
@@ -32,32 +32,68 @@ from jsonschema import (validate, ValidationError)
 __credits__ = ["MXCuBE collaboration"]
 
 
+@unique
+class ProcedureState(IntEnum):
+    """
+    Defines the valid Shutter states
+    """
+    RUNNING = 0
+    FAILED = 1
+    SUCCESFUL = 2
+    IDLE = 4
+
+
 class AbstractProcedure(ConfiguredObject):
+
+    __content_roles = []
+
     _ALLOW_PARALLEL = False
 
     # JSONSchema used for input and output validation and possibly
     # form generation.
     # https://json-schema.org/
+    # Needs to be defined by each subclass
     _ARG_SCHEMA = None
     _RESULT_SCHEMA = None
 
-    def __init__(self):
-        self._is_running = None
-        self._ready_event = None
-        self._msg = None
-        self._failed = None
-        self._results = None
+    def __init__(self, name):
+        super(AbstractProcedure, self).__init__(name)
+
+        self._msg=None
+        self._results=None
         self._ready_event = gevent.event.Event()
-        self._task = None
+        self._task=None
+        self._state = ProcedureState.IDLE
 
         # YML configuration options
-        self.category = ""
-        self.name = "Procedure"
-        self.extra = {}
+        # Category that the Procedure belongs to, configurable through
+        # YAML file and used by for listing and displaying the procedure
+        # in the right context.
+        self.category=""
+
+    def _init(self):
+        pass
+
+    def init(self):
+        pass
 
     def _execute(self, data_model):
         """
-        Task logic
+        Override to implement main task logic
+
+        Args:
+            data_model: Immutable data model, frozen dict in
+            Python 2.7 and Data class in Python 3.7. Input validated
+            by schema defined in _ARG_SCHEMA
+
+        Returns:
+        """
+        pass
+
+    def _post_execute(self, data_model):
+        """
+        Override to implement post execute task logic
+
         Args:
             data_model: Immutable data model, frozen dict in
             Python 2.7 and Data class in Python 3.7. Input validated
@@ -73,8 +109,7 @@ class AbstractProcedure(ConfiguredObject):
         Returns:
 
         """
-        self.procedure_failed = False
-        self.is_running = True
+        self._state = ProcedureState.RUNNING
         dispatcher.send(self, "procedureStarted")
 
     def _set_successful(self):
@@ -83,7 +118,7 @@ class AbstractProcedure(ConfiguredObject):
         Returns:
 
         """
-        self._is_running = False
+        self._state = ProcedureState.SUCCESFUL
         dispatcher.send(self, "procedureSuccessful", self.results)
 
     def _set_failed(self):
@@ -92,7 +127,7 @@ class AbstractProcedure(ConfiguredObject):
         Returns:
 
         """
-        self._is_running = False
+        self._state = ProcedureState.FAILED
         dispatcher.send(self, "procedureFailed", self.msg)
 
     def _set_stopped(self):
@@ -101,7 +136,7 @@ class AbstractProcedure(ConfiguredObject):
         Returns:
 
         """
-        self._is_running = False
+        self._state = ProcedureState.FAILED
         dispatcher.send(self, "procedureStopped", self.results)
 
     def _start(self, data_model):
@@ -109,14 +144,24 @@ class AbstractProcedure(ConfiguredObject):
         Internal start, for the moment executed in greenlet
         """
         try:
+            # The _pre_execute have been removed and can be done within
+            # execute if needed
             self._execute(data_model)
         except Exception as ex:
-            self._failed = True
+            self._state = ProcedureState.FAILED
             self._msg = "Procedure execution failed (%s)" % str(ex)
             logging.getLogger("HWR").error(self._msg)
         finally:
+            try:
+                self._post_execute(data_model)
+            except Exception as ex:
+                self._state = ProcedureState.FAILED
+                self._msg = "Procedure post_execute failed (%s)" % str(ex)
+                logging.getLogger("HWR").error(self._msg)
+
             self.ready_event.set()
-            if self._failed:
+
+            if self._state ==  ProcedureState.FAILED:
                 self._set_failed()
             else:
                 self._set_successful()
@@ -149,22 +194,13 @@ class AbstractProcedure(ConfiguredObject):
         return self._msg
 
     @property
-    def running(self):
+    def state(self):
         """
         Execution state
         Returns:
-            True if running otherwise false
+             ProcedureState: The current state of the procedure
         """
-        return self._is_running
-
-    @property
-    def failed(self):
-        """
-        Execution state
-        Returns:
-            True if task exectuion failed false otherwise
-        """
-        return self._failed
+        return self._state
 
     @property
     def results(self):
--- a/HardwareObjects/abstract/AbstractProcedure.py
+++ b/HardwareObjects/abstract/AbstractProcedure.py
@@ -50,6 +50,11 @@ class AbstractProcedure(ConfiguredObject):
         self._ready_event = gevent.event.Event()
         self._task = None
 
+        # YML configuration options
+        self.category = ""
+        self.name = "Procedure"
+        self.extra = {}
+
     def _execute(self, data_model):
         """
         Task logic
@@ -115,6 +120,24 @@ class AbstractProcedure(ConfiguredObject):
                 self._set_failed()
             else:
                 self._set_successful()
+
+    @property
+    def argument_schema(self):
+        """
+        Schema for argument passed to start
+        Returns:
+            str (JSONSchema)
+        """
+        return self._ARG_SCHEMA
+
+    @property
+    def result_schema(self):
+        """
+        Schema for result
+        Returns:
+            str (JSONSchema)
+        """
+        return self._RESULT_SCHEMA
 
     @property
     def msg(self):
--- a/HardwareObjects/abstract/AbstractProcedure.py
+++ b/HardwareObjects/abstract/AbstractProcedure.py
@@ -67,6 +67,13 @@ class AbstractProcedure(HardwareObject):
         finally:
             self.ready_event.set()
 
+    def stop_procedure(self):
+        """
+        Stops the execution of procedure
+        Returns:
+        """
+        self.set_procedure_stopped()
+
     def pre_execute(self, data_model):
         """
         Pre execute task
@@ -130,3 +137,12 @@ class AbstractProcedure(HardwareObject):
         """
         self.is_running = False
         self.emit("procedureFailed", self.failed_msg)
+
+    def set_procedure_stopped(self):
+        """
+        Emits procedureStoped signal
+        Returns:
+
+        """
+        self.is_running = False
+        self.emit("procedureStopped", self.procedure_results)
--- a/HardwareObjects/abstract/AbstractProcedure.py
+++ b/HardwareObjects/abstract/AbstractProcedure.py
@@ -1,0 +1,127 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
+
+"""
+AbstractProcedure
+"""
+
+import logging
+import gevent
+
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+
+
+__credits__ = ["MXCuBE collaboration"]
+
+
+class AbstractProcedure(HardwareObject):
+
+    def __init__(self):
+        self.is_running = None
+        self.ready_event = None
+        self.failed_msg = None
+        self.procedure_failed = None
+        self.procedure_results = None
+
+    def init(self):
+        self.ready_event = gevent.event.Event()
+
+    def start_procedure(self, data_model):
+        """
+        Starts procedure
+        Args:
+            data_model: data model defined in the queue_model_objects
+
+        Returns: None
+
+        """
+        try:
+            self.pre_execute(data_model)
+            self.execute(data_model)
+        except Exception as ex:
+            self.procedure_failed = True
+            msg = "Procedure execution failed (%s)" % str(ex)
+            logging.getLogger("HWR").error(msg)
+        finally:
+            self.post_execute(data_model)
+            self.ready_event.set()
+
+    def pre_execute(self, data_model):
+        """
+        Pre execute task
+        Args:
+            data_model: data model defined in the queue_model_objects
+
+        Returns: None
+
+        """
+        self.set_procedure_started()
+
+    def execute(self, data_model):
+        """
+        Actual exection task
+        Args:
+            data_model: data model defined in the queue_model_objects
+
+        Returns:
+
+        """
+        pass
+
+    def post_execute(self, data_model):
+        """
+        Post exectute
+        Args:
+            data_model: data model defined in the queue_model_objects
+
+        Returns:
+
+        """
+        if self.procedure_failed:
+            self.set_procedure_failed()
+        else:
+            self.set_procedure_successful()
+
+    def set_procedure_started(self):
+        """
+        Emits procedureStarted signal
+        Returns:
+
+        """
+        self.procedure_failed = False
+        self.is_running = True
+        self.emit("procedureStarted")
+
+    def set_procedure_successful(self):
+        """
+        Emits procedureSuccessful signal
+        Returns:
+
+        """
+        self.is_running = False
+        self.emit("procedureSuccessful", self.procedure_results)
+
+    def set_procedure_failed(self):
+        """
+        Emits procedureFailed signal
+        Returns:
+
+        """
+        self.is_running = False
+        self.emit("procedureFailed", self.failed_msg)
--- a/HardwareObjects/abstract/AbstractProcedure.py
+++ b/HardwareObjects/abstract/AbstractProcedure.py
@@ -16,101 +16,53 @@
 #
 #  You should have received a copy of the GNU General Public License
 #  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
-
-"""
-AbstractProcedure
-"""
-
 import logging
 import gevent
 
-from HardwareRepository.BaseHardwareObjects import HardwareObject
+from HardwareRepository.BaseHardwareObjects import ConfiguredObject
+from HardwareRepository.dispatcher import dispatcher
+
+# Using jsonschma for validating the JSCONSchemas
+# https://json-schema.org/
+# https://github.com/Julian/jsonschema
+
+from jsonschema import (validate, ValidationError)
 
 
 __credits__ = ["MXCuBE collaboration"]
 
 
-class AbstractProcedure(HardwareObject):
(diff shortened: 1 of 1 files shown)


HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        try:
+            self.pre_execute(data_model)
+            self.execute(data_model)
+        except Exception as ex:
+            self.procedure_failed = True
+            msg = "Procedure execution failed (%s)" % str(ex)
+            logging.getLogger("HWR").error(msg)
+        finally:
+            self.post_execute(data_model)
```
jordiandreu: I could have sense to protect also this post_execute call with a try/except block?

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -0,0 +1,127 @@
...
+import logging
+import gevent
+
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+
+
+__credits__ = ["MXCuBE collaboration"]
+
+
+class AbstractProcedure(HardwareObject):
```
jordiandreu: Is is necessary to inherit from HardwareObject? Only the _emit_ method is used and I'm not sure if is necessary to expose all the HardwareObject API in a procedure.

IvarsKarpics: I agree, but I also think it would be good to define signals and data structure emitted by these signals.

HardwareObjects/abstract/AbstractProcedure.py
IvarsKarpics: Done

HardwareObjects/abstract/AbstractProcedure.py
jordiandreu: I agree with defining signals and data emitted. In fact, not only the _emit_ but also the _connect_ and _disconnect_ methods are required to register the receivers/signals.
Shall we define the set of signals with associated data in some sort of module of definitions?

marcus-oscarsson:
Hi @IvarsKarpics , thanks for this nice work. I have a few additions to make could I maybe make them directly on this branch or would that be confusing ?

Marcus

marcus-oscarsson:
Hi

So I added a commit with a few updates with ideas that we have discussed since the last web meeting.

- Inheriting ConfiguredObject and I'm using the dipsatcher to emit the signals.
- Using JSONSchema and jsonschema python module to define input and outout schemas for validation and forms.
- Assuming that the tasks will run as greenlets for the moment, but this could be generalized to
 sub processes as well (could perhaps be interesting)
- And a few other minor things.

All this have to go with a few methods on the Beamline object, for listing procedures perhaps by some kind of category, running and stopping (or at least get a reference to perform those actions)

Please let me know what you think,
Marcus

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -0,0 +1,211 @@
...
+    _ARG_SCHEMA = None
+    _RESULT_SCHEMA = None
+
+    def __init__(self):
+        self._is_running = None
+        self._ready_event = None
+        self._msg = None
+        self._failed = None
+        self._results = None
+        self._ready_event = gevent.event.Event()
```
rhfogh: _ready_event() is defined twice

marcus-oscarsson:
Thanks for taking a look @rhfogh 

> As you can see I think this needs a fair bit of thought.
> Several comments, mostly general, so I put them here:
> 

So it seems like there are a few bits missing for ConfiguredObject so I will try to take care of those. There is maybe also a misunderstanding on how AbstractProcedure is supposed to be used. So ill try to explain how I imagine this is:

- A Procedure inherits AbstractProcedure and can be configured via a YAML file
- A Procedure can be be executed in parallel or sequentially, (sequentially by default)
- A Procedure defines the logic to be exectued and schemas for input and output
- AbstractProcedure defines two attributes that can be used to classify the Procedure
category and extra (could maybe be changed to a single attribute called meta)
- To clarify there is no need to have several instances of the same Procedure 

> * All subclasses of ConfiguredObject should contain their own __content_roles class attribute and all_roles property. __content_roles may be empty, but they should be there to remind people of the machinery.

AbstractProcedure does for the moment not have any roles, and I would expect ConfiguredObject all_roles to return an empty list if nothing is defined ?

Each Procedure implementation could however contain roles but most objects would be accessed through the Beamline object.

> * Subclasses of ConfiguredObject should take name as an init parameter, and pass it onto to the superclass __init__ - it is used to set up the tree of HardwareObjects. The name has to be unique within the container, so you cannot call all procedures 'Procedure'.

Sure, missed that.

> * the category is no longer part of ConfiguredObject - we agreed to remove it.
> * All subclasses of ConfiguredObject should have or inherit init() and _init(). This is my fault - I should have put dummy init() and _init() into ConfiguredObject myself - but it should be done with this PR, I think.

The category is not meant to be part of ConfiguredObject but to make it possible to categorize a procedure so that it can be retrieved based on a category. We could also or instead imagine an arbitrary meta data attribute. There is no need for it to be part of ConfiguredObject but its usefll for a Procedure

> * As the system currently works, ConfigfuredObjects are given as individual properties in the parent. There is exactly one 'kappa_motor' within a given container. If you want multiple procedures you must either add them this way one by one and then make a get_procedures' collection that collects them all, or you must extend the basic ConfiguredObject machinery to support collections of unnamed objects.

I'm not sure I understand what you mean, there seems to be some misunderstanding on how this is supposed to be used. As I mentioned above there is only on instance of each Procedure, is that what you meant ?

> * When should we use properties and when functions? I thought we had agreed on generally using functions.
>   For argument_schema and result_schema OK , those are attributes, but another point is raised. Presumably you make them properties to make them read-only (?). But are they immutable? If not maybe the property shoud return a copy?

Exactly, I made all the attributes properties to make them read only. This also applies to things like msg, results and so on. The schemas are defined when the Procedure is implemented and should never be changed during runtime. I don't see the need to return a copy of the schema ?

>   For msg, running, failed, and results would not a function be more consistent, unless we want to use properties generally?

Ok, we can make them methods if that is more consistent, but why are not the attributes on Beamline methods then ?

> * Should running and failed be independent, or should we just return a state? What do we do elsewhere?

They could be combined into something called state

> * start() and stop() do not seem to have any connection with the various _something functions. Is start the same as _execute(), BTW?

_execute is executed by _start that is in itself executed by start. _execute is the method to implement
by the concrete implementation.

> * There should be two sets of functions:
>   Those that start with '_' (underscore) should be called from other functions in AbstractProcedure, and should not be called from anywhere else (except subclasses).
>   Those that you want  to call from other functions should not start with underscore - and should between them describe all the core functionality of the class. Do we have the set of these functions, and what are they?

Start and Stop are the public methods and the rest are protected and of those _execute is the one that is to be implemented by a concrete implementation. Is that what you meant ?

> * I am a bit in doubt about this one:
>   ConfiguredObject takes a yaml configuration file (which allows comments).
>   The schemas and validation use JSON (which, btw do not allow comments).
>   I realise that there are external constraints on this, so maybe we do have to support two different (but similar) file types, even in this close proximity, but has the point been considered?

The schemas would be defined in code with the procedure, I don't really see any need to pass the schemas as configuration. As the schemas are defined in code you can also add comments if you need.

Ill update the PR with the fixe to __init__

Cheers,
Marcus

marcus-oscarsson:
@rhfogh 

I've made a commit with a few updates fixing the issues related to ConfiguredObject. We should really look at _init, init _content_roles and all_roles for ConfiguredObject. It should not be necessary to implement the first two nor define the last two if they are not needed. Inheriting from ConfiguredObject should be enough to as you say "remind people of the machinery", I don't understand  why one would need to define an extra line of code just to be used as a reminder of a functionality provided by a subclass ?f

I also removed the extra attribute and added _post_execute (that was previously removed)

Cheers,
Marcus

rhfogh:
@marcus-oscarsson 

Thanks for the changes, I like this.
I cannot figure out how to interleave answers, so I make a separate one.

- The one important point is this one:
It is not that there could be more than one instance of each procedure, but that each must have a unique name, that name must be passed up to ConfiguredObject.\_\_init\_\_ (see comments to the code for more details). Further, that name must be defined as a property in the container, Beamline in this case, so that the procedures you can add are predefined (that might be one reason to have site-specific BeamlineObjects, BTW). Either that, or we need to define procedures as a special case, if we do not want to specify all the procedure names in the container.

- You are quite right that we do not need empty init and _init() functions here. They need to be in the ConfiguredObject class (where I should have put them myself in the first place) so that there will always be an _init() to call. 

- As for __content_roles and all_roles, empty ones are clearly not needed in the sense that the code will work the same without them. The question is whether it is more pedagogical to put them there so people who subclass AbstractProcedure can see all the machinery directly, or at least to put a comment mentioning them, or whether it is enough (as well as less cluttered, of course) to rely on programmers checking the superclass of the superclass when they add a new Procedure. But either would do.

- The category is fine, I was just (wrongly) suspicious that it might have been a hold-over from an earlier ConfiguredObject version

- I had missed the call to _excecute from _start, hence my comment. As they are now I fully agree with the _functions.

- Functions v. properties? Well, we are not going to be fully consistent either way, as long as BeamlineObject uses properties - and I think people like them better in that contect, I certainly do. Personally I prefer properties for anything that can be called an attribute, but you made a convincing argument, once, that if you set a property and misspell the name you get no warning, so you preferred functions. I think we should have a clear (if slightly arbitrary) policy on when to use one and when to use the other, and I could not say why result and msg fall on the property side. If you can state such a policy I shall most likely agree.

- The point about returning a copy of the schema is that if the schema is mutable, people could modify it. Thereby modifying the schema for the entire class. If you used a function to gert the schema it would definitely have to return a  copy, but as a property it may be enough to leave the hole open and rely on people understanding that as 'the' schema that is inded a 'property' of the class.

- The point about yaml v. json is simply whether it could be a little confusing to use two similar but different formats in so close proximity. We need yaml for the configuration, and quite likely we need  JSON for the schemas for other reasons, so (as I suggested) this one probably cannot be avoided.

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -0,0 +1,248 @@
...
+    _ALLOW_PARALLEL = False
+
+    # JSONSchema used for input and output validation and possibly
+    # form generation.
+    # https://json-schema.org/
+    # Needs to be defined by each subclass
+    _ARG_SCHEMA = None
+    _RESULT_SCHEMA = None
+
+    def __init__(self):
```
rhfogh: Should be 
def \_\_init\_\_(self, name):

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -0,0 +1,248 @@
...
+        self._ready_event = gevent.event.Event()
+        self._task=None
+        self._state = ProcedureState.IDLE
+
+        # YML configuration options
+        # Category that the Procedure belongs to, configurable through
+        # YAML file and used by for listing and displaying the procedure
+        # in the right context.
+        self.category=""
+        self.name="Procedure"
```
rhfogh: This line musty be removed, the name is set elsewhere (and must anyway be unique).

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def _init(self):
```
rhfogh: The two init() functions can go - they should be defined (and empty) in the superclass instead.

marcus-oscarsson:
@rhfogh 


> @marcus-oscarsson
> 
> Thanks for the changes, I like this.
> I cannot figure out how to interleave answers, so I make a separate one.

You can quote a reply by pressing the little ... icon in the top right corner,

> 
> * The one important point is this one:
>   It is not that there could be more than one instance of each procedure, but that each must have a unique name, that name must be passed up to ConfiguredObject.__init__ (see comments to the code for more details). Further, that name must be defined as a property in the container, Beamline in this case, so that the procedures you can add are predefined (that might be one reason to have site-specific BeamlineObjects, BTW). Either that, or we need to define procedures as a special case, if we do not want to specify all the procedure names in the container.

Ok, it was maybe not clear from the beginning but this is why I said that we need a small addition to the Beamline object. We need to add a mechansim that makes it possible to register a set of Procedures from configuration, so that there is no need to predefine or hardcode there names in the BeamlineObject.

For instance:

```
procedures:
    - quick-realign: qucikrealign.xml
    - procedure-a: procedure-a.xml
```

> * You are quite right that we do not need empty init and _init() functions here. They need to be in the ConfiguredObject class (where I should have put them myself in the first place) so that there will always be an _init() to call.

Ok

> * As for __content_roles and all_roles, empty ones are clearly not needed in the sense that the code will work the same without them. The question is whether it is more pedagogical to put them there so people who subclass AbstractProcedure can see all the machinery directly, or at least to put a comment mentioning them, or whether it is enough (as well as less cluttered, of course) to rely on programmers checking the superclass of the superclass when they add a new Procedure. But either would do.

Hm, I'm not sure about this one, we can maybe discuss this on our next meeting. There are plenty of things one could add for pedagogical reasons but I don't think that's what we normally do when we implement something.

> * The category is fine, I was just (wrongly) suspicious that it might have been a hold-over from an earlier ConfiguredObject version

Ok

> * I had missed the call to _excecute from _start, hence my comment. As they are now I fully agree with the _functions.

Ok

> * Functions v. properties? Well, we are not going to be fully consistent either way, as long as BeamlineObject uses properties - and I think people like them better in that contect, I certainly do. Personally I prefer properties for anything that can be called an attribute, but you made a convincing argument, once, that if you set a property and misspell the name you get no warning, so you preferred functions. I think we should have a clear (if slightly arbitrary) policy on when to use one and when to use the other, and I could not say why result and msg fall on the property side. If you can state such a policy I shall most likely agree.

Not sure that it was me who said that. In this case I made msg and results a property so that they are not changed by mistake, a bit far fetched maybe but it might happen by accident. I have no general
policy for when to use properties or attributes. 

> * The point about returning a copy of the schema is that if the schema is mutable, people could modify it. Thereby modifying the schema for the entire class. If you used a function to gert the schema it would definitely have to return a  copy, but as a property it may be enough to leave the hole open and rely on people understanding that as 'the' schema that is inded a 'property' of the class.

If the schema is mutable it would not matter if was returned by a property or a function, it would still be mutable. As it is right now the schema is normal python dict but one could maybe make it a immutable dict if really necessary.

> * The point about yaml v. json is simply whether it could be a little confusing to use two similar but different formats in so close proximity. We need yaml for the configuration, and quite likely we need  JSON for the schemas for other reasons, so (as I suggested) this one probably cannot be avoided.

Ok

Thanks for the input @rhfogh

rhfogh:
> @rhfogh
> 
> > @marcus-oscarsson
> > Thanks for the changes, I like this.
> > I cannot figure out how to interleave answers, so I make a separate one.
> 
> You can quote a reply by pressing the little ... icon in the top right corner,
> 
> > * The one important point is this one:
> >   It is not that there could be more than one instance of each procedure, but that each must have a unique name, that name must be passed up to ConfiguredObject.**init** (see comments to the code for more details). Further, that name must be defined as a property in the container, Beamline in this case, so that the procedures you can add are predefined (that might be one reason to have site-specific BeamlineObjects, BTW). Either that, or we need to define procedures as a special case, if we do not want to specify all the procedure names in the container.
> 
> Ok, it was maybe not clear from the beginning but this is why I said that we need a small addition to the Beamline object. We need to add a mechansim that makes it possible to register a set of Procedures from configuration, so that there is no need to predefine or hardcode there names in the BeamlineObject.
> 
> For instance:
> 
> ```
> procedures:
>     - quick-realign: qucikrealign.xml
>     - procedure-a: procedure-a.xml
> ```

I agree about that. We just need to think about how to do it. Also, would we want a core shared set of procedure names that people can then add to? If so we could actually do it with the current system and minor tweaks - such as bringing categories back and having a category called 'procedure' ;-) Or do we want everybody to add their own independent set of procedures with no correlation between sites?

> > * You are quite right that we do not need empty init and _init() functions here. They need to be in the ConfiguredObject class (where I should have put them myself in the first place) so that there will always be an _init() to call.
> 
> Ok
> 
> > * As for __content_roles and all_roles, empty ones are clearly not needed in the sense that the code will work the same without them. The question is whether it is more pedagogical to put them there so people who subclass AbstractProcedure can see all the machinery directly, or at least to put a comment mentioning them, or whether it is enough (as well as less cluttered, of course) to rely on programmers checking the superclass of the superclass when they add a new Procedure. But either would do.
> 
> Hm, I'm not sure about this one, we can maybe discuss this on our next meeting. There are plenty of things one could add for pedagogical reasons but I don't think that's what we normally do when we implement something.
> 
> > * The category is fine, I was just (wrongly) suspicious that it might have been a hold-over from an earlier ConfiguredObject version
> 
> Ok
> 
> > * I had missed the call to _excecute from _start, hence my comment. As they are now I fully agree with the _functions.
> 
> Ok
> 
> > * Functions v. properties? Well, we are not going to be fully consistent either way, as long as BeamlineObject uses properties - and I think people like them better in that contect, I certainly do. Personally I prefer properties for anything that can be called an attribute, but you made a convincing argument, once, that if you set a property and misspell the name you get no warning, so you preferred functions. I think we should have a clear (if slightly arbitrary) policy on when to use one and when to use the other, and I could not say why result and msg fall on the property side. If you can state such a policy I shall most likely agree.
> 
> Not sure that it was me who said that. In this case I made msg and results a property so that they are not changed by mistake, a bit far fetched maybe but it might happen by accident. I have no general
> policy for when to use properties or attributes.
> 
> > * The point about returning a copy of the schema is that if the schema is mutable, people could modify it. Thereby modifying the schema for the entire class. If you used a function to gert the schema it would definitely have to return a  copy, but as a property it may be enough to leave the hole open and rely on people understanding that as 'the' schema that is inded a 'property' of the class.
> 
> If the schema is mutable it would not matter if was returned by a property or a function, it would still be mutable. As it is right now the schema is normal python dict but one could maybe make it a immutable dict if really necessary.

If we wanted to be rigid about it we would talk about encapsulation - it should be *impossible* to change object internals (inclusing the content of the Schema)  except by an explicit change command. And that includes not making the mutable internal collections accessible to the outer world at all. 
More realistically it is a matter of the principle of least surprise. If I get hold of myProcedure.Schema, it is not unreasonable to expect that modifying it might modify the schema in place. If you get hold of a schema dictionary via a function call, including a call like myProcedure.get_schema() you might reasonably expect that you could modify it without causing side effects. Aftrer all, how do you know where it came from originally?
 
> > * The point about yaml v. json is simply whether it could be a little confusing to use two similar but different formats in so close proximity. We need yaml for the configuration, and quite likely we need  JSON for the schemas for other reasons, so (as I suggested) this one probably cannot be avoided.
> 
> Ok
> 
> Thanks for the input @rhfogh

marcus-oscarsson:
> > @rhfogh
> > > @marcus-oscarsson
> > > Thanks for the changes, I like this.
> > > I cannot figure out how to interleave answers, so I make a separate one.
> > 
> > 
> > You can quote a reply by pressing the little ... icon in the top right corner,
> > > * The one important point is this one:
> > >   It is not that there could be more than one instance of each procedure, but that each must have a unique name, that name must be passed up to ConfiguredObject.**init** (see comments to the code for more details). Further, that name must be defined as a property in the container, Beamline in this case, so that the procedures you can add are predefined (that might be one reason to have site-specific BeamlineObjects, BTW). Either that, or we need to define procedures as a special case, if we do not want to specify all the procedure names in the container.
> > 
> > 
> > Ok, it was maybe not clear from the beginning but this is why I said that we need a small addition to the Beamline object. We need to add a mechansim that makes it possible to register a set of Procedures from configuration, so that there is no need to predefine or hardcode there names in the BeamlineObject.
> > For instance:
> > ```
> > procedures:
> >     - quick-realign: qucikrealign.xml
> >     - procedure-a: procedure-a.xml
> > ```
> 
> I agree about that. We just need to think about how to do it. Also, would we want a core shared set of procedure names that people can then add to? If so we could actually do it with the current system and minor tweaks - such as bringing categories back and having a category called 'procedure' ;-) Or do we want everybody to add their own independent set of procedures with no correlation between sites?

I don't think bringing category back for this reason is a good idea. I think its enough to create a mechanism that makes it possible  to register Procedures as I just described above. There might be a set of shared Procedures but its very likely that the implementations is going to be different from site to site. I would leave it up to each site to name these procedures as they see fit. 

> 
> > > * You are quite right that we do not need empty init and _init() functions here. They need to be in the ConfiguredObject class (where I should have put them myself in the first place) so that there will always be an _init() to call.
> > 
> > 
> > Ok
> > > * As for __content_roles and all_roles, empty ones are clearly not needed in the sense that the code will work the same without them. The question is whether it is more pedagogical to put them there so people who subclass AbstractProcedure can see all the machinery directly, or at least to put a comment mentioning them, or whether it is enough (as well as less cluttered, of course) to rely on programmers checking the superclass of the superclass when they add a new Procedure. But either would do.
> > 
> > 
> > Hm, I'm not sure about this one, we can maybe discuss this on our next meeting. There are plenty of things one could add for pedagogical reasons but I don't think that's what we normally do when we implement something.
> > > * The category is fine, I was just (wrongly) suspicious that it might have been a hold-over from an earlier ConfiguredObject version
> > 
> > 
> > Ok
> > > * I had missed the call to _excecute from _start, hence my comment. As they are now I fully agree with the _functions.
> > 
> > 
> > Ok
> > > * Functions v. properties? Well, we are not going to be fully consistent either way, as long as BeamlineObject uses properties - and I think people like them better in that contect, I certainly do. Personally I prefer properties for anything that can be called an attribute, but you made a convincing argument, once, that if you set a property and misspell the name you get no warning, so you preferred functions. I think we should have a clear (if slightly arbitrary) policy on when to use one and when to use the other, and I could not say why result and msg fall on the property side. If you can state such a policy I shall most likely agree.
> > 
> > 
> > Not sure that it was me who said that. In this case I made msg and results a property so that they are not changed by mistake, a bit far fetched maybe but it might happen by accident. I have no general
> > policy for when to use properties or attributes.
> > > * The point about returning a copy of the schema is that if the schema is mutable, people could modify it. Thereby modifying the schema for the entire class. If you used a function to gert the schema it would definitely have to return a  copy, but as a property it may be enough to leave the hole open and rely on people understanding that as 'the' schema that is inded a 'property' of the class.
> > 
> > 
> > If the schema is mutable it would not matter if was returned by a property or a function, it would still be mutable. As it is right now the schema is normal python dict but one could maybe make it a immutable dict if really necessary.
> 
> If we wanted to be rigid about it we would talk about encapsulation - it should be _impossible_ to change object internals (inclusing the content of the Schema) except by an explicit change command. And that includes not making the mutable internal collections accessible to the outer world at all.
> More realistically it is a matter of the principle of least surprise. If I get hold of myProcedure.Schema, it is not unreasonable to expect that modifying it might modify the schema in place. If you get hold of a schema dictionary via a function call, including a call like myProcedure.get_schema() you might reasonably expect that you could modify it without causing side effects. Aftrer all, how do you know where it came from originally?
> 

I guess it boils down to a discussion on how one would ideally would like to design software, in reality the concepts you mention tends to be imposed by the language we use. If I in C return a pointer to a struct I know that manipulating it will modify that struct. Its maybe less transparent in C because you have to de-reference the pointer, so it becomes more explicit.

For this reason, and I guess that's what you are also saying, its probably better to use a property since its makes it more evident that the data returned actually belongs to the object in question.

Anything that is considered to be a unusual behavior for the language and paradigm used should in be documented to avoid confusion.

> > > * The point about yaml v. json is simply whether it could be a little confusing to use two similar but different formats in so close proximity. We need yaml for the configuration, and quite likely we need  JSON for the schemas for other reasons, so (as I suggested) this one probably cannot be avoided.
> > 
> > 
> > Ok
> > Thanks for the input @rhfogh

Cheers
Marcus

marcus-oscarsson:
@rhfogh 

The procedure concept provides means to implement logic that performs a sequence of actions involving basically any instrument or component available on the beamline. A procedure could for
instance be realign, aneal or a custom action needed on a beamline. It's been used to conveniently execute these "procedures" that are normally triggered manually by the user and not part of the normal data collection routines. There are other "procedures" like centring that could with advantage also be implemented as a procedure.

The parameters are in the proposed implementation is defined by a JSONSchema and there are currently only one parameter `data_model` but one could of course imagine n of those or that data_model in that case is contains a list. This gives us a very easy way to provide procedures with generic interfaces to the application.

The procedures as we have discussed them would be executed and available through the Beamline object. The reason that there is no addition to Beamline in this very PR is that it's based on commits before the Beamline object was introduced. I think the idea was to make one PR for the AbstractProcedure itself and then another one for making it available through Beamline (alternatively rebase this one). This is one of the complex situations we can end up in while doing parallel work. 

> I was wondering whether there might be an alternative approach.

What is the alternative that you are thinking about ?

We have already been thinking about synergy between the queue entries and the procedures, something that would be good if possible (which I believe it is), we have to investigate exactly how that can be done when we look at the Queue.

> Could we say, that centring, acquisition, sample changing, and in general all the core procedures you might want to excecute from a workflow 

This is maybe a completely different discussion, the Procedure mechanism is a step in that direction and it would be good if we could share as much as possible. I think its already great if we can get the mechanism in place and then we can look at exactly what to call core procedures.

I will try to get the time to make some addition to Beamline before the Berlin meeting.

PR #417, opened 2019-09-25
Beamline Object - Updates

marcus-oscarsson:
Dear all,

So me and Antonia spent bigger part of Monday and Tuesday to get the Beamline object to work. We have finally arrived at something that at least superficially seems to work like before :)

When we did this we had to adapt a few things to make it work for us, roughly what we did:

There were a few settings missing that we added:
- available_methods 
- click_centrings_num_click

We also added access to the following HardwareObjects:
- Beamstop
- Capilary
- Cryostream
- SampleChangerMaintnance
- Workflow
- XMLRPCServer

The last one should eventually become a part of the SampleChanger HardwareObject in our opinions.

We renamed Graphics to Microscope as we briefly discussed previously. Microscope contains Shapes (Regions of interest), Camera, Focus, Zoom and Lights. It further contains utility methods related to the microscope view like taking snapshots. Its in this way not very different from the Qt4GraphicsManager object except for that certain things are delegated to shapes directly.

We did the corresponding updates to the rest of the site specific parts of the code as well. We did not change all the Graphics to Microscope calls entirely since it was difficult to know what parts could be delegated to a corresponding Shapes object for the Qt version.

We have added AbstractMicroscope class that needs to be completed with the necessary methods. We have no strong opinions about the name Microscope so this can be changed if someone has a more appropriate suggestion.

We also moved get_default_characterisation_parameters to DataAnalysis, these are the default parameters for EDNA. The Bemline object was missing this but it was moved from the BeamlineSetup object, that now will become obsolete. 

Finally, detector_distance was moved from Detector to Resolution

Cheers,
Marcus and Antonia

### DIFF
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -590,8 +590,10 @@ class Beamline(ConfiguredObject):
         osc_start = motor_positions.get("phi", params["osc_start"])
         acq_parameters.osc_start = round(float(osc_start), 2)
         kappa = motor_positions.get("kappa", 0.0)
+        kappa = kappa if kappa else 0.0
         acq_parameters.kappa = round(float(kappa), 2)
         kappa_phi = motor_positions.get("kappa_phi", 0.0)
+        kappa_phi = kappa_phi if kappa_phi else 0.0
         acq_parameters.kappa_phi = round(float(kappa_phi), 2)
 
         try:
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -197,6 +197,16 @@ class Beamline(ConfiguredObject):
     __content_roles.append("beam")
 
     @property
+    def beam_definer(self):
+        """Beam-definer Hardware object
+
+        Returns:
+            Optional[HardwareObject]:
+        """
+        return self._objects.get("beam_definer")
+    __content_roles.append("beam_definer")
+
+    @property
     def hutch_interlock(self):
         """Hutch Interlock Hardware object
 
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -299,6 +299,17 @@ class Beamline(ConfiguredObject):
     __content_roles.append("sample_changer")
 
     @property
+    def  sample_changer_maintenance(self):
+        """Sample Changer Maintnance Hardware object
+
+        Returns:
+            Optional[AbstractMaintnanceSampleChanger]:
+        """
+        return self._objects.get("sample_changer_maintenance")
+
+    __content_roles.append("sample_changer_maintenance")
+
+    @property
     def plate_manipulator(self):
         """Plate Manuipulator Hardware object
         NBNB TODO REMOVE THIS and treat as an alternative sample changer instead.
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -83,6 +83,12 @@ class Beamline(ConfiguredObject):
         # List[str] of advanced method names
         self.advanced_methods = []
 
+        # List[str] of available methods
+        self.available_methods = []
+
+        # int number of clicks used for click centring
+        self.click_centring_num_clicks = 3
+
         # bool Is wavelength tunable
         self.tunable_wavelength = False
 
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -327,6 +327,17 @@ class Beamline(ConfiguredObject):
     __content_roles.append("session")
 
     @property
+    def cryo(self):
+        """Cryostream Hardware object.
+
+        Returns:
+            Optional[Cryo]:
+        """
+        return self._objects.get("cryo")
+
+    __content_roles.append("cryo")
+
+    @property
     def lims(self):
         """LIMS client object.
 
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -329,7 +329,7 @@ class Beamline(ConfiguredObject):
         """Microscope object. Includes defined shapes.
 
         Returns:
-            Optional[AbstractMmicroscope]:
+            Optional[AbstractMicroscope]:
         """
         return self._objects.get("microscope")
 
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -359,7 +359,6 @@ class Beamline(ConfiguredObject):
 
     __content_roles.append("lims")
 
- 
     @property
     def microscope(self):
         """Microscope object. Includes defined shapes.
@@ -563,7 +562,6 @@ class Beamline(ConfiguredObject):
     # NB Objects need not be HardwareObjects
     # We still categorise them as'hardware' if they are not procedures, though
     # The attribute values will be given in the config.yml file
-
     def get_default_acquisition_parameters(self, acquisition_type="default"):
         """
         :returns: A AcquisitionParameters object with all default parameters for the
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -429,6 +429,28 @@ class Beamline(ConfiguredObject):
     __content_roles.append("imaging")
 
     @property
+    def xml_rpc_server(self):
+        """XMLRPCServer for RPC
+
+        Returns:
+            Optional[XMLRPCServer]:
+        """
+        return self._objects.get("xml_rpc_server")
+
+    __content_roles.append("xml_rpc_server")
+
+    @property
+    def workflow(self):
+        """Standarad EDNA workflow procedure.
+
+        Returns:
+            Optional[Workflow]:
+        """
+        return self._objects.get("workflow")
+
+    __content_roles.append("workflow")
+
+    @property
     def gphl_workflow(self):
         """Global phasing data collection workflow procedure.
 
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -307,16 +307,17 @@ class Beamline(ConfiguredObject):
 
     __content_roles.append("lims")
 
+ 
     @property
-    def graphics(self):
-        """Graphics/OAV object. Includes defined shapes.
+    def microscope(self):
+        """Microscope object. Includes defined shapes.
 
         Returns:
-            Optional[AbstractGraphics]:
+            Optional[AbstractMmicroscope]:
         """
-        return self._objects.get("graphics")
+        return self._objects.get("microscope")
 
-    __content_roles.append("graphics")
+    __content_roles.append("microscope")
 
     @property
     def queue_manager(self):
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -223,6 +223,30 @@ class Beamline(ConfiguredObject):
 
     __content_roles.append("fast_shutter")
 
+
+    @property
+    def beamstop(self):
+        """Beamstop Hardware object
+
+        Returns:
+            Optional[AbstractNState]:
+        """
+        return self._objects.get("beamstop")
+
+    __content_roles.append("beamstop")
+
+    @property
+    def capillary(self):
+        """Capillary Hardware object
+
+        Returns:
+            Optional[AbstractNState]:
+        """
+        return self._objects.get("capillary")
+
+    __content_roles.append("capillary")
+
+
     @property
     def diffractometer(self):
         """Diffractometer Hardware object
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -197,16 +197,6 @@ class Beamline(ConfiguredObject):
     __content_roles.append("beam")
 
     @property
-    def beam_definer(self):
-        """Beam-definer Hardware object
-
-        Returns:
-            Optional[HardwareObject]:
-        """
-        return self._objects.get("beam_definer")
-    __content_roles.append("beam_definer")
-
-    @property
     def hutch_interlock(self):
         """Hutch Interlock Hardware object
 
@@ -238,30 +228,6 @@ class Beamline(ConfiguredObject):
         return self._objects.get("fast_shutter")
 
     __content_roles.append("fast_shutter")
-
-
-    @property
-    def beamstop(self):
-        """Beamstop Hardware object
-
-        Returns:
-            Optional[AbstractNState]:
-        """
-        return self._objects.get("beamstop")
-
-    __content_roles.append("beamstop")
-
-    @property
-    def capillary(self):
-        """Capillary Hardware object
-
-        Returns:
-            Optional[AbstractNState]:
-        """
-        return self._objects.get("capillary")
-
-    __content_roles.append("capillary")
-
 
     @property
     def diffractometer(self):
@@ -309,7 +275,7 @@ class Beamline(ConfiguredObject):
     __content_roles.append("sample_changer")
 
     @property
-    def  sample_changer_maintenance(self):
+    def sample_changer_maintenance(self):
         """Sample Changer Maintnance Hardware object
 
         Returns:
@@ -346,17 +312,6 @@ class Beamline(ConfiguredObject):
         return self._objects.get("session")
 
     __content_roles.append("session")
-
-    @property
-    def cryo(self):
-        """Cryostream Hardware object.
-
-        Returns:
-            Optional[Cryo]:
-        """
-        return self._objects.get("cryo")
-
-    __content_roles.append("cryo")
 
     @property
     def lims(self):
--- a/HardwareObjects/EMBL/EMBLDoorInterlock.py
+++ b/HardwareObjects/EMBL/EMBLDoorInterlock.py
@@ -165,7 +165,7 @@ class EMBLDoorInterlock(Device):
            as long as it is a one char
         """
         if HWR.beamline.diffractometer is not None:
-            detector_distance = HWR.beamline.detector.detector_distance
+            detector_distance = HWR.beamline.resolution.detector_distance
             if HWR.beamline.diffractometer.in_plate_mode():
                 if detector_distance  is not None:
                     if detector_distance .getPosition() < 780:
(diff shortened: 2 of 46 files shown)


marcus-oscarsson:
So with this done we had a few things that could be interesting to discuss,

- How do we handle an arbitrary amount of sample hangers in an elegant way.
- Same as above but for workflow engines.
- Would it be a good idea to pass the Beamline Object as a member to each HardwareObject
  so that you do not have to perform the global import ?
- It would be useful if we could register the procedures to the Beamline object so a mechanism on the Beamline object for managing (list, execute, stop, state) procedures would perhaps be good ?

Cheers
Marcus and Antonia

IvarsKarpics:
Thanks for the work @marcus-oscarsson  and @beteva !
I did try the new beamline object on the qt graphics and with small adjustments it was running (thank you @rhfogh for the gigantic work!) Some thoughts from my side:
* I would suggest to keep a single sample changer per mxcube session. I do not know if there will be a need for two sample changers that operate in parallel. It would be enough to define - sample_changer: path_to_xml and gui would do the rest. I made a quick fix for the qt TreeBrick and it is durable.
* Maybe for workflows we need a more generic container.
* Would not the passing the beamline object create a circular refs? 
* I am not in favor of name "microscope". For me thats a device rather than a concept that represents what we need. What was wrong with graphics?

Thanks!
Ivars

rhfogh:
I shall look more closely at the code tomorrow, but a few reactions just to the comments, for now. 

- Microscope is fine.

- Presumably you mean that it is the 'SampleChangerMaintenance' that should eventually be part of the SampleChanger.

- Do Beamstop, Capillary, and Cryostream really deserve their very own top-level objecst? Might it not be better to make them part of the Diffractometer?

- "Finally, detector_distance was moved from Detector to Resolution"
  I would argue quite strongly against that. First because Resolution is not an object, it is notional. Second because something called detector_distance surely belongs to something called detector, not somwehere else. If anything, the Resolution could be part of the Detector? Third because detector_distance has exactly the same relation to Resolution as wavelength. Why make one part of Resolution and not the other?

rhfogh:
On the discussion topics, again quickly:

- Passing the beamline around does nto strike me as a good idea. Most of all, what do you actually gain? Second, I think it is important that there is a clear, single access point for Beammline and its contents. We do not want the access to depend on already having a HardwareObject to hand, so we need another access anyway. Having multiple access mechanisms is a recipe for confusion, even if you manage to ensure that they all point ot the same object in the end. And, as Ivars points out, we need to deal with the circularity of Beamline containg HardwareObjwects, which contain the beamline.

- Do we really need a way to handle 'arbitrary amounts of workflow engines'? How many are we ever going to get? And, unlike SampleChangers, several of them can be active and ready to go at the same time. If someone can come up with an AbstractWorkflowEngine that might be useful - though I think there would be very little commonality one could exploit and the user interfaces would all be unique - but what would be the point of accessing 'a workflow' and calling functions on it without considering which workflow you call?

- Will there be individual procedure objects - if so stop, state, and probably execute probably belong there? If not, where should the code for the procedures live?

marcus-oscarsson:
@IvarsKarpics 

> Thanks for the work @marcus-oscarsson and @beteva !
> I did try the new beamline object on the qt graphics and with small adjustments it was running (thank you @rhfogh for the gigantic work!) Some thoughts from my side:
> 
> * I would suggest to keep a single sample changer per mxcube session. I do not know if there will >be a need for two sample changers that operate in parallel. It would be enough to define - >sample_changer: path_to_xml and gui would do the rest. I made a quick fix for the qt TreeBrick and >it is durable.

You are right that there are probably no need to operate two sample changers simultaneously. It would be nice is if we could remove the separate attribute for the plate manipulator. So that the rest is handled by checking head type or what kind of sample changer that is used.

> * Maybe for workflows we need a more generic container.

Its at least worth discussing it.

> * Would not the passing the beamline object create a circular refs?

No, I don't think so, its just passing at as a argument to __init__ that then adds it
as an attribute.

> * I am not in favor of name "microscope". For me thats a device rather than a concept that represents what we need. What was wrong with graphics?

Well we tried to find something that really reflects what it is MicroscopeView would perhaps be better then. Zoom, Focus and Light does not seem to belong to something called Graphics that's one of
the reasons that we wanted a different name.

> 
> Thanks!
> Ivars

marcus-oscarsson:
> On the discussion topics, again quickly:
> 
> * Passing the beamline around does not strike me as a good idea. Most of all, what do you actually gain? Second, I think it is important that there is a clear, single access point for Beammline and its contents. We do not want the access to depend on already having a HardwareObject to hand, so we need another access anyway. Having multiple access mechanisms is a recipe for confusion, even if you manage to ensure that they all point ot the same object in the end. And, as Ivars points out, we need to deal with the circularity of Beamline containg HardwareObjwects, which contain the beamline.

There would still be a clear single point of access to the various objects, through the Beamline object. One would gain that each HardwareObject would be more self contained, as it is know we make them dependent on the global Beamline, which I guess is fine in our context. It was in anyways just airing an idea. It seems like neither you nor Ivars thinks its a good idea so, we can drop it :)

> * Do we really need a way to handle 'arbitrary amounts of workflow engines'? How many are we ever going to get? And, unlike SampleChangers, several of them can be active and ready to go at the same time. If someone can come up with an AbstractWorkflowEngine that might be useful - though I think there would be very little commonality one could exploit and the user interfaces would all be unique - but what would be the point of accessing 'a workflow' and calling functions on it without considering which workflow you call?

Hm, I see well I just wanted to avoid adding workflow_one, workflow_two ... to the Beamline object but we can maybe keep the hard coded like this, if you don't think its any use.

> * Will there be individual procedure objects - if so stop, state, and probably execute probably belong there? If not, where should the code for the procedures live?

Well my idea was that you we write procedure objects inheriting AbstractProcedure. Each procedure object has a well defined schema for its arguments and return values. AbstractProcedure also defines the common events emitted like started, progress, failure and ended. We can add which procedures we are currently supporting to the YML file. That would register them to the Beamline object that can provide a list of which procedures that exists and a way to run them.

You are right that each Procedure also have to defines the logic for what Stop and Execute would, as defined by AbstractProcedure.

Like this the UI can list which procedures exists and an easy way to display a generic Dialog. We are actually doing something like this in MXCuBE3 and I would really like to generalize it and make it a part of HardwareRepository.

rhfogh:
@marcus-oscarsson 

- I do see your point about workflow_1, workflow_2 etc. I do not think there is that much to be gained, unless we get to worrkflow_3 at some point, but if somebody thinks it is worth doing I would certainly not object. 

- We would definitely need a way to list which procedures are available / running from the Beamline. Depending on how it is organised one might also want a function on the beamline to start a procedure executing, yes - maybe called launch rather than execute. I would vote for having state, stop, and execute in the Procedure object and not on the beamline - but one way or the other I look forward to seeing a proposal.

marcus-oscarsson:
@rhfogh 

Forgot these ones sorry

> * I have no problems with Microscope. It is not ideal, but neither is anything else. Viewer?? SampleViewer?? 

Why not ...

> * Presumably you mean that it is the 'SampleChangerMaintenance' that should eventually be part of the SampleChanger.

Yes that's what I meant to say

> * Do Beamstop, Capillary, and Cryostream really deserve their very own top-level objecst? Might it not be better to make them part of the Diffractometer?

This can of course be discussed I don't think we have any real strong opinion about any of them, more than that they do not necessarily have to belong to the diffractomter.

> * "Finally, detector_distance was moved from Detector to Resolution"
>   I would argue quite strongly against that. First because Resolution is not an object, it is notional. >Second because something called detector_distance surely belongs to something called detector, >not somwehere else. If anything, the Resolution could be part of the Detector? Third because >detector_distance has exactly the same relation to Resolution as wavelength. Why make one part of >Resolution and not the other?

Hm well the detector distance does not really belong more to the detector than to the beamline or resolution either for that matter. I think any one of those three would work and are logical depending on how one look at it. Actually everybody here was of different opinion  so we moved it to where its most frequently used.

rhfogh:
@marcus-oscarsson 
On detector distances. I could argue the 'does not belong to the detector' - does the beamstop position not belong to the beamstop - it seems a parallel case? But the clinching argument, I think, is that if you see the beamline as a collection of objects or assemblies, detector_distance would obviously seem to belong to the detector. Alsom from a purely semantic point of view the obvious way to address something called 'detector_distance' it is 'detector.distance'. Only if you see the beamline as a collection of abstract concepts (and if you routinely think in terms of 'resolution' as something divorced from the actual objects on the beamline)  does it make sense to put the parameter under 'Resolution'. And at that point you might as well put it under MxCollect - it is a data collection parameter after all.

HardwareObjects/Beamline.py
```diff
@@ -190,6 +196,16 @@ def beam(self):
 
     __content_roles.append("beam")
 
+    @property
```
rhfogh: What is a beam_definer? Can there be more than one, as for slits? What is its class - AbstractMotor?? Should it not be under Beam rather than under Beamline?

HardwareObjects/Beamline.py
```diff
@@ -296,6 +347,17 @@ def session(self):
 
     __content_roles.append("session")
 
+    @property
```
rhfogh: SHould taht be N-state? AbstractCryo?, ...

HardwareObjects/Beamline.py
```diff
@@ -308,15 +370,15 @@ def lims(self):
...
 
     @property
-    def graphics(self):
-        """Graphics/OAV object. Includes defined shapes.
+    def microscope(self):
+        """Microscope object. Includes defined shapes.
 
         Returns:
-            Optional[AbstractGraphics]:
+            Optional[AbstractMmicroscope]:
```
rhfogh: Typo

HardwareObjects/Beamline.py
```diff
@@ -517,8 +600,10 @@ def get_default_acquisition_parameters(self, acquisition_type="default"):
         osc_start = motor_positions.get("phi", params["osc_start"])
         acq_parameters.osc_start = round(float(osc_start), 2)
         kappa = motor_positions.get("kappa", 0.0)
+        kappa = kappa if kappa else 0.0
```
rhfogh: Why this? It only makes a difference if 'kappa' is in the motor_positions dictionary with a value that evaluates to False (e.g. None, 0, or "") Is taht something we need to sup[port?

HardwareObjects/EMBL/EMBLDoorInterlock.py
```diff
@@ -165,7 +165,7 @@ def unlock_door_interlock(self):
            as long as it is a one char
         """
         if HWR.beamline.diffractometer is not None:
-            detector_distance = HWR.beamline.detector.detector_distance
+            detector_distance = HWR.beamline.resolution.detector_distance
```
rhfogh: See general discussion

HardwareObjects/abstract/AbstractMicroscope.py
```diff
@@ -0,0 +1,98 @@
...
+    def save_scene_animation(self, filename, duration=1):
+        """ Take snapshots and create an animation.
+        Args:
+            filename (str): Filename.
+            duration (int): Duration time [s].
+        """
+        pass
+
+    @property
+    def camera(self):
```
rhfogh: WHen somethign is defined for access as a property, I think it is *very* important to discourage accessing it in other ways. Therefore the underlying attributes should start with underscore. I would recommend simply '_camera', '_shapes', and leve off the 'hwobj' suffix.

HardwareObjects/abstract/AbstractMicroscope.py
```diff
@@ -0,0 +1,98 @@
...
+
+import abc
+
+
+class AbstractMicroscope(object):
+    """ Abstract Mictoscope Class """
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self):
+        self.camera_hwobj = None
```
rhfogh: Should be '_shapes', '_focus', etc. See comment below.

HardwareObjects/Beamline.py
marcus-oscarsson: Beam definer is something that defines the beam size and shape, is it the same as your beam ?

HardwareObjects/Beamline.py
marcus-oscarsson: I guess it could inherit AbstractNState if you mean that it can be In or Out ?

HardwareObjects/Beamline.py
marcus-oscarsson: Oh, will take care of that one, Thanks !

HardwareObjects/Beamline.py
marcus-oscarsson: Well, yes because `None` is apparently what we get from our current diffractometer implementation when we have the plates installed, that is when no Kappa Goniometer is installed. I have not looked into the details of where it comes from yet but I though it would be good to handle that case.

HardwareObjects/EMBL/EMBLDoorInterlock.py
marcus-oscarsson: Rasmus, I'm not going to get into a long debate on this one. I personally think it should be on the beamline as a whole but I have nothing against having it on either resolution or detector if that makes people happier. You are right one could probably argue the same for other things as well

HardwareObjects/abstract/AbstractMicroscope.py
marcus-oscarsson: Yes, I agree. I will fix that

HardwareObjects/abstract/AbstractMicroscope.py
marcus-oscarsson: yes ofcourse

HardwareObjects/Beamline.py
rhfogh: Cool.

HardwareObjects/Beamline.py
rhfogh: It is not the same as a beam, but it might be located lower down, as in 'Beamline.beam.beamdefiner'. A more important question is whether there is only one of them. I had a vague idea that several different slits and apertures might all be different kind of beamdefiners. In that case we would need something more complex, to handle multiple definers. On the other hand if it is a kind of high level pseudomotor, one might consider using beam.size and beam.shape directly  as public properties, rather than having a separate object for them.

marcus-oscarsson: It seems that you got it about right, the important thing is that we can set the beam size, shape, and retrieve a list of possible sizes and shapes regardless of hardware involved, slits, aperture or other mechanism. So we need some common interface for this.

It could be part of the beam object if, I would not mind. We can start with that and see were it gets us ?

marcus-oscarsson:
So we will try to take care of the various comments sometime next week and update this PR.

Thanks !

HardwareObjects/Beamline.py
rhfogh: Sounds good. And if it looks like a bad fit there, we can see what to do nect.

rhfogh:
@marcus-oscarsson 
Great. Thanks!

marcus-oscarsson:
@rhfogh 

Alright we have taken care of the changes we discussed :)

Cheers,
Antonia and Marcus

HardwareObjects/Resolution.py
```diff
@@ -163,7 +163,7 @@ def dtoxPositionChanged(self, pos):
         self.update_resolution(self.dist2res(pos))
 
     def getLimits(self):
-        low, high = HWR.beamline.detector.detector_distancegetLimits()
+        low, high = HWR.beamline.detector.detector_distance.getLimits()
```
rhfogh: Should this be detector.detector_distance or detector.distance? I would vote for the latter, but do not let it hold up the PR, we can look at that later.

marcus-oscarsson: Yes distance is probably a better name for it, but i think it was detector_distance before right ? In anyways I don't mind changing it to distance

jordiandreu:
> @marcus-oscarsson
> On detector distances. I could argue the 'does not belong to the detector' - does the beamstop position not belong to the beamstop - it seems a parallel case? But the clinching argument, I think, is that if you see the beamline as a collection of objects or assemblies, detector_distance would obviously seem to belong to the detector. Also from a purely semantic point of view the obvious way to address something called 'detector_distance' it is 'detector.distance'. Only if you see the beamline as a collection of abstract concepts (and if you routinely think in terms of 'resolution' as something divorced from the actual objects on the beamline) does it make sense to put the parameter under 'Resolution'. And at that point you might as well put it under MxCollect - it is a data collection parameter after all.
> 
> Byut I am probably getting carried away here. Should we see what other people think to decide the issue?

I would like to point out here that detector_distance is, in fact, detector-sample_distance. This means that this value does depend on values out of the detector properties, for instance, it depends on the alignment between the sample and the detector surface.

In practice, the detector(-sample)_distance is use to calculate the resolution at a given energy (wavelength) or the detector position to achieve a target resolution at a given energy. Thus, maybe it has more sense to think about these two magnitudes (resolution/detector position) the ones to be exposed by some hardware object(s)?

rhfogh:
@jordiandreu 
It looks like there are several people (including myself) who differ on this and feel strongly enough to argue. It sure sounds like there is more than one reasonable solution. Would it make sense to take this up at the next web meeting, and settle it once and for all by consensus?

jordiandreu:
> @jordiandreu
> It looks like there are several people (including myself) who differ on this and feel strongly enough to argue. It sure sounds like there is more than one reasonable solution. Would it make sense to take this up at the next web meeting, and settle it once and for all by consensus?

I think is a good idea...or during the next MXCuBE meeting, as you wish.

marcus-oscarsson:
@jordiandreu Yes I agree with you and that is why I personally wanted to put detector_distance on the beamline as a whole. However in the end it does not really matter that much, at least not to me, to which object detector_distance belongs as long as we have a well defined place for it and it its somewhat logic. 

Its after all not so strange to find detector distance on Detector, it could also have been called detector_sample_distance it would then perhaps felt less strange to people to find it on Beamline.

I think we can keep it on detector for now and reiterate this on the meeting if you still think its pertinent.

marcus-oscarsson:
We can maybe merge this then and have a discussion around the detector distance on our next meeting ?

rhfogh:
I vote for merge, yes.

marcus-oscarsson:
So it seems like we are ready to merge this then :)

rhfogh:
Two weeks passed, much discussion, no objections. Time to merge

PR #419, opened 2019-10-18
Added support for procedures

rhfogh:
With this PR, you can set up and configure procedures in the same way as any other hardware object.
To register the object as a procedure you add its name to
Beamline._procedure_names.

The ConfiguredObject.procedures property will give you a dictionary of all configured procedures - notice that the resulting dictionary is a copy and modifying it will not modify the object internals. 

The same machinery could be used to add procedures to other classes (like DIffractometer?) if desired.

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -26,6 +26,14 @@ class ConfiguredObject(object):
 
         self._objects = OrderedDict((role, None) for role in self.all_roles)
 
+    def _init(self):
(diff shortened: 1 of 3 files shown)


rhfogh:
ALso added no-op init and _init functions to ConfiguredObject

marcus-oscarsson:
@rhfogh Sorry didnt get the time to look at this today. I will take a look on Monday. Thanks !

PR #420, opened 2019-10-18
[WIP] Abstract Actuator - updated

rhfogh:
Replaces the previous AbstractActuator PR.

- The previous AbstractActuator class has been renamed AbstractTwoState, with subclasses modified accordingly. It should be refactored by someone in the longer term, probably based on AbstractNState

- AbstractActuator is now the new supercalass for aall motors and objects with a value. It still needs to be given a set of generally aspplicable states - can anyone help?

- AbstractMockActuator is an abstract mock object that contains all necessary code for a mock actuator except for setting the mock initial values. Transmission Mockup shows how that will look in practice

- ExporterMotorExample is a proposal for  how ExporterMotor could be written with the new superclass, as an example. I  am not competent to rewrite such a hardware-linked class, and I have no way of testing it. Could someone please help out? 

- There is some minor refactoring in BaseHardwareObjects to facilitate use of ConfiguredObject in the future.

### DIFF
--- a/HardwareObjects/abstract/AbstractActuator.py
+++ b/HardwareObjects/abstract/AbstractActuator.py
@@ -1,0 +1,262 @@
+#! /usr/bin/env python
+# encoding: utf-8
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Standard abstract HardwareObject with a'value' (e.g. motor, transmission, ...)
+"""
+
+from __future__ import division, absolute_import
+from __future__ import print_function, unicode_literals
+
+import logging
+from abc import ABCMeta, abstractmethod
+import gevent.event
+
+from HardwareRepository.BaseHardwareObjects import HardwareObject, HardwareObjectNode
+
+# We should make a standard set of states and use that wherever possible
+from somewhere import GeneralState
+
+__credits__ = [" Copyright © 2019 by MXCuBE collaboration. All rights reserved"]
+__license__ = "LGPLv3+"
+__category__ = "General"
+__author__ = "rhfogh"
+__date__ = "20190607"
+
+
+class AbstractActuator(HardwareObject, HardwareObjectNode):
+    """Abstract base class for Abstract actuators"""
+
+    # NB we should separate HardwareObjectNode and HardwareObject for future refactoring
+
+    __metaclass__ = ABCMeta
+
+    # Stete enumeration
+    STATE = GeneralState
+
+    # Tuple of states that count as ready
+    READY_STATES = (STATE.READY,)
+
+    # Default lower, upper limits
+    # Limit handling can be overridden in subclasses
+    _limits = (None, None)
+
+    def __init__(self, name):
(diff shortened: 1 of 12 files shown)


marcus-oscarsson:
Thanks, we will try to take a look at this sometime next week.

HardwareObjects/abstract/AbstractTwoState.py
```diff
@@ -0,0 +1,84 @@
+"""
```
marcus-oscarsson: So this is just a rename from what was previously called AbstractActuator to AbstractTwoState ?

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -1,84 +1,262 @@
+#! /usr/bin/env python
```
marcus-oscarsson: Do we really need an AbstractActutator when we have AbstractMotor ?. I realize that its maybe more correct from a modelling point of view but I'm not sure that its needed ?

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -1,84 +1,262 @@
...
-from warnings import warn
-from HardwareRepository.BaseHardwareObjects import Device
-from HardwareRepository.TaskUtils import task
+from abc import ABCMeta, abstractmethod
+import gevent.event
+
+from HardwareRepository.BaseHardwareObjects import HardwareObject, HardwareObjectNode
+
+# We should make a standard set of states and use that wherever possible
+from somewhere import GeneralState
```
marcus-oscarsson: This line seems a bit "sketchy" , I would preferably that we mark PR's with this kind of code as WIP so that they can
be discussed in order to have a more complete solution for the actual PR.

The base class of an object should in my opinion define its states. If two objects inhering different base classes have the same states it might be a good idea to look at why the bases are different. We could of course share a state definition If it turns out that we have two objects that have exactly the same states but completely different logic.

I think a more common problem is that sub classes often define a different set of states, some in addition and some overlapping. The additional states are of course not a problem but the ones that overlap should be the same as the ones in the super class.

HardwareObjects/abstract/AbstractTwoState.py
rhfogh: See general comment

HardwareObjects/abstract/AbstractActuator.py
rhfogh: Se general comment

HardwareObjects/abstract/AbstractActuator.py
rhfogh: 1) I agree. But then I did mark it as WIP when I made it.

2) My main concern about states is that the number of different enumeration should be as small as possible - but no smaller (as Einstein might have said). Curently there are lots, and it is very unclear to me which ones can or should be merged, and also which of the large number of possible states in each set are actually necessary, or used. I see no independent benefit in having different stets of states for different abstract classes if the two sets of states are, or could reasonably be made, the same.
BTW I think that additional states are not necessarily harmless - they mean that any code that checks on what to do in different states may break, because there are more states than were envisaged. 
Anyway, I think we need a set of states that can be used as widely as possible, and that contains the smallest number of states that is practical. Since I am not competent to write one, I would welcome any proposal along those lines.

rhfogh:
@marcus-oscarsson 
AbstractActuator, AbstractMotor, AbstractTwostate, etc.
The ideas behind this PR is to have a single abstract superclass for any hardware object that has a 'value'. That could be extended to include NState or not, if we should so decide, and would certainly include all motors (replacing AbstractMotor).  As for the name, I originally proposed StandardHardwareObject. That was voted down, and it was proposed that AbstractActuator be used instead. Which of course meant that I had to rename the pre-existing AbstractActuator (which has a different function), to make room. I have deliberately not tried to use this new abstract class in any of the many places where I think it belongs, since refactoring all the motors does not make sense before we have actually agreed on the principle (and I am not sure I am the right person to do it anyway).
What matters is the function and scope of the class. If that is turned down we would have to think again. If we can agree on it, I would be happy to change to whatever names people think are most expressive.

HardwareObjects/abstract/AbstractTwoState.py
marcus-oscarsson: I, see I was just wondering if this was actually the old AbstractActuator, I guess its a YES

rhfogh: It is indeed a YES.

marcus-oscarsson:
Superseded by #436

PR #421, opened 2019-10-19
Bug fixes to make detector.distance and microscope links work

rhfogh:
Bug fixes - changes required to make program run under Qt with recent changes (graphics->microscope and detector.detector_distance -> detector,distance).
Goes together with parallel PR in mxcube

WARNING - access to microscope and microscope.shapes is *inconsistent*!
Functions that live in QtGraphics manager are sometimes addressed through HWR.beamline.microscope, sometimes through HWR.beamline.microscope.shapes. 
This needs REFACTORING, so that the HWO organisation is the same in the Qt and Web branches, and so that the functions can be accessed in a uniform manner for both UIs.

For now the PR contains a TEMPORARY UGLY HACK, setting
QtGraphicsManager.shapes = self.

### DIFF
--- a/configuration/xml-qt/beamline_config.yml
+++ b/configuration/xml-qt/beamline_config.yml
@@ -38,7 +38,7 @@ _objects:
     # NBNB TODO remove plate_manipulater and treat as another smaple changer
     - plate_manipulator: plate-manipulator.xml
     - diffractometer: mini-diff-mockup.xml
-    - graphics: graphics.xml
+    - microscope: graphics.xml
     - lims: lims-client-mockup.xml
     - queue_manager: queue.xml
     - queue_model: queue-model.xml
--- a/HardwareObjects/QtGraphicsManager.py
+++ b/HardwareObjects/QtGraphicsManager.py
@@ -82,6 +82,13 @@ class QtGraphicsManager(HardwareObject):
         """
         HardwareObject.__init__(self, name)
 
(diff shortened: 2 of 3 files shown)


marcus-oscarsson:
I'm fine with this, @IvarsKarpics or one of the other Qt developers should maybe have the final say on this one since this will only effect the Qt version. We can maybe merge it for now as it would not work without this fix and take another look at how this could be solved in a more elegant way ?

configuration/xml-qt/beamline_config.yml
```diff
@@ -38,7 +38,7 @@ _objects:
     # NBNB TODO remove plate_manipulater and treat as another smaple changer
     - plate_manipulator: plate-manipulator.xml
     - diffractometer: mini-diff-mockup.xml
-    - graphics: graphics.xml
+    - microscope: graphics.xml
```
IvarsKarpics: could we keep graphics please?

marcus-oscarsson: I guess we did not find a name that everybody was happy with but we concluded that graphics was not the name to go for. Now we did the change already a few PRs ago so it would be a bit strange to go back on just the name. The problem with shapes would still remain even if the name changes. I would have nothing against changing the name to something else but maybe not to graphics. shapes could perhaps be renamed to graphics so that it becomes microscope.graphics ... instead of microscope.shapes even if they are not doing the exact same thing.

PR #423, opened 2019-10-21
Cleanup201910

rhfogh:
Removed now unneeded classes. Removed references to BeamlineSetup, get_beam_size.

Also other function and function name clean-ups.

NB Site-specific configuration files have *not* been cleansed - each site will have to take care of their own.


marcus-oscarsson:
It looks like the two first commits are somehow the same as the 3rd and 4th ?

HardwareObjects/XMLRPCServer.py
```diff
@@ -472,11 +468,14 @@ def get_cp(self):
         return json_cplist
 
     def beamline_setup_read(self, path):
-        try:
-            return self.beamline_setup_hwobj.read_value(path)
-        except Exception as ex:
-            logging.getLogger("HWR").exception(str(ex))
-            raise
+        raise NotImplementedError(
```
marcus-oscarsson: This has quite big consequences, as the XMLRPCServer is used by workflows and read_value is used to get a few
values like energy and transmission. We need to be able to export the Beamline object in a similair fashion. So this needs to be replaced by a Beamline object compatible version

HardwareObjects/base_queue_entry.py
```diff
@@ -295,9 +294,6 @@ def pre_execute(self):
         """
         msg = "Calling pre_execute on: " + str(self)
         logging.getLogger("queue_exec").info(msg)
-        self.beamline_setup = self.get_queue_controller().getObjectByRole(
```
marcus-oscarsson: Have you double checked so that self.beamline_setup is not used by any other QueueEntry ?

marcus-oscarsson: Made a quick check myself, seems to only be the case for the EDNA inputs and you seem to have taken care of those, thanks ....

marcus-oscarsson:
Thanks, there is quite alot of nice improvements here :+1: . This also touches code from Soleil, EMBL and MAXIV so @MartinSavko , @IvarsKarpics  and @JieNanMAXIV  should probably have a look.

I'm mostly concerned about `beamline_setup_read` which is used by the EDNA Workflows, Im checking with @olofsvensson  which attributes that are used so that we can find a suitable way to replace this method.

HardwareObjects/queue_entry.py
```diff
@@ -1436,19 +1430,20 @@ class GenericWorkflowQueueEntry(BaseQueueEntry):
...
         BaseQueueEntry.__init__(self, view, data_model)
         self.rpc_server_hwobj = None
-        self.workflow_hwobj = None
         self.workflow_running = False
         self.workflow_started = False
 
     def execute(self):
         BaseQueueEntry.execute(self)
 
+        workflow_hwobj = HWR.beamline.workflow
```
IvarsKarpics: Is there a need for this? Maybe use in the code HWR.beamline.workflow? It would be good to reduce variable like *_hwobj

IvarsKarpics:
Could you please revert the StateMachine.py. I will fix it later when the pr has been accepted. Otherwise I am fine to merge the PR.

HardwareObjects/queue_entry.py
rhfogh: I agree - just wanted to keep the changed limited for a start.

rhfogh:
I thought I had put the state machine back

rhfogh:
@IvarsKarpics Now I *have* added the Statenmachine.py back (as I thought I had done earlier). Apologies for the somewhat confused commit history.

marcus-oscarsson:
This PR ave a merge conflict I guess its because MAXIV/old have been removed, could you take a look @rhfogh ?

PR #424, opened 2019-10-24
ESRF Cleanup - removal of unused files

marcus-oscarsson:
Dear all,

We have gone through the repository and removed some files that we believe are not used any longer. We have kept some of the older files, like the ones related to SPEC is anyone still using them ?

Please have a look at the list of removed files to double check if there is anything you are using (that we somehow missed).

I would be great if you could do the same exercise to go through all the files and remove the
ones you know are not used any more.

We will follow up this PR with some more cleanups, but it easier to do them when we have
removed the unused files.

Cheers,
Marcus and Antonia


HardwareObjects/InstanceServer.py
```diff
@@ -1,1289 +0,0 @@
-from HardwareRepository.BaseHardwareObjects import Procedure
```
IvarsKarpics: Qt version is still using the InstanceServer. Please revert it. Later I will move it to the gui part.

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -37,10 +37,10 @@ class ProcedureState(IntEnum):
     """
     Defines the valid Shutter states
```
rhfogh: I guess that should be Procedure States ;-)

HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -37,10 +37,10 @@ class ProcedureState(IntEnum):
...
     """
-    RUNNING = 0
-    FAILED = 1
+    FAILED = 0
```
rhfogh: Just curious: Is 'Idle' the same as 'Ready'? Or what is behind that name?

marcus-oscarsson: In a sense yes but ready would also somehow indicate that the procedure have at least run once. Where IDLE just indicates that its not doing anything. But we could rename IDLE to READY if that's what you are asking

HardwareObjects/InstanceServer.py
marcus-oscarsson: OK, Done !

IvarsKarpics:
Two approvals, I guess it is ready to be merged

PR #425, opened 2019-10-24
TangoLimaVideo and AbstractDetector

marcus-oscarsson:
Hi,

 - Removed Qub and Qt dependency from TangoLimaVideo
 - Added some documentation to AbstractDetector

Two completely unrelated commits by the way ...

Marcus


HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -1,179 +1,179 @@
...
-    def get_distance(self):
-        """
-        Descript. :
-        """
-        return
+        self._distance_motor_hwobj = None
 
     @abc.abstractmethod
-    def get_distance_limits(self):
+    def set_distance(self, value, timeout=None):
```
rhfogh: Should we not get rid of get_distance and set_distance altogether - if not now, then soon? We can (and, I think, should) handle the distance as my_detector.distance.get_value() and my_detector.distance.set_value(a_value). If that has the same effect, why have two differnet methods to make clutter? If it does not have teh saem effect, how do we ensure that people konw the difference and use the right alternative in tghe right contect?

marcus-oscarsson: Sure, its just that I did not know what you wanted to do with this comment

https://github.com/mxcube/HardwareRepository/blob/9756d2665469853cfdf9a2768958dda918dfe45f/HardwareObjects/abstract/AbstractDetector.py#L67

marcus-oscarsson:
I will merge this since the conflict is now resolved and we said that it was OK to merge it during the meeting.

PR #426, opened 2019-10-28
[WIP] Update to AbstractCollect and introduction of a  DataObject

marcus-oscarsson:
Dear all,

We introduced a DataObject that is immutable and validated against a JSONSchema. The object currently used as argument(s) for the AbstractProcedure can also be replaced with this object.

We made some smaller updates to the AbstractCollect
-  Attempt to make the standard "sequence" a bit more modular
-  Removed the current_* attributes on self as this can lead to that we "leak" state between collects
-  The introudcution of the beamline object will make alot of the set_/get_ methods unnecessary

Cheers,
Marcus

### DIFF
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -112,6 +112,9 @@ class Beamline(ConfiguredObject):
         # int Starting run number for path_template
         self.run_number = 1
 
+        # Undelators
+        self.undelators = []
+
     def init(self):
         """Object initialisation - executed *after* loading contents"""
 
--- a/HardwareObjects/ISPyBClient.py
+++ b/HardwareObjects/ISPyBClient.py
@@ -897,16 +897,13 @@ class ISPyBClient(HardwareObject):
             logging.exception("Could not store data collection")
             return (0, 0, 0)
 
-    def _store_data_collection(self, mx_collection, beamline_setup=None):
+    def _store_data_collection(self, mx_collection):
         """
         Stores the data collection mx_collection, and the beamline setup
         if provided.
 
         :param mx_collection: The data collection parameters.
         :type mx_collection: dict
-
-        :param beamline_setup: The beamline setup.
-        :type beamline_setup: dict
 
         :returns: None
 
@@ -925,27 +922,27 @@ class ISPyBClient(HardwareObject):
             )
 
             detector_id = 0
-            if beamline_setup:
-                lims_beamline_setup = ISPyBValueFactory.from_bl_config(
-                    self._collection, beamline_setup
-                )
 
-                lims_beamline_setup.synchrotronMode = data_collection.synchrotronMode
+            lims_beamline_setup = ISPyBValueFactory.get_beamline_setup(
+                self._collection
+            )
 
-                self.store_beamline_setup(
-                    mx_collection["sessionId"], lims_beamline_setup
-                )
+            lims_beamline_setup.synchrotronMode = data_collection.synchrotronMode
 
-                detector_params = ISPyBValueFactory().detector_from_blc(
-                    beamline_setup, mx_collection
-                )
+            self.store_beamline_setup(
+                mx_collection["sessionId"], lims_beamline_setup
+            )
 
-                detector = self.find_detector(*detector_params)
-                detector_id = 0
+            detector_params = ISPyBValueFactory().detector_from_blc(
+                mx_collection
+            )
 
-                if detector:
-                    detector_id = detector.detectorId
-                    data_collection.detectorId = detector_id
+            detector = self.find_detector(*detector_params)
+            detector_id = 0
+
+            if detector:
+                detector_id = detector.detectorId
+                data_collection.detectorId = detector_id
 
             collection_id = self._collection.service.storeOrUpdateDataCollection(
                 data_collection
@@ -1877,9 +1874,6 @@ class ISPyBClient(HardwareObject):
         :param mx_collection: The data collection parameters.
         :type mx_collection: dict
 
-        :param beamline_setup: The beamline setup.
-        :type beamline_setup: dict
-
         :returns: None
         """
         if self._disabled:
@@ -2057,9 +2051,9 @@ class ISPyBValueFactory:
     """
 
     @staticmethod
-    def detector_from_blc(bl_config, mx_collect_dict):
+    def detector_from_blc(mx_collect_dict):
         try:
-            detector_manufacturer = bl_config.detector_manufacturer
+            detector_manufacturer = HWR.beamline.detector.getProperty("manufacturer")
 
             if isinstance(detector_manufacturer, string_types):
                 detector_manufacturer = detector_manufacturer.upper()
@@ -2067,18 +2061,18 @@ class ISPyBValueFactory:
             detector_manufacturer = ""
 
         try:
-            detector_type = bl_config.detector_type
+            detector_type = HWR.beamline.detector.getProperty("type")
         except BaseException:
             detector_type = ""
 
         try:
-            detector_model = bl_config.detector_model
+            detector_model = HWR.beamline.detector.getProperty("model")
         except BaseException:
             detector_model = ""
 
         try:
             modes = ("Software binned", "Unbinned", "Hardware binned")
-            det_mode = int(mx_collect_dict["detector_mode"])
+            det_mode = HWR.beamline.detector.get_binning_mode()
             detector_mode = modes[det_mode]
         except (KeyError, IndexError, ValueError, TypeError):
             detector_mode = ""
@@ -2086,7 +2080,7 @@ class ISPyBValueFactory:
         return (detector_type, detector_manufacturer, detector_model, detector_mode)
 
     @staticmethod
-    def from_bl_config(ws_client, bl_config):
+    def get_beamline_setup(ws_client):
         """
         Creates a beamLineSetup3VO from the bl_config dictionary.
         :rtype: beamLineSetup3VO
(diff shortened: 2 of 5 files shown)


HardwareObjects/ISPyBClient.py
```diff
@@ -2097,36 +2091,40 @@ def from_bl_config(ws_client, bl_config):
...
         try:
-            synchrotron_name = bl_config.synchrotron_name
-            beamline_setup.synchrotronName = synchrotron_name
+            beamline_setup.synchrotronName = HWR.beamline.session.getProperty("synchrotron_name", "UNKNOWN")
         except (IndexError, AttributeError):
-            beamline_setup.synchrotronName = "ESRF"
+            beamline_setup.synchrotronName = "UNKNOWN"
 
-        if bl_config.undulators:
+        if HWR.beamline.undelators:
```
IvarsKarpics: undulators

HardwareObjects/ISPyBClient.py
```diff
@@ -2057,36 +2051,36 @@ class ISPyBValueFactory:
...
+            detector_manufacturer = HWR.beamline.detector.getProperty("manufacturer")
 
             if isinstance(detector_manufacturer, string_types):
                 detector_manufacturer = detector_manufacturer.upper()
         except BaseException:
             detector_manufacturer = ""
 
         try:
-            detector_type = bl_config.detector_type
+            detector_type = HWR.beamline.detector.getProperty("type")
```
IvarsKarpics: I would prefer not to use this construct (in a try call getProperty). The property should be initialized in the init of hwobj and here HWR.beamline.detector.get_type()
I can stay as it is and I can create a new PR to fix this.

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -31,14 +31,18 @@
...
 from HardwareRepository.TaskUtils import task
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 from HardwareRepository.ConvertUtils import string_types
 from HardwareRepository import HardwareRepository as HWR
 
+from HardwareRepository.utils.dataobject import DataObject
 
 __credits__ = ["MXCuBE collaboration"]
 
+log = logging.getLogger("user_level_log")
```
IvarsKarpics: maybe better use user_level_log and hwr_level_log

HardwareObjects/ISPyBClient.py
marcus-oscarsson: oh yes ... Ill fix that :)

HardwareObjects/ISPyBClient.py
marcus-oscarsson: Yes you are right, it would be better if detector had a get_type()

HardwareObjects/abstract/AbstractCollect.py
marcus-oscarsson: yes, its clearer

HardwareObjects/Beamline.py
```diff
@@ -112,6 +112,9 @@ def __init__(self, name):
         # int Starting run number for path_template
         self.run_number = 1
 
+        # Undelators
+        self.undelators = []
```
rhfogh: undulators here too.

utils/dataobject.py
```diff
@@ -0,0 +1,104 @@
...
+    popitem = _immutable
+    __setitem__ = _immutable
+    __setattr__ = _immutable
+    __delitem__ = _immutable
+    __getattr__ = dict.__getitem__
+    
+    def __hash__(self):
+        return id(self)
+
+    def dangerously_set(self, key, value):
```
rhfogh: Thias is OK, but might force_set be better?

rhfogh:
This looks good Still, I think we need some discussion on different ways of making DataObjects, for example DataObject (Python 3.7 and up), and attrs (Python 2.7 and up, but more complex). See relevant issue.

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -69,280 +73,122 @@ class AbstractCollect(HardwareObject, object):
...
-    def set_beamline_configuration(self, **configuration_parameters):
-        """Sets beamline configuration
-
-        :param configuration_parameters: dict with config param.
-        :type configuration_parameters: dict
-        """
-        self.bl_config = BeamlineConfig(**configuration_parameters)
-
-    def collect(self, owner, dc_parameters_list):
+    def collect(self, owner, cp_dict):
```
rhfogh: I have a problem with the archirtecture of this class - quite apart from the general concerns in Issue #433 (of which this is a good example). 

- The cp_dict is passed around as a function parameter. This is *the* way colelction parameters are handled. SInce it is an immutable class, you are entitled to assume that it does not change when passed as input to another function. Yet, it is changed all over teh place, using 'dangerously_set'. Worse, it is used for holding status information. 

    - Using an immutable obejct routinely to hold changing information should be a big no-no. If you need this kind of games, the function that modifies the object should return a modified copy and leave the inputm object unchanged. 
    - Status parameters and current settings do not belong in a parameter that is passed around between internal functions. They should instead be put in a class attribute, e.g. self.cp_dict. It makes sense that self.cp_dict is immutable, since it represents the internal state of the Collect object that cannot be changed from the outside, You can then change the contents on the current self.cp_dict by calling functions *on AbstractCollect*. ALternatively you can put the necessary attributes on AbstractCollect directly, adn ionly use data objectas when you are tranferring information into and out of the class.

marcus-oscarsson:
@rhfogh, @IvarsKarpics, as this PR is still WIP and in a quite early state its by no means intended to present a complete solution for all the issues that we would like to address with AbstracCollect and the DataObject. I'll try to address the comments above but its also a bit complicated as there is a separate issue for the DataObject. I would actually prefer if we could conduct the discussion here instead.

marcus-oscarsson:
@rhfogh 

In general:

The idea behind the DataObject was to provide an object that was compatible with the current way of passing data so that it would be easy to adapt to these changes and then in a second pass change this to a more suitable solution (if it would have a clear advantage). I would personally prefer Python 3 Data Classes but we cant use those as we have said that we want to be Python 2.7 compatible. I further, for obvious reasons, wanted to use the same object for both Python 2.7 and 3.

So what I wanted to achieve was an object that could be almost transparently used as it is today and that lets us know when its used improperly so that the code could be adapted accordingly. I also wanted to provide a way to actually set an attribute directly without making copy so that code that would be complicated to change immediately would still work without to much work. The goal would be to have no dangerously_set left in the code and my intention was to rewrite much of the code so that its not needed but I did not have the time yet. Doing this involves as you mention above splitting whats used today into slightly different groups of parameters or variables, this is still in progress. As I mentioned above this is still in quite an early stage

marcus-oscarsson:
> The clean-up and breaking up into many smaller functions (like _prepare and _acquire) is very good and necessary work.
> 
> Unfortuntely I have a problem with the architecture of this class - quite apart from the general concerns in Issue #433 (of which this is a good example - where do I go to check that the information available inside the cp object is what I think it is?):
> 
> The cp_dict is passed around as a function parameter. This is _the_ way collection parameters are handled. Since it is an immutable class, you are entitled to assume that the object contents do not change when passed as input to another function. Yet, they are changed all over the place, using 'dangerously_set'. Worse, the object is used for holding status information.

See my comment above, the goal would be to not use dangerously_set but at the same time provide a convenient way to stay compatible with the old way of passing data. I would be happy if it could be removed as soon as possible.

> 
> * Using an immutable object routinely to hold changing information is actively misleading. If you need this kind of games, the function that modifies the object should return a modified copy and leave the input object unchanged.

Sure we could do it this way instead and it would as you say be more correct, we can simply remove the dangerously_set and create a way to return a copy with the value updated but I did not see this as necessary as the intention for dangerously_set was to eventually disappear.

> * Status parameters and changeable current settings do not belong in a parameter that is passed around between internal functions. They should instead be put in a class attribute, e.g. self.cp_dict. It makes sense that self.cp_dict is immutable, since it represents the internal state of the Collect object that cannot be changed from the outside, You can then change the contents on the current self.cp_dict by calling functions _on AbstractCollect_. Alternatively you can put the necessary attributes on AbstractCollect directly, and only use data objects when you are tranferring information into and out of the class.

I agree, the parameters should be split into different groups I have simply not have the time to do it yet. I would personally prefer if we could avoid keeping parameters as instance variables as this keeps a state of something that is supposed to be valid for only one execution and there is hence a risk that these are not cleared properly.

rhfogh:
@marcus-oscarsson 
Your point is well taken that this is still WIP, and that it can evolve as we go through it. That deals with a lot of my comments, for the time being at least. And we do indeed need to distinguish between different kinds of parameters that are now there, I had missed that point.

One thing I would still recommend for now is to make the cp_dict into a class attribute instead of passing it around, *between internal functions*. It would still be good to separate the parameters that are passed around from those that only make sense as Collect attributes (like status), And it would also make sense to use the same data structure internally in Collect as you pass to or from other objects. But I think that little is gained by passing it as a parameter, internally. Conceptually they are a unit: 'the parameters used for collection'. It can get clunky to pass it between functions, and if there are ever any changes needed in the values it is hard to keep track of what changes where. Leaking of data can be dealt with in other ways: if this is an in principle immutable object we can avoid external objects modifying it; We can create it de novo when we call 'collect' and lean it up when we are finished - or just leave it as the parameters for the last collect, why not?

Anyway, let us discuss it later, either at the next meeting or before then?

marcus-oscarsson:
Hi @ALL,

Data classes, (https://docs.python.org/3/library/dataclasses.html) together with for instance Pydantic (https://pydantic-docs.helpmanual.io/) and Dataclasses JSON Schema (https://github.com/s-knibbs/dataclasses-jsonschema) could provide a more scalable solution. The question is if we can use these since Data classes are only supported in Python 3 and there is as far as I know no backport.

Would it unreasonable to think that most people are now moving to Python 3 or and will be using the new things (Collect and Procedure) we introduce with Python 3 once we are finished with refactoring. The older versions could potentially coexist until everybody is satisfied with the transition. We could thus say that the DataObject and the new features we introduce will be introduced exclusively for Python 3.

That essentially means that those that wants to stay on Python 2.7 needs to stay with the "old collect"  .Which of-course also needs to be adapted to the rest of the API changes but to pass data with dictionaries. Like this we stay compatible with Python 2.7 but introduce a few new concepts only for Python 3.

Do we stay true to our original commitment to stay Python 2.7 compatible through the entire refactoring ?. If this is a case a solution that @rhfogh suggested with .attr (https://www.attrs.org/en/stable/) or Marshmallow (https://marshmallow.readthedocs.io/en/stable/) could be used for both Python 2.7 and 3.

Is there perhaps a third good alternative that we missed ?

What do you think ?

Cheers,
Marcus

rhfogh:
That would be a brilliant solution! After as lightning-quick look I would very much vote for DataClasses, with JSONSchema and/or Pydantic. But it does depend on the people who have problems moving to PYthon 3 (Ivars. Martin, ...) and what they say.

IvarsKarpics:
It is an important decision and I would suggest to come up with an agreement during the next face-to-face meeting.

marcus-oscarsson:
Sure that sounds good, It would be good to go trough other potential options in the meantime. So let us know if you have an alternative solution ?

marcus-oscarsson:
Closing this one for the moment, to be discussed later

PR #428, opened 2019-11-04
Minor updates for MXCuBE3 mock

marcus-oscarsson:
Hi,

A few changes needed to get MXCuBE3 mockups to run with the latest Python3 and Hardware Repository work.

Cheers,
Marcus

### DIFF
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -112,8 +112,8 @@ class Beamline(ConfiguredObject):
         # int Starting run number for path_template
         self.run_number = 1
 
-        # List of undelators
-        self.undelators = []
+        # List of undulators
+        self.undulators = []
 
     def init(self):
         """Object initialisation - executed *after* loading contents"""
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -112,6 +112,9 @@ class Beamline(ConfiguredObject):
         # int Starting run number for path_template
         self.run_number = 1
 
+        # List of undelators
+        self.undelators = []
+
     def init(self):
         """Object initialisation - executed *after* loading contents"""
 
--- a/HardwareObjects/DataAnalysis.py
+++ b/HardwareObjects/DataAnalysis.py
@@ -41,7 +41,16 @@ class DataAnalysis(AbstractDataAnalysis.AbstractDataAnalysis, HardwareObject):
         self.edna_default_file = self.getProperty("edna_default_file")
 
         if not os.path.isfile(self.edna_default_file):
-            raise ValueError("File %s not found" % self.edna_default_file)
+            hwr = HWR.getHardwareRepository()
+            _default_file = os.path.join(hwr.serverAddress[0], self.edna_default_file)
+
+            if os.path.isfile(_default_file):
+                logging.getLogger("HWR").warning("DataAnalysis: File %s not found" % self.edna_default_file)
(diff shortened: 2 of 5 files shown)


HardwareObjects/Beamline.py
```diff
@@ -112,6 +112,9 @@ def __init__(self, name):
         # int Starting run number for path_template
         self.run_number = 1
 
+        # List of undelators
+        self.undelators = []
```
IvarsKarpics: undulators. Could the undulators be part of energy?

HardwareObjects/DataAnalysis.py
```diff
@@ -41,7 +41,16 @@ def init(self):
...
+
+            if os.path.isfile(_default_file):
+                logging.getLogger("HWR").warning("DataAnalysis: File %s not found" % self.edna_default_file)
+                logging.getLogger("HWR").warning("DataAnalysis: Using %s instead" % _default_file)
+                self.edna_default_file = _default_file
+            else:
+                raise ValueError("File %s not found" % self.edna_default_file)
+
         with open(self.edna_default_file, "r") as f:
             self.edna_default_input = "".join(f.readlines())
```
IvarsKarpics: maybe f.close() at the end

HardwareObjects/GenericDiffractometer.py
```diff
@@ -162,6 +162,10 @@ def __init__(self, name):
         self.centring_motors_list = None
         self.front_light_switch = None
         self.back_light_switch = None
+        self.aperture = None
```
IvarsKarpics: Did I miss this or we agreed that all these things are part of diffractometer?

HardwareObjects/Beamline.py
marcus-oscarsson: It could perhaps, I don't see why not. This is actually part of a much bigger refactoring, which I already started the one for the collect. Collect today have a list of undulators and a few other things that should be moved, I moved them to beamline when I started. The PR is WIP but we need this to get it to run at the moment. The final location should maybe change, please take a closer look at the WIP PR ...

HardwareObjects/DataAnalysis.py
marcus-oscarsson: The context manager handles that

HardwareObjects/Beamline.py
IvarsKarpics: No problem. I can make a new issue (move undulators to energy) to discuss this.

HardwareObjects/GenericDiffractometer.py
marcus-oscarsson: I think for this one we actually did say that beamstop, aperture, capillary and cryo would be part of the diffractometer. That does not mean that we can continue to discuss it but we had at least a short discussion about it when we initially added them directly on beamline.

IvarsKarpics: ok

marcus-oscarsson:
Can this be merged while we wait for the decision on where to keep undulators have been taken. I think that we in reality only need the undulator gaps. The most elegant solution would probably be to create a AbstractUndulator through which we can get the gaps. The very simplest solution would simply be a method that returns the lost with pairs, (name, gap). Id probably vote for the first one, it does not really matter to me if we keep them on Beamline or Energy, although I have a slight preference for Beamline :)

marcus-oscarsson:
Does the approval mean that we can merge it or do you still want some changes ?

PR #430, opened 2019-11-05
Fixed tests and CI with travis

IvarsKarpics:
With the new beamline object the definition of tests needed a change:
1. There is no call of getHardwareObject in the test routines, but members are accessed directly, for example, HWR.beamline.energy
2. It was not possible to use the existing configuration xml-qt because the beamline object initializes Qt graphics that makes core dump (seg fault) because the main Qt app was not initialized. I created a new directory called `test` that will contain minimum configuration for testing. Current configuration directories:
* xml-qt 
* xml-web
* test


HardwareRepository.py
```diff
@@ -155,7 +158,13 @@ def load_from_yaml(configuration_file, role, _container=None, _table=None):
                 (role, class_name, configuration_file, "%.1d" % load_time, msg1)
             )
             msg0 = "Done loading contents"
-        for role1, config_file in _objects.items():
+
```
rhfogh: I totally do not understand this change. As originally coded, the yaml files had to give a dictionary of objects, and the loader code then assumed that it got a dictionary. With the change, as I understand it, it is now supported also to give the objects as a list of tuple of pairs. Why the added complication? What is the use case? Considering that those yaml files are not written yet, why do we need to support alternative  formatting?

IvarsKarpics: I was not able to install ruamel.yaml for python2 so I made a fix to use existing yaml to load the files.

rhfogh: Does standard yaml not support dictionaries?

rhfogh: Actually, that is strange. It is supposed to work for PYthon 2.7, and I installed mine simply with
conda install ruamel.yaml

IvarsKarpics: Well yes, using conda is easy, but bot everyone is using it. Also the installation documentations is not based on the conda. I am not in favor to add a documentation on how to install a single package. Till now I could not find a easy way (pip, apt-get, yum) to install ruamel.yaml.

rhfogh: I double checked, and I agree that there seems no better way of supporting PyYaml (yaml 1.0) than the one you are proposing.

It is strange, though, that you cannot install ruamel.yaml. The code below simply uses pip, and works in my docker (mx3_deb9_qt4 is the standard docker):

FROM mx3_deb9_qt4
RUN apt-get update
RUN apt-get install python-pip -y
RUN pip install --upgrade pip
# To get rid of error, that the ps command cannot be found
RUN apt-get install procps -y
# Needed for new configuration from July 2019
RUN pip install ruamel.yaml

IvarsKarpics: This one worked: sudo python -m pip install ruamel.yaml
Ok, I will revert the file and update .travis.yaml

IvarsKarpics: Done!

IvarsKarpics:
For some reasons when I run pylint -E HardwareObjects/mockup/ or pylint -E HardwareObjects/EMBL I get recursion error. For time being I removed it from travis. I suggest to merge this PR and look at the recursion issue later.

IvarsKarpics:
Thanks!

PR #431, opened 2019-11-06
Minor fixes after cleanup

marcus-oscarsson:
Hi,

Some minor fixes needed after the various cleanup PR's.

Marcus


HardwareObjects/MiniDiff.py
```diff
@@ -123,6 +123,9 @@ def __init__(self, *args):
         self.lightWago = None
         self.currentSampleInfo = None
         self.aperture = None
+        self.cryo = None
```
rhfogh: Should that not be 'cryostream'? That is the role given below.

marcus-oscarsson: oh, yes

PR #432, opened 2019-11-07
Fixed test config and AbstractDetector

IvarsKarpics:
Removed swp file and keep just one detector config file


IvarsKarpics:
I noticed that travis failid due to: 
```
TypeError: Error when calling the metaclass bases
309    a new-style class can't have only classic bases
```
I fixed AbstractDetector to avoid the error. Could we maybe rewrite abstract classes as pure python classes and not inherit them from HardwareObjects?

rhfogh:
@IvarsKarpics 
I think that abstract classes must inherit either HardwareObject of HardwareObjectYaml, since we need to specify at the Abstract level which of the two configuration systems must be used. The solution to this problem could be to make sure HardwareObjectMixin (in the current class system) inherits from object.

IvarsKarpics:
Do @all agree that abstract class must inherit HardwareObject? For me it is fine, because then we can define signals.

marcus-oscarsson:
Me and @beteva had this discussion before making the PR as well and she reminded me that we already had a discussion about this once and we said that the Abstract classes should inherit HardwareObject. So its fine for us :)

marcus-oscarsson:
I think your problem would be solved if HardwareObject and HardwareObjectYaml inherits object, its not necessary in Python 3 but it wont, as far as i know, have any effect as all classes are "new style" in Python 3

rhfogh:
As a purely technicasl comment, I think abstract hardware objects should inherit either HardwareObject or HardwareObjectYaml. Both of these should inherit HardwareObjectMixin, and HardwareObjectMixin should inherit all shared superclasses, i.e. object and CommandContainer.
But this is easy to sort out.

marcus-oscarsson:
@rhfogh, sure. I took a quick look and it looks like there indeed are a few things to look at related to the inheritance chain. Some classes are inherited several times it seems ... Also just by curiosity why is it a mixin and not simply a base class ? or was it like that already ? 

We need to inherit object ultimately for it to work for Python 2.7 so its all the same if its done in the mixin or a common base class.

@IvarsKarpics, @rhfogh do we fix that in this PR or do we make a new one exclusively for that fix ?

rhfogh:
@marcus-oscarsson 
The inheritance is really messed up at tha moment, The Mixin class should not inherit HardwareObjectNode, for instance, but we can fix that later - no need to do it for this PR.

I added the HardWareObjectMixin class in kind of a hurry. as I saw it as mopstly WIP for discuxssion at that point. Quite likely it is the wrong name.

marcus-oscarsson:
Okey, I agree :)

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -20,19 +20,16 @@
...
 import abc
 
-from HardwareRepository.BaseHardwareObjects import HardwareObject
-
 __copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
 __license__ = "LGPLv3+"
 
 
-class AbstractDetector(HardwareObject):
+class AbstractDetector(object):
```
marcus-oscarsson: Is this final ?, it seems that we all agree that we should inherit from HardwareObject but its still changed here ?

IvarsKarpics: Hi, I will fix it

IvarsKarpics:
Hi @marcus-oscarsson : is the PR ok now?

marcus-oscarsson:
Hey @IvarsKarpics,

Thanks, I think that's fine, we should then make sure that the "BaseHardwreObject" which ever it is inherits object so that we don't have to do that everywhere ...

IvarsKarpics:
Ok

PR #436, opened 2019-11-25
[WIP] Motors

beteva:
AbstarctMotor cleanup. MotorStates is Enum.
Change BlissMotor, ExporterMotor, Resolution to correspond to the new AbstractMotor.

### DIFF
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -99,7 +99,7 @@ class BlissMotor(AbstractMotor):
         self._velocity = self.motor_obj.velocity
         return self._velocity
 
-    def move(self, position, wait=True, timeout=None):
+    def _move(self, position, wait=True, timeout=None):
         """Move motor to absolute position. Wait the move to finish.
         Args:
             position (float): target position
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -41,16 +41,16 @@ class BlissMotor(AbstractMotor):
 
     def __init__(self, name):
         AbstractMotor.__init__(self, name)
-        self.motor = None
+        self.motor_obj = None
 
     def init(self):
         """Initialise the motor"""
         AbstractMotor.init(self)
         cfg = static.get_config()
-        self.motor = cfg.get(self.motor_name)
-        self.connect(self.motor, "position", self.update_position)
-        self.connect(self.motor, "state", self.update_state)
-        self.connect(self.motor, "move_done", self.update_state)
+        self.motor_obj = cfg.get(self.motor_name)
+        self.connect(self.motor_obj, "position", self.update_position)
+        self.connect(self.motor_obj, "state", self.update_state)
+        self.connect(self.motor_obj, "move_done", self.update_state)
 
     def get_state(self):
         """Get the motor state.
@@ -58,7 +58,7 @@ class BlissMotor(AbstractMotor):
             (list): list of 'MotorStates'.
         """
         states = []
-        _state = self.motor.state.current_states_names
+        _state = self.motor_obj.state.current_states_names
 
         # convert from bliss states to MotorStates
         try:
@@ -74,7 +74,7 @@ class BlissMotor(AbstractMotor):
         Returns:
             float: Motor position.
         """
-        return self.motor.position
+        return self.motor_obj.position
 
     def get_limits(self):
         """Returns motor low and high limits.
@@ -85,7 +85,7 @@ class BlissMotor(AbstractMotor):
         # for some GUI components (like MotorSpinBox), so
         # instead we return very large value.
 
-        _low, _high = self.motor.limits
+        _low, _high = self.motor_obj.limits
         _low = _low if _low else -1e6
         _high = _high if _high else 1e6
         self._limits = (_low, _high)
@@ -96,7 +96,7 @@ class BlissMotor(AbstractMotor):
         Returns:
             (float): velocity [unit/s]
         """
-        self._velocity = self.motor.velocity
+        self._velocity = self.motor_obj.velocity
         return self._velocity
 
     def move(self, position, wait=True, timeout=None):
@@ -106,7 +106,7 @@ class BlissMotor(AbstractMotor):
             wait (bool): optional - wait until motor movement finished.
             timeout (float): optional - timeout [s].
         """
-        self.motor.move(position, wait=wait)
+        self.motor_obj.move(position, wait=wait)
         if timeout:
             self.wait_move(timeout)
 
@@ -117,11 +117,11 @@ class BlissMotor(AbstractMotor):
         """
         if timeout:
             with Timeout(timeout, RuntimeError("Execution timeout")):
-                self.motor.wait_move()
+                self.motor_obj.wait_move()
 
     def stop(self):
         """Stop the motor movement"""
-        self.motor.stop(wait=False)
+        self.motor_obj.stop(wait=False)
 
     def name(self):
         """Get the motor name. Should be removed when GUI ready"""
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -41,6 +41,7 @@ class BlissMotor(AbstractMotor):
 
     def __init__(self, name):
         AbstractMotor.__init__(self, name)
+        self.motor = None
 
     def init(self):
         """Initialise the motor"""
@@ -49,34 +50,23 @@ class BlissMotor(AbstractMotor):
         self.motor = cfg.get(self.motor_name)
         self.connect(self.motor, "position", self.update_position)
         self.connect(self.motor, "state", self.update_state)
-        self.connect(self.motor, "move_done", self.is_ready)
-
-    def connectNotify(self, signal):
-        """Check who neweds this"""
-        if signal == "positionChanged":
-            self.update_position(self.get_position())
-        elif signal == "stateChanged":
-            self.update_state(self.get_state())
-        elif signal == "limitsChanged":
-            self.update_limits()
+        self.connect(self.motor, "move_done", self.update_state)
 
     def get_state(self):
         """Get the motor state.
         Returns:
-            (enum 'MotorStates'): Motor state.
+            (list): list of 'MotorStates'.
         """
-        state = self.motor.state
+        states = []
+        _state = self.motor.state.current_states_names
 
         # convert from bliss states to MotorStates
         try:
-            state = MotorStates.__members__[state.upper()]
-        except AttributeError:
-            # check here for the double state (READY | LIMPOS...)
-            state = MotorStates.UNKNOWN
+            for stat in _state:
+                states.append(MotorStates.__members__[stat.upper()])
         except KeyError:
-            state = MotorStates.UNKNOWN
-
-        self.state = state
+            states.append(MotorStates.UNKNOWN)
+        self.state = states
         return self.state
 
     def get_position(self):
@@ -106,11 +96,8 @@ class BlissMotor(AbstractMotor):
         Returns:
             (float): velocity [unit/s]
         """
-        try:
-            self._velocity = self.motor.velocity
-            return self._velocity
-        except NotImplementedError:
-            raise
+        self._velocity = self.motor.velocity
+        return self._velocity
 
     def move(self, position, wait=True, timeout=None):
         """Move motor to absolute position. Wait the move to finish.
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -1,111 +1,141 @@
-from HardwareRepository.BaseHardwareObjects import Device
-import gevent
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU General Lesser Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+"""
+Example xml file:
+<device class="BlissMotor">
+  <username>Detector Distance</username>
+  <motor_name>dtox</motor_name>
+  <tolerance>1e-2</tolerance>
+</device>
+"""
+
+from gevent import Timeout
 from bliss.config import static
+from HardwareRepository.HardwareObjects.abstract.AbstractMotor import (
+    AbstractMotor,
+    MotorStates,
+)
+
+__copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
 
 
-class BlissMotor(Device):
-    (NOTINITIALIZED, UNUSABLE, READY, MOVESTARTED, MOVING, ONLIMIT) = (0, 1, 2, 3, 4, 5)
+class BlissMotor(AbstractMotor):
+    """Bliss Motor implementation"""
 
     def __init__(self, name):
-        Device.__init__(self, name)
+        AbstractMotor.__init__(self, name)
 
     def init(self):
-        self.motorState = BlissMotor.NOTINITIALIZED
-        self.username = self.motor_name
-
+        """Initialise the motor"""
+        AbstractMotor.init(self)
         cfg = static.get_config()
         self.motor = cfg.get(self.motor_name)
-        self.connect(self.motor, "position", self.positionChanged)
-        self.connect(self.motor, "state", self.updateState)
-        self.connect(self.motor, "move_done", self._move_done)
+        self.connect(self.motor, "position", self.update_position)
+        self.connect(self.motor, "state", self.update_state)
+        self.connect(self.motor, "move_done", self.is_ready)
 
     def connectNotify(self, signal):
+        """Check who neweds this"""
         if signal == "positionChanged":
-            self.emit("positionChanged", (self.getPosition(),))
+            self.update_position(self.get_position())
         elif signal == "stateChanged":
-            self.updateState()
+            self.update_state(self.get_state())
         elif signal == "limitsChanged":
-            self.motorLimitsChanged()
+            self.update_limits()
 
-    def _move_done(self, move_done):
-        if move_done:
-            self.updateState("READY")
-        else:
-            self.updateState("MOVING")
+    def get_state(self):
+        """Get the motor state.
+        Returns:
+            (enum 'MotorStates'): Motor state.
+        """
+        state = self.motor.state
 
-    def updateState(self, state=None):
-        if state is None:
-            state = self.motor.state
-        # convert from grob state to Hardware Object motor state
-        if state == "MOVING":
-            state = BlissMotor.MOVING
-        elif state == "READY":
-            state = BlissMotor.READY
-        elif state == "LIMPOS" or state == "LIMNEG":
-            state = BlissMotor.ONLIMIT
-        else:
-            state = BlissMotor.UNUSABLE
+        # convert from bliss states to MotorStates
+        try:
+            state = MotorStates.__members__[state.upper()]
+        except AttributeError:
+            # check here for the double state (READY | LIMPOS...)
+            state = MotorStates.UNKNOWN
+        except KeyError:
+            state = MotorStates.UNKNOWN
 
-        self.setIsReady(state > BlissMotor.UNUSABLE)
+        self.state = state
+        return self.state
 
-        if self.motorState != state:
-            self.motorState = state
-            self.emit("stateChanged", (self.motorState,))
-
-    def getState(self):
-        self.updateState()
-        return self.motorState
-
-    def motorLimitsChanged(self):
-        self.emit("limitsChanged", (self.getLimits(),))
-
-    def getLimits(self):
-        # no limit = None, but None is a problematic value
-        # for some GUI components (like MotorSpinBox), so
-        # in case of None it is much easier to return very
-        # large limits
-        ll, hl = self.motor.limits
-        return ll if ll is not None else -1e6, hl if hl is not None else 1e6
-
-    def positionChanged(self, absolutePosition):
-        # print self.name(), absolutePosition
-        self.emit("positionChanged", (absolutePosition,))
-
-    def getPosition(self):
+    def get_position(self):
+        """Read the motor position.
+        Returns:
+            float: Motor position.
+        """
         return self.motor.position
 
-    def getDialPosition(self):
-        return self.getPosition()
+    def get_limits(self):
+        """Returns motor low and high limits.
+        Returns:
+            (tuple): two floats tuple (low limit, high limit).
+        """
+        # no limit = None, but None is a problematic value
+        # for some GUI components (like MotorSpinBox), so
+        # instead we return very large value.
 
-    """@task
-    def _wait_ready(self):
-        while self.motorIsMoving():
-            time.sleep(0.02)
-    """
+        _low, _high = self.motor.limits
+        _low = _low if _low else -1e6
+        _high = _high if _high else 1e6
+        self._limits = (_low, _high)
+        return self._limits
 
-    def move(self, position):
-        # self._wait_ready(timeout=15)
-        self.motor.move(position, wait=False)  # .link(self.updateState)
+    def get_velocity(self):
+        """Read motor velocity.
+        Returns:
+            (float): velocity [unit/s]
+        """
+        try:
+            self._velocity = self.motor.velocity
+            return self._velocity
+        except NotImplementedError:
+            raise
 
-    def moveRelative(self, relativePosition):
-        self.move(self.getPosition() + relativePosition)
+    def move(self, position, wait=True, timeout=None):
+        """Move motor to absolute position. Wait the move to finish.
+        Args:
+            position (float): target position
+            wait (bool): optional - wait until motor movement finished.
+            timeout (float): optional - timeout [s].
+        """
+        self.motor.move(position, wait=wait)
+        if timeout:
+            self.wait_move(timeout)
 
-    def syncMoveRelative(self, relative_position, timeout=None):
-        return self.syncMove(self.getPosition() + relative_position)
-
-    def waitEndOfMove(self, timeout=None):
-        with gevent.Timeout(timeout):
-            self.motor.wait_move()
-
-    def syncMove(self, position, timeout=None):
-        self.move(position)
-        self.waitEndOfMove(timeout)
-
-    def motorIsMoving(self):
-        return self.motorState == BlissMotor.MOVING
-
-    def getMotorMnemonic(self):
-        return self.motor_name
+    def wait_move(self, timeout=None):
+        """Wait until the end of move ended, using the application state.
+        Args:
+            timeout(float): Timeout [s].
+        """
+        if timeout:
+            with Timeout(timeout, RuntimeError("Execution timeout")):
+                self.motor.wait_move()
 
     def stop(self):
+        """Stop the motor movement"""
         self.motor.stop(wait=False)
+
+    def name(self):
+        """Get the motor name. Should be removed when GUI ready"""
+        return self.motor_name
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -1,3 +1,4 @@
+# encoding: utf-8
 #
 #  Project: MXCuBE
 #  https://github.com/mxcube.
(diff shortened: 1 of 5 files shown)


IvarsKarpics:


```diff

Issues
======
+ Solved 132
- Added 15
           

Complexity increasing per file
==============================
- HardwareObjects/Resolution.py  1
- HardwareObjects/abstract/AbstractMotor.py  1
         

Complexity decreasing per file
==============================
+ HardwareObjects/BlissMotor.py  -3
+ HardwareObjects/ExporterMotor.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/HardwareRepository/HardwareRepository/pullRequest?prid=4561630&bid=15338304)

HardwareObjects/BlissMotor.py
```diff
@@ -1,111 +1,141 @@
...
+        """Initialise the motor"""
+        AbstractMotor.init(self)
         cfg = static.get_config()
         self.motor = cfg.get(self.motor_name)
-        self.connect(self.motor, "position", self.positionChanged)
-        self.connect(self.motor, "state", self.updateState)
-        self.connect(self.motor, "move_done", self._move_done)
+        self.connect(self.motor, "position", self.update_position)
+        self.connect(self.motor, "state", self.update_state)
+        self.connect(self.motor, "move_done", self.is_ready)
```
IvarsKarpics: maybe better update_state or set_state?

HardwareObjects/BlissMotor.py
```diff
@@ -1,111 +1,141 @@
...
 
     def connectNotify(self, signal):
+        """Check who neweds this"""
```
IvarsKarpics: needs this?

HardwareObjects/BlissMotor.py
```diff
@@ -1,111 +1,141 @@
...
+        AbstractMotor.__init__(self, name)
 
     def init(self):
-        self.motorState = BlissMotor.NOTINITIALIZED
-        self.username = self.motor_name
-
+        """Initialise the motor"""
+        AbstractMotor.init(self)
         cfg = static.get_config()
         self.motor = cfg.get(self.motor_name)
```
IvarsKarpics: I missed this notation. Why there is a motor inside the motor?

HardwareObjects/BlissMotor.py
```diff
@@ -1,111 +1,141 @@
...
-
-    def getPosition(self):
+            self.update_limits()
+
+    def get_state(self):
+        """Get the motor state.
+        Returns:
+            (enum 'MotorStates'): Motor state.
+        """
+        state = self.motor.state
```
IvarsKarpics: I would suggest to use more readable code like:
```
state = MotorStates.UNKNOWN
try:
    state = MotorStates.__members__[state.upper()]
except:
    pass
```

HardwareObjects/ExporterMotor.py
```diff
@@ -1,145 +1,244 @@
...
+        self._wait_ready(timeout)
+
+    def wait_motor_move(self, timeout=20):
+        """Wait until the end of move ended using the motor state.
+        Args:
+            timeout(float): Timeout [s]. Default value is 20 s
+        Raises:
+            RuntimeError: Execution timeout.
+        """
+        with Timeout(timeout, RuntimeError("Execution timeout")):
```
IvarsKarpics: "Execution timeout" is not so informative. Maybe better: Timeout waiting motor ready

HardwareObjects/ExporterMotor.py
```diff
@@ -1,145 +1,244 @@
...
-        )
-        self.chan_all_motor_states.connectSignal(
-            "update", self.all_motor_states_changed
+        """Initialise the motor"""
+        AbstractMotor.init(self)
+        _exporter_address = self.getProperty("exporter_address")
+        _host, _port = _exporter_address.split(":")
+        self._exporter = Exporter(_host, int(_port))
+
+        self.__position = self.add_channel(
```
IvarsKarpics: maybe better self.__position_channel?

HardwareObjects/ExporterMotor.py
```diff
@@ -1,145 +1,244 @@
...
-        self.cmd_get_limits = self.add_command(
-            {"type": "exporter", "name": "get_limits"}, "getMotorLimits"
-        )
-        self.cmd_get_max_speed = self.add_command(
-            {"type": "exporter", "name": "get_max_speed"}, "getMotorMaxSpeed"
+        if self.__position:
+            self.__position.connectSignal("update", self.update_position)
+            self.get_position()
+
+        self.__motor_state = self.add_channel(
```
IvarsKarpics: motor_state_channel

HardwareObjects/Resolution.py
```diff
@@ -1,178 +1,293 @@
...
-        by = float(HWR.beamline.detector["beam"].getProperty("by"))
+        try:
+            ttheta = arctan(radius / distance)
+            if ttheta:
+                return _wavelength / (2 * sin(ttheta / 2))
+        except (TypeError, ZeroDivisionError):
+            logging.getLogger().exception("error while calculating resolution")
+        return None
+
+    def dist2res(self, distance=None):
```
IvarsKarpics: PEP please.  `def distance_to_resolution`

HardwareObjects/Resolution.py
```diff
@@ -1,178 +1,293 @@
...
-            else:
-                return None
-        except Exception:
+            ttheta = arctan(radius / distance)
+            if ttheta:
+                self.position = _wavelength / (2 * sin(ttheta / 2))
+                self.emit("positionChanged", (self.position,))
+                self.emit("valueChanged", (self.position,))
+        except (TypeError, ZeroDivisionError):
             logging.getLogger().exception("error while calculating resolution")
```
IvarsKarpics: getLogger("HWR")

HardwareObjects/ExporterMotor.py
```diff
@@ -1,145 +1,244 @@
...
 class ExporterMotor(AbstractMotor):
+    """API for motor using the Exporter protocol, based on AbstractMotor"""
+
     def __init__(self, name):
         AbstractMotor.__init__(self, name)
-
-        self.motor_name = None
-        self.motor_resolution = None
-        self.motor_pos_attr_suffix = None
+        self.username = None
```
IvarsKarpics: is the username used somewhere?

HardwareObjects/BlissMotor.py
beteva: I've already changed this.

HardwareObjects/BlissMotor.py
beteva: needs of course. We have a lot of connectNotify all over the HardwareObjects. We should see how to change it.

HardwareObjects/BlissMotor.py
beteva: Because I need to use a motor object which is known by the bliss configuration only.  In general, when you use Bliss, we cannot use the command/channel model (as for SPEC or Exporter), but we do have access to a python object.

HardwareObjects/BlissMotor.py
beteva: This is the state reported by Bliss, which afterwards is translated to MotorStates. The Bliss motor state is a class itself and not a string. Bliss is changing and this function will be changhed soon.

HardwareObjects/ExporterMotor.py
beteva: OK. I just wanted to be consice

HardwareObjects/BlissMotor.py
IvarsKarpics: I was complaining about the typo :)

HardwareObjects/BlissMotor.py
IvarsKarpics: maybe then a name self.bliss_motor_obj is better?

HardwareObjects/ExporterMotor.py
beteva: Do we need to specify that it is a channel?

HardwareObjects/ExporterMotor.py
beteva: Same as above - it only maked the name longer, but do we really care that it is a channel?

HardwareObjects/Resolution.py
beteva: OK

HardwareObjects/Resolution.py
beteva: OK

HardwareObjects/ExporterMotor.py
beteva: Could be used in an application. I thought that we could have a standard way of getting the user name, which is more explicit when you want to show the users some information, rather than the cryptic motor name

HardwareObjects/BlissMotor.py
beteva: Got it,  but wanted to start a discussion about the method signature and usage ;-)

IvarsKarpics: True. I was also wondering if wee need this method. Btw it should also be in snake_case

HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -17,262 +17,173 @@
...
 
+__copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
 
-class MotorStates(object):
-    """Enumeration of the motor states
-    """
+@unique
+class MotorStates(IntEnum):
```
rhfogh: I really like both the IntEnum and the shortness of the list.I hope we will not need more, but just a question:
HOME, LIMPOS, LIMNEG, and READY means the motor is ready, not moving, and at (one of various) position(s). INITIALIZING and MOVING, means not ready and in action. UNKNOWN and FAULT are clear. Do we need any kind of state for a motor that is off or not yet initialised (hopefully not)? What state is the motor in immediately as you start the program? Or is UNKNOWN used for both?

HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -17,262 +17,173 @@
...
         Returns:
-            bool: True if ready, otherwise False
+            (bool): True if ready, otherwise False.
         """
-        return self.__state == self.motor_states.READY
-
-    def set_ready(self, task=None):
-        """Sets motor state to ready"""
-        self.set_state(self.motor_states.READY)
+        return self.state == MotorStates.READY
```
rhfogh: Is the motor not ready when it is in OME, LIMPOS or LIMNEG?

rhfogh:
@beteva 
What is the status of thie WIP, compared to my AbstractActuator? Is it a tremporary fix pending a permanent solutoin, or is it meant to be used instead of SAbstractActuator?
In the latter case, could you comment on the major differences between the two:

- using 'position' instead of get/set value?
- storing position and state locally, instead of getting them from the hardware when you need them?
- Not using ready_event for checking readiness?

If we have to incompatible proposals for handling abstract motors, it would be useful to discuss which one is better, and why.

HardwareObjects/abstract/AbstractMotor.py
beteva: @rhfogh I've thought that UNKNOW is the initial state. Unless you really want to set some other state (e.g. you've connected to the hardware), but then you change in the implementation of a particular motor. LIMPOS, LIMNEG and HOME are only for information. That means that the state can be an or of ready + any of the three above. The motor is ready to move, but you want to give additional information about limits or you have a home switch or you want to give the information that homing was executed.... As you said, INITIALIZING and MOVING, means not ready and in action. Not initialized can very well be UNKNOWN. As for OFF, not quite sure. This can be like limits - information only. BTW, one can expand the Enum in a particular implementation, if need be.

HardwareObjects/abstract/AbstractMotor.py
beteva: You are right @rhfogh, better to test  MotorStates.READY in self.state

beteva:
@rhfogh The motors are a very particular case, as they have several meaningful values (like position, limits) an no discrete value. That's why I would not like to have get_position and move, rather than position.get/set value. The AbstractActuator or whatever we decide to call it has a finite and discrete values so there get/set is more appropriate. I suggest further discussion on the Actuator subject a during the December meeting.

HardwareObjects/abstract/AbstractMotor.py
rhfogh: Excellent. I am happy that we do not need more states than the original .list.

rhfogh:
The AbdstractActuator was actually my proposal for abstract motors - it got a misleading name because I thought that was what people wanted, but it should be the abstract class for anything with a single value.But anyway, you can read it as an AbstactMotor proposal. 

As I remember the discussions at the last face-to-race meeting, we did actually agree to use get_value/set_value rather than position in all cases. The clinching argument was, I believe 1) that the motor position is clearly the main 'value', even if there are other things in the class, like limits. 2) it avoids a messy discussion about what is a motor (and has a 'position') and what is not, (and has a 'value'). Resolution, for instance, does that get a position of a value? Focus? 

I do think that get/set_value is better than move/position, and that we already agreed that once. But we are of course free to change our mind. Let us see what other people say.

IvarsKarpics:


```diff

Issues
======
+ Solved 140
- Added 10
           

Complexity increasing per file
==============================
- HardwareObjects/Resolution.py  1
- HardwareObjects/abstract/AbstractMotor.py  1
         

Complexity decreasing per file
==============================
+ HardwareObjects/BlissMotor.py  -4
+ HardwareObjects/ExporterMotor.py  -1
         

Clones removed
==============
+ HardwareObjects/BlissMotor.py  -5
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/HardwareRepository/HardwareRepository/pullRequest?prid=4561630&bid=15338304)

marcus-oscarsson:
We are not in principle against the idea with using set/and get value, but it sometimes difficult to apply it consistently. For instance motor has move and move_relative which in the case of set value have to be set and set_relative then I guess ? For Bliss motors, and in SPEC, there is a way to set an offset indirectly by setting the position. So setting a position would not move the motor but set an offset so that the actual position is offset + current position. There might be similar concepts in other control systems I imagine. One might think that this is slightly awkward (including my self) its however how it is and this makes it complicated to apply set_value since it's a bit confusing, in this context move and set_position is much clearer. This is one reason for no using set and get_value for motors.

The BaseHardwareObject could have a .value attribute and a set_value() method that one would need to override where its applicable. In the case of motor .value would return the position and set_value would be move. This would maybe both provide a clear interface and a generic way to set and get the value of a HardwareObject. The only drawback would be that there is two ways of doing the same thing, which I don't think is a big issue in this case.

Apart from the set and get value discussion I think that this is ready to merged, or are there something more to discuss ?

rhfogh:
As I see it we have two different proposals for abstract motors, this one, and AbstractActuator. Except for the states, that are handled only here, they are about equivalent. I would be reluctant to accept one of them and discard the other before we have some kind of agreeement - once one is merged, what is the likelihood that the other will be even considered, let alone implemented? 

Back to specifics, could you explain in a bit more detail how this 'offset-setting' works? You seem to be saying that calling 'set_position' would not actually move the motor, i.e. it would leave the mechanical state of the beamline unchanged, but it would change the value of the attributes that represent the motor position, and supposedly reflect the beamlien state. That sounds so confusing that I am bound to have misunderstood it.

beteva:
Motors tend to have two different positions - the real one, coming from the hardware and a "meaningful" one - the one which a meaning for the scientists, e.g. this is the position of my slits when the beam is at the centre. Very often, when a beamlime is aligned, this position is set, but the motor does not move. Only an offset is set. This is what in some control systems (like SPEC. Bliss an probably Sardana) there is the concept of user_position and dial_position. In the AbstractMotor, we do not want to introduce this complexity, as it might be confusing for people who do not use this king of control. But we also want to stay open for those who need it. Hence the position/move as opposed to get/set value.

rhfogh:
OK, that makes sense. Do I understand it correctly that there are (in these cases) two operations:

- One that moves the motor to a new position, which you then can read out again

- One that changes the offset, or zero-point, without moving the motor, so that the same actual motor position now corresponds to a different number for the value, or 'position'. This would be a kind of recalibration. 

If this is correct, it would follow that once you call the recalibration function all stored values become wrong, so you had better have it clear when that happens, and not mix the operations.

Also, the problem would become that you have two different operations, which would require different functions with clerarly different names. Have I got this right, and if so, which names would correspond to which of the two operations?

marcus-oscarsson:
@rhfogh I think you got it about right. About the two proposals for AbstractMotors, I think we should continue to build on this one since it contain concrete implementations of ExporterMotor and BlissMotor that are tested. We only need to settle the remaining questions and I was wondering of the set and get value was the only thing remaining to be discussed after the last fixes or if there were something else ?

HardwareObjects/ExporterMotor.py
IvarsKarpics: well, I think yes. For me self.__motor_state sounds like motor state and not a channel.

rhfogh:
@marcus-oscarsson, @beteva  I think this one needs to be discussed at the upcoming face-to-face meeting before it is merged. 

On the function names, I seem to remember that we discussed this quite intensely at the Hamburg meeting, and in the end agreed on get/set_value when third parties (to wit: Martin Savko) decided to opt for get/set_value rather than move. We can certainly change our minds, but I think we all need to have an input first.

But it is not just that. I think (of course, I wrote it ;-) ) that my proposal #420 is better in a number of ways, and I think we ought to compare or merge or at least discuss the ideas, instead of simply going with an alternative by default. Specifically:

- #420 uses a _ready_event instead of 'looping-till-state-is-ready', which ought to be both quicker and simpler.

- #420 is set up so that you get a functional motor after implementing only three methods (get_value, _set_value, and _set_limits (the latter optional). That seems similar in the two proposals, but in #420 the Abstract class takes care of emitting all the somethingChanged signals, so you do not need to do that in the subclasses

- #420 does not store a local 'position'; or 'value' attribute, which seems to make more sense when you would generally ask the motor for its position rather than go with a cached copy of the value.

- #420 explicitly allow for more than one state to count as 'ready'. In this WIP the is_ready function now says "return MotorStates.READY in self.state", which I do not think is merge-ready. Is self.state a tuple, now? Where does it say so?

Finally, I looked through the code of the abstract_motor branch, to find out how you handled the problem of setting the position versus setting the offset. But I could find only one function in any of the classes that did any position changing, to wit 'move' (and 'move_relative', which calls it). There was notyhing anywhere that I could see that had the effect of changing the position offset. So, if the problem of the position offset was the main argument for using move/position instead of 'get/set_value', I would like to see how that problem is actually dealt with (and what the relevant functions are called), which I cannot seem to find in this WIP.

marcus-oscarsson:
@rhfogh 

It would be great if it was possible to continue with this and merge it before the upcoming face to face meeting. I suggest that we do the necessary changes to this PR inorder to work from something that is already in working state. We can definitely merge some of the good ideas expressed in #420, see my comments below. The important thing is that we can make this evolve into something that we are satisfied with and it seems like we are quite close.

> On the function names, I seem to remember that we discussed this quite intensely at the Hamburg meeting, and in the end agreed on get/set_value when third parties (to wit: Martin Savko) decided to opt for get/set_value rather than move. We can certainly change our minds, but I think we all need to have an input first.

Yes and I don't think we disagree here, our suggestion was that we add the get and set_value to a Base class perhaps HardwareObject and then override them in the subclass. In this case AbstractMotor. I think the idea of AbstractMotor was to keep a clear interface to avoid confusion and that's why get and set_value was not used from the beginning. The question is if we would allow for both both get/set_value and move to exist ?

> 
> But it is not just that. I think (of course, I wrote it ;-) ) that my proposal #420 is better in a number of ways, and I think we ought to compare or merge or at least discuss the ideas, instead of simply going with an alternative by default. Specifically:
> 
> * #420 uses a _ready_event instead of 'looping-till-state-is-ready', which ought to be both quicker and simpler.

Sure I think we could use an event instead of a busy wait loop, it needs to be double checked with @beteva but I would not see any immediate issue with it. 

> * #420 is set up so that you get a functional motor after implementing only three methods (get_value, _set_value, and _set_limits (the latter optional). That seems similar in the two proposals, but in #420 the Abstract class takes care of emitting all the somethingChanged signals, so you do not need to do that in the subclasses

It looks like the way that signals are emitted can be improved in both proposals, both the AbstractMotor in this PR and the AbstractActuator. I did not find anything, maybe i missed it, that emits a value update on _set_value in any of the default implementations ? I think that we could improve this part.

> * #420 does not store a local 'position'; or 'value' attribute, which seems to make more sense when you would generally ask the motor for its position rather than go with a cached copy of the value.

Its true that its not necessary to store the position in the abstract class , I think this was done because of the offset feature that we tried to explain above. Its a detail that should be handled in the subclass

> * #420 explicitly allow for more than one state to count as 'ready'. In this WIP the is_ready function now says "return MotorStates.READY in self.state", which I do not think is merge-ready. Is self.state a tuple, now? Where does it say so?
> 

I think this is a good idea, I think @beteva wanted to add this as well.

> Finally, I looked through the code of the abstract_motor branch, to find out how you handled the problem of setting the position versus setting the offset. But I could find only one function in any of the classes that did any position changing, to wit 'move' (and 'move_relative', which calls it). There was notyhing anywhere that I could see that had the effect of changing the position offset. So, if the problem of the position offset was the main argument for using move/position instead of 'get/set_value', I would like to see how that problem is actually dealt with (and what the relevant functions are called), which I cannot seem to find in this WIP.

This functionality is handled in the underlying BLISS motor, so a simple call to set_position would be enough.

Cheers,
Marcus

rhfogh:
@marcus-oscarsson 

- "it seems like we are quite close.*"? That is not my impression

- "*The question is if we would allow for both both get/set_value and move to exist ?*" Why would you want two different functions that do the same thing but have to be maintained separately? And if they are to do different things, what are they exactly? It would be a lot easier for me to understand if somebody could point to an example of the function in the subclass that is supposed to change the offset.

- If all the stuff about setting the offset etc. is a matter that has to be handled in specific subclasses, why does it determine  what the function names should be in the superclass? 

- The signals are emitted from _set_value not from set_value. That is deliberate, it means that you can reimplement _set_value in a subclass without having to bother about the signals.

It would be great if we could *agree on the way forward and merge that in* before the next face-to-face meeting. That would allow us to move ahead in the agreed direction and be more advanced when we meet. I do not see the great advantage in merging something that has to go through major changes anyway - nor do I think it is likely that those major changes will be seriously considered afterwards, once we have accepted this WIP and started moving ahead with it.After all, if nobody looked seriously at #420 when it was a new WIP, why should they do it once we have merged in something else? 

This is a matter for consensus, I think, not for individual veto. But for what it is worth you have my vote.

rhfogh:
@marcus-oscarsson 
Re-reading your pot:
Are you saying that there is (to be) another function, called 'set_position', that is not in any of the classes in the WIP. but will be in their subclasses, and that has the effect of *not* moving the motor, but changing the offset so that the apparent value is different even though the motor is in the same place?? That would be so massively confusing that I surely cannot have got it right?

marcus-oscarsson:
> @marcus-oscarsson
> 
> * "_it seems like we are quite close._"? That is not my impression

Oh, I see but it seems to that we only need to settle the four points you mentioned above. So it should in my opinion not be to complicated.

> * "_The question is if we would allow for both both get/set_value and move to exist ?_" Why would you want two different functions that do the same thing but have to be maintained separately? And if they are to do different things, what are those things exactly? It would be a lot easier for me to understand if somebody could point to an example of the function in the subclass that is supposed to change the offset.

It would be to provide a clear interface with logical method names and one that could be used generically when appropriate. There would be no need to maintain two methods, it would simply be a mapping. I honestly don't have a strong opinion about it and I think only get/set would be fine, but I can understand that someone would like to have meaning full method names. I would however be against only having move and not set/get_value 

> * If all the stuff about setting the offset etc. is a matter that has to be handled in specific subclasses, why does it determine  what the function names should be in the superclass?
> * The signals are emitted from set_value, not from _set_value. That is deliberate, it means that you can reimplement _set_value in a subclass without having to bother about the signals.
> 

That sounds like a good solution, but I did not find any signals emitted in set_value either, I'm maybe looking in the wrong place. It's in anyways not super important. I do think its great idea and we should definitely do that.

> It would be great if we could _agree on the way forward and merge that in_ before the next face-to-face meeting. That would allow us to move ahead in the agreed direction and be more advanced when we meet. I do not see the great advantage in merging something that has to go through major changes anyway - nor do I think it is likely that those major changes will be seriously considered afterwards, once we have accepted this WIP and started moving ahead with it.After all, if nobody looked seriously at #420 when it was a new WIP, why should they do it once we have merged in something else?

I think we need to consider both PR's and that's what we currently are doing. @beteva have had these changes pending for sometime and finally made PR for them. They are tested on the beamline and works for both BLISS and Exporter Motors which is an important point. There are quite alot of overlap between the two PR's but the most important parts seems to be the points we have discussed above. Apart from those the two PRs are quite similar, so I propose that we address those and make the necessary changes in this PR. 

So would it be alright for you if we continue with this if we can conclude on:

 - Emitting singals fron the base class
 - Using at least set/get_value 
 - More than one ready state through is_ready
 - Using a ready event

> 
> This is a matter for consensus, I think, not for individual veto. But for what it is worth you have my vote.

What do you think about that ?

rhfogh:
@marcus-oscarsson 
I'd say that we need to get to a PR that we agree about more before we can get to the fun stuff (as you say) for the meeting. And we should try to get there during the upcoming week.Your summary sounds like a god approach. If you do not have time I can make a proposal, but I happy enough to leave it to you and Antonia.

On the function names:
I checked quickly, and I did not find any call to set_position that went to functions in external objects or did anything to offsets. But, BTAIM, if set_position has a known and different meaning we clearly cannot use it as a synonym for 'move'. One solution would be to stick with the 'move'/'get_position'/'set_position' model, but that has several drawbacks: It is messy (as you said yourself), set_position will (wrongly)  be taken as the equivalent setter for get_position, 'move' is not appropriate for many things that wlll use the same interface (resolution, energy, focus, ...), it directs people to think in terms of underlying offsets, which we would rather not force them to do - and we already agreed to do something else.

As an alternative, how about getting rid of 'position' altogether? That should avoid the confusion. The mapping would be:

'move' -> set_value

'get_position' -> get_value

'set_position' -> (re)set_position_offset (or something, I do not understand the function well enough to suggest a gog name. Does ti take a value, for instance? But anyway a name that shows this is a recalibration kind of function)

What would be teh downside to that?

Till tomorrow,

Rasmus

IvarsKarpics:


```diff

Issues
======
+ Solved 136
- Added 10
           

Complexity increasing per file
==============================
- HardwareObjects/Resolution.py  1
- HardwareObjects/abstract/AbstractMotor.py  1
         

Complexity decreasing per file
==============================
+ HardwareObjects/BlissMotor.py  -4
+ HardwareObjects/ExporterMotor.py  -1
         

Clones removed
==============
+ HardwareObjects/BlissMotor.py  -5
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/mxcube/HardwareRepository/pullRequest?prid=4561630&bid=15338304)

beteva:
Implemented:
    Emitting signals from the base class
    More than one ready state through is_ready
    Using a ready event as defined in @rhfogh #420.

rhfogh:
Issues to discuss at meeting 20191209

Comparing this PR with my old #420  (which is my reference),  I have the following points for #436

Minor points:
==========

- 'motor_name' should be just 'name'
- The 'state' attribute should be '_state' - there is a get_state function
- What does 'wait_move' do, and what is it needed for?
- Missing in #436: is_valid_value. Might that be useful?
- Added in #436: stop, abort, home, new State enum, velocity. *Excellent* improvements.

Major points:
==========

- naming: 'position' and 'move' versus get/set_value

- Signals handling:
I would propose:

- get rid of update_limits and update_state, and emit stateChanged from set_state. 
state and limits should *only* be changed through the relevant setter commands, so there should be no need for separate update_ functions.

- Keep update_position / update_value, (missing in #420) so that implementation code can call it.

IvarsKarpics:


```diff

Issues
======
+ Solved 136
- Added 10
           

Complexity increasing per file
==============================
- HardwareObjects/Resolution.py  1
- HardwareObjects/abstract/AbstractMotor.py  1
         

Complexity decreasing per file
==============================
+ HardwareObjects/BlissMotor.py  -4
+ HardwareObjects/ExporterMotor.py  -1
         

Clones removed
==============
+ HardwareObjects/BlissMotor.py  -5
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/mxcube/HardwareRepository/pullRequest?prid=4561630&bid=15338304)

IvarsKarpics:


```diff

Issues
======
+ Solved 136
- Added 10
           

Complexity increasing per file
==============================
- HardwareObjects/Resolution.py  1
- HardwareObjects/abstract/AbstractMotor.py  1
         

Complexity decreasing per file
==============================
+ HardwareObjects/BlissMotor.py  -4
+ HardwareObjects/ExporterMotor.py  -1
         

Clones removed
==============
+ HardwareObjects/BlissMotor.py  -5
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/mxcube/HardwareRepository/pullRequest?prid=4561630&bid=15338304)

HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -66,13 +66,17 @@ def __init__(self, name):
...
         self.motor_name = None
         self.username = None
+        self.read_only = False
+        self.default_value = None
         self._ready_event = None
 
     def init(self):
         """Initialise some parametrs."""
         self.motor_name = self.getProperty("motor_name")
         self.username = self.getProperty("username") or self.motor_name
```
IvarsKarpics: maybe we could use syntax `self.username = self.getProperty("username", self.motor_name)` ?

marcus-oscarsson:
@rhfogh, @beteva 

Are you planning on adding a AbstractActuator to this PR or in another PR ?. I think It would be great if we could get the updated AbstractMotor and AbstractActuator into the code base even if there are some client code that needs to be changed. This as some other PRs  needs AbstractActuator. We can make another issue simply saying adapt code using AbstractMotor ...

What do you think ?

rhfogh:
I was planning to get onto sorting out the ABstractActuator, Motor, NState starting today, and make the PR without waiting for the client classes. I'll get to it ASAP.

Issue #457, opened 2019-12-11
Add detector attributes


beteva:
Add radius, width and height and corresponding methods to read them.
Previously implemented in Resolution, now moved to detector.

marcus-oscarsson:
We can close this one right ?, please reopen if you think its still relevant @beteva

PR #458, opened 2019-12-12
AbstractDetector new attributes and methods

beteva:
Resolves issue #457

### DIFF
--- a/HardwareObjects/Resolution.py
+++ b/HardwareObjects/Resolution.py
@@ -1,178 +1,293 @@
+# -*- coding: utf-8 -*-
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU General Lesser Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+"""Resolution as motor."""
+
 import logging
-import math
-from HardwareRepository.HardwareObjects.abstract.AbstractMotor import AbstractMotor
+from scipy import arcsin, arctan, sin, sqrt, tan
+from scipy.constants import h, c, e
 from HardwareRepository import HardwareRepository as HWR
+from HardwareRepository.HardwareObjects.abstract.AbstractMotor import (
+    AbstractMotor,
+)
+
+__copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
 
 
 class Resolution(AbstractMotor):
-    def __init__(self, *args, **kwargs):
-        AbstractMotor.__init__(self, name="Resolution")
+    """Resolution as motor"""
 
-        # self.get_value = self.getPosition
-        self.valid = True
+    def __init__(self, name):
+        AbstractMotor.__init__(self, name=name)
+
         self.det_width = None
         self.det_height = None
+        self.det_radius = None
+        self.position = None
+        self._hwr_detector = None
+        self._hwr_energy = None
 
     def init(self):
-        self.currentResolution = None
-        detector = HWR.beamline.detector
+        """Initialise the motor"""
+        AbstractMotor.init(self)
+        self._hwr_detector = HWR.beamline.detector
+        self._hwr_energy = HWR.beamline.energy
 
-        if detector:
-            self.det_width = float(detector.getProperty("width"))
-            self.det_height = float(detector.getProperty("height"))
+        if self._hwr_detector:
+            self.det_width = float(self._hwr_detector.getProperty("width"))
+            self.det_height = float(self._hwr_detector.getProperty("height"))
         else:
-            self.valid = False
             logging.getLogger().exception("Cannot get detector properties")
             raise AttributeError("Cannot get detector properties")
 
+        self.connect(self._hwr_detector.distance, "stateChanged", self.update_state)
         self.connect(
-            detector.distance,
-            "stateChanged",
-            self.dtoxStateChanged
+            self._hwr_detector.distance, "positionChanged", self.update_position
         )
-        self.connect(
-            detector.distance,
-            "positionChanged",
-            self.dtoxPositionChanged
+        self.connect(self._hwr_energy, "valueChanged", self.update_energy)
+
+    def is_ready(self):
+        """Check if the distance motor state is READY.
+        Returns:
+            (bool): True if ready, otherwise False.
+        """
+        try:
+            return self._hwr_detector.distance.is_ready()
+        except AttributeError:
+            return False
+
+    def get_state(self):
+        """Get the state of the distance motor.
+        Returns:
+            (enum 'MotorStates'): Motor state.
+        """
+        return self._hwr_detector.distance.get_state()
+
+    def get_position(self):
+        """Read the position.
+        Returns:
+            (float): position.
+        """
+        if self.position is None:
+            dtox_pos = self._hwr_detector.distance.get_position()
+            self.get_detector_radius(dtox_pos)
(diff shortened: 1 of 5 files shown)


IvarsKarpics:


```diff

Issues
======
+ Solved 138
- Added 10
           

Complexity increasing per file
==============================
- HardwareObjects/abstract/AbstractMotor.py  1
- HardwareObjects/Resolution.py  1
- HardwareObjects/abstract/AbstractDetector.py  2
         

Complexity decreasing per file
==============================
+ HardwareObjects/ExporterMotor.py  -1
+ HardwareObjects/BlissMotor.py  -4
         

Clones removed
==============
+ HardwareObjects/BlissMotor.py  -5
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/mxcube/HardwareRepository/pullRequest?prid=4664174&bid=15614405)

HardwareObjects/Resolution.py
```diff
@@ -1,178 +1,288 @@
...
+    def __init__(self, name):
+        AbstractMotor.__init__(self, name=name)
 
-        # self.get_value = self.getPosition
-        self.valid = True
         self.det_width = None
         self.det_height = None
+        self.det_radius = None
+        self._nominal_value = None
+        self._hwr_detector = None
```
IvarsKarpics: Should it be self.detector as a property? Same for energy?

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -36,40 +36,43 @@ def __init__(self, name):
...
+    def init(self):
+        """Initialise some common paramerters"""
+        self.width = self.getProperty("width")
+        self.height = self.getProperty("height")
+        try:
+            self._det_metadata = self.getProperties("beam")
+        except KeyError:
+            pass
+
+        self._pixel_size = (self.getProperty("px"), self.getProperty("py"))
```
IvarsKarpics: some attributes have _ and some not. I suggest to have all with _

HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -17,262 +18,223 @@
...
+        HardwareObject.__init__(self, name)
 
-    def get_motor_mnemonic(self):
-        return self.motor_name
+        self._state = MotorStates.UNKNOWN
+        self._nominal_value = None
+        self._limits = (None, None)
+        self._velocity = None
+        self._tolerance = None
+        self.motor_name = None
```
IvarsKarpics: also with _ ?

marcus-oscarsson:
@beteva is not in today and will be back next week. I think she simply forgot to rebase this to only include the changes related to the detector.

@IvarsKarpics, has a point. I think we said that "value" attributes in general should be protected and that we add a property to access them, right ? (@all)

HardwareObjects/Resolution.py
rhfogh: I woiuld say no, but neither should it be self._hwr_detector. All hardware objects should live in only *one* place and be accessed there; for the detector that place is beamline.detector. Having multiple links in different places just means more things to maintain. YOu can have ain internal attribute (with underscore) for practical reasons, though I woudl avoid it myaself, but if it is there it should nto be public.

HardwareObjects/abstract/AbstractDetector.py
rhfogh: Attributes that have an '_' must have get_ and set_ functions. But yes, we should be consistent one way or the other.

rhfogh:
@IvarsKarpics 
As I remeber it we never used properties fora ttributes; thery just ahd get_ and set_ functions

marcus-oscarsson:
@rfogh, @all the last statement was not really clear. I think the use of properties needs to be more clearly stated and agreed !. I have nothing against either as long as properties are not used for something that involving and underlying operation/process (i.e moving hardware) or something that takes an argument like `timeout`. 

We can otherwise, in my opinion, use properties, although I don't really see the necessity of adding a property to public attribute (if its both readable and settable).

@rhfogh, @IvarsKarpics, @beteva: I agrree with @rhfogh  that things that are available through `beamline` should be accessed from `beamline`

rhfogh:
@marcus-oscarsson 
Sorry for the spelling of my last message. 
I thought the agreement was that properties were used only for contained HardwareObjects. For atttributes we would then use get_xyz and set_xyz, or just naked attributes where that feels appropriate. 

I actually prefer properties myself, but the decision went the other way, and I think it is better to be  consistent, so you never have to wonder whether something should be written or accessed as a property or through a getter function.

Apart from contained HardwareObjects, can anyone point to a general context where properties can/should be used instead of getters?

marcus-oscarsson:
@rhfogh, @all, Ok, I remember that now yes, thanks. Lets stick with that then, if no one else have any objections.

HardwareObjects/abstract/AbstractDetector.py
beteva: There is a get_pixel_size method, hence the _. Not sure that we need set_pixel_size. Should be implemented only if somebody needs it. Otherwise it can be dangerous, as to my knowledge the pixel size is unique and immutable for each detector.
The width and height, I've implemented **get** and removed it. It was a bit too much. That's why they do not have _. I think the same should be  applied for humidity and temperature and in general for all values which are only  _information_  and have no set functionality.

IvarsKarpics: I agree. Pixel size, temperature and humidity are like `internal` values and are not set from outside.

rhfogh: I disagree, unfortunately.

- For one thing, whatever we do should be consistent. Treat width, height, and pixel_size the same, and do it the same way in other HOs.

- If temperature and humidity change, should they not be read-only actuators? Even if we use plain, writable attributes for fixed values (which I would not recommend), we shold surely not use it for values that change over time?

- If we leave e.g. width and height as plain attributes, they are settable, according to the code. How sure are we that people will know enough not to set them by mistake? Having a get_width function (or a read-only property, but we were avoiding those). What does it cost us, to write a couple of extra functions once?

- As a separte point, the \_det_radius should not be an attribute. It should be calculated afresh each tome from the position, width, and height.

marcus-oscarsson: > I disagree, unfortunately.
> 
> * For one thing, whatever we do should be consistent. Treat width, height, and pixel_size the same, and do it the same way in other HOs.

I think that, that is a good idea in general

> * If temperature and humidity change, should they not be read-only actuators? Even if we use plain, writable attributes for fixed values (which I would not recommend), we shold surely not use it for values that change over time?

I don't think we should make things more complicated than they need to be and I don't really think It would be strictly necessary for temperature and humidity to be read-only Actuators. It would be elegant and consistent, but lets first see how well it fits with the current implementation and then decide on that ? 

> * If we leave e.g. width and height as plain attributes, they are settable, according to the code. How sure are we that people will know enough not to set them by mistake? Having a get_width function (or a read-only property, but we were avoiding those), would ensure that people knew these were read-only. What does it cost us, to write a couple of extra functions once?

All those properties should definitely be read-only for client code, either through a property or simply get_<attribute> method. For consistency I would say that they should be getter methods
as some of them actually involve communication with hardware and hides a more complex operation than simply reading a value.

> * As a separate point, the _det_radius should not be an attribute. It should be calculated afresh each time from the position, width, and height.

with the above reasoning it should then be a method called `get_det_radius` that performs the calculation

Marcus

rhfogh: @marcus-oscarsson
I agree with your comments.
Can we summarise it as saying that we should always use getter methods, except:

- 1 for contained hardwareobjects
 - 2 maybe for class level attributes, that can be implemented as a  simple attribute on the class (*not* on the instance).

marcus-oscarsson: @rhfogh, @all, As I mentioned in a previous conversation, I would say:

- getters and setters for "complex" operations or operations involving hardware
- properties for "simple" operations or directly reading or setting values.

Which I think includes the two cases you mention above, both being the "simple" case.

rhfogh: @marcus-oscarsson 
As per our previoous conversation I agree - with the following three points:
- Anything that may move hardware or trigger signals (also if only possibly on other beamlines) is a complex operation
- Anything that can be called an 'operation' at all needs functions - you only use properties when all you are doing is getting or setting a value.
- You use defined *properties*, not bare attributes, for anything that is read-only.

beteva: @ALL, first there is get_radius method, I will remove the corresponding attribute.
The width, height, temperature, humidity, they can be even properties (with read only interface) for completeness. This makes the interface consistent, but a bit heavy.

PR #462, opened 2020-01-24
Added TangoLimaVideoLoopback

marcus-oscarsson:
Hi

Added TangoLimaVideoLoopback that writes the from the camera to a loopback device and starts a process that processes the stream, i.e converting it to MPEG

Marcus


marcus-oscarsson:
Reviewed and 15 days have passed so merging

PR #464, opened 2020-01-28
Using QGraphicsScene coordinates rather than QGraphicsView ones

natxo14:
Bug discovered using CameraBrick.py
If HWR.beamline.microscope <=> QtGraphicsManager then
The widget to display images is :
QtGraphicsManager.graphics_view and the scene, the GraphicsView.graphics_scene 

When the widget is 'small' and fits the scene surface (and not more), their coordinates ( QGraphicsView's and QGraphicScene's ) are visually the same, but if the view is bigger than the scene, there's a delta between them ( ex: GraphicsItemCentringLines drawn far from the mouse position ).
One solution to fix this: all the positions related to scene's coordinates

### DIFF
--- a/HardwareObjects/QtGraphicsLib.py
+++ b/HardwareObjects/QtGraphicsLib.py
@@ -2759,7 +2759,7 @@ class GraphicsCameraFrame(QtImport.QGraphicsPixmapItem):
         :param event:
         :return:
         """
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
Merging since its been approved and more than 15 days have passed.

PR #465, opened 2020-01-30
Abstract actuator

marcus-oscarsson:
Dear all,
this is the proposed implementation for the new abstract actuator as decided at the last MXCuBE developers' meeting. Feel free to use it ;-)
Marcus and Antonia

### DIFF
--- a/HardwareObjects/abstract/AbstractActuator.py
+++ b/HardwareObjects/abstract/AbstractActuator.py
@@ -1,84 +1,128 @@
-"""
-Actuators have two positions - 'IN' and 'OUT'.
-"""
+# encoding: utf-8
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU General Lesser Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
 
-import logging
-from warnings import warn
-from HardwareRepository.BaseHardwareObjects import Device
-from HardwareRepository.TaskUtils import task
+"""Abstract Actuator"""
+
+import abc
+from enum import IntEnum, unique
+from gevent import Timeout
+from gevent.event import Event
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+
+__copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
 
 
-class AbstractActuator(Device):
+class AbstractActuator(HardwareObject):
+    """Abstract actuator"""
 
-    (UNKNOWN, IN, OUT, MOVING) = ("unknown", "in", "out", "moving")
+    __metaclass__ = abc.ABCMeta
+
+    unit = None
 
     def __init__(self, name):
-        Device.__init__(self, name)
-        self.actuator_state = AbstractActuator.UNKNOWN
-        self.username = "unknown"
-        # default timeout - 3 sec
-        self._timeout = 3
-        self.states = {True: "IN", False: "OUT"}
+        HardwareObject.__init__(self, name)
+        self._nominal_value = None
+        self._limits = (None, None)
+        self.actuator_name = None
+        self.read_only = False
+        self.default_value = None
+        
+    def init(self):
+        """Initialise some parameters."""
+        self.actuator_name = self.getProperty("actuator_name")
+        self.read_only = self.getProperty("read_only") or False
+        self.default_value = self.getProperty("default_value")
 
-    def connectNotify(self, signal):
-        if signal == "actuatorStateChanged":
-            self.value_changed(self.get_actuator_state(read=True))
-
-    def value_changed(self, value):
-        self.actuator_state = self.states.get(value, AbstractActuator.UNKNOWN)
-        self.emit("actuatorStateChanged", (self.actuator_state.lower(),))
-
-    def get_actuator_state(self, read=False):
-        """Return the state value
-           Args:
-             read (bool): read the hardware if True
-           Returns:
-             (string): the state
+    @abc.abstractmethod
+    def get_value(self):
+        """Read the actuator position.
+        Returns:
+            float: Actuator position.
         """
-        logging.getLogger().info("%s state: %r" % (self.username, self.actuator_state))
-        return self.actuator_state
+        return None
 
-    @task
-    def actuator_in(self, wait=True, timeout=3):
-        """Set the actuator 'in'. Update the status
-           Keyword Args:
-             wait (bool): wait for the movement to finish
-             timeout (float): movement expires after timeout [s]
+    def get_limits(self):
+        """Return actuator low and high limits.
+        Returns:
+            (tuple): two floats tuple (low limit, high limit).
         """
-        self.value_changed(self.actuator_state)
+        return self._limits
 
-    @task
-    def actuator_out(self, wait=True, timeout=3):
-        """Set the actuator 'out'. Update the status
-           Keyword Args:
-             wait (bool): wait for the movement to finish
-             timeout (float): movement expires after timeout [s]
+    def set_limits(self, limits):
+        """Set actuator low and high limits.
+        Args:
+            limits (tuple): two floats tuple (low limit, high limit).
         """
-        self.value_changed(self.actuator_state)
+        self._limits = limits
+        self.emit("limitsChanged", (self._limits,))
 
-    @task
-    def actuator_toggle(self, wait=True, timeout=3):
-        """Toggele the actuator.
-           Keyword Args:
-             wait (bool): wait for the movement to finish
-             timeout (float): movement expires after timeout [s]
+    @abc.abstractmethod
+    def _set_value(self, value, wait=True, timeout=None):
+        """Move actuator to absolute position. Wait the move to finish.
+        Args:
+            value (float): target value
+            wait (bool): optional - wait until actuator movement finished.
+            timeout (float): optional - timeout [s].
         """
-        if self.actuator_state == AbstractActuator.IN:
-            self.actuator_out(wait, timeout)
-        elif self.actuator_state == AbstractActuator.OUT:
-            self.actuator_in(wait, timeout)
 
-    def getActuatorState(self, read=False):
-        warn(
-            "getActuatorState method is deprecated, use get_actuator_state",
-            DeprecationWarning,
-        )
-        return self.get_actuator_state(read)
+    def set_value(self, value, wait=True, timeout=None):
+        """Move actuator to absolute value. Wait the move to finish.
+        Args:
+            value (float): target value
+            wait (bool): optional - wait until actuator movement finished.
+            timeout (float): optional - timeout [s].
+        """
+        self._set_value(value)
+        self.update_value()
 
-    def actuatorIn(self, wait=True, timeout=3):
-        warn("actuatorIn method is deprecated, use actuator_in", DeprecationWarning)
-        self.actuator_in(wait, timeout)
+        if wait:
+            self.wait_ready(timeout)
 
-    def actuatorOut(self, wait=True, timeout=3):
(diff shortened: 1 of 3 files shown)


IvarsKarpics:


```diff

Issues
======
+ Solved 15
- Added 11
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4916806&bid=16272218)

IvarsKarpics:


```diff

Issues
======
+ Solved 15
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4916806&bid=16272218)

marcus-oscarsson:
@rhfogh We have made some adjustments since our last discussion per mail, please have a final look. We decided to keep the set_limits method since it might happen other actuator apart from Motors have limits.

Cheers,
Antonia and Marcus

IvarsKarpics:


```diff

Issues
======
+ Solved 15
           

Complexity increasing per file
==============================
- HardwareObjects/abstract/AbstractActuator.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4916806&bid=16272218)

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -1,84 +1,117 @@
...
-
-    def actuatorIn(self, wait=True, timeout=3):
-        warn("actuatorIn method is deprecated, use actuator_in", DeprecationWarning)
-        self.actuator_in(wait, timeout)
-
-    def actuatorOut(self, wait=True, timeout=3):
-        warn("actuatorOut method is deprecated, use actuator_out", DeprecationWarning)
-        self.actuator_out(wait, timeout)
+
+    def set_value(self, value, wait=True, timeout=None):
```
rhfogh: The 'wait parameter should be removed - it is redundant and begs thequestion what happens if wait=False and timeout==1.0.. 
If timeout == 0 you do not wait, otherwise you simply pass the timeout parameter on to wait_ready.

IvarsKarpics:


```diff

Issues
======
+ Solved 15
- Added 2
           

Complexity increasing per file
==============================
- HardwareObjects/abstract/AbstractActuator.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4916806&bid=16272218)

marcus-oscarsson:
@rhfogh I think you made a comment at the same time as we made a fix so we can only see your comment in the mail. The wait parameter is in anyways removed now ...

marcus-oscarsson:
Morning !

> All the details are fine, now. 

Great :)

>But I have a serious problem with the handling of limits and, in a similar vein, velocity. It may be that there is a use case for set_limits that I am simply missing; if so please let me know, and bear with the absolutist language in this comment. But this is how I see the situation for now:
> 
> If there is a function called set_limits in the abstract class, you are entitled to assume that 1) you can set the limits, at least in the normal case 2) the limits you set will be honored. The same goes for set_velocity, in AbstractMotor.

I think there are several things to consider here, one could make the assumption as you say that all methods should be implemented and their function should be "honored", which in one way maybe is elegant from a modelling point of view. However, I think that we also have to keep this model light weight and easy to use, so we have to somehow limit our model so that it remains practical for our purpose.

It would of course be possible to crate abstract classes for limits and then for velocity maybe AbstractActuatorWithLimits and AbstractActuatorWithVelociy or maybe simply WithLimits, WithVelociy. One than needs to create a few more classes and also inherit three classes to
implement an Actuator with Limits and Velocity. This is just an extreme example of where I personally think we don’t really want to go.

> 
> But the reality, as I understand it, is that:
> 
>     1. in many  of cases, limits are undefined (AbstratNPosition) nonexistant, or immutable.

Limits would not necessarily be undefined for AbstratNPosition imagine that you have 4 positions
and you’d like to toggle between its extremities, you could use get_limits to do that. Its actually not unusual that we have three state objects were we often only use the two extremes (IN, OUT).

Limits would be used for other actuators that are not  AbstratNPosition or AbstractMotor, such as Energy and Resolution
> 
>     2. Where limits exist and are settable, they are generally handled closer to the hardware, and it is not obvious that people will remember (or want) to have an HO.set_limits function that actually sets the hardware limits.

Its true that we are not often using set_limits from our application, its maybe not so bad to have it there for completeness, We don't have to make this method abstract so that it can be implemented of needed.
> 
>     3. Even where limits are settable, this is not something that will (or should) happen often in normal program operation. The typical case will be that limits are set on configuration only (if at all).

Se answer above.

> 
> 
> In short, the normal case will be that objects do not have settable limits, or, at best that limit setting would be a rare and non-obvious thing to do. With the current code I think there is a high risk that default object behaviour will be exactly as coded here, i.e. that objects will have a set_limits that will set the nominal_limits attribute but will have _exactly zero_ effect on the instrument behaviour. That is asking for trouble, IMHO.

I guess I've already answered this above, I would not mind removing set_limits as I agree with that we are not using it from the application. All tough I'd rather see that it stays for the future.

> 
> If we cannot agree on removing limits from the abstract class altogether, I would prefer that we implemented only get_limits, and set it to return a null value by default. It would then be up to people coding objects that actually have defined limits to decide how to set them and to implement them properly. If you feel strongly about it, we could have a set_limits function that raised NotImplementedError, 

I think what you are suggesting above is fine implementing; get_limits and throwing NotImplementedError from set_limits. To be double checked with @beteva 

>which would at least reflect the most normal case and not invite misuse. Or, if we want, we could make set_limits an abstract function, which would force people to decide what limit handling they wanted
> 
> AbstractMotor.set_velocity has pretty much the same problems. How often is the velocity settable at all, how often does it make sense to set it from the program, and what does the settable _velocity attribute do in the cases where setting the velocity is not meaningful?

@beteva might have some more background on this. The only comment I have on velocity is that the same reasoning as for limits apply. We could throw NotImplementedError from the default implemetnation if thats better ?

Cheers,
Marcus

beteva:
Dear @rhfogh,
 in addition to what  @marcus-oscarsson commented, here is my contribution:
set_limits - true we normally do not set them from the application, but we do use them usually when we do homing, so there is a need that the method exists. As @marcus-oscarsson said, we may want to have it in the AbstractActuator for completeness and raise NotImplementedError, rather than make the method abstract.
set/get velocity - It is often changed when doing scans. This is done internally for the microdiff, but not necessarily for other type of diffractometers or actuators like energy, so both get and set are needed for the motors. Actually in my experience I've seen more often that the hardware accepts changing of velocity, than having it immutable. If such is the case, as @marcus-oscarsson suggests, we could just raise NotImplementedError.

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -1,84 +1,121 @@
...
-             timeout (float): movement expires after timeout [s]
+        Set actuator to absolute value.
+        
+        Args:
+            value (float): target value
+            timeout (float): optional - timeout [s],
+                             If timeout == 0: return at once and do not wait;
+                             if timeout is None: wait forever.
+        """
+        self._set_value(value)
```
rhfogh: Proposal 1:
Add the following:

    if self.read_only:
        raise RuntimeError("Attemp[rt to set value on read_only object")

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -1,84 +1,121 @@
...
-             wait (bool): wait for the movement to finish
-             timeout (float): movement expires after timeout [s]
+        Set actuator to absolute value.
+        
+        Args:
+            value (float): target value
+            timeout (float): optional - timeout [s],
+                             If timeout == 0: return at once and do not wait;
+                             if timeout is None: wait forever.
+        """
```
rhfogh: Proposal 1
Add the following at the start of the function (we could define a special Exception, but this is simpler):

    if self._read_only:
        raise AttributeError("Attempt to set vaue on a read_only %s object" % self.__class__.__name__)

rhfogh: Proposal 2:
Add the following at the start of the function:

    lowlim, hilim = self.get_limits()
    if lolim is not None and and value < lolim:
        raiseValueError("Value %s is less than limit %s" % (value, lolim))
    if hilim is not None and and value > hilim:
        raiseValueError("Value %s is greater than limit %s" % (value, hilim))

rhfogh:
@marcus-oscarsson @beteva 

It looks like we could agree to have set_limits raise NotImplementedError in the abstract class,

I have a further clarification in mind - which might matter for AbstractNPosition (that I have to write next) even if it does not here. My Proposal 2, in the comment to set_value is an example of how I think limits ought to work. Basically, if there are limits that you can set, the code must raise an error if you try to set a value outside the permitted range. (Ideally it should also guarantee that the result of get_value() is within the range, but that is less crucial). One consequence is that limits only make sense for numerical values. For enumerated numbers (like attenuation or zooms) they are irrelevant but harmless. For non-value enumerations (like 'IN' 'OUT', 'LOCKED') they make no sense, and, IMHO, should be avoided.

Could we agree on that?

HardwareObjects/abstract/AbstractActuator.py
beteva: I would rather leave this to the particular implementation of  a motor, as we do not want to mess with checking the limits if there are None of +_ Inf. It might be consistent, though, to add a method check_limits(), which can implement your suggested code. We can call it in set_value and have it empty by default. Signed @beteva and @marcus-oscarsson

beteva: OK ;-) Signed @beteva and @marcus-oscarsson

beteva:
Why don't we simply not implement set_limits in AbstractNPosition and leave the AbstractActuator raise the NotImplementedError. We would like, though to have the get_limits even if the values are non-numerical. Then for AbstractNPosition we simply do not implement check_limits (see comment above).

rhfogh:
@marcus-oscarsson @beteva 

OK, looks like we are there:

- Define get_limits in AbstractActuator and have it return (None, None) by default (by whatever implementation)

-   Define set_limits in AbstractActuator and have it raise NotImplementedError by default.

- Treat AbstractMotor.(gs)et_velocity the same (I assume)

- Make a function to check a value against limits, have it return True by default, and call it from set_value. I would like to call it validate_value(), or is_valid_value() rather than check_limits() because 1) there might be other things to check, like data type or contect-dependent limits, or just enmumeration membership. 2) in objects with several values, like Energy/Wavelength it is a little cleaner (limits of what?).

The only thing I still dislike is the idea of having limits on an enumerated, non-numerical value. But since that will be done on a case-by-case basis and so will not have implications for the abstract classes or general behaviour, we do not have to agonise about it.

beteva:
Youpii, we have a deal ;-)
@rhfogh, you are right, we should have the freedom of performing different checks, depending on the particular class. How about calling the method is_valid rather than set_limits. This comes in line of is_ready. As you said, it should return True by default.

rhfogh:
Not sure about 'is_valid'. I had in mind a function that took a value and checked whether that was a valid value; is_valid sounds like it is checking whether the *entire object* is valid, rather than the value passed in.

e.g. def is_valid_value(value):

beteva:
@rhfogh, I will then vote for validate_value, with possibility to check  one value (of any type) or a list of values, returning True/False

rhfogh:
validate_value is fine. But what is the use case for the list?

beteva:
I was thinking if we have such a method than it can be used when moving few motors synchronously than we can feed all the values at once. We do not need to specify in the Abstract class what the input value is.

rhfogh:
I cannot quite see how that would work. If you are moving several motors, you would have to call the validate_value function for each of them separately, since they will have different limits. As I see it, it comes down to a choice between 
'all(HO.validate_value(x) for x in values_list)' in the calling code
versus
'If isinstance(value, (list, tuple)) then ... else ...' in the function code
and I find the former version cleaner than overloading the function parameter.
But as you say, we do not need to specify that imemdiately.

beteva:
This is really am implementation detail bit in microdiff you cannot move all the motors, so there is an extra method with only one call to the underlying exporter. I thought that we can use just one validate as we already prepare a list of motor1,value1, ... so we could use it in validate value. Bur we really do not have to choose in the Abstract class.

rhfogh:
I would suggest to go ahead and merge the PR. As you say we do not need to choose now.

Out of curiosity, which object in microdiff has the method that takes a list of motor1, value1, etc.?

beteva:
Hi @rhfogh, as we've come to an agreement, could you, please, merge this PR.
As for the microdiff objetc, it is the move_sync_motors, which used to take a dictionary and mow a list and we construct ouf of name and position a string, which is sent to startSimultaneousMove of the jlib

rhfogh:
Merged. 

I still  do not see why any single motor would want to validate a list of values, though. The diffractometer, maybe yes, but that is not an AbstractActuator (and you would name the function diffrently). AFAICS an AbstractActuator would only validate positions for itself, not for otehr actuators, and would do them one at a time.

But we can look at that another time, if we think it is worth the trouble..

PR #466, opened 2020-02-12
Abstract actuator

beteva:
Exporter and Bliss Motor implementation


IvarsKarpics:


```diff

Issues
======
+ Solved 8
- Added 7
           

Complexity increasing per file
==============================
- HardwareObjects/abstract/AbstractNState.py  2
- Command/exporter/ExporterStates.py  1
- HardwareObjects/ExporterNState.py  4
- HardwareObjects/BlissMotor.py  1
- HardwareObjects/BlissNState.py  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4989763&bid=16473522)

rhfogh:
@beteva 

Comments, part 1

General
======

- Which classes should have a 'username'? If individual classes only, why put it in AbstractNState? If it is more general, why not put it in AbstractActuator? What is the relation between username and actuator_name and what they are for? Maybe this information should be put in comments in the code?

AbstractActuator
=============

- I would propose that validate_value should take only a value as a parameter, and should check against internal limits of the object. This could be coded already in the abstract class Also the documentation should be changed to say that there may be checks against other things than just the limits. For AbstractNState it should check against the predefined)_values.

- Should set_value not call validate_value?

- Why does _set_value need a timeout? Part of my idea behind set-value was that all the timeout-handling could be done at that level.

AbstractNState
============

- How should limit-specific functions behave in these classes? Should we have a function in the abstract class that raise NotImplemented (which could then be overridden in subclasses)?

- Should not get_predefined_values be removed? Those vslures are only set at initiation, if not harcoded, no? At thge very least the function should be renamed _get_predefined_values.

- In general, how should the predefined values be done?
Using dictionaries means there is no standardisation, no guards against modification, and the values are defined afresh in every config file for every synchrotron. We do not want
people to select in each individual case whether it should be  {'in':0, 'out':1} or {'IN':1, 'OUT':0}, or ...
I would recommend using Enums (but namedtuples or DataObjects could also do). 
You could still define the values in configuration, but you could also define e.g.
InOutEnum () so that all in/out NPosition clases used the same enum. but we can discuss the possibilities.

ISPyB calls. More romorrow,

Rasmus

beteva:
> @beteva
> 
> Comments, part 1
> # General
> 
>     * Which classes should have a 'username'? If individual classes only, why put it in AbstractNState? If it is more general, why not put it in AbstractActuator? What is the relation between username and actuator_name and what they are for? Maybe this information should be put in comments in the code?
Username is just a name for display. We may or may not need it. It is up to us to see if we put it in the AbstractActuator.
actuator_name is the generic representation of the ID of the object. It will be needed to refer to the object in the protocol used by the different control systems. Like in bliss it will give the object, in Tango it will be the "device name"...
> 
> 
> # AbstractActuator
> 
>     * I would propose that validate_value should take only a value as a parameter, and should check against internal limits of the object. This could be coded already in the abstract class Also the documentation should be changed to say that there may be checks against other things than just the limits. For AbstractNState it should check against the predefined)_values.
The validate_value may check for any valid parameter and not just for low/high value. I reckon I should change the limits to values, but would keep the method signature.
> 
>     * Should set_value not call validate_value?
It might. Have not thought yet.
> 
>     * Why does _set_value need a timeout? Part of my idea behind set-value was that all the timeout-handling could be done at that level.
Timeout is an artifact of a previous iteration. Will remove it.
> 
> 
> # AbstractNState
> 
>     * How should limit-specific functions behave in these classes? Should we have a function in the abstract class that raise NotImplemented (which could then be overridden in subclasses)?
I would rather prefer to leave it as it is now - return (None, None). We could eventually implement a has_limits, if needed.
> 
>     * Should not get_predefined_values be removed? Those valures are only set at initiation, if not hard-coded, no? At the very least the function should be renamed _get_predefined_values.
The predefined values are the possible states. They are either configured in an xml file, or given by the hardware if such a functionality exists. We do want the method.
> 
>     * In general, how should the predefined values be done?
>       Using dictionaries means there is no standardisation, no guards against modification, and the values are defined afresh in every config file for every synchrotron. We do not want
>       people to select in each individual case whether it should be  {'in':0, 'out':1} or {'IN':1, 'OUT':0}, or ...
You have a point.
>       I would recommend using Enums (but namedtuples or DataObjects could also do).
Agree for Enum
>       You could still define the values in configuration, but you could also define e.g.
>       InOutEnum () so that all in/out NPosition clases used the same enum. but we can discuss the possibilities.
> 
The predefined_values is very versatile, but we need to be more rigorous. We agree to configure Enums with predefined position (IN/OUT...) and when initialising the specific implementation of AbstractNState, we check which of these Enums we need, but we use the dictionary with the predefined values. We keep it this way because in some cases (e.g. aperture) we do not have well defined positions - they come out of the hardware.
We also need a way to attribute different values (like a factor) to a predefined position. Again the example is aperture, where in order to calculate the beam flux, you need to know what factor to multiply with, depending on the aperture value. Mind you, aperture is not the only example, but it is easy to explain.
> 
> ISPyB calls. More romorrow,
> 
> Rasmus

@marcus-oscarsson and @beteva

beteva:
The PR only contains 2 states Enums.  Rest to be defined next if we agree on this type of implementation.

IvarsKarpics:


```diff

Issues
======
+ Solved 9
- Added 11
           

Complexity increasing per file
==============================
- Command/exporter/ExporterStates.py  1
- HardwareObjects/BlissNState.py  6
- HardwareObjects/ExporterNState.py  4
- HardwareObjects/BlissMotor.py  1
- HardwareObjects/abstract/AbstractNState.py  5
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4989763&bid=16473522)

IvarsKarpics:


```diff

Issues
======
+ Solved 9
- Added 10
           

Complexity increasing per file
==============================
- Command/exporter/ExporterStates.py  1
- HardwareObjects/abstract/AbstractNState.py  4
- HardwareObjects/BlissNState.py  6
- HardwareObjects/BlissMotor.py  1
- Command/exporter/ExporterClient.py  4
- HardwareObjects/ExporterNState.py  4
         

Clones added
============
- Command/exporter/ExporterClient.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4989763&bid=16473522)

IvarsKarpics:


```diff

Issues
======
+ Solved 30
- Added 14
           

Complexity increasing per file
==============================
- HardwareObjects/abstract/AbstractNState.py  6
- HardwareObjects/ExporterNState.py  4
- Command/exporter/ExporterClient.py  4
- HardwareObjects/BlissNState.py  6
- Command/exporter/ExporterStates.py  1
- HardwareObjects/BlissMotor.py  1
- HardwareObjects/abstract/AbstractEnergy.py  1
         

Clones added
============
- Command/exporter/ExporterClient.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=4989763&bid=16473522)

PR #472, opened 2020-03-07
[WIP] HWR API Updates

marcus-oscarsson:
Dear all,

We are sorry for this big PR, we realize that it might be complicated to digest. We have made the initial changes to make things work with the new HWR API discussed during the last developers meeting. We have done this during the various discussions and that's way this have been growing for a while. 

Most of these changes are API updates but there are also a few bug fixes. I've tried to keep the "generic" method name replaces to a minimal but I've made the straightforward ones for everybody when I needed it myself. This as we wanted to make a another PR with all those "generic" updates following this one.

Its marked as WIP but is ready to be merged, I marked it as WIP in case you found the commit history to complicated. If it is the case Ill rework it but I did not want to spend to much time doing that. In anyways, just let me know :)

Cheers,
Marcus and Antonia


marcus-oscarsson:
Also, note that the last commit is a black formatting commit so the majority of the file changes comes from that ...

IvarsKarpics:
Uhhh, this is a BIG PR :)  Thank you for this massive work. If no one is against, then I will merge it.

Issue #442, opened 2019-12-09
Define AbstractBeam (partly from BemInfo)


marcus-oscarsson:
- Define a AbstractBeam,
- Rename BeamInfo to Beam

marcus-oscarsson:
Done

PR #476, opened 2020-03-11
Added previously missing file AbstractBeam

marcus-oscarsson:
Dear all,

I forgot to add the AbstractBeam that Antonia started to work on, this will probably create some conflicts with @IvarsKarpics latest PR #442, it would be good if we could merge these two.

Cheers,
Marcus

### DIFF
--- a/HardwareObjects/abstract/AbstractBeam.py
+++ b/HardwareObjects/abstract/AbstractBeam.py
@@ -1,0 +1,130 @@
+# encoding: utf-8
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""
+Beam abstract class - methods to define the size and shape of
+the beam, its presence.
+"""
+
+__copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+import abc
+from enum import Enum, unique
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+
+
+@unique
+class BeamShape(Enum):
+    """ Beam shape definitions """
+
+    UNKNOWN = "unknown"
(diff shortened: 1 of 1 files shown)


IvarsKarpics:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- HardwareObjects/abstract/AbstractBeam.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/mxcube/HardwareRepository/pullRequest?prid=5159440&bid=16941894)

HardwareObjects/abstract/AbstractBeam.py
```diff
@@ -0,0 +1,130 @@
...
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, name):
+        HardwareObject.__init__(self, name)
+        self.beam_width = None
+        self.beam_height = None
+        self.beam_shape = BeamShape.UNKNOWN
+        self.beam_label = None
+        self.default_beam_divergence = None
```
IvarsKarpics: Should it be a tuple?

HardwareObjects/abstract/AbstractBeam.py
```diff
@@ -0,0 +1,130 @@
...
+    def get_value(self):
+        """ Get the size (width and heigth) of the beam and its shape.
+        Retunrs:
+            (float, float, Enum, str): Width, heigth, shape and label.
+        Raises:
+            NotImplementedError
+        """
+        return self.beam_width, self.beam_heigth, self.beam_shape, self.beam_label
+
+    def get_availble_size(self):
```
IvarsKarpics: typo

HardwareObjects/abstract/AbstractBeam.py
```diff
@@ -0,0 +1,130 @@
...
+        """Get the beam dicergence.
+        Returns:
+            (tuple): Beam divergence (horizontal, vertical) [μm]
+        """
+        if self.beam_definer:
+            return self.beam_definer.get_divergence()
+        else:
+            return self.default_beam_divergence
+
+    def get_value(self):
```
rhfogh: get_value must return the same kind of value as you pass in with set_value. Here get_value returns a complex tuple while set_value takes a simple size tuple.

rhfogh:
I think this PR should be marked WIP - but have no objection to having it merged in for further work downstream.

The class is a subclass of HardwareObject (I agree) but it is coded as if it was an AbstractActuator, and here I do not agree. An AbstractActuator has a value, and it should be reasonably clear what that value is. Here the value could equally well be the size, the position, or the complete parameters dictionary, so using get/set_value would ikely cause confusion. I think it is better to code this with separate attributes for size, position, shape, and label, and to use 'parameters' or 'properties' or something similar to refer to the lot at once. Do we actually need a function that returns the lot at once? And if we do, should we maybe use a data type? These packed tuples of heterognous data are not ideal IMHO.

marcus-oscarsson:
@rhfogh  @IvarsKarpics, Antonia is on vacation so she can't comment right away. I cant really say why the size, position, shape, and label are not separate attributes. There is probably a reason for it, it could also be that its still in the draft stage. Antonia told me that she had a few things to finish up. So I agree with both of you that we can merge this as its partly used by the previous PR that were merged and then keep on improving on it.

PR #479, opened 2020-03-12
Renamed getLimits->get_limits getState->get_state; refactored TINEmotor

rhfogh:
Renamed getLimits to get_limits
Renamed getState to get_state
Refactored TINEMotor to match AbstractMotor
replaced set_limits with update_limits (it is not changing tghe hardware, only the nominal limits)

Equivalent pull request for mxcube


marcus-oscarsson:
Looks good :)

PR #480, opened 2020-03-12
[WIP] Simplified version of AbstractNState

marcus-oscarsson:
Dear all,

This is a simplified version of the AbstractNState object, it simply provides the means to define
N discrete states that needs to be defined in the subclass. I've made an example implementation for the micro diffractometer zoom (testing with the BZoom). 

It's not as flexible and configurable as the original version but is simpler and provides perhaps a more well defined interface. Objects like shutters or beam stops then have to be implemented as separate objects with clearly defined open/closed in/out states the values however needs to be mapped to the actual hardware.

Marcus

### DIFF
--- a/HardwareObjects/Microdiff.py
+++ b/HardwareObjects/Microdiff.py
@@ -430,7 +430,7 @@ class Microdiff(MiniDiff.MiniDiff):
             "phiz": float(self.phizMotor.get_value()),
             "sampx": float(self.sampleXMotor.get_value()),
             "sampy": float(self.sampleYMotor.get_value()),
-            "zoom": self.zoomMotor.get_value(),
+            "zoom": self.zoomMotor.get_value().value,
             "kappa": float(self.kappaMotor.get_value())
             if self.in_kappa_mode()
             else None,
--- a/HardwareObjects/ExporterNState.py
+++ b/HardwareObjects/ExporterNState.py
@@ -25,14 +25,6 @@ Example xml file:
   <exporter_address>wid30bmd2s:9001</exporter_address>
   <cmd_name>FluoDetectorIsBack</cmd_name>
   <state_definition>InOutEnum</state_definition>
-  <predefined_value>
-      <name>IN</name>
-      <value>True</value>
-  </predefined_value>
-  <predefined_value>
-      <name>OUT</name>
-      <value>False</value>
-  </predefined_value>
 </device>
 """
 from gevent import Timeout, sleep
@@ -58,7 +50,8 @@ class ExporterNState(AbstractNState):
     def init(self):
         """Initialise the device"""
         AbstractNState.init(self)
-        value_channel = self.getProperty("cmd_name")
+        value_channel = self.getProperty("value_channel_name")
+        state_channel = self.getProperty("state_channel_name")
 
         _exporter_address = self.getProperty("exporter_address")
         _host, _port = _exporter_address.split(":")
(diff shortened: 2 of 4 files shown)


HardwareObjects/Microdiff.py
```diff
@@ -430,7 +430,7 @@ def get_positions(self):
             "phiz": float(self.phizMotor.get_value()),
             "sampx": float(self.sampleXMotor.get_value()),
             "sampy": float(self.sampleYMotor.get_value()),
-            "zoom": self.zoomMotor.get_value(),
+            "zoom": self.zoomMotor.get_value().value,
```
IvarsKarpics: what is this?

marcus-oscarsson: get_value for the AbstractNState returns a enum so to get the actual value which it represents so one has to do .value. I simply kept the zoomMotor name for the moment but it should perhaps be renamed to simply zoom or zoom_hwobj if you prefer.

set_value and get_value of AbstractNState takes an enum variable for the moment, but we can discuss that if you prefer something else ...

IvarsKarpics: Ah ok. Thanks for the explanation. Yes, maybe we can talk about during the web meeting

rhfogh:
I like the simplification, but Antonia convinced me that the value returned by an NState ought to be a string, not an Enum member. As she told it, the integer or whatever that is returned from the hardware may differ from class to class, even if both have states IN and OUT. 
So instead of defining 
class MyEnum:
  IN="IN"
  OUT="OUT"
and then a separate mapping to track the returned state values returned locally, you could define e.g.
class MyLocalEnum:
  IN=1
  OUT=5
or whatever.

You would need two things:
- A way of saying that a given class was required to have values "IN" and "OUT" (that was the idea of teh state_definition, though it would surely be done better, maybe by making the list of names part of the class code)),
- A reasonably simple way of going from name to value and back; it would only be needed internally, since the the Enum member would never be exported, but I shall look for something bettern than accessing __members__

If we are thinking in these terms anyway, we should maybe do the same for state and specific_state as well?
class MyLocalEnum

rhfogh:
OK, I would like to withdraw the previous comment. Following our discussion it clearly makes sense to pass around enum instances, so you have both name and value available when you need them.
Presumably that would mean that you would normally define enums like
class MyLocalEnum:
IN=1
OUT=5
and external code would check on the names rather than the values of the enum instances?

Is there a way to know/ensure that all local implementations of e.g. a shutter will have the same names, or lest a shared subset? And is there a way to know/ensure that all Zooms will have values that are floats, and that large values correspond to zoomed-in (or is it zoomed-out)?

 This may not be necessary across the board - it is probably rare that generic code will manipulate shutters directly (though the GPhL workflow code does need to do zoom-to-maximum). But if we can produce some homogeneity it might be useful. Any ideas?

marcus-oscarsson:
> OK, I would like to withdraw the previous comment. Following our discussion it clearly makes sense to pass around enum instances, so you have both name and value available when you need them.
> Presumably that would mean that you would normally define enums like
> class MyLocalEnum:
> IN=1
> OUT=5
> and external code would check on the names rather than the values of the enum instances?
> 

Yes, knowing what kind of object we are dealing with and that it has N states with well defined symbolic names should be enough. The values of those symbols can of course change depending on configuration. 

> Is there a way to know/ensure that all local implementations of e.g. a shutter will have the same names, or lest a shared subset? And is there a way to know/ensure that all Zooms will have values that are floats, and that large values correspond to zoomed-in (or is it zoomed-out)?
> 

Not without specifically defining that a certain subclass have a certain behavior.   

> This may not be necessary across the board - it is probably rare that generic code will manipulate shutters directly (though the GPhL workflow code does need to do zoom-to-maximum). But if we can produce some homogeneity it might be useful. Any ideas?

This is why I included limits for the MicrodiffZoom, limits = (min, max). But another concept might be more suitable but this is also very dependent on what kind of hardware we are implementing. for A zoom it makes sense. We could even implement a zoom subclass that has a method that returns the zoom levels in order [LEVEL1 ... LEVELN-1 LEVELN] which corresponds to increasing zoom. But I believe it would be difficult to apply any general rule for all NStates

rhfogh:
> Not without specifically defining that a certain subclass have a certain behavior.   

Is there a way to do that? Or at least to encourage harmonisation?

jordiandreu:
Hi everyone,
I think it would be interesting a more dynamic way to create the enum class fro configuration, which can be done by using the functional API from Enum module:

MyAbstractStates = Enum('MyEnumClass', {name1:value1, name2:value2})

This facilitates the definitons of different NState objects.

In the other hand, I think that setting or requesting the certain states of interest (maximum zoom, minimum zoom, open, close, ...) maybe should be implemented by an specific API for the derived class and do not force a common state definition for each actuator.

I can propose a PR or add the changes to this (as is in in WIP).

rhfogh:
@jordiandreu 
Sounds like a good idea to me - but it is probably Marcus and Antonia who are most on top of this for now.

marcus-oscarsson:
@jordiandreu, I believe that's to a large extent  what we are already doing. see for instnace line:

https://github.com/mxcube/HardwareRepository/blob/8993662039ae85a7de8a2804bb778ba56e5396d7/HardwareObjects/MicrodiffZoom.py#L59

@beteva is back from vacation today so she will continue with this as soon as she can. We have a very special situation at ESRF today (and for some weeks to come) as the site is now closed for an unknown period of time.

We spoke briefly today though and our aim is to sort this out this week.

Cheers,
Marcus

marcus-oscarsson:
Sorry I managed to close this PR with some clumpy git commands, I re-opened it in #484 (see above)

marcus-oscarsson:
@jordiandreu, @all: The idea was to fix the enum names for a particular class so that we know what to expect in terms of names for the enum. Its clear that the values needs to be configurable somehow and it would be good if we could avoid an additional layer of objects just to do that. That's why we decided to only make it possible to configure the values in this way.

I'm not necessarily against also making it possible to define the names by configuration, but that would mean that we need to define some kind of convention for the different types. Which I'm fairly sure will be broken over time. But there is maybe a solution to that ?

I used the ast.eval_literal since it does exactly what I would like to. While eval evaluates any code and then throws an exception if it fails. I dont think its a big issue but I like to avoid a full eval when I can.

jordiandreu:
> I'm not necessarily against also making it possible to define the names by configuration, but that would mean that we need to define some kind of convention for the different types. Which I'm fairly sure will be broken over time. But there is maybe a solution to that ?

My idea is to hide the states behind the API for the given Hardware Object instead. I feel like it is not necessary to have to manage the Enum itself but the member instead: The name could be interesting for the Front end part while the value for the control of the instrument.  Instead, you can define an API for the top level Abstract Hardware objects i.e.

* AbstractNState provides the mechanism to configure the state enum.
* AbstractZoom(AbstractNState) provides an API to query values required by the main application, let say is_zoom_max, is_zoom_min, etc.

Is this the additional layer you mention you prefer to avoid?

With your implementation, maybe one will tend to copy/paste  the configuration code from class to class.

marcus-oscarsson:
I see, I think the each subclass with each with its particularities needs specific implementation like you mention above. The values for a symbol in, out, open and so on needs to be configured depending on hardware (as you know). Its here I would like to avoid an additional mapping and keep it directly configurable. This configuration can't be copied unless you have two zooms of the same kind of course.

In anyways, @beteva is just back from vacation and will merge this with her ideas. It seems like you have some ideas as well, so I suggest that we try to synchronize these with @beteva's ideas. So maybe you can push your ideas ones she has made her updates, if there are still things to add ?

PR #481, opened 2020-03-16
Get loggers at the HO level

jordiandreu:
Get the HWR logger and the user logger at the HO
level in a hierarchical fashion.


marcus-oscarsson:
What does it mean that its hierarchical ?

jordiandreu:
Hierarchical in the sense explained in the issue (issue https://github.com/mxcube/HardwareRepository/issues/482). The logger name includes also the class from which the log message is sent.

No one is forced to use this but it simplifies the logging and the method print_log from HardwareObject class could be removed (not in the PR).

marcus-oscarsson:
Sounds good to me, I'm fine with merging this, @all any other comments ?

BaseHardwareObjects.py
```diff
@@ -575,6 +575,8 @@ class HardwareObject(HardwareObjectNode, HardwareObjectMixin):
     def __init__(self, rootName):
         HardwareObjectNode.__init__(self, rootName)
         HardwareObjectMixin.__init__(self)
+        self.logger = logging.getLogger('HWR').getChild(self.__class__.__name__)
```
IvarsKarpics: maybe it could be self.hwr_logger and self.gui_logger. Just to be clear.

jordiandreu: I would prefer shorter names. We do have only two loggers, the standard logger to file, intended for developers and to track the  application performance and an special logger for notifications of interest for the users (GUI).

For me, _self.log_ and _self.user_log_ would be also valid options...maybe even better than my first choice!  What do you think about?

marcus-oscarsson: I see, I guess its something that we use quite a lot so a shorter name would be nice. I would personally avoid gui_logger since we could confuse that with log messages coming from the UI while the intention of the logger is to send messages to the user :). Otherwise log or hwr_log or hwr_logger all works for me.

marcus-oscarsson:
Whats the status on this one, is still WIP ?

jordiandreu:
@ALL we should agree on the names  for these two loggers. Once fixed, the implementation is backward compatible and everyone is free to use it.

I vote for short names:
* self.log (report to log file)
* self.user_log (report to log widget)

marcus-oscarsson:
Fine for me +1

IvarsKarpics:
Fine also for me

jordiandreu:
@ALL, I will update the PR with the latest naming convention we agreed

marcus-oscarsson:
I'm not really working today :), but in anyways ... I guess that means we can remove the WIP and merge ?

jordiandreu:
> I'm not really working today :), but in anyways ... I guess that means we can remove the WIP and merge ?

Yes, it can be merged now.

PR #483, opened 2020-03-16
Refactor move

rhfogh:
- Changed moveRelative and syncMoveRelative to set_value_relative, and syncMove to set_value
- Refactored existing set_value, setValue, and getValue commands
- changed CommandContainer commands to snake_case
- removed some deprecated synonym functions
- Some refactoring of state_changed, value_changed, limits_changed

More later - still done nothing about 'move' itself, for instance.


HardwareObjects/ALBA/ALBACatsMaint.py
```diff
@@ -11,11 +11,11 @@ def init(self):
         CatsMaint.init(self)
 
         # load ALBA attributes and commands from XML
-        self._chnAtHome = self.getChannelObject("_chnAtHome")
-        self.super_abort_cmd = self.getCommandObject("super_abort")
+        self._chnAtHome = self.get_channel_obect("_chnAtHome")
```
jordiandreu: This should read _object_ instead of _obect_, right?

rhfogh: Indeed.I shall fix it.

marcus-oscarsson:
Nicely done !

PR #484, opened 2020-03-16
[WIP] Simplified version of AbstractNState

marcus-oscarsson:
Dear all,

This is a simplified version of the AbstractNState object, it simply provides the means to define
N discrete states that needs to be defined in the subclass. I've made an example implementation for the micro diffractometer zoom (testing with the BZoom).

It's not as flexible and configurable as the original version but is simpler and provides perhaps a more well defined interface. Objects like shutters or beam stops then have to be implemented as separate objects with clearly defined open/closed in/out states the values however needs to be mapped to the actual hardware.

Marcus


marcus-oscarsson:
Sorry made a small git mistake so the old PR #480 was closed, here is the same PR again since I could not revert or reopen the old one.

beteva:
Dear @ALL, there was a lot going around the NStates in just one week. Here are my comments:
Maybe first i should remove the confusion between states and values. Sorry if I repeat myself, this is just to fix some bases for the terms we use
1. The values are IN/OUT, OPEN/CLOSE, zoomlevel1/.../zoomleveln etc. These values are read with get_value, set with set_ value methods and are updated with update_value, 
2. The states are the one read with get_state and updated with update_state. The minimum set of states is the one, defined in BaseHardwareObject.
All the above methods are nherited from AbstractActuator.
In the AbstarctNState implementation what we want to fix is the "label" of the value. The value itself (1,2,10, "PARK" or whatever is only local for the specific class, but its name is fixed.
Sometimes on top of the specific value, we need to introduce some other parameters (e.g. a factor for calculating the flux, when the aperture changes).
There are two different ways of obtaining all the above - Enums or dictionaries.
In case of Enums, the extra parameters can be added (like AP1 = 10, 22e-7). The type of the Enum is knows from the setup. The client knows that if you handle aperture, the value is ENUM.AP1.value[0].
In case of dictionary, we can fix "name" and "value" keys , so the dictionary looks like {"name": "AP1", "value": 10, "factor":  22e-7}.
I will try push this afternoon the code for adding extra parameters with the Enum solution. The dictionary solution is the one in the master. We can discuss afterwards.
Cheers

marcus-oscarsson:
@rhfogh: this is really off-topic. are you reading your mail ?

rhfogh:
@beteva I was off line for a day or so, because of a problem with our broadband. Back on now, and will get to MXCuBE after lunch.

jordiandreu:
@beteva, I've having a look at this PR which I understand still in WIP. Here are my comments:

1. I think the method name **get_values** is a bit confusing since we are inheriting the methods **get_value** and **set_value** from AbstractActuator.  In the other hand, the **get_state** method is the one  inherited from the HardwareObject  class.  I would propose to change the method to something like **get_state_values** or **get_enum_values**.

2. Is there a reason to overwrite the  methods **update_limits** and **get_limits** inherited from AbstractActuator?

3. Should the **value_to_enum** method be private? In principle, as  I  understood from your last comment, the name of the enum would be used by other objects to operate/monitor the  state values and the values of the Enum class will be only used for internal operation of the N-State Actuator?

4. I would suggest to implement the **get_values** method only for the abstract class i a general  way. Then it will  not be needed to re-implement this method in derived classes.

5. The name of the enum class created is hardcoded in the abstract class as 'ValueEnum'. We could pass the name as part of the configuration, then you could use the _isisntance_ method to distinguish between different enum state values.

6. The BaseValueEnum class definition imposes a string value  for the UNKNOWN state value. Maybe the value for the UNKNOWN state should be defined in the configuration somehow, according to the type used  for the enum values.

rhfogh:
@jordiandreu @beteva 
I have not quite finished looking, but this would be a good time to add my comments:

1.  The name get_values is indeed confusing. And you do not need a getter anyway, since you can always get the enumeration as cls.VALUES, or instance.VALUES. What you do need is a way to initialise the VALUES enum, as indeed this function does. You could get rid of the function and put the code inside init() if you wanted, but there is nothing wrong with having the function, provided you use with a different name. 
Someting like initialise_enum(), setup_values, ... etc. 
In practice it would be called only from init, but with such a name it would not hurt to have the function public

2. get_limits is not overwritten. set_limits and update_limits are - the reason is that it is bad to have a set_limits function if setting the limits do not have any effect on what happens subsequently. 

3. Might be a good idea to make it _value_to_enum, yes.

4. Good idea - but AFAICS that is what is already being done, in AbstractNState.

5. No need to change the name. You can get hold of the enumeration as cls.VALUES or  obj.VALUES, and do isinstance anyway.

6. As I understand it, the values in the enum are only for mapping from the states returned by the motors to the sttes shown in MXCuBE. If there is a motor state that matches 'UNKNOWN' you would need a different value, but then you could override the use of BaseValueEnum. If not it might not be a problem if the type was different.

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -59,63 +46,24 @@ class AbstractNState(AbstractActuator):
...
-        for value in self.getProperty("predefined_value", ()):
-            try:
-                predefined_values.update(
-                    {value.getProperty("name"): value.getProperty("value")}
-                )
-            except AttributeError:
-                pass
-
-        return predefined_values
+        return value.name in self.VALUES.__members__
```
rhfogh: Your code returns True for a different enum with the same name.
I think that is problematic - if we use enums as the values, we ought to check on object identity - and in that case it could be done simpler as:

       return value in self.VALUES

If we only want to compare names, then it might be better to have functions that pass around only strings.

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -130,3 +78,38 @@ def update_limits(self, limits=None):
...
+            values = ast.literal_eval(self.getProperty("values"))
+            self.VALUES = Enum(
+                "ValueEnum",
+                dict(values, **{item.name: item.value for item in BaseValueEnum},),
+            )
+            return self.VALUES
+        except (ValueError, TypeError):
+            return self.VALUES
+
+    def value_to_enum(self, value):
```
rhfogh: Two minor  things:
1) Would it not be more readable if the loop starts with

    if value == enum_var.value:
        return enum_var

i.e. you do that test first, whether the value is a tuple or not?

2) I know ther is a case, but could you explain to me when you need a value that is a tuple but testing only on the first element fo teht tuple?

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -130,3 +78,38 @@ def update_limits(self, limits=None):
...
+    def get_values(self):
+        """Get the predefined valies. Create the VALUES Enum
+        Returns:
+            (Enum): "ValueEnum" with predefined values.
+        """
+        try:
+            values = ast.literal_eval(self.getProperty("values"))
+            self.VALUES = Enum(
+                "ValueEnum",
+                dict(values, **{item.name: item.value for item in BaseValueEnum},),
```
rhfogh: The comma after 'BaseValueEnum}' is illegal in Python 2

rhfogh:
I rather like the way this is looking (all the comments are for naming and other minor matters).

It would really have been great to define e.g. 'InOutNState' classes, so that you could know what the state enum was for a given kind of object across all implementations. Regrettably it looks like this falls in the 'too hard' basket. So this simpler and cleaner approach is (unfortunately) likely to be the best possible, certainly for now.

HardwareObjects/BlissNState.py
```diff
@@ -76,33 +71,34 @@ def init(self):
...
-            self._bliss_obj.move(self.predefined_values[value], wait=False)
+            self._bliss_obj.move(value, wait=False)
         elif self.device_type == "actuator":
-            _attr = "set_" + self.predefined_values[value].lower()
+            _attr = "set_" + value.lower()
             _cmd = getattr(self._bliss_obj, _attr)
             _cmd()
-            self.__saved_state = value
 
     def get_state(self):
```
rhfogh: The code in this functoin is very complex to analyse. It looks to me that _state is first set to a string, and later reset to an enum value; also that the function can return a string in some situations, specifically when the state is a specific sate (so self.SPECIFIC_STATES.__members__[_state].value[0] does not raise an error). That looks a bit worrying to me.

rhfogh:
@beteva 
BlissNState._set_value takes either a string or an enum value as parameter. In general I think it is better to insist on passing in values in one form rather than letting people choose several and disambiguating in the function; that makes for cleaner code. I this specific case there may well be a good reason; could you explain what situation you have in mind

jordiandreu:
@rhfogh, 
> The name get_values is indeed confusing. And you do not need a getter anyway, since you can always get the enumeration as cls.VALUES, or instance.VALUES. What you do need is a way to initialise the VALUES enum, as indeed this function does. You could get rid of the function and put the code inside init() if you wanted, but there is nothing wrong with having the function, provided you use with a different name.
> Someting like initialise_enum(), setup_values, ... etc.
> In practice it would be called only from init, but with such a name it would not hurt to have the function public

 Agree

> get_limits is not overwritten. set_limits and update_limits are - the reason is that it is bad to have a set_limits function if setting the limits do not have any effect on what happens subsequently.

Indeed, it does not sound good to me to raise a _NotImplementedError_ when the method belongs to a parent Abstract class and it has a valid implementation.
 
> Might be a good idea to make it _value_to_enum, yes.

> Good idea - but AFAICS that is what is already being done, in AbstractNState.

Not really, since I think it is also implemented in BlissNState class.
 
> No need to change the name. You can get hold of the enumeration as cls.VALUES or obj.VALUES, and do isinstance anyway.

I don't know if I understand your point here..doing _isinstance(self.VALUES) would return always ValueEnum, isn't it?_

> As I understand it, the values in the enum are only for mapping from the states returned by the motors to the sttes shown in MXCuBE. If there is a motor state that matches 'UNKNOWN' you would need a different value, but then you could override the use of BaseValueEnum. If not it might not be a problem if the type was different.

Well, I think that the idea is that in the Microdiffzoom class you can use the values to store the magnification factors, for instance, or any other value or set of values. Could you then overwrite the UNKNOWN.value in a derived class?

HardwareObjects/abstract/AbstractNState.py
beteva: Thanks for checking it. The coma was added by black. I will remove it.

HardwareObjects/BlissNState.py
beteva: We have two distinct bliss implementation, which we make look the same way in the HardwareObject. The bliss InOut object returns IN or OUT only, which is at the same time state and value. The __saved_state only serves to check if the value we want to set corresponds to the bliss state once done. The string is for bliss, the Enum is for the HardwareObject. The bliss MultiplePositions has distinct value and state. The state is handled by the  SPECIFIC_STATES part. It does not rise any error, because the try/except is only a short way to check  if what is back from the bliss object Enum is a tuple or a string and return the appropriate HardwareObjectState. If there is an error, it is handled with the try/except of the beginning of the get_state.

rhfogh:
@jordiandreu 

Point 2:
The problem is that AbstractNState objects do not actually *have* limits, except maybe in a few special cases. I think it is very bad to have a set_limits function on an object where the limits you set do not actually have any effect. Well, is sets the _nominal_limits attribute, but that attribute is never used for anything, because limits are not defined for e.g. an IN/OUT actuator. Less deceptive to make set_limits raise NoImplemented than to have settable limits and let people think they make a difference, when they do not. One could also argue that the limits should be moved out of AbstractActuator and into AbstractMotor (or a new subckass) but other people preferred this solution, and I am OK with it.

Point 4.
get_values in AbstractNState is a good general function, but in BlissNState and MicroDiffZoom the situation s different so they need a different function. So they override. I do not think you can have a single function that fits all three cases.

Point 5
I was thinking of: "isinstance(aValue, self.VALUES)". Does that not do what we want, already?

Point 6.
You can override the 'UNKONWN' value in a subclass - but only because BaseValueEnum is never actually subtyped. You cannot add new enum values in a subclass, unless the superclass has none. What the code does is to make anew enum and copy the name,value pairs from BaseValueEnum into the new enum, and at this step you could easily add a new value instead of 'UNKNOWN', if you need one.

HardwareObjects/BlissNState.py
rhfogh: OK, I think II understand - thanks for the explanation.

But in that case, should .the second try: statement not start:

   try:
      _state = self.STATES.__members__[_state]

As it is, the _state variable is not changed from the original string when the try: succeeds.

rhfogh:
@beteva 
In BLissNState.init, it says:

        if "MultiplePositions" in self._bliss_obj.__class__:

Are you sure that should not be:

        if "MultiplePositions" in self._bliss_obj.__class__.__name__:

HardwareObjects/abstract/AbstractNState.py
rhfogh: OK, I understand the tuple now.

A niggle:
As written, the code would map value='U' to an enum_var with value 'UNKNOWN', because

                if value == enum_var.value[0]:

would be True.
Better to guard against that?

beteva:
Thanks @jordiandreu and @rhfogh for being so reactive on the code. First I would like to say that it has to be seen as one all, because the particular implementation shows the reasoning behind the abstract code (and what's behind my female mind ;-) As there a a lot of cross conversations, i will explain my ideas in a separate one, using @jordiandreu numbering.

1. **get_values may** be confusing. it is the method, which gives back the predefined values, which are to be used with each specific object. As afterwards these values will be red with **get_value**, I thought that the method should have somehow similar name. I am ready to change it to a one which is less confusing, but agreed by everyone. I did not put the code inside init(), because I wanted to use it in different inherited classes (one example is BlissNState).

2. get/set/update limits - in the case of AnstractNState, we want to be sure that we do not do anything with the limits, unless is needed fro a particular implementation. Hence the NotImplementedError. In a particular implementation (like MicroiffZoom) we want to get the limits in order to display them in the GUI. If my reasoning is wrong, feel free to provide the new code.

3. **value_to_enum** is used in different classes. i thought that it is better to be public.

4. **get_values** is implemented in AbstractNState. I need to overload it in BlissNState, but not in ExporterNState. What is wrong?

5. I thought that we can have a hard coded name ValueEnum, because  we can always test against ValueEnum if there is a need. You can, anyhow overwrite or re-use it when creating you own Enum. As @rhfogh pointed out, we will only use self.VALUES or obj.VALUES in the GUI.

6. @rhfogh is right, in MicrodiffZoom the values are the magnification factors, the state of a motor is a state, not a value. If you motor has moved somewhere, where you do not have a magnification factor, than you will have a value of VALUES.UNKNOWN. The state of the motor can be STATES.READY or STATES.UNKNOWN.

HardwareObjects/abstract/AbstractNState.py
beteva: I am puzzled, as I thought that the enum is the one used in the class. Could you, please, explain me when you get a different enum from the one with the value you need to check.

You are right, We can simply do
<code>return value in self.VALUES</code

As you know, the validate value is used in set_value, so it cannot take anything else, but the value, hence an Enum and not string.

rhfogh:
@beteva 

On point 1 I think it would be better with a name based on 'setup', 'init', or 'configure'. In current code the function is never actually used to get values, and the main effect is to set self.VALUES from the configuration. I would say that as coded, the fact that you return the new self.VALUES is a side effect, if not unnecessary, which makes a name starting with 'get' a bit misleading.

On point 3 you could do it either way.

For the rest of your points I definitely agree with you.

HardwareObjects/abstract/AbstractNState.py
beteva: You are right, I'll change the code.

HardwareObjects/abstract/AbstractNState.py
rhfogh: @beteva 
The enum can only ever be the one in the class. But the value could be anything - people might call the function from their own code (e.g. GPhL code). If they did not know any better, they might use an IN or OUT enum value from the enum of a different class, or one they had made up for themselves, and in that case it is better to check on the enum object rather than just its name. 

But I was getting overcomplicated here, wondering if there was some subtle purpose behind that particular coding.

    return value in self.VALUES

clearly does what we want.

jordiandreu:
> Point 2:
> The problem is that AbstractNState objects do not actually have limits, except maybe in a few special cases. I think it is very bad to have a set_limits function on an object where the limits you set do not actually have any effect. Well, is sets the _nominal_limits attribute, but that attribute is never used for anything, because limits are not defined for e.g. an IN/OUT actuator. Less deceptive to make set_limits raise NoImplemented than to have settable limits and let people think they make a difference, when they do not. One could also argue that the limits should be moved out of AbstractActuator and into AbstractMotor (or a new subckass) but other people preferred this solution, and I am OK with it.

Ok, it is a matter of taste, then...if everyone agree on that, I do so.

> 
> Point 4.
> get_values in AbstractNState is a good general function, but in BlissNState and MicroDiffZoom the situation s different so they need a different function. So they override. I do not think you can have a single function that fits all three cases.

My feeling is that _get_values_ is the initialization method for the Enum class in any child class of AbstractNState. The getProperty method is not enough to digest the data structure needed in the configuration file. The implementation of the get_values depends directly on how the information in the configuration file is organized.  If we could agree on this, there is no need to overwrite this method and the Enum classes will have a common  structure  shared among all classes and sites.

> 
> Point 5
> I was thinking of: "isinstance(aValue, self.VALUES)". Does that not do what we want, already?
> 
Yes, you are right.

> Point 6.
> You can override the 'UNKONWN' value in a subclass - but only because BaseValueEnum is never actually subtyped. You cannot add new enum values in a subclass, unless the superclass has none. What the code does is to make anew enum and copy the name,value pairs from BaseValueEnum into the new enum, and at this step you could easily add a new value instead of 'UNKNOWN', if you need one.

Yes, but this approach forces you to overwrite the get_values method, which is in contradiction in my general approach, but is  ok  if we do agree  on the current implementation.

beteva:
> > Point 4.
> > get_values in AbstractNState is a good general function, but in BlissNState and MicroDiffZoom the situation s different so they need a different function. So they override. I do not think you can have a single function that fits all three cases.
> 
> My feeling is that _get_values_ is the initialization method for the Enum class in any child class of AbstractNState. The getProperty method is not enough to digest the data structure needed in the configuration file. The implementation of the get_values depends directly on how the information in the configuration file is organized. If we could agree on this, there is no need to overwrite this method and the Enum classes will have a common structure shared among all classes and sites.
> 
Maybe you can suggest how this function should look like. for me it is difficult ot see how we can make this match avery case.

From my side, let me explain why I am overloading the method:
for MicrodiffZoom, one can only get the number of levels from the application, and with the B-Zoom this can be dynamic. As for the BlissNState, the values, in the case of a "motor" are red from the bliss object, and it is less error prone to read them, rather than define them in an xml file.

jordiandreu:
> Maybe you can suggest how this function should look like. for me it is difficult ot see how we can make this match avery case.

> From my side, let me explain why I am overloading the method:
for MicrodiffZoom, one can only get the number of levels from the application, and with the B-Zoom this can be dynamic. As for the BlissNState, the values, in the case of a "motor" are red from the bliss object, and it is less error prone to read them, rather than define them in an xml file.

Maybe you could stripe this code from get_values:

`values = {val.upper(): val for val in self.bliss_obj.positions_list}`
 `self.VALUES = Enum(
        "ValueEnum",
        dict(values, **{item.name: item.value for item in BaseValueEnum},) )`

and insert it in the initialization, right inside the clause:

 `if "MultiplePositions" in self._bliss_obj.__class__:`

Then, protect the the loading of the configuration when done from the xml file, in the get_values method from the abstract class.

HardwareObjects/BlissNState.py
beteva: This part should be changed once we clarify how to handle the SPECIFIC_STATE. In any case it is not the `_state = self.STATES.__members__[_state]`

rhfogh: @beteva 
You are right, that was a booboo on my part. I meant:

       _state = self.SPECIFIC_STATES.\_\_members\_\_[_state].value[0]

beteva: :-)

beteva:
Dear @ALL, I suggest to rename **get_values** to **init_values**. Is this a name which you find reasonable. And indeed it does not need to return anything, only define the self.VALUES Enum

rhfogh:
@beteva 
init_values sounds good. Maybe initialise_values? (init_values could have been an attribute)

beteva:
> @beteva
> init_values sounds good. Maybe initialise_values? (init_values could have been an attribute)

OK for initialise_values

rhfogh:
@beteva 
Looking good, but (as ou might have guessed)  I still have some comments:

BlissEnergy, AbstractEnergy:

- Should we not make AbstractEnergy an AbstractMotor? Then we would have the implementation of  validate_value and set_value for free, and would not have to duplicate them here. What would be the downside?

- I think it would be better _not_ to update _nominal_value, _nominal_limits, _state in the respective get_ functions (get_value, get_state etc.). As I had understood it, get_ functions should only get, and should address the motors directly. All updating should be handled in the update_ funcitions only (update_value, update_state, etc.). On general principles it is better to have separate getter and updater functions with clearly defined tasks. More specifically, if e.g. update_state is supposed to do something (send a signal, change the ready_event, ...) every time the state changes, it is a potential bug if _state (etc.) can be changed without calling update_state.

- AbstractNState..value_to_enum:
  The code is OK, but the code below is simpler and does the same job:

        for enum_var in self.VALUES.__members__.values():
            if value == enum_var.value:
                return enum_var
            if isinstance(enum_var.value, tuple) and value == enum_var.value[0]:
                return enum_var

- A general point, for discussion: Is it useful to store self.default_value as an attribute, or would it be nough to just set it as teh starting value of _nominal_value?

beteva:
> AbstractNState..value_to_enum:
> The code is OK, but the code below is simpler and does the same job:
> 
> ```
>     for enum_var in self.VALUES.__members__.values():
>         if value == enum_var.value:
>             return enum_var
>         if isinstance(enum_var.value, tuple) and value == enum_var.value[0]:
>             return enum_var
> ```

@rhfogh Thanks for optimising the code.  I've changed it several times (with the discussions going on) and did not clean it up. I'll commit this change and you can merge the PR (the conflicts are solved now). Thanks

beteva:
@rhfogh 
> BlissEnergy, AbstractEnergy:
> 
> Should we not make AbstractEnergy an AbstractMotor? Then we would have the implementation of  validate_value and set_value for free, and would not have to duplicate them here. What would be the downside?

For me the down side is that the set_value for the energy is not just moving a motor. It can be a complicated procedure (as on all ESRF tunable beamlines), which involves changing of gap, realignment ...  that's why in BlissEnergy we not only define _set_value, but have to overload the set_value. For me this was easier. Otherwise I would had to add to AbstractEnergy/update_value a code, which might be ESRF need only. It is true that the code is repeated, but I thought that treating it as a motor could be misleading.

PR #485, opened 2020-03-16
Fix mockup py27qt4

jordiandreu:
Several fixes to be able to  run the mockup qt4-py2 application.
It comes together with https://github.com/mxcube/mxcube/pull/382.
Some change will be required once the AbstractNState class gets accepted.


HardwareObjects/Beamline.py
```diff
@@ -631,14 +632,15 @@ def get_default_acquisition_parameters(self, acquisition_type="default"):
...
 
-        try:
-            acq_parameters.detector_mode = self.detector.get_detector_mode()
-        except:
-            logging.getLogger("HWR").warning(
-                "get_default_acquisition_parameters: "
-                "Could not get detector_mode, setting to ''"
-            )
-            acq_parameters.detector_mode = ""
+        # TODO: REVIEW, get_detector_mode not in the AbstractClass
```
IvarsKarpics: I think we agreed that this will be roi_mode. So: self.detector.get_roi_mode:

https://github.com/mxcube/HardwareRepository/blob/eec64835420a37b21980293baffdced6997035a1/HardwareObjects/abstract/AbstractDetector.py#L114

HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -84,6 +84,13 @@ def get_wavelength_limits(self):
         )
         return self._wavelength_limits
 
+    def get_energy_limits(self):
```
IvarsKarpics: hmm. Do we need this? We have def get_limits()

HardwareObjects/mockup/MotorMockup.py
```diff
@@ -63,7 +63,8 @@ def init(self):
         self.set_limits(limits)
 
         # self.set_state(self.motor_states.READY)
-        self.set_value(float(self.getProperty("start_position", DEFAULT_POSITION)))
+        #self.set_value(float(self.getProperty("start_position", DEFAULT_POSITION)))
+        self.set_value(float(DEFAULT_POSITION))
```
IvarsKarpics: We still need to get the default value from xml. Maybe rename start_position to default_position and change all xmls where it appears, like detector-distance-mockup.xml

HardwareObjects/queue_model_objects.py
```diff
@@ -1660,7 +1660,7 @@ def __init__(self):
         self.camera_hw_roi = 0
         self.camera_write_data = True
 
-        self.detector_distance = float()
+        self.detector_distance = 0.0# float()
```
IvarsKarpics: hmm. 0.0 is a valid value. Maybe None?

HardwareObjects/Beamline.py
jordiandreu: Fine for me

HardwareObjects/abstract/AbstractEnergy.py
jordiandreu: I see...the get_limits is inhereted from the AbtractActuator...so, the nominal value and limits correspond to energy and wavelenght is an extra attribute of the Energy object, right?

IvarsKarpics: get_value and get_limits corresponds to energy in keV. If one needs values in A then get_wavelenght and get_wavelength_limits are available.

HardwareObjects/mockup/MotorMockup.py
jordiandreu: I agree.

jordiandreu: I could not find any appearance of _start_position_ in the xml files, only _start_value_ in *detector-distance-mockup.xml*

HardwareObjects/queue_model_objects.py
jordiandreu: Sorry, float() is fine for me. It was just a test value when debugging an invalid validator I found in XRayParametersBrick (see PR in mxcube repo)

HardwareObjects/Beamline.py
jordiandreu: I've seen that in acquisition parameter both attributes exists: _detector_mode_ and _detector_roi_mode_. What should then be the detector mode value?

HardwareObjects/abstract/AbstractEnergy.py
jordiandreu: Fine. Changes done.

HardwareObjects/mockup/MotorMockup.py
jordiandreu: Done

HardwareObjects/queue_model_objects.py
jordiandreu: Done

HardwareObjects/Beamline.py
IvarsKarpics: Hmm, we need both :) Binning mode for historical reasons is stored in ispyb and roi mode is used the acquisition parameters. We should have self.detector_binning_mode and self.detector_roi_mode

jordiandreu: So, shall I replace _detector_mode_  by _detector_binning_mode_ in the acquisition parameters defined in the queue_model_object?

marcus-oscarsson: detector_roi_mode is the ROI mode and the detector_mode is simply the binning. It would actually be good to as you say replace detector_mode by detector_binning_mode. I would vote for that

marcus-oscarsson:
I'm fine with these changes I think that update renaming detector_mode would be a plus but its maybe not necessary in this PR

jordiandreu:
So, I would propose to add the _detector_roi_mode_ and maintain the _detector_mode_ as it is.
Then, I would open a new PR proposing the rename of the _detector_mode_ to _detector_binning_mode_.

HardwareObjects/Beamline.py
jordiandreu: I have added the _detector_roi_mode_ and kept the _detector_mode_ as  it is.

marcus-oscarsson:
Sounds like a good solution to me

HardwareObjects/Beamline.py
```diff
@@ -631,15 +632,25 @@ def get_default_acquisition_parameters(self, acquisition_type="default"):
             )
             acq_parameters.shutterless = False
 
+        # TODO: Rename detector_mode to detector_binning_mode(?).
         try:
-            acq_parameters.detector_mode = self.detector.get_detector_mode()
+            acq_parameters.detector__mode = self.detector.get_mode()
```
marcus-oscarsson: Ohh, wait en extra underscore here ....

jordiandreu: yes...fixed!

marcus-oscarsson:
@IvarsKarpics are you fine with merging this ?

HardwareObjects/Beamline.py
beteva: I also vote for detector_binning_mode. We should probably ask the ISpyB team to change a bit their model so we can have both roi and binning, but only if used.

IvarsKarpics:
Looks good to me

PR #487, opened 2020-03-18
Adapted some mockups to new HWR API

marcus-oscarsson:
Dear all,

Here are a few updates to some of the mockup objects, its just the minimal changes inorder to get the mockup version to work with the new HWR API. The idea is to complete this work during our next code camp. But, of course, let me know if there is anything that you would like to address immediately and we will sort it out, only less work for later :).

Cheers,
Marcus

### DIFF
--- a/HardwareObjects/mockup/BeamInfoMockup.py
+++ b/HardwareObjects/mockup/BeamInfoMockup.py
@@ -20,11 +20,12 @@ beamInfoChanged
 
 import logging
 from HardwareRepository.BaseHardwareObjects import Equipment
+from HardwareRepository.HardwareObjects.abstract.AbstractBeam import (AbstractBeam, BeamShape)
 
 
-class BeamInfoMockup(Equipment):
+class BeamInfoMockup(AbstractBeam):
     def __init__(self, *args):
-        Equipment.__init__(self, *args)
+        super(BeamInfoMockup, self).__init__(*args)
 
         self.beam_size_slits = [9999, 9999]
         self.beam_size_aperture = [9999, 9999]
@@ -33,6 +34,7 @@ class BeamInfoMockup(Equipment):
         self.beam_info_dict = {}
 
     def init(self):
+        super(BeamInfoMockup, self).init()
         self.aperture_hwobj = self.getObjectByRole("aperture")
         if self.aperture_hwobj is not None:
             self.connect(
@@ -84,6 +86,38 @@ class BeamInfoMockup(Equipment):
             float(self.beam_info_dict["size_x"]),
             float(self.beam_info_dict["size_y"]),
         )
+
+    def get_value(self):
+        """ Get the size (width and heigth) of the beam and its shape.
+        Retunrs:
+            (float, float, Enum, str): Width, heigth, shape and label.
+        Raises:
+            NotImplementedError
+        """
+        _i = self.evaluate_beam_info()
+        shape = "ELIPTICAL" if _i["shape"] == "ellipse" else "RECTANGULAR"
+
+        return (
+            _i["size_x"],
+            _i["size_y"],
+            BeamShape.__members__[shape],
+            "%sx%s" % (_i["size_x"], _i["size_y"])
+        )
+
+    def get_available_size(self):
+        """ Get the available predefined beam definers configuration.
+        Returns:
+            (dict): Dictionary {"type": (list), "values": (list)}, where
+               "type": the definer type
+               "values": List of available beam size difinitions,
+                         according to the "type".
+        Raises:
+            NotImplementedError
+        """
+        return {
+            "type": ["aperture"],
+            "values": self.aperture_hwobj._diameter_size_list
+        }
 
     def get_beam_shape(self):
         self.evaluate_beam_info()
--- a/HardwareObjects/mockup/BeamInfoMockup.py
+++ b/HardwareObjects/mockup/BeamInfoMockup.py
@@ -20,7 +20,8 @@ beamInfoChanged
 
 import logging
 from HardwareRepository.BaseHardwareObjects import Equipment
-from HardwareRepository.HardwareObjects.abstract.AbstractBeam import (AbstractBeam, BeamShape)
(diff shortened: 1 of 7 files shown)


HardwareObjects/mockup/BeamInfoMockup.py
```diff
@@ -20,11 +20,12 @@
...
 import logging
 from HardwareRepository.BaseHardwareObjects import Equipment
+from HardwareRepository.HardwareObjects.abstract.AbstractBeam import (AbstractBeam, BeamShape)
 
 
-class BeamInfoMockup(Equipment):
+class BeamInfoMockup(AbstractBeam):
     def __init__(self, *args):
-        Equipment.__init__(self, *args)
+        super(BeamInfoMockup, self).__init__(*args)
```
jordiandreu: BeamInfoMockup must implement abstract method self._set_value() from AbstractBeam class.

HardwareObjects/mockup/EDNACharacterisationMockup.py
```diff
@@ -29,7 +29,7 @@
 __license__ = "LGPLv3"
 
 
-class EDNACharacterisationMockup(AbstractCharacterisation):
+class EDNACharacterisationMockup(EDNACharacterisation):
     def __init__(self, name):
         super(EDNACharacterisationMockup, self).__init__(name)
```
jordiandreu: I would have sense that all mockup inherit from Abstract classes?

marcus-oscarsson: EDNACharacterisation inherits AbstractCharacterisation and provides some implementation that is useful for the mockup.

HardwareObjects/mockup/BeamInfoMockup.py
marcus-oscarsson: Thanks, Ive added the method :)

HardwareObjects/mockup/EDNACharacterisationMockup.py
jordiandreu: Ok, we can go ahead with this, but:
* In this case, _get_html_report_ is implemented in both classes and could be removed from the mockup class.
* _self._run()_ is never used (?)

I'm in favor of restructuring this class following the idea that mockup classes may inherit directly from abstract.

Just an idea...maybe we could define and AbstractEDNACharacterisation class with most of the methods are implemented in. The current implementation depends basically on EDNA plugin objects and on the queue model. Only an abstract method related to how the characterization run is implemented from site-to-site will be implemented  from site to site and the mockup class would use as input/output the data in _edna_test_data.xml_

marcus-oscarsson: I see, its an interesting idea, So what you mean is that the mockups should never inherit a concrete implementation but always an abstract class ? 

If its the case, I don't really have anything against doing what you suggest but it seems a bit unnecessary to introduce an additional class AbstractEDNACharacterisation just to not inherit from EDNACharacterisation in the mockup. What would be the benefit of this strategy ? 

Why not just take the concrete implementation and override the methods that needs to be overridden to create the mockup. ? 

As I said I'm not against what you are saying just trying to understand what the benefit would be ?

HardwareObjects/mockup/MotorMockup.py
```diff
@@ -45,12 +45,16 @@
 
 
 class MotorMockup(AbstractMotor):
+    _actuator_set_value = AbstractMotor.set_value
```
rhfogh: I am wondering why this is necessary. The original idea was that AbstractMotor.set_value should be handling all the extraneous stuff,, so that only _set_value was needed in the subclasses. Why is that not possible here?
As a second point, I think it would be better and more maintainable  to use super().set_value instead of setting a special attribute to hold the link

marcus-oscarsson: Its not possible because of the asynchronous (simulation) move, there is maybe another way to solve the problem, if there is just let me know and Ill try to implement it.

For the special attribute Its a matter of style, I did it this way, I don't think it changes the maintainability or are better or worse than super().set_value. I don't mind either

marcus-oscarsson:
> can discuss further during the code camp about the AbstractCharacterisation class.
> @marcus-oscarsson , Just a change is missing to approve the PR:
> 
> > in characterise() method, _self.run_edna_ shoud read _self._run_edna_.

I'm not sure I follow, in which file ?

HardwareObjects/mockup/MotorMockup.py
rhfogh: Good point. I shall see if I can think of something.

As for the special attribute, my personal taste is to avoid attributes and even additional variables where you can. From experience of refactoring, every time you see an attribute you need to check if it is used elsewhere, why, and for what. But,as you say, it is a matter of style

jordiandreu:
@marcus-oscarsson , sorry, the bug is the EDNACharacterization class. I will push a fix in a different PR.

marcus-oscarsson:
@rhfogh or @jordiandreu could you merge if you are happy with the PR ?

jordiandreu:
I will but I'm not authorized...

marcus-oscarsson:
Really that's strange ...

IvarsKarpics:
Hi @jordiandreu  I just changed your role to owner. You shall be able to merge the PR

marcus-oscarsson:
Should  work now ...

jordiandreu:
I can merge now, however the CI build has failed and Codacy is still hanged an no status has been reported yet.

marcus-oscarsson:
I see, well i guess we can ignore the Codacy checks for now. I ll try to see why the motor test is failing.

jordiandreu:
@marcus-oscarsson I fixed the motor test last week by changing the default position value in the xml configuration...maybe is related to this.

marcus-oscarsson:
Ahh, yes I re-based your changes on top of mine so that was fine. It was actually that I accidentally changed the MotorMockup to be async by default and the test expected that it was synchronous

PR #489, opened 2020-03-20
Refactored move, userName()

rhfogh:
Replaced def move with def _set_value
Replaced move calls with set_value calls
Replaced userName() calls with use of username attribute
Replaced zoom_supported with read_only

A few mockups and complex motors still not complete.

### DIFF
--- a/HardwareObjects/EMBL/EMBLPiezoMotor.py
+++ b/HardwareObjects/EMBL/EMBLPiezoMotor.py
@@ -23,7 +23,7 @@ EMBLPiezoMotor
 
 import logging
 
-HardwareRepository.HardwareObjects.EMBL.TINEMotor import TINEMotor
+from HardwareRepository.HardwareObjects.EMBL.TINEMotor import TINEMotor
 
 __credits__ = ["EMBL Hamburg"]
 __license__ = "LGPLv3+"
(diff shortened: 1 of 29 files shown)


HardwareObjects/EMBL/TINEMotor.py
```diff
@@ -147,29 +147,16 @@ def abort(self):
...
-        :param target: float
-        :param wait: int
-        :param timeout: boolean
+        :param value: float
         :return:
         """
         if self.chan_state is not None:
             self.update_state(self.STATES.BUSY)
             self.chan_state.setOldValue("moving")
-        if target == float("nan"):
```
IvarsKarpics: could you please revert this if?

HardwareObjects/EMBL/EMBLPiezoMotor.py
```diff
@@ -47,21 +47,12 @@ def init(self):
...
-        :type target: float
-        :param wait: wait finishing the move
-        :type wait: bool
-        :param timeout: timeout is seconds
-        :type timeout: float
+        :param value: target position
+        :type value: float
         :return: None
         """
-        if target == float("nan"):
```
IvarsKarpics: Could you pleas revert the if

IvarsKarpics: Ah ignore this... I see the check in the AbstractMotor

HardwareObjects/EMBL/TINEMotor.py
IvarsKarpics: same here

HardwareObjects/EMBL/EMBLPiezoMotor.py
rhfogh: Just out of interest, I am not sure the original would have worked - I think that 'nan == nan' evaluates to False. But anyway, it is in AbstractMotor, as you say..

marcus-oscarsson:
Looks good to me ! :+1:

rhfogh:
With two approvals that ought to be mergeable. but

1) Can aonyone guess why the Codacy check is still hanging - the commit 
was three days ago?

2) Who should be doing the merging?

marcus-oscarsson:
I think we should discuss the usage of Codacy and see if we cant replace it with something else. Maybe a black pre-commit hook and pylint would be more useful ?

I wanted to give the other sites an opportunity to have a look at the sites specific changes before I merged.

rhfogh:
@marcus-oscarsson 
Makes sense. I just did not want it to fall between the cracks, but clearly you have it in hand.

PR #490, opened 2020-03-21
Hardware Objects for ESRF ID13's hardware

natxo14:
Volpi : a device to illuminate the sample. Able to control light intensity
Turret : device to switch from one microscope camera objective to another.


HardwareObjects/ESRF/BlissTurret.py
```diff
@@ -0,0 +1,49 @@
+from HardwareRepository.BaseHardwareObjects import Device
+from bliss.config import static
+
+class BlissTurret(Device):
```
IvarsKarpics: I suggest to use HardwareObject instead of Device.

rhfogh: Indeed - I would say that Device should be deprecated adn phased out. There is very little functionality in it, and all of it can be replaced by Hardwarobject.

marcus-oscarsson:
Thanks @natxo14 well done, as the guys @IvarsKarpics suggested its better to inherit HardwareObject. Doing that, we should also take a look at get/set position and intensity. If we were at ESRF I would of course stop by your office :). In general the "main" property of the HardwareObject should be accessed with get/set value, so that would mean to change from set/get_postion to set/get_value for the turret and same for intensity for the volpi

natxo14:
Thanks for your feedback!
Changes done

marcus-oscarsson:
:+1: Thanks !

PR #492, opened 2020-03-26
Simplified MotorMockup

rhfogh:
Simplified MotorMockup

@marcus-oscarsson I know it is a bit irritating that I am refactoring what you just put in. My version is a bit simpler. But the main reason is that (from experience) I think that people will be using the mockups, and especially this one, which is nice and understandable,  as templates to understand and write their own. That makes it important that this is following the letter and spirit of the AbstrctActuator. And that is that set_value should be left unchanged, and that _set_value should be the function that takes care of the actual motor moves.

Anyway, feel free to reject it if you like.

### DIFF
--- a/HardwareObjects/mockup/MotorMockup.py
+++ b/HardwareObjects/mockup/MotorMockup.py
@@ -45,7 +45,6 @@ DEFAULT_POSITION = 10.124
 
 
 class MotorMockup(AbstractMotor):
-    _actuator_set_value = AbstractMotor.set_value
 
     def __init__(self, name):
         AbstractMotor.__init__(self, name)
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
I thought we were going to take a look at all this during the code camp, but its good with a head start. Are you by the way sure that the move works ?

In anyways, it looks good, so I'm fine with this.

rhfogh:
@marcus-oscarsson Thanks.
I do still need to test it - I am in the middle of a big refactoring so testing is tricky right now. I'll get there.

marcus-oscarsson:
@rhfogh okey so I will merge this now as it is a improvement and we can test it later and make the necessary fixes. Its not entirely clear to me if the state management will behave correctly ...

beteva:
Just a comment: it is OK for a Mockup, but setting the velocity in the init is not anything that should be done by default - some motors (like the front end at ESRF) do not accept it.

HardwareObjects/mockup/MotorMockup.py
```diff
@@ -55,30 +54,25 @@ def init(self):
...
             limits = DEFAULT_LIMITS
+        self.update_limits(limits)
 
-        self.set_limits(limits)
-
-        try:
-            self.set_value(float(self.getProperty("default_position")))
-        except BaseException:
-            self.set_value(DEFAULT_POSITION)
+        start_position = self.getProperty("default_position", DEFAULT_POSITION)
```
IvarsKarpics: I think this is confusing. Maybe lets keep one name `default_position`

rhfogh: This ws in Marcus' version -so the code just adapts to teh expected config. t would make sense to have only one name, though.

marcus-oscarsson: In all honesty if we are "blaming", it was there before i touched it as well :). So we are assigning The MotorMockup to Rasmus then ?

PR #493, opened 2020-03-26
MotorMockup fix

rhfogh:
Added specific states and  handling of move timeouts to MotorMockup.

Expanded MotorStates.

MotorStates is still not settled, but I think that as a minimum we will need two additional states, READY and UNKNOWN. Further additions can be looked at later, if required.

### DIFF
--- a/HardwareObjects/abstract/AbstractMotor.py
+++ b/HardwareObjects/abstract/AbstractMotor.py
@@ -36,9 +36,6 @@ __license__ = "LGPLv3+"
 class MotorStates(Enum):
     """Motor states definitions."""
 
-    UNKNOWN = HardwareObjectState.UNKNOWN, 0
-    READY = HardwareObjectState.READY, 1
-
     HOME = HardwareObjectState.READY, 5
     LOWLIMIT = HardwareObjectState.READY, 6
     HIGHLIMIT = HardwareObjectState.READY, 7
--- a/HardwareObjects/abstract/AbstractMotor.py
+++ b/HardwareObjects/abstract/AbstractMotor.py
@@ -36,6 +36,9 @@ __license__ = "LGPLv3+"
 class MotorStates(Enum):
     """Motor states definitions."""
 
+    UNKNOWN = HardwareObjectState.UNKNOWN, 0
+    READY = HardwareObjectState.READY, 1
+
     HOME = HardwareObjectState.READY, 5
     LOWLIMIT = HardwareObjectState.READY, 6
     HIGHLIMIT = HardwareObjectState.READY, 7
--- a/HardwareObjects/mockup/MotorMockup.py
+++ b/HardwareObjects/mockup/MotorMockup.py
@@ -104,7 +104,7 @@ class MotorMockup(AbstractMotor):
         elif  value == _high:
             self.update_specific_state(self.SPECIFIC_STATES.HIGHLIMIT)
         else:
-            self.update_specific_state(self.SPECIFIC_STATES.READY)
+            self.update_specific_state(self.STATES.READY)
 
     def abort(self):
         """Imediately halt movement. By default self.stop = self.abort"""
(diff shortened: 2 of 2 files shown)


HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -36,6 +36,9 @@
 class MotorStates(Enum):
     """Motor states definitions."""
 
+    UNKNOWN = HardwareObjectState.UNKNOWN, 0
+    READY = HardwareObjectState.READY, 1
+
```
beteva: The idea behing the MotorStates is only to add the one, missing in the HardwareObjectState !!!

rhfogh: @beteva 
OK, I did not realise that.
Does it follow that get_specific_state should return a mixture of MotorStates and HardwareObjectStates?

I am not sure I like that idea, but for now I shall adapt the code to it; we can argue about changes later. Watch this space

beteva:
> Added specific states and handling of move timeouts to MotorMockup.
> 
> Expanded MotorStates.
> 
> MotorStates is still not settled, but I think that as a minimum we will need two additional states, READY and UNKNOWN. Further additions can be looked at later, if required.

@rhfogh, I am afraid I disagree. We should not redefine states, which are already defined by the  HardwareObjectState

rhfogh:
Looks like we need a discussion at some point, then. The way  I see it, states and specific_states are two different things, reported  by two different functions, which should have two different enums.

But we can look at that later. I have already  checked in a fix that restores AbstractMotor to where it used to be, and sets  the specific state in MotorMockup to STATES.READY and STATES.UNKNOWN where needed. If that is the wrong way to do it, could you let me know what you prefer? Otherwise, would it be OK if we leave it where it is now and discuss it at the code camp?

HardwareObjects/abstract/AbstractMotor.py
beteva: I would say no. I thought we agreed on the fact that get_state only return one of the states, defined in  HardwareObjectStates and the get_specific state is an addition, in which one can find an extra information, like the motor is ready, but also on a limit switch. At least this is what I retained. maybe I'm wrong ;-(

jordiandreu:
There was already an issue with the States discussion..maybe we could restart from it on the code camp:
https://github.com/mxcube/HardwareRepository/issues/451

rhfogh:
@jordiandreu 
I agree. For now I just want to get to something that is relatively uncontroversial, to make it a better starting point or discussion. I would not have started all this, except that I thought I understood how this was supposed to work. And I was mistaken.

HardwareObjects/abstract/AbstractMotor.py
rhfogh: OK, what do you think get_specific_state should return when the motor is neither HOME, HIGHLIMIT, or LOWLIMIT?

marcus-oscarsson:
Not really working today, but in anyways ... it's interesting to see that the states can be the source of so much discussion :). I think its a good idea to make the non controversial changes now and then continue to discuss during or after the code camp.

I found the title of the PR a bit confusing as it says "Refactor mockups", and there is only one file changed. Is the intention to add more or did I missing something ?

rhfogh:
@marcus-oscarsson 
The name was wrong, and I have renamed it. I was in a branch where i was going to refactor a lot of mockups, and forgot to rename when I did something else instead.

Just FYI, my next task is actually to refactor the mockups on top of my latest PR #494 until at least I can open the program without errors. After that I shall pause until the we meeting.

marcus-oscarsson:
Sounds good, and also very reasonable :)

marcus-oscarsson:
Looks good to me, @beteva ill let you merge when you had a final look.

rhfogh:
Thanks.

PR #494, opened 2020-03-27
Refactor wrapper functions

rhfogh:
Refactored/removed wrapper functions, and further renamed to new-style function names.

The changes are extensive enough that people will likely be a need to make some extra adjustments, but I cannot see another way of getting this done.

This PR does not test correct because the mockup objects need refactoring too  I should have that fixed before too long.

### DIFF
--- a/HardwareObjects/ALBA/ALBAFlux.py
+++ b/HardwareObjects/ALBA/ALBAFlux.py
@@ -53,4 +53,4 @@ class ALBAFlux(Device, AbstractFlux.AbstractFlux):
 
 
 def test_hwo(hwo):
-    print(hwo.get_flux())
+    print(hwo.get_value)
--- a/HardwareObjects/ALBA/ALBAFlux.py
+++ b/HardwareObjects/ALBA/ALBAFlux.py
@@ -53,4 +53,4 @@ class ALBAFlux(Device, AbstractFlux.AbstractFlux):
 
 
 def test_hwo(hwo):
-    print(hwo.get_value)
+    print(hwo.get_value())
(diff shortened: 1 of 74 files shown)


HardwareObjects/ALBA/ALBAFlux.py
```diff
@@ -53,4 +53,4 @@ def get_dose_rate(self, energy=None):
 
 
 def test_hwo(hwo):
-    print(hwo.get_flux())
+    print(hwo.get_value)
```
IvarsKarpics: get_value()

HardwareObjects/Resolution.py
```diff
@@ -257,7 +240,7 @@ def update_energy(self, value):
         _wavelength = (h * c / e) / value * 10e6
 
         distance = self._hwr_detector.distance.get_value()
-        radius = self.get_detector_radius(distance)
+        radius = sself._hwr_detector.get_radius(distance)
```
marcus-oscarsson: Watch out, typo ! sself

rhfogh: Indeed. It is fixed in the next PR.
Thanks, Rasmus

marcus-oscarsson:
Yes Indeed, really well done !, just one comment for the typo ... otherwise fine !

marcus-oscarsson:
There is also a conflict ...

rhfogh:
Yeah, I do not know what happened here. I am working on merging/rebasing on to the newest master, but it looks like it will need some manual checking before i can push that.

HardwareObjects/Resolution.py
beteva: Thanks @rhfogh for the huge work.
just one more  - _set value signature has changed
from  def _set_value(self, value, wait=False, timeout=None):
to def _set_value(self, value):

rhfogh:
Thanks

PR #495, opened 2020-03-29
snake_case, mockup refactoring, function names, fixes

rhfogh:
NB, this PR is made on top of #494 (which it includes).
It would be simpler to merge #494 first, and then evaluate this one.

After this PR, MXCuBE opens without errors using mockups

Contents
-----------

 - snake_case:etState, isReady, waitReady, getCurrentPositionName, getActuatorState, acceptCentring, startCentringMethod takeSnapshots, getPositions, moveMotors, moveSyncMotors

 - Changed to actuator function names: get_current_energy, get_position, getPosition, get_energy , move_energy, getCurrentState
 - Refactored ShutterMockup (with its config)
 -  Refactored ResolutionMockup (with its config)
 - Minor changes to Resolution
 - move_motors commands no longer modify their input dictionary
 - Changed :DetectorMockup.get_beam_centre to get_beam_postion. *NB* these names need sorting out.
 - Sundry fixes for Python 2 compatibility and minor bugs


rhfogh:
@All
This is the last PR I am doing before the code camp.It should pretty much cover snake_case and renaming functions to the AbstractActuator names.

We still need some refactoring to make use of the new abstract classes (including Antonias AbstractNState) - I shall probably come up with a WIP for discussion before the camp.

marcus-oscarsson:
Looks good to me, I see that you are using double underscores in some places, I'm personally not a fan of that but it's Python standard so I'm fine with it if the other are ...

rhfogh:
I generally try to avoid the double underscores as well. Their only justification is if you want to protect an attribute from access by subclasses. That does not happen often, but it is exactly what I think is needed here, to avoid the temptation of subclasses to write directly to _nominal_value.

rhfogh:
@All
I have merged master into tis PR now, so there should be only the new changes to look at.

BaseHardwareObjects.py
```diff
@@ -753,11 +750,8 @@ def __deviceNotReady(self):
...
             self.emit("equipmentNotReady")
 
-    def isReady(self):
-        return self.isValid() and self.__ready
-
     def is_ready(self):
-        return self.isReady()
+        return self.isValid() and self.__ready
 
     def isValid(self):
```
IvarsKarpics: if there is so many changes, then maybe also we can change isValid to is_valid?

jordiandreu:
@rhfogh , mxcube PR389 together with HWR PR 494 works fine, but this PR prevents the mockup(qt4/qt5) application to open without exceptions.

rhfogh:
@jordiandreu 
That is not good. It worked when I tested it (I fixed a couple of errors, too), but I should have tested it again at the very end. Will do..

rhfogh:
@jordiandreu 
Fixed. Program now opens (on my computer at least). Sorry for the hiccup.

jordiandreu:
@rhfogh, I still getting a couple of exceptions but I will better upload the changes right after this pull request gets merged.

jordiandreu:
Thanks for the work!

rhfogh:
Thanks!

PR #496, opened 2020-03-30
WIP Abstract classes clarifications - for discussion at code camp

rhfogh:
This WIP changes AbstractActuator, AbstractMotor, and  HardwareObjectMixin. I have made the consequential changes in a number of other classes, to show how this would work, but if we agreed on this we would have to extend it to many other classes. The PR is applied on top of PR #495.

Currently it it is not always clear (certainly to me) which function and which class a certain behaviour belongs in. These changes should make the resulting code more uniform and so easier to maintain, which could be important, since we are now about to make a major refactoring. This is to some extent going back on changes I insisted on earlier, for which I apologise.

The general idea is that:

 - The internal attribtues should be managed *only* in AbstractActuator and HardwareObjectMicxin, where they are defined.
 - Subclasses should have *sole* responsibility for implementing get_value and \_set_value, 
 - Internal attributes are modified only by caliling the update_something functions, and it is the responsibility of the subclasses to call those when needed.

Specifically the changes are:
- Make the following attribute '__double_underscore:
   - HardwareObjectMixin.__ready_event
   - HardwareObjectMixin.__state
   - HardwareObjectMixin.__specific_state
   - AbstractActuator.__nominal_value
   - AbstractActuator.__nominal_limits

 - Remove the call to self.update_value() from AbstractActuator.set_value()

 - Add AbstractActuator.get_nominal_value

 - Replace direct access to internal attributes with function calls

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -382,9 +382,9 @@ class HardwareObjectMixin(CommandContainer):
         self.connect_dict = {}
 
         # event to handle waiting for object to be ready
-        self._ready_event = event.Event()
-        self._state = self.STATES.UNKNOWN
-        self._specific_state = None
+        self.__ready_event = event.Event()
+        self.__state = self.STATES.UNKNOWN
+        self.__specific_state = None
 
     def __bool__(self):
         return True
@@ -418,7 +418,7 @@ class HardwareObjectMixin(CommandContainer):
         Returns:
             HardwareObjectState
         """
-        return self._state
+        return self.__state
 
     def get_specific_state(self):
         """ Getter for specific_state attribute. Override if needed.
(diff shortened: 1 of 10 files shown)


marcus-oscarsson:
Looks like this also contains previous PR's, lets get the other PR's merged first. So that we can rebase this, because It looks like we are reviewing the same code over and over.

rhfogh:
You are right this is on top of previous PRs(I think I mentioned it somewhere). It is not ideal, but the changes in the other PRs were so pervasive that I thought it would be impossible to make three separate PRs that could be reviewed and merged separate;y.

rhfogh:
Discussed at code camp, and agreed this is too rigid and should not be progressed.

PR #497, opened 2020-03-31
Protect from nested raise exception

jordiandreu:
If widget is an objectof type pyqtgraph.widgets.PlotWidget.PlotWidget
and the target attribute in hasattr does not exists a NameError exception
is raised.

### DIFF
--- a/HardwareObjects/InstanceServer.py
+++ b/HardwareObjects/InstanceServer.py
@@ -65,9 +65,13 @@ class InstanceServer(Procedure):
 
     def initialize_instance(self):
         for widget in QtImport.QApplication.allWidgets():
-            if hasattr(widget, "configuration"):
-                self.guiConfiguration = widget.configuration
-                break
+            try:
+                if hasattr(widget, "configuration"):
+                    self.guiConfiguration = widget.configuration
+                    break
+            except NameError:
+                logging.getLogger().warning("Widget {} has no attribute {}"
+                                            .format(widget, "configuration"))
 
         self.emit("instanceInitializing", ())
         if self.isLocal():


HardwareObjects/InstanceServer.py
```diff
@@ -65,9 +65,13 @@ def init(self):
 
     def initialize_instance(self):
         for widget in QtImport.QApplication.allWidgets():
-            if hasattr(widget, "configuration"):
-                self.guiConfiguration = widget.configuration
-                break
+            try:
```
rhfogh: Actually, there is something wrong here.
hasattr(widget, "configuration") should not raise an error.
hasattr(float, 'Ka-BOOM!') just returns False.
\\Could the error come from somethings else, like self.guiConfiguration?

jordiandreu: The error comes from _pyqtgraph_. The __getattr__ method from PlotWidget raises the exception:

>     def __getattr__(self, attr):  ## implicitly wrap methods from plotItem
>        if hasattr(self.plotItem, attr):
>            m = getattr(self.plotItem, attr)
>            if hasattr(m, '__call__'):
>               return m
>        raise NameError(attr)
Any other workaroud/solution is welcome.

rhfogh: Very strange indeed. PlotWidget should have raised AttributeError instead. No better ideas to offer.

marcus-oscarsson:
Hello, I guess this does not really go here and I can make an issue out if instead but I think its somehow related. I'm trying to use the latest PR's with MXCuBE3 and I get a few error messages on is this Ka-BOOM exception ...

Its from ShutterMockup, which now expects a default value, for its state. Is that really needed ?. And it would be really nice if we could use some meaningful exceptions so that we now roughly whats going on, even if the exception name is nifty :)

rhfogh:
@marcus-oscarsson 
That was debug code that I forgot to remove after use, hence the silly text. Has been fixed in a PR. Apologies.

PR #498, opened 2020-04-01
Made ShutterMockup an AbstractNState

rhfogh:
This was debugging code, that I just forgot to clean up. Apologies

### DIFF
--- a/HardwareObjects/mockup/ShutterMockup.py
+++ b/HardwareObjects/mockup/ShutterMockup.py
@@ -44,8 +44,6 @@ class ShutterMockup(AbstractActuator.AbstractActuator):
 
     def init(self):
         super(ShutterMockup, self).init()
-        if self.default_value is None:
-            raise Exception("Ka-BOOM!")
         self._nominal_value = getattr(self.VALUES, self.default_value)
         self._state = self.STATES.READY
 
--- a/HardwareObjects/mockup/ShutterMockup.py
+++ b/HardwareObjects/mockup/ShutterMockup.py
@@ -44,7 +44,7 @@ class ShutterMockup(AbstractActuator.AbstractActuator):
 
     def init(self):
         super(ShutterMockup, self).init()
-        self._nominal_value = getattr(self.VALUES, self.default_value)
+        self._nominal_value = getattr(self.VALUES, self.default_value or "UNKNOWN")
         self._state = self.STATES.READY
 
     def get_value(self):
--- a/HardwareObjects/mockup/ShutterMockup.py
+++ b/HardwareObjects/mockup/ShutterMockup.py
@@ -19,33 +19,18 @@
 
 import time
 import random
-from enum import Enum, unique
 from HardwareRepository.HardwareObjects.abstract import AbstractNState
-# Temporarily, pending a stable AbstactNState:
-from HardwareRepository.HardwareObjects.abstract import AbstractActuator
 
 
-class ShutterMockup(AbstractActuator.AbstractActuator):
+class ShutterMockup(AbstractNState.AbstractNState):
     """
     ShutterMockup for simulating a simple open/close shutter.
-
-    TYhis is a temporary vresion, pending a stable AbstractNState
     """
-
-    @unique
-    class VALUES(Enum):
-        UNKNOWN = "UNKNOWN"
-        OPEN = "OPEN"
-        CLOSED = "CLOSED"
-
-    def __init__(self, name):
-        super(ShutterMockup, self).__init__(name)
-        # self.current_state = ShutterMockup.STATE.OPEN
 
     def init(self):
         super(ShutterMockup, self).init()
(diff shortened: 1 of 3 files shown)


HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -44,8 +44,6 @@ def __init__(self, name):
 
     def init(self):
         super(ShutterMockup, self).init()
-        if self.default_value is None:
-            raise Exception("Ka-BOOM!")
         self._nominal_value = getattr(self.VALUES, self.default_value)
```
marcus-oscarsson, line 47: Why does the shutter mockup need a default value from configuration ?

rhfogh:
@marcus-oscarsson I am not sure if it actually *needs* one. But I thought you might want to configure whether your shutter should default to open or closed. I have checked in an additional fix so that the state defaults to 'UNKNOWN' if no default_value is set.

But if you think this is not particularly useful we can just set it to UNKNOWN regardless, it was a quick thought, not the fruit of prolonged requirements analysis.

marcus-oscarsson:
I see, I don't think a default value for the mockups from configuration is really useful.

I think the configuration of the mockups should be as similar as possible to its real counter part (if possible). It will otherwise only requires us to add additional configuration

rhfogh:
@marcus-oscarsson 
OK. it should be removed, then
Actually the whole class needs rewriting. I implemented it as a subclass of AbstractActuator because the new version of AbstractNState was not yet merged at the time and I needed something that did not cause errors. But this should obviously be an AbstractNState. Do you want me to do that now, or should we wait and do it in the meeting?

marcus-oscarsson:
I think its as you like, in some way the more we do today the less we need to do tomorrow ;)

rhfogh:
This PR has been completely repurposed. Instead of a small fix, it now changes the ShutterMockup from (temporarily) being an AbstractActuator to being an AbstractNState.

The program opens without errors (in my hands)

configuration/xml-qt/fast-shutter-mockup.xml
```diff
@@ -1,5 +1,5 @@
 <device class ="ShutterMockup">
     <actuator_name>fast_shutter</actuator_name>
     <user_name>FastShutter</user_name>
```
beteva, line 3: The property is actually username, not user_name, but maybe we should change it in AbstractActuator to the one with underscore?

HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -19,35 +19,18 @@
...
-        # self.current_state = ShutterMockup.STATE.OPEN
-
     def init(self):
         super(ShutterMockup, self).init()
-        if self.default_value is None:
-            raise Exception("Ka-BOOM!")
-        self._nominal_value = getattr(self.VALUES, self.default_value)
-        self._state = self.STATES.READY
+        self._nominal_value = self.VALUES.UNKNOWN
+        self._state = self.STATES.UNKNOWN
```
beteva, line 33: The default self._state is already UNKNOWN, but maybe it is a good idea to explicitly define it.

beteva:
Thanks @rhfogh, it will be quite useful for the ode camp

HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -19,35 +19,18 @@
...
-        super(ShutterMockup, self).__init__(name)
-        # self.current_state = ShutterMockup.STATE.OPEN
-
     def init(self):
         super(ShutterMockup, self).init()
-        if self.default_value is None:
-            raise Exception("Ka-BOOM!")
-        self._nominal_value = getattr(self.VALUES, self.default_value)
-        self._state = self.STATES.READY
+        self._nominal_value = self.VALUES.UNKNOWN
```
marcus-oscarsson, line 32: Question, should we use update_value and update_state in init ?

marcus-oscarsson:
Looks good to me :)

HardwareObjects/mockup/ShutterMockup.py
rhfogh: Answer: yes we should. I have made another Booboo. You could argue, though, that  this is an example of how using the attributes directly is so much more obvious, that unless we make it hard to do (like making it __state etc.) people will just keep doing it

HardwareObjects/mockup/ShutterMockup.py
rhfogh: No, that should not have been set. I looked, but not thoroughly enough.

rhfogh:
Now it is merged, I ouwould propose we leave it for now, and fix the errors in teh next round.

Please keep pointing out my mistakes, btw. We should be getting better code that way.

marcus-oscarsson:
Agree, lets do it at the next round. I did not consider it to be a big problem, so preferred to merge.

PR #501, opened 2020-04-01
Motor mockup and testing fix

rhfogh:
Fixed bug in MotorMockup._move, and another one in detector_test.

pytest now passes

Added more informatoin to test messages (it is easier if you can see how the value is wrong)


rhfogh:
As it stands, this PR should work correctly if _set_value forks a separate process (surely the normal case). If _set_value only returns when the move is finished (the case for e.g. MotorMockup), the timeout handling is probably incorrect. Further improvements very welcome.

PR #504, opened 2020-04-02
[CODE CAMP] AbstractTransmission

jordiandreu:


rhfogh:
@jordiandreu 
Have you considered whether this could be made into (or replaced by) a MotorMockup? Near as I can seethe new functions here are already in MotorMockup, with some bells and whistles. AbstractTransmission is currently an AbstractMotor, and Transmission is  Device and will one way or the other be refactored.

jordiandreu:
@rhfogh  yes, I did. However, Motor interface brings some extra attributes like velocity that, in my opinion, not need to be accessed by the application and I think (maybe wrong) that the API from Actuator is enough to manage the transmission from the application perspective.

test/pytest/test_transmission.py
```diff
@@ -0,0 +1,40 @@
...
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+
+def test_transmission_attributes(beamline):
+    assert (
+        not beamline.energy is None
+    ), "Energy hardware objects is None (not initialized)"
+
+    transmission = beamline.transmission.get_value()
```
IvarsKarpics, line 26: maybe rename to transmission_value or just value?

jordiandreu: Changed.

jordiandreu:
Last commit rewrites an assertion to suggested notation.

test/pytest/test_transmission.py
IvarsKarpics: thanks

test/pytest/test_transmission.py
```diff
@@ -0,0 +1,40 @@
...
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+
+def test_transmission_attributes(beamline):
+    assert (
+        not beamline.energy is None
+    ), "Energy hardware objects is None (not initialized)"
```
IvarsKarpics, line 24: Energy -> Transmission

test/pytest/test_transmission.py
```diff
@@ -0,0 +1,40 @@
...
+
+    value = beamline.transmission.get_value()
+    limits = beamline.transmission.get_limits()
+
+    assert isinstance(value, float), "Transmission value has to be float"
```
IvarsKarpics, line 29: (float, int)

jordiandreu: I was wrong...this is required, thanks @IvarsKarpics . Done!

jordiandreu:
@ALL I have finished this PR and can be reviewed.

IvarsKarpics:
Looks good

HardwareObjects/abstract/AbstractTransmission.py
```diff
@@ -17,45 +18,22 @@
...
 __license__ = "LGPLv3+"
 
 
-class AbstractTransmission(AbstractMotor):
-    """
-    Base class for transmission control by filters, slits, apertures or other means
-    """
-
-    __metaclass__ = abc.ABCMeta
+class AbstractTransmission(AbstractActuator):
```
rhfogh, line 31: Just a suggestion, but I would start like this - just for documentation purposes:


class AbstractTransmission(AbstractActuator):
    """Abstract Transmission
    
    The value of transmission is in % (float or int). 
    If transmission is not continuously variable, 
    beamlines should provide the nearest achievable value"""
    
    unit = "%"
    __metaclass__ = abc.ABCMeta

marcus-oscarsson: :+1:

marcus-oscarsson:
Looks good to me, we could maybe add @rhfogh's suggestion for documentation

HardwareObjects/abstract/AbstractTransmission.py
beteva: I would suggest adding validate_value. AbstractActuator returns always True, but we do want to check if it is within the limits.
Otherwise looks good to me.

beteva: :+1:

jordiandreu: > Just a suggestion, but I would start like this - just for documentation purposes:
> 
> ```
> class AbstractTransmission(AbstractActuator):
>     """Abstract Transmission
> 
>     The value of transmission is in % (float or int). 
>     If transmission is not continuously variable, 
>     beamlines should provide the nearest achievable value"""
> 
>     unit = "%"
>     __metaclass__ = abc.ABCMeta
> ```

I do agree, except for "%" as unit.

jordiandreu: Done

jordiandreu:
> Approved, subject to considering the comment(s)

Done

jordiandreu:
> Implement validate_value.

@beteva many thanks, done.

HardwareObjects/abstract/AbstractTransmission.py
rhfogh: @jordiandreu 
'%' is kind of non-standard, but I cannot think of anything better or more informative. What would be your preferred solution, just to skip the unit?

beteva:
Thanks @jordiandreu

HardwareObjects/abstract/AbstractTransmission.py
jordiandreu: @rhfogh I would just let unit = None, since it is what it really is. The docstring already contains the meaning for the value attribute .

PR #505, opened 2020-04-02
[CODE CAMP] BaseHardwareObjects

rhfogh:
Documented relevant classes

Also added fixes to satisfy pylint, and a few TODO suggestions (hence additional files)

### DIFF
--- a/BaseHardwareObjects.py
+++ b/BaseHardwareObjects.py
@@ -13,7 +13,7 @@ from HardwareRepository.ConvertUtils import string_types
 
 @enum.unique
 class HardwareObjectState(enum.Enum):
-    """Enumeration of ommon states, shared between all HardwraeObjects"""
+    """Enumeration of common states, shared between all HardwareObjects"""
 
     UNKNOWN = 0
     WARNING = 1
@@ -528,8 +528,6 @@ class HardwareObjectMixin(CommandContainer):
         """Clear gevent tasks, called when disconnecting a HardwareObject.
 
         Override in subclasses as needed.
-i
-        Returns:
 
         """
         self.update_state(self.STATES.UNKNOWN)
@@ -702,7 +700,7 @@ class HardwareObject(HardwareObjectNode, HardwareObjectMixin):
 class HardwareObjectYaml(ConfiguredObject, HardwareObjectMixin):
     """Yaml-configured hardware object.
 
-    For use when we move confiugratoin out of xml and into yaml.
+    For use when we move confiugration out of xml and into yaml.
 
     The class is needed only to provide a single superclass
     that combines ConfiguredObject and HardwareObjectMixin"""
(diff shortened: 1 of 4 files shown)


BaseHardwareObjects.py
```diff
@@ -13,6 +13,8 @@
 
 @enum.unique
 class HardwareObjectState(enum.Enum):
+    """Enumeration of ommon states, shared between all HardwraeObjects"""
```
IvarsKarpics, line 16: common, HardwareObjects

BaseHardwareObjects.py
```diff
@@ -706,17 +785,38 @@ def getDevice(self, deviceName):
...
+    And that use looks like it could be replaced by something else"""
+
     pass
 
 
 class Equipment(HardwareObject, DeviceContainer):
+    """Equipment class -old style
+
+    NB This class needs refactoring. Since many (soon: all??) contained
+     objects are no longer of class Devicem the code in here is unlikely to work."""
```
marcus-oscarsson, line 818: typo Device ...

BaseHardwareObjects.py
```diff
@@ -600,43 +659,53 @@ def __getattr__(self, attr):
...
         from .HardwareRepository import getHardwareRepository
 
         return getHardwareRepository().xml_source[self.name()]
 
 
 class HardwareObjectYaml(ConfiguredObject, HardwareObjectMixin):
-    """Yaml-configured hardware object"""
+    """Yaml-configured hardware object.
+
+    For use when we move confiugratoin out of xml and into yaml.
```
marcus-oscarsson, line 705: confiugratoin -> configuration

BaseHardwareObjects.py
```diff
@@ -468,47 +488,62 @@ def update_state(self, state=None):
...
 
     # Moved from HardwareObjectNode
     def clear_gevent(self):
-        """Clear gevent tasks. Override in subclasses
-
-        NB Not sure why this is needed, except maybe for object reloading?
+        """Clear gevent tasks, called when disconnecting a HardwareObject.
 
+        Override in subclasses as needed.
+i
```
marcus-oscarsson, line 531: oops :)

BaseHardwareObjects.py
```diff
@@ -657,6 +728,16 @@ def GUI(self, parent):
 
 
 class Device(HardwareObject):
+    """Old superclass for devices
```
marcus-oscarsson, line 731: We should maybe as one of the lasts steps of all this, remove all usage of Device and Equipment and used HardwareObject instead.

marcus-oscarsson:
Looks good to me, thanks !

BaseHardwareObjects.py
rhfogh: Seconded!

rhfogh: @marcus-oscarsson Actually I would love to get rid of those. Do you think I could/should do it now, leaving the current Equipment instances a bit dangling?Or even roughly refactoring them? Or might it it be better to let people refactor those first and remove Equipment after that?

PR #506, opened 2020-04-02
[CODE CAMP] AbstractDetector

MartinSavko:
Clean up of AbstractDetector and DetectorMockup

  AbstractDetector:
      -- removing _det from _det_radius and _det_metadata
      -- adding _ to height and width and adding respective getter methods
      -- keep arguments of wait_ready consistent with the super class
      -- add list of signals to the file header
      -- add _threshold_energy attribute and respective getter nad setter method
      -- remove self._radius -- it is not used anywhere
      -- add self.re_emit_values method
      -- emit signal when changing roi_mode
      -- remove statusChanged signal, we have stateChanged and
         specificStateChanged instead

DetectorMockup:
      -- remove commented self.distance
      -- emit detectorRoiModeChanged instead detectorModeChanged. This
         is what the brick expects
      -- remove last_image_saved(), stop_acquisition and wait_ready methods 
        as those are in the parent class already and are not abstract         
      -- remove set_roi_mode from and add emit to the method in the supperclass
      -- remove stop_acquistion -- already in the superclass
      -- move re_emit_values method to the superclass
      -- add underscore to actual_frame_rate for consistency with other
         internal parameters
      -- call self.status self._state instead

AbstractDetector and DetectorMockup:
      -- add optional wavelength parameter to get_beam_position method,
         access the value through HWR.beamline.energy if not specified
 
Access to detector_distance is maintained through detector, the main argument is, as in the case of beam_position, catering for the situation when two detectors are used at the same time


HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -42,29 +55,30 @@ def __init__(self, name):
...
-        self.height = self.getProperty("height")
 
         try:
-            self._det_metadata = self["beam"]
+            self._metadata = self["beam"]
         except KeyError:
             pass
-
+        
         try:
```
beteva, line 74: Maybe you want to remove this, as we said that the detector distance motor comes from HWR.beamline

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -42,29 +55,30 @@ def __init__(self, name):
...
-            self._distance_motor_hwobj = self.getObjectByRole("detector_distance")
-        except KeyError:
+            self._distance_motor_hwobj = self.getObjectByRole("detector_distance")	
+        except KeyError:
```
beteva, line 76: Maybe you want to remove this, as we said that the detector distance motor comes from HWR.beamline

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -42,29 +55,30 @@ def __init__(self, name):
...
-
+        
         try:
-            self._distance_motor_hwobj = self.getObjectByRole("detector_distance")
-        except KeyError:
+            self._distance_motor_hwobj = self.getObjectByRole("detector_distance")	
+        except KeyError:	
             pass
-
+
```
beteva, line 78: Maybe you want to remove this, as we said that the detector distance motor comes from HWR.beamline

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -167,21 +181,33 @@ def set_binning_mode(self, value):
...
         Returns:
             tuple(float, float): Beam position x,y coordinates [pixel].
         """
+        # Following is just an example of calculating beam center using
+        # simple linear regression,
+        # One needs to overload the method in case more complex calculation
+        # is required, e.g. using other detector positioning motors and
+        # wavelength
+
         distance = distance or self._distance_motor_hwobj.get_value()
```
beteva, line 198: Should this  be changed to HWR.beamline.detector_distance.get_value() ?

MartinSavko: Indeed that's what I was going to do but then I realized that this would imply that there is only one detector on the beamline and the whole reason for having get_beam_position() as part of the detector would be invalidated. I think the best place to access detector_distance is the detector itself.

HardwareObjects/abstract/AbstractDetector.py
MartinSavko: Please see my comment below.

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -17,9 +17,22 @@
...
+
+"""
+Detector API
+
+emits signals:
+    detectorRoiModeChanged
+    temperatureChanged
+    humidityChanged
+    expTimeLimitsChanged
+    frameRateChanged
```
IvarsKarpics, line 29: I was wondering if we could combine temp., hum. and frame rate in one signal?

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -42,29 +55,30 @@ def __init__(self, name):
...
+    
+        self._threshold_energy = None
         self._distance_motor_hwobj = None
-        self.width = None  # [pixel]
-        self.height = None  # [pixel]
-        self._det_radius = None
-        self._det_metadata = {}
+        self._width = None  # [pixel]
+        self._height = None  # [pixel]
+        self._radius = None
```
rhfogh, line 63: Is it not better to do without this attribute? The radius is recalculated every time anyway, as it depends on the beam position.I do not thine _radius is actually used for anything

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -106,7 +120,7 @@ def stop_acquisition(self):
         Stops acquisition
         """
 
-    def wait_ready(self):
+    def wait_ready(self, timeout=None):
```
rhfogh, line 123: wait_ready is implemented in superclass (HardwareObjectMixin). As long s update_state is called when teh state changes, the existing wait_ready should be reliable.

HardwareObjects/mockup/DetectorMockup.py
```diff
@@ -23,7 +23,6 @@ def init(self):
         """
         AbstractDetector.init(self)
 
-        # self.distance = 500
         self._temperature = 25
         self._humidity = 60
         self.actual_frame_rate = 50
```
rhfogh, line 28: The frame_rate is not mentioned in the AbstractDetector (except s a signql in the header comment).
If it is important enough to be in the mockup, should it not also be in the AbstractDetector?

HardwareObjects/mockup/DetectorMockup.py
```diff
@@ -32,28 +31,28 @@ def init(self):
...
+        xval, yval = super(DetectorMockup, self).get_beam_position(
+            distance=distance, wavelength=wavelength
+        )
         if None in (xval, yval):
             # default to Pilatus values
             xval = self.getProperty("width", 2463) / 2.0 + 0.4
             yval = self.getProperty("height", 2527) / 2.0 + 0.4
         return xval, yval
 
     def update_values(self):
```
rhfogh, line 54: Maybe this function should be moved to the AbstractDetector, considering that it is known that these signals are sent, already at the AbstractDetector level.

HardwareObjects/mockup/DetectorMockup.py
```diff
@@ -32,28 +31,28 @@ def init(self):
         self._exposure_time_limits = [0.04, 60000]
         self.status = "ready"
```
rhfogh, line 32: Is status something different from state and specific_state, that are both implemented in HardwareObject? Possibly the use of status could be replace by the use od (e.g.) state, and update_state.

marcus-oscarsson:
Thanks, looks good in general. It would be nice if it was possible to use the state functionality provided in the BaseHardwareObject, like this we can even remove wait_ready as @rhfogh suggest.

marcus-oscarsson:
Just added issue #519 that could be of general interest

HardwareObjects/abstract/AbstractDetector.py
MartinSavko: Agree.

marcus-oscarsson:
Hello, there was conflict with master which I solved so don't forget to make new PULL when you continue to work on this

Whats the status by the way, do you still have pending changes ?

rhfogh:
@marcus-oscarsson 
Looking through the comments, I had a few proposals (over a month ago!), which no one seems to have adressed. If that is because they have been considered and people disagree or want to postpone, that is fine, but it would be nice if there was some indication that someone had actually read them.

HardwareObjects/abstract/AbstractDetector.py
beteva: You are right, it is not used anywhere. We can safely remove it.

HardwareObjects/mockup/DetectorMockup.py
beteva: The frame rate seems to be used by EMBL only.

HardwareObjects/abstract/AbstractDetector.py
MartinSavko: Agree, we can remove it.

HardwareObjects/mockup/DetectorMockup.py
MartinSavko: I would keep for now and move it to the AbstractDetector class.

HardwareObjects/mockup/DetectorMockup.py
MartinSavko: It is already part of the AbstractDetector -- here the method of the superclass is called and alternative calculation is executed only as a fallback.

HardwareObjects/mockup/DetectorMockup.py
MartinSavko: I think this would most closely correspond to specific_state -- looking at how it is currently used in EMBLDetector.py. I will modify the PR accordingly.

MartinSavko: Well, actually not quite sure now -- "ready" seems not very specific -- I would call it self._state instead.

HardwareObjects/abstract/AbstractDetector.py
MartinSavko: I would vote for keeping it separate.

MartinSavko:
Hi Marcus,
I have now attempted to address all of the comments. I do not have any pending changes.

MartinSavko:
I see your point -- it is indeed not nice. Would the following be acceptable?

def init(self)
     ...
     ...

rhfogh:
@MartinSavko 
Indeed, that is *exactly* what I would have proposed.

marcus-oscarsson:
Wow, I did not even have the time to check !

Thanks alot @MartinSavko, nicely done !

PR #507, opened 2020-04-02
[CODE CAMP] AbstractActuator

beteva:


HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -90,30 +97,29 @@ def validate_value(self, value):
...
         Raises:
             ValueError: Value not valid or attemp to set write only actuator.
         """
         if self.read_only:
             raise ValueError("Attempt to set value for read-only Actuator")
-        elif self.validate_value(value):
+        if self.validate_value(value):
             self._set_value(value)
+            self.update_value()
             if timeout or timeout is None:
```
IvarsKarpics, line 121: is this correct? Should it not just be if timeout:

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -90,30 +97,29 @@ def validate_value(self, value):
...
-        Implementation of specific set actuator logic.
+        """Implementation of specific set actuator logic.
         Args:
             value: target value
         """
 
     def set_value(self, value, timeout=0):
-        """
-        Set actuator to absolute value.
+        """ Set actuator to absolute value.
```
rhfogh, line 106: I cannot think of a better way of putting this,but my first reaction was that this meant
'set actuator to abs(value)'
Does anyone have a better word?

HardwareObjects/abstract/AbstractActuator.py
rhfogh: @IvarsKarpics 
If I may: timeout=None is supposed to mean 'wait forever'.
timeout=0 means 'do not wait'

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -136,6 +143,7 @@ def update_limits(self, limits=None):
         if limits is None:
             limits = self.get_limits()
 
-        if all(limits):
+        # All values are not None nor NaN
```
rhfogh, line 148: If I read this right, the limits (0, 100) will not be set, since all((0,100)) evaluates as false. Is that not a problem? Are you not allowed to set a limit to zero?

HardwareObjects/abstract/AbstractActuator.py
beteva: You are right, I'll change the text.

HardwareObjects/abstract/AbstractActuator.py
IvarsKarpics: Ok, maybe we do not need this check, but just call wait_ready with the argument

rhfogh: How about putting 
   if timeout == 0: return
into wait_ready, and then getting rid of the test?

I may well be that using a gevent timeout of 0 has the same resul, but is both complex and opaque.

IvarsKarpics: sounds good

beteva:
@rhfogh , I'll review all PR and will let you know when ready. ;-)

rhfogh:
@beteva 
We seem to be undoing each others' changes. My PRs were meant as proposals not commands, and quite possibly your way is better. Still, would it not make make sense if we exchanged a bit of information? I have two questions:

- Why do you want to have a function like init() that does nothing but call the same function in the superclass? AFAIK leaving out the subclass init() means the superclass init() gets called anyway. Am I wrong? Or do you think it gives better clarity? Or what am I missing?

-Why do you prefer to have a _wavelength_value attribute? Wavelength and energy are by definition equivalent, one is a direct mathematical function of the other. To me it makes sense to track only the energy, save a bit of duplication, and convert between energy and wavelength when you need to deal with wavelengths. If you only store one attribute they can never get out of sync, not even by mistake. Again what am I missing? What would be your arguments?

beteva:
Dear @rhfogh 
>     * Why do you want to have a function like init() that does nothing but call the same function in the superclass? AFAIK leaving out the subclass init() means the superclass init() gets called anyway. Am I wrong? Or do you think it gives better clarity? Or what am I missing?
It is maybe me who is  wrong. I thought that he only guaranty we have is that `__init__` is called by the super function. I just wanted to be sure that init is executed, the `__init__` is for completeness, not that is is needed. I am ready to remove it, I should clarify this.
> 
>     * Why do you prefer to have a _wavelength_value attribute? Wavelength and energy are by definition equivalent, one is a direct mathematical function of the other. To me it makes sense to track only the energy, save a bit of duplication, and convert between energy and wavelength when you need to deal with wavelengths. If you only store one attribute they can never get out of sync, not even by mistake. Again, what am I missing? What would be your arguments?
 Sometimes the wavelength is implemented as a separate motor, not a calculation from the energy (at least not in the HardwareObject). I thought, that having `_wavelength_value` and `_wavelength_limits`, could be used by this type of implementation.

rhfogh:
@beteva 
For the wavelength that makes sense. Anyway, you would know that better than I. 

For the init() I am sure it is unnecessary (if harmless), and I think it is better to avoid no-op functions if we can.

Anyway, no more comments on this from my side.

Your patience is appreciated, as always.

beteva:
Ok, I will remove what is unnecessary. Thanks for your work and comments, always appreciated.

marcus-oscarsson:
Nicely done @beteva !, thanks for all this work and merging the different ideas and opinions !

marcus-oscarsson:
Also thanks to @rhfogh for the all the good ideas :)

PR #502, opened 2020-04-02
[CODE CAMP] AbstractSampleChanger

marcus-oscarsson:

PR #508, opened 2020-04-02
Add some empty test method 

jbflo:
Empty test method to be check and fill later


marcus-oscarsson:
@jbflo I've added the tests to #502, so you can continue from there :)

Just check do, git checkout -b origin cc-AbstractSampleChanger and you should be able to push directly to that branch. 

I will start to add some documentation to AbstractSampleChanger

Cheers,
Marcus

jbflo:
> @jbflo I've added the tests to #502, so you can continue from there :)
> 
> Just check do, git checkout -b origin cc-AbstractSampleChanger and you should be able to push directly to that branch.
> 
> I will start to add some documentation to AbstractSampleChanger
> 
> Cheers,
> Marcus

Hi,
when i tried to push directly to cc-AbstractSampleChanger
i got access denied.
Cheers
JB

marcus-oscarsson:
Oh, yes , should work now :)

PR #510, opened 2020-04-02
[Fix] bug in check limits

jordiandreu:
It fails to check limits containing the value 0.

### DIFF
--- a/HardwareObjects/abstract/AbstractActuator.py
+++ b/HardwareObjects/abstract/AbstractActuator.py
@@ -26,6 +26,7 @@ Emits signals valueChanged and limitsChanged.
 """
 
 import abc
+import math
 from HardwareRepository.BaseHardwareObjects import HardwareObject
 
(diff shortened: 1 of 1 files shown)


jordiandreu:
@beteva I propose this change to protect the limits from values like None or NaN.

beteva:
Well done, thanks

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -142,6 +143,7 @@ def update_limits(self, limits=None):
         if limits is None:
             limits = self.get_limits()
 
-        if all(limits):
+        # All values are not None nor NaN
```
rhfogh, line 146: @beteva @jordiandreu 
Actually, what happens if one of the limits is zero? I thought someone mentioned that in the discussion

jordiandreu: If one of the limit values was 0, the nominal limits were not updated and no limitsChange signal were emitted.

PR #513, opened 2020-04-02
[CODE CAMP] test_hardware_objects_generic

rhfogh:
First try of testing Abstract class functions for all HardwareObjects.
Will need developing, and testing, once we have more classes in test config.


test/pytest/test_hardwareobjects_generic.py
```diff
@@ -0,0 +1,158 @@
...
+                    "%s.update_state(HardwareObjectState.%s) is not reflected in result"
+                    % (role, name)
+                )
+
+            obj.update_values()
+
+
+        if isinstance(obj, AbstractActuator.AbstractActuator):
+
+            default = obj.get_value()
```
IvarsKarpics, line 74: rename to default_value?

marcus-oscarsson:
it would be nice if we could get these through a fixture so that we could choose when to call them. We could for instance create one method for each functionality i.e. test_set_state and test_set_value. There is maybe a better way of doing this in pyTest but using a fixture would be one way and it would give us the possibility to decide when these are used without duplicating to much code.

marcus-oscarsson:
@IvarsKarpics made a fixture you can look at:

   https://github.com/mxcube/HardwareRepository/blob/master/test/pytest/conftest.py

But there is maybe a better way of doing this with pyTest ?

rhfogh:
@marcus-oscarsson I'll get onto it.

rhfogh:
@All
- Refactored and expanded generic testing
- run tests and fixed problems
- changed start_position property to default_value, for consistency. 
- Some refactoring and clean-up, also to remove (potential) problems.

NB As it stands the suite fails 8 tests. I am on to that, but fixing it will require refactoring the set_value functions, and (I think) adding an AbstractMockActuator class. It seemed better to do that in a separate PR (coming up).  meanwhile please review this one.

HardwareObjects/abstract/testing/TestAbstractActuatorBase.py
```diff
@@ -0,0 +1,154 @@
...
+import time
+import gevent
+import pytest
+
+from HardwareRepository.HardwareObjects.abstract.testing import TestHardwareObjectBase
+
+test_object = TestHardwareObjectBase.test_object
+
+
+class TestAbstractActuatorBase(TestHardwareObjectBase.TestHardwareObjectBase):
```
marcus-oscarsson, line 39: Thanks for all this nice work :), it would be nice with some comment for the tests so that one knows what they are supposed to test

HardwareObjects/abstract/testing/TestAbstractActuatorBase.py
```diff
@@ -0,0 +1,154 @@
...
+            "Staring valuee %s evaluates invalid" % start_val
+        )
+        default_limits = test_object.get_limits()
+        if test_object.read_only:
+            assert default_limits == (start_val, start_val), (
+                "read_only default limits %s do not match starting value %s"
+                % (default_limits, start_val)
+            )
+
+    def test_setting_timeouts_1(self, test_object):
```
marcus-oscarsson, line 135: For instance this test, what is it intended to test ?

HardwareObjects/abstract/testing/TestAbstractMotorBase.py
```diff
@@ -0,0 +1,146 @@
...
+
+test_object = TestAbstractActuatorBase.test_object
+
+
+class TestAbstractMotorBase(TestAbstractActuatorBase.TestAbstractActuatorBase):
+    """Tests for AbstractMotor subclasses"""
+
+    __metaclass__ = abc.ABCMeta
+
+    def test_valocity(self, test_object):
```
marcus-oscarsson, line 41: typo i guess :)

marcus-oscarsson:
Looks very nice :), what do you think about also testing the signal emition ?

rhfogh:
@marcus-oscarsson 
SIgnals emission will be next, but I wanted to get this to a state where the tests passed. I'll get to the signals, but I do need to spend some time on Glpobal Phasing business too, now.

marcus-oscarsson:
I see, sounds good :)

HardwareObjects/abstract/testing/TestAbstractMotorBase.py
rhfogh: Now fixed in the #529 PR

HardwareObjects/abstract/testing/TestAbstractActuatorBase.py
rhfogh: @marcus-oscarsson 
I added them, but in the #529 PR.

PR #514, opened 2020-04-02
[CODE CAMP] for AbstractActuator: Bug Fixes, and tested NStateMockup

rhfogh:
@beteva 
With this PR the program opens, as far as the Mockup shutters are concerned. Which was the important point, right?

Except for the config files, the changes are bug fixes - feel free to ignore them or copy them across any way you want

AbstractMotor.update_value got into trouble comparing with tolerance  when _nominal_value was None. Up to you Hardware types to consider if we really need to check for NaN.

MotorMockup gave an error on the non-ascii copyright symbol until I added the encoding line.

MicroDiffZoomMockup still throws an error on program opening because of the code in init and incomplete config files. But that one is clearly still in progress, and it is all yours.


rhfogh:
Added a few fixes.

NB part of them are about allowing for value to be (et as ) None. Alternative we could make that illegal, which would (have) require(d) a different set of fixes.

rhfogh:
@beteva 
Regrettably the cpommits in this PR are repreated in #516. This one should be dealt with first, I think. See #516 for a couple of questions

HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -115,15 +117,17 @@ def update_value(self, value=None):
...
+        if self._nominal_value is None:
+            if value is None:
+                return
+
+        elif value is not None and self._tolerance:
             if abs(value - self._nominal_value) <= self._tolerance:
                 return
 
         self._nominal_value = value
-        self.emit("valueChanged", (self._nominal_value,))
```
beteva, line 129: @rhfogh, The desired behaviour is that the value changes constantly, while the motor is moving, the GUI is updated and the user can see that something happens. I am not sure if this is the case any more.

marcus-oscarsson: @rhfogh, hm whats the behavior you get in the mockup version ... continuously updated values as the motor moves ?

marcus-oscarsson:
Hm, I guess @beteva needs to merge this with her work/branch as this is request to merge mxcube:cc_AbstractActuator from rhfogh:cc_AbstractActuator to be able to continue and then we can merge mxcube:cc_AbstractActuator

HardwareObjects/abstract/AbstractMotor.py
rhfogh: @marcus-oscarsson @beteva 
Good point. For the real motors, the idea (I believe) is that the value channel will send signals as the value changes, which will trigger calls to update_value. If that works fully, we could strictly speaking remove the update_value call from set_value altogether. It probably makes sense to have it anyway as a safety mechanism, but in that case I think it should be at the end, when the object is ready and the value is final, not while we are still waiting for completion.

I checked MotorMockup. Barring bugs, there are two cases: if timeout is non-zero or None, the value changes slowly, depending on the velocity (as you coded it, Marcus) and you call update_value every 0.02s.If timeout is zero the task is forked until it completes and does not do regular updates (not what we want, but I could not see how to code it better without forking separate, intercommunicating, processes). Either way you get a call to _set_value when the task finishes or times out, which calls self,update_value() using the last achieved value.

PR #516, opened 2020-04-03
[CODE CAMP] AbstractEnergy and Mockup

rhfogh:
Made EnergyMockup, and proposed simplification of AbstractEnergy.

Also added AbstractEnergy.validate_value


rhfogh:
@beteva 
This PR and #514 share some commits. #514 should be dealt with first. 
I had a few minor problems getting the merging right, and some errors may have crept in. It has gotten confused enough by now (mea culpa) that you may have to check a few things when (if?) this is merged. Meanwhile it is all in your hands.

A couple of questions:

- AbstractActuator.set_limits documentation.: Are the limits not floats (or maybe ints) always? If so should the documentation not say 'floats' rather than 'elements'? If not, what could they be?

- AbstractActuator.set_value: If you do update_value() after _set_value() but before wait_ready() you might be calling get_value while the relevant motor is still moving, so the value you will get will not be the final one. Is this not a potential bug? What is the reason for calling update_value before wait_ready()?

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -72,7 +72,7 @@ def get_value(self):
     def get_limits(self):
         """Return actuator low and high limits.
         Returns:
-            (tuple): two elements (low limit, high limit) tuple.
+            (tuple): two floats (low limit, high limit).
```
beteva, line 75: I've specially changed this, as the values are not always floats. Tey are actually Enums in the case of NState.

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -110,8 +110,7 @@ def set_value(self, value, timeout=0):
         Args:
             value: target value
             timeout (float): optional - timeout [s],
-                             If timeout == 0: return at once and do not wait
-                                              (default);
+                             If timeout == 0: return at once and do not wait (default);
```
beteva, line 113: ?

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -143,7 +142,7 @@ def update_limits(self, limits=None):
         Args:
             limits (tuple): two floats tuple (low limit, high limit).
         """
-        if limits is None:
+        if not limits:
```
beteva, line 145: Actually bot are not correct. We should check against a tuple of (None,None).

HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -33,16 +34,19 @@
...
-        """Get the proprrties read_only, default_value"""
-        AbstractActuator.init(self)
+    # These are not actually necessary. Do wewant them anyway, for clarity?
+    #
+    # def __init__(self, name):
+    #     AbstractActuator.__init__(self, name)
+    #
+    # def init(self):
+    #     """Get the proprrties read_only, default_value"""
+    #     AbstractActuator.init(self)
```
beteva, line 50: Yes we do and not just for clarity.  It is in the init that you define the read_only and this way we guarantee that init is executed.

HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -113,9 +111,27 @@ def _calculate_energy(self, wavelength=None):
             (float): Energy [keV]
         """
         hc_over_e = h * c / e * 10e6
-        wavelength = wavelength or self._wavelength_value
+        wavelength = wavelength or self.get_wavelength()
+        if wavelength is None:
+            return
         return hc_over_e / wavelength
 
+    def validate_value(self, value):
```
beteva, line 119: You are right, we should implement validate_value for this one

HardwareObjects/abstract/AbstractEnergy.py
rhfogh: @beteva Not sure that is correct, technically. If there is no init in this class, an init()  call will automatically execute to the init of the superclass. Biut again, it does not hurt

HardwareObjects/abstract/AbstractActuator.py
rhfogh: True, but when the values are not floats, what are the limits? Can you do 'value < highlim' for enums? limits are mostly disabled for NState, and where they are present, as for the zooms, are the limits not floats?

HardwareObjects/abstract/AbstractActuator.py
rhfogh: @beteva This one slipped in during the editing and merging process. Not sure if it should be on one line or two, but whatever black tells us, I guess. Ignore as you wish.

HardwareObjects/abstract/AbstractActuator.py
rhfogh: Good point - I think you found a bug, but maybe not just the one you thought ;-) ?

Thinking about it, if the limits passed in are None, we clearly need to call get_limits(). 
If the limits passed in are (None, None), OTOH, I think we should set them. (None,None) means 'do not use limits', it is the default value, and it must be possible to set the value back to 'do not use limits'. 

My 'if limits' only make a difference if the limits passed in are 0 or False or () (empty tuple).
0 and False are arguably bugs, and should raise an error. If someone passes in an empty tuple the code would wok in either case, but I think it is probalbly better to call get_limits, than to silently do nothing.

How do you think this ought to look?

HardwareObjects/abstract/AbstractEnergy.py
rhfogh: I believe that is in the next PR, #516

marcus-oscarsson:
I suggest that we get the #507 and #514  merged before and then have a look at this as there are quite a few changes to for instance AbstractActuator in different PR's. I think @beteva have started to merge #507 and #514 ...

rhfogh:
@marcus-oscarsson I agree.

PR #518, opened 2020-04-03
[WIP][CODE CAMP?] AbstractMachineInfo issue #448

### LINKED ISSUE #448: Create AbstractMachineInfo
Use MachineInfo as inspiration for AbstractMachineInfo

Add to the already existing methods:
- get_undulator_gaps()
- get_machine_current()
- get_machine_state_text()

Finally make sure that MachineInfo inherits AbstractMachineInfo

laispc:
Hi! This class is not exactly in April's Code Camp agenda, but it has been done during it.
Closes #448. 

The progress until now is below. The attributes used by mxcube2 were already added and tested with this GUI. The same is being done for mxcube3.

- [x] Machine current attribute
- [x] Message attribute
- [x] Hutch temperature attribute
- [x] Hutch humidity attribute
- [x] Flux attribute
- [x] Disk space attribute
- [ ] Transmission attribute
- [ ] Beam size attribute
- [ ] Lifetime attribute
- [ ] Topup remaining attribute
- [ ] Finish mockup class
- [x] Test with mxcube2
- [ ] Test with mxcube3
- [ ] Review docstrings
- [ ] Write test (test_mach_info)

Any feedback is welcome! :)
Thanks,
Laís

### DIFF
--- a/HardwareObjects/abstract/AbstractMachineInfo.py
+++ b/HardwareObjects/abstract/AbstractMachineInfo.py
@@ -35,6 +35,7 @@ class AbstractMachineInfo(HardwareObject):
         self._temperature = None
         self._humidity = None
         self._flux = None
+        self._disk = None
         self._values = {}
 
     def init(self):
@@ -87,6 +88,15 @@ class AbstractMachineInfo(HardwareObject):
         """
         return None
 
+    def update_current(self, value=None):
+        """Check if current has changed.
+        Args:
+            value: value
+        """
+        if value is None:
+            value = self.get_current()
+        self._current = value
+
     @abc.abstractmethod
     def get_message(self):
         """Read message.
@@ -94,6 +104,15 @@ class AbstractMachineInfo(HardwareObject):
             value: Message.
         """
         return None
+
+    def update_message(self, value=None):
+        """Check if message has changed.
+        Args:
+            value: value
+        """
+        if value is None:
+            value = self.get_message()
+        self._message = value
 
     @abc.abstractmethod
     def get_temperature(self):
@@ -103,6 +122,15 @@ class AbstractMachineInfo(HardwareObject):
         """
         return None
 
+    def update_temperature(self, value=None):
+        """Check if temperature has changed.
+        Args:
+            value: value
+        """
+        if value is None:
+            value = self.get_temperature()
+        self._temperature = value
+
     @abc.abstractmethod
     def get_humidity(self):
         """Read hutch humidity.
@@ -111,6 +139,15 @@ class AbstractMachineInfo(HardwareObject):
         """
         return None
 
+    def update_humidity(self, value=None):
+        """Check if humidity has changed.
+        Args:
+            value: value
+        """
+        if value is None:
+            value = self.get_humidity()
+        self._humidity = value
+
     @abc.abstractmethod
     def get_flux(self):
         """Read flux.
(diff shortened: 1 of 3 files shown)


IvarsKarpics:
I would suggest to keep abstract class very simple. I would use MachineInfoMockup.py as base containing self.values_ordered_dict and abstract get_current, get_message methos (maybe there was some other for filling ispyb). Flux should not be here. Whats inside the ordered_dict is up to the site specific implementation.
There are MachineInfo and MachInfo, MachineInfoMockup and MachInfoMockup. Maybe we can clear this.

marcus-oscarsson:
Thanks for all the nice work, its a good start :). I agree with @IvarsKarpics that it's maybe better to keep it simple and remove a few of the things that is not directly associated with the machine info.

To make it easier :), there are of course two mockup classes, MachInfoMockup and MachineInfoMockup. It looks like you have looked at the later, MachineInfoMockup which contains more attributes like disk space and humidity. This hardware object is in my opinion a bit too tightly coupled with whats displayed in the brick. It would be interesting to have the information about the diskspace and humidity but we should move those to where the better belong.

So to summarize :), 
- It would be better to look at MachInfoMockup (https://github.com/mxcube/HardwareRepository/blob/master/HardwareObjects/mockup/MachInfoMockup.py). Which is a bit simpler compared to what you implemented here :).

- We should the rename it to MachineInfoMockup and then remove the old one.

- We can then discuss where to put humidity and disk space

laispc:
Hi! Thank you very much for the feedback! I agree that this class can be much more simple indeed. :)
Actually I tried to combine what I found in both MachineInfoMockup and MachInfoMockup. But it started to become too complex for an abstract class.

So from now I am looking more at MachInfoMockup as a base and removing the unnecessary parts.

I am implementing all suggestions made above and uploading a new, reviewed PR.
So I think we can close this and continue the discussing on the new one, ok?

Before going to the other PR, some questions:

- Should we keep the update methods?
- Besides current and message, is there other attributes that should be added? Top up remaining, lifetime?
Does temperature remain here?

marcus-oscarsson:
> Hi! Thank you very much for the feedback! I agree that this class can be much more simple indeed. :)
> Actually I tried to combine what I found in both MachineInfoMockup and MachInfoMockup. But it started to become too complex for an abstract class.
> 
> So from now I am looking more at MachInfoMockup as a base and removing the unnecessary parts.
> 
> I am implementing all suggestions made above and uploading a new, reviewed PR.
> So I think we can close this and continue the discussion on the new one, right?
>
Sure, its as you like. Whatever you think its the most practical :)

> Before going to the other PR, some questions:
> 
> * Should we keep the update methods?

The update methods indeed seems a bit unnecessary for an abstract class, there is an update_values method inherited from BaseHardwareObjects which role is to re-emit relevant values. I think that method could be used instead. It looks like the update_* methods here are mostly used internally to update a protected attribute.

> * Besides current and message, is there other attributes that should be added? Top up remaining, lifetime?

I guess it could be nice to have those and they are directly related to the information coming from the machine so it makes sense

>   Does temperature remain here?

Temperature should in my opinion be removed

marcus-oscarsson:
@laispc, So would you like that I close this one and you open a new one ?

laispc:
@marcus-oscarsson , thanks for the answer! Ok, I just sent the new PR and I am closing this one.

PR #522, opened 2020-04-07
Removed motor- bug fix.

beteva:
Bug fix.


IvarsKarpics:
Hi @beteva 

Could you please remove also test_motor.py? I guess all motors will be tested in the generic test made by Rasmus.

beteva:
Done

PR #526, opened 2020-04-08
Minor bug fixes

marcus-oscarsson:
Dear all,

Fixed a few smaller bugs.

Cheers,
Marcus

### DIFF
--- a/HardwareObjects/mockup/EnergyMockup.py
+++ b/HardwareObjects/mockup/EnergyMockup.py
@@ -28,10 +28,12 @@ from HardwareRepository.HardwareObjects.abstract.AbstractEnergy import AbstractE
 # Default energy value (keV)
 DEFAULT_VALUE = 12.4
 # Default energy limits (keV)
-DEFAULT_LIMITS = "(4, 20)"
+DEFAULT_LIMITS = (4, 20)
+
 
 class EnergyMockup(AbstractEnergy):
     """Energy Mockup class"""
+
     def __init__(self, name):
         AbstractEnergy.__init__(self, name)
         self.__move_task = None
@@ -39,9 +41,11 @@ class EnergyMockup(AbstractEnergy):
     def init(self):
         """Initialise default values"""
         self.update_value(self.default_value or DEFAULT_VALUE)
-        self.update_limits(
-            ast.literal_eval(self.getProperty("energy_limits") or str(DEFAULT_LIMITS))
-        )
+
+        limits = self.getProperty("energy_limits", None)
+        limits = DEFAULT_LIMITS if limits is None else ast.literal_eval(limits)
+
+        self.update_limits(limits)
         self.update_state(self.STATES.READY)
 
     def _move(self, value):
--- a/HardwareObjects/mockup/EnergyMockup.py
+++ b/HardwareObjects/mockup/EnergyMockup.py
@@ -28,8 +28,7 @@ from HardwareRepository.HardwareObjects.abstract.AbstractEnergy import AbstractE
 # Default energy value (keV)
 DEFAULT_VALUE = 12.4
 # Default energy limits (keV)
-DEFAULT_LIMITS = (4, 20)
-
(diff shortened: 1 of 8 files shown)


beteva:
Hello @marcus-oscarsson, It looks like the changes I've made on EnergyMockup and ShutterMockup are not taken into account.

HardwareObjects/mockup/EnergyMockup.py
```diff
@@ -28,23 +26,21 @@
 # Default energy value (keV)
 DEFAULT_VALUE = 12.4
 # Default energy limits (keV)
-DEFAULT_LIMITS = (4, 20)
-
+DEFAULT_LIMITS = "(4, 20)"
```
beteva, line 29: I have left it as a tuple, not a string, because this way it corresponds to _nominal_limits.

HardwareObjects/mockup/EnergyMockup.py
```diff
@@ -53,12 +49,12 @@ def _move(self, value):
...
             step = -1 if value < start_pos else 1
-            for _val in range(int(start_pos) + step, int(value) + step, step):
-                self.update_value(_val)
+            for ii in range(int(start_pos) + step, int(value) + step, step):
+                self.update_value(ii)
                 time.sleep(0.2)
         return value
 
-    def set_value(self, value, timeout=None):
+    def set_value(self, value, timeout=0):
```
beteva, line 57: The timeout has to be None or the test has to change in order to wait until the value is reached. Otherwise the test fails.

HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -29,8 +29,8 @@ class ShutterMockup(AbstractNState.AbstractNState):
 
     def init(self):
         super(ShutterMockup, self).init()
-        self._nominal_value = self.VALUES.UNKNOWN
-        self._state = self.STATES.UNKNOWN
+        self._nominal_value = self.VALUES.CLOSED
+        self._state = self.STATES.READY
```
beteva, line 33: The test_shutter has to be changed accordingly or the value has to be set back to UNKNOWN. Otherwise the test will fail.

beteva:
Well done, we have forgotten about  the move_relative.

HardwareObjects/mockup/EnergyMockup.py
marcus-oscarsson: It cant be a tuple if we are going to pass it to ast.literal_eval

HardwareObjects/mockup/EnergyMockup.py
marcus-oscarsson: Sorry, yes seems like I don't have the latest changes ...

HardwareObjects/mockup/ShutterMockup.py
marcus-oscarsson: Ok, so ill do that

HardwareObjects/mockup/EnergyMockup.py
rhfogh: I do not think it should be passed to literal_eval. It all started with a bug of mine, that Antonia fixed, but the place it is needed (line 44 in EnergyMockup.init) could and should simply be refactored, so you evaluate only the configured property, and use the default tuple if there is no property.

marcus-oscarsson:
Updated with master and not passing DEFAULT_LIMITS to literal_eval :)

HardwareObjects/mockup/EnergyMockup.py
beteva: This was actually my bug fix. In the code only the value, coming from the property is under ast.

marcus-oscarsson:
Okey so I think I got it right this time :), sorry for the small mess I forgot to update with latest master ... 

I'm not working this afternoon but ill try to do any smaller updates to the PR's if needed so that we can continue and get things merged :)

beteva:
Looks good to me. Thanks @marcus-oscarsson

PR #528, opened 2020-04-09
MicrodiffZoom handles B-Zoom and MotorisedZoom.

beteva:
Mockup handles specific values format.

### DIFF
--- a/HardwareObjects/MicrodiffZoom.py
+++ b/HardwareObjects/MicrodiffZoom.py
@@ -21,14 +21,16 @@
 MicrodiffZoom
 
 Example xml file:
-<device class="MicrodiffZoom">
+<object class="MicrodiffZoom">
   <username>zoom</username>
   <exporter_address>wid30bmd2s:9001</exporter_address>
   <value_channel_name>CoaxialCameraZoomValue</value_channel_name>
   <state_channel_name>ZoomState</state_channel_name>
-</device>
+  <!-- if levels do not corresponf to values -->
+  <values>{"LEVEL1": 1, "LEVEL2": 2, "LEVEL3": 3, "LEVEL4": 4, "LEVEL5": 5, "LEV
+EL6": 6}</values>
+</object>
 """
-
 
 from enum import Enum
 from HardwareRepository.HardwareObjects.abstract.AbstractNState import BaseValueEnum
@@ -40,6 +42,7 @@ __license__ = "LGPLv3+"
 
 class MicrodiffZoom(ExporterNState):
     """MicrodiffZoom class"""
+
     def __init__(self, name):
         ExporterNState.__init__(self, name)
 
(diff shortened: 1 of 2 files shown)


HardwareObjects/mockup/MicrodiffZoomMockup.py
```diff
@@ -1,38 +1,64 @@
...
     def init(self):
         """Initialize the zoom"""
         AbstractNState.init(self)
 
-        try:
-            values = ast.literal_eval(self.getProperty("values")).values()
-        except AttributeError:
-            limits = (0, 10)
+        self.initialise_values()
+        _len = self.VALUES.__members__.__len__() - 1
```
rhfogh, line 29: _len = len(self.VALUES) - 1
should give the same result

HardwareObjects/MicrodiffZoom.py
```diff
@@ -40,51 +42,66 @@
...
-            if _high == float("inf"):
-                _high = 10
-
-            self.set_limits((_low, _high))
-        except ValueError:
-            self.set_limits((1, 10))
-
         self.initialise_values()
+        # check if we have values other that UKNOWN
+        _len = self.VALUES.__members__.__len__() - 1
```
rhfogh, line 55: _len = len(self.VALUES) - 1
should give the same result

rhfogh:
Looks good.

The len(self.VALUES) stuff is about clarity, not funcionality.

One question: It looks from the code that it is assumed that the limits are integers (or maybe None). Should this not be reflected in the function documentation, so it mentioned tuples of 'int' instead of tuples of 'elements'.

If this is so, it might reflect also on the documentation of AbstractActuator - if limits are either ints of floats or None (and can they be anything else), maybe htat should be rflected in the documentation also there.

One idea:
We could use IntEnum for AbstractZoom.VALUE, or maybe even for AbstractNState.VALUE in general. An intEnum works arithmetically a an integer, so the limits could be the actual enumeration members, and you could do comparison, tests and sorting on them as if they were inegers.

HardwareObjects/mockup/MicrodiffZoomMockup.py
beteva: OK

HardwareObjects/MicrodiffZoom.py
beteva: OK*2

beteva:
> One question: It looks from the code that it is assumed that the limits are integers (or maybe None). Should this not be reflected in the function documentation, so it mentioned tuples of 'int' instead of tuples of 'elements'.

You are right. i did change, but apparently not everywhere. 
> If this is so, it might reflect also on the documentation of AbstractActuator - if limits are either ints of floats or None (and can they be anything else), maybe htat should be rflected in the documentation also there.
> 
This is specific fro the MD2 Zoom. I would only document it in the MicrodifZoom.py
> One idea:
> We could use IntEnum for AbstractZoom.VALUE, or maybe even for AbstractNState.VALUE in general. An intEnum works arithmetically a an integer, so the limits could be the actual enumeration members, and you could do comparison, tests and sorting on them as if they were inegers.

Unfortunately not, as we have the "UNKNOWN" value, which is a string and I would say , better stay this way. Otherwise we have to think about what would be a value, which represents the UNKNOWN, but this is tricky.

rhfogh:
@beteva 
Makes sense.

marcus-oscarsson:
Wonderful ! :)

PR #530, opened 2020-04-13
Added missing update_values functions to abstract classes

rhfogh:
Added missing AbstractActuator.update_values and AbstractNState,update_values

PR #531, opened 2020-04-15
Added the possibility to configure the MotorMockup to wrap

marcus-oscarsson:
Added the possibility to configure the MotorMockup to wrap


HardwareObjects/abstract/AbstractMotor.py
```diff
@@ -108,6 +108,7 @@ def validate_value(self, value):
         if math.isnan(value) or math.isinf(value):
             return False
         limits = self._nominal_limits
+
```
beteva, line 111: Why change?

marcus-oscarsson: Ohh, thats a misstake

HardwareObjects/mockup/MotorMockup.py
```diff
@@ -60,6 +63,12 @@ def init(self):
         velocity = self.getProperty("velocity", DEFAULT_VELOCITY)
         self.set_velocity(velocity)
 
+        try:
+            wr = self.getProperty("wrap_range", None)
+            self._wrap_range = DEFAULT_WRAP_RANGE if not wr else ast.literal_eval(wr)
+        except (ValueError, SyntaxError):
+            self._wrap_range = DEFAULT_WRAP_RANGE
+
```
beteva, line 71: I think in order to get self._wrap_range to be None, it has to be
wr = self.getProperty("wrap_range", "None"). Otherwise we will always have a value. 
Or maybe it is OK as it is, but the DEFAULT_WRAP_RANGE should be = 1
Needs a check as well that wr is not 0, because of ast.literal_eval("0") is 0.

beteva:
Thanks @marcus-oscarsson for the  useful feature.

HardwareObjects/mockup/MotorMockup.py
rhfogh: @beteva 
As I read the code, getProperty will return None if there is no such property. You should not even need to write 
self.getProperty("wrap_range", None)
it would be enough with
self.getProperty("wrap_range")

rhfogh:
@marcus-oscarsson 
That does indeed look nice. it will need merging (or modifying one of them), though, to combine it with  #529.

HardwareObjects/mockup/MotorMockup.py
marcus-oscarsson: @rhfogh the None default value is indeed not necessary but I think its clearer this way. I can remove it if its disturbing :)

@beteva as it is know it will get the config value and, if there is one , run literal_eval on it otherwise return the DEFAULT_WRAP_RANGE.

marcus-oscarsson:
@rhfogh, @all No problem, I actually thought of adding it as a concept on AbstractMotor, something like soft wrapping but I don't think anyone is using that. This is in anyways needed to simulate the behavior of omega, I'm happy to make any updates once #529 is merged or if you prefer to do the opposite ( merging this first and then do the change). Does not matter to me ...

HardwareObjects/mockup/MotorMockup.py
rhfogh: @marcus-oscarsson 
No reason to make changes to the function call since it iw correct wither way.

rhfogh:
@marcus-oscarsson 
Let us just merge things in teh orde rthey come naturaly, then.

HardwareObjects/mockup/MotorMockup.py
beteva: @marcus-oscarsson, maybe it is me who does not get it right. If there is no value, then it will be the DEFAULT_WRAP_RANGE. But then the DEFAULT_WRAP_RANGE = 360. That means that in _move ans set_value will always be a self._wrap_range. **Is this what we want?** That's why as an alternative to the None I've suggested DEFAULT_WRAP_RANGE = 1.

beteva: @rhfogh I know that the default value of getProperty is None. I've suggested to change it to the string "None" in order to get None when ast.literal_eval is executed.

marcus-oscarsson: Ohh, I see well I did that way since the default limits were configured to +-360 so the already existing default config indicating that it was an omega mockup (by default). But I dont mind to change to DEFAULT_WRAP_RANGE = 1.

rhfogh: @beteva 
1) By Golly, you are right! We would not want wrapping at 360 when you are not setting a wrap_range property, we would want no wrapping.
2) My taste would be to not run the eval if there was no value, not to set a default value which goes through eval the way we want - why make it more complex than you have to? But sure, either will work.

beteva: Whatever you like: DEFAULT_WRAP_RANGE = 1 or wr = self.getProperty("wrap_range", "None"). The result will be either self._warp_range = 1 or self._warp_range = None, which does not matter for the final result. But the feature quite useful and we may want to add it to AbstractMotor. If this is the case, than we should probably choose the "None" default value.

marcus-oscarsson: Okey, so:
 
1. Ive changed the limits to -10000 and 10000 and set default wrap range to 1.
2. Eval does not run unless there is value in the config ...

rhfogh: 1) If you look at the code that actually uses wrap_range, it only wraps if val evaluates to true. Which is also what I would recommend. So, the simplest option is to set wrap_range to None if you do not want wrapping.

2) A completely independent question is whether you want wrapping on or off by default. It sound like we would agree to have it off off. If so the simplest way of doing it is to get rid of DEFAULT_WRAP_RANGE entirely. I owuld propose the following code:

        wr = self.getProperty("wrap_range", None)
        if wr:
            try:
                wr =  ast.literal_eval(wr)
            except (ValueError, SyntaxError):
                pass
        self._wrap_range = wr

marcus-oscarsson:
@rhfogh could we merge this one before #530 ?

PR #533, opened 2020-04-16
Fixed BlissMotor issues to work with MotorSpinBoxESRFBrick

natxo14:
Get status from bliss motor on init.

Fixed issues with states : make update_state function work with different types of 'state' input parameter ( bliss states, HWRObject states, specific_states ... )

### DIFF
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -58,7 +58,6 @@ class BlissMotorStates(enum.Enum):
     DISABLED = 7
     UNKNOWN = 8
 
-
 class BlissMotor(AbstractMotor):
     """Bliss Motor implementation"""
 
@@ -78,7 +77,7 @@ class BlissMotor(AbstractMotor):
     def __init__(self, name):
         AbstractMotor.__init__(self, name)
         self.motor_obj = None
-
+        
     def init(self):
         """Initialise the motor"""
         AbstractMotor.init(self)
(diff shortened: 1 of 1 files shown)


HardwareObjects/BlissMotor.py
```diff
@@ -98,11 +100,21 @@ def update_state(self, state=None):
...
+                # if 'state' is already a HardwareObjectState no need to convert it.
+                self._specific_state = state
+                AbstractMotor.update_state(self, state)  
+                return
+            else:
+                # state comming from bliss (with current_states_names attribute)
+                try:
+                    state = state.current_states_names[0]
+                except (AttributeError, KeyError):
+                    state = "UNKNOWN"
```
beteva, line 117: update_state was meant to receive bliss motor state, not HardwareOblectState. The command current_states_names is a bliss command, which gives the name of the state, reported from the bliss motor. Current changes break the execution.

natxo14: The problem is that in MotorSpinBoxBrick (MSBB), on update_gui function there's a call to 
`self.motor_hwobj.update_values()`
In class HardwareObjectMixin update_values calls BlissMotor update_state() with state=None.
And the way BlissMotor was done before, it meant that the motor state was going to be set to UNKNOWN, which I thought was wrong.

In the new BlissMotor logic, maybe i'm wrong calling get_state() inside a function to update the state of the HWRObject, but the logic I implemented does not change the execution like you said:

My logic in BlissMotor update_state is:
'state' can be either: None, a boolean, or a bliss state.
* If boolean: already done
* if the state is None, then I get it from the HWRObject parent class. And as it is a HardwareObjectState, I don't need it to be 'translated' like if it was a specific_state
* if it is a bliss state (nor a boolean nor a HardwareObjectState), then I transform it with bliss' current_states_names and go on 

I've tried this implementation with a MSBBrick and it worked. Did you found trouble with any brick on your side??

beteva: The code look good, only as @rhfogh pointed out, we should overload the abstract methods only if really necessary. It was my fault to use update_state instead of a specific method. This will come with the next PR.

PR #535, opened 2020-04-16
Improved generic tests v3

rhfogh:
11 minutes ago

Replaces #529 

For some reason the previous PR showed more differences than it should, so I am replacing it with a new cleaner version

    @IvarsKarpics I have moved the abstract test classes to test/pytest (I made some silly errors with relative imports, which is why I could not get them to work there until IK showed me how).

    @marcus-oscarsson I have added tests for emission of stateChanged, valueChanged, and limitsChanged.

    @beteva
    I removed the velocity-setting from AbstractMotor. But removing the changes in AbstractActuator caused the tests to fail. Maybe we need to change the tests too, or to change AbstractActuator so they pass in another way. but I think it is better that I make the PR so that the test pass, and leave it to you to make the modifications - after all you are the one who knows what we need here.

I only have use of this hotspot at short, irregular intervals; sorry for being mostly offline, still.


HardwareObjects/mockup/MockActuator.py
```diff
@@ -0,0 +1,121 @@
...
+import time
+import random
+import gevent
+from HardwareRepository.HardwareObjects.abstract import AbstractActuator
+
+__copyright__ = """ Copyright © 2010-2020 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+class MockActuator(AbstractActuator.AbstractActuator):
```
IvarsKarpics, line 38: maybe rename class and file to ActuatorMockup (like all mockups in the mockup dir)

beteva: I do agree that we should swap the two. @IvarsKarpics it was actually you who started with the name MockupActuator ;-)

rhfogh:
@beteva 
My internet is back, and I'll make the name change right now.

rhfogh:
Replaced by #536

PR #536, opened 2020-04-17
Improved generic tests v4

rhfogh:
Replaces #535 

 - Renamed MockActuator to ActuatorMockup
- Removed unused MockupActuator
- For some reason I get zillions of duplicate commits and mergings when trying to add to an old PR and push the result. Hence I have made a new PR replacing the old one, with the additions as shown


HardwareObjects/mockup/ActuatorMockup.py
```diff
@@ -0,0 +1,121 @@
...
+import time
+import random
+import gevent
+from HardwareRepository.HardwareObjects.abstract import AbstractActuator
+
+__copyright__ = """ Copyright © 2010-2020 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+class ActuatorMockup(AbstractActuator.AbstractActuator):
```
marcus-oscarsson, line 38: This is a really nice idea @IvarsKarpics  and @rhfogh  ! :+1:

test/pytest/TestAbstractActuatorBase.py
```diff
@@ -116,39 +122,53 @@ def test_limits_setting(self, test_object):
...
         startval = test_object.get_value()
+        test_object.set_value(startval, timeout=0)
 
         test_object._nominal_value = None
-        with pytest.raises(BaseException):
+        with pytest.raises(RuntimeError):
             test_object.set_value(startval, timeout=0)
             test_object.wait_ready(timeout=1.0e-6)
+
+    def test_signal_value_changed(self, test_object):
```
marcus-oscarsson, line 162: Great, nicely done !

test/pytest/TestAbstractActuatorBase.py
```diff
@@ -116,39 +122,53 @@ def test_limits_setting(self, test_object):
...
+        catcher = TestHardwareObjectBase.SignalCatcher()
+        val = test_object.get_value()
+        test_object._nominal_value = None
```
marcus-oscarsson, line 165: We should maybe add comment here, explaining why ._nominal_value is set to None I'm not sure that one
would immediately think about it when one reads the code again in a while ...

test/pytest/TestHardwareObjectBase.py
```diff
@@ -78,18 +89,42 @@ def test_update_state(self, test_object):
...
+            # Timeout to guard against waiting foreer if signal is not sent)
+            with gevent.Timeout(1.0):
+                catcher.async_result.get()
+        finally:
+            test_object.disconnect("stateChanged", catcher.catch)
+
+
+
+class SignalCatcher(object):
+    """Utility class to test emissoi of signals
```
marcus-oscarsson, line 120: Small typo ;)

test/pytest/TestAbstractActuatorBase.py
rhfogh: @marcus-oscarsson I'll add a comment.
The reason is that 1) the value needs to change to get teh effect, so there has to be two different values, 2) at the AbstractActuator level the only values you know for sure are valid are None and the starting value.

marcus-oscarsson: Yes I realized that but do you remeber that in one year when you look at the code ?

rhfogh:
Obsolete - replaced by #543

PR #538, opened 2020-04-17
Blissmotor states

natxo14:
A minor change in BlissMotor : 
on update_state function, checking whether 'state' is a HardwareObjectState the following didn't work as expected:
`if state in HardwareObjectState:`

changed to:
`if isinstance(state, HardwareObjectState):`

Hope not to touch this anymore until Antonia finishes on her side.


marcus-oscarsson:
Thanks @natxo14, I think @beteva is looking into this at the same time right ?

beteva:
@marcus-oscarsson , this PR was replaced by #540.

rhfogh:
@natxo14 
'sinstance 'is fully as good as 'in', so the change is fine.

But, for my education, can you give your code where it does not work?
The following works for me, in Python 2.7:

  >>> from enum import Enum
  >>> class EE(Enum):
  ...   a = 1
  ...   b = 2
  ...   c = 3
  ... 
  >>> 'a' in EE
  False
  >>> EE.a in EE
  True
  >>> isinstance(EE.a, EE)
  True

marcus-oscarsson:
@natxo14 did #540 solve your issue ?

natxo14:
@marcus-oscarsson , yes it did
@beteva , thanks a lot!

@rhfogh 
Maybe I draw it too fast, I've been trying to reproduce the issue I thought I had with precedent file versions but I don't get it. So I guess that 'in' was good from the beginning.

PR #539, opened 2020-04-17
Fix update values v2

rhfogh:
Replaces #530 .
Due to multiple additional commits and merges in 530 this is a new, clean PR with the same content.


marcus-oscarsson:
Thanks @rhfogh

Issue #537, opened 2020-04-17
state and specific_state handling - particularely for BlissMotor


rhfogh:
@beteva, @natxo14 
I had somecomments on the just accepted BlissMotor PR #533, but I saw afterwards that it is anyway planned to refactor it in a new PR, so I have removed them again. Please disregard the previous version.. Sorry for being inattentive.

natxo14:
Hi Rasmus!

No need to apologise, feedback it's always welcome :) . 
I hope if you don't mind if I put the message you left on the PR, I think it can help to track the issues you found in BlissMotor:

"
The changes to BlissMotor state handling in the recently accepted #533 PR raises some questions.
Specifically, the behavlour of the update_state function seems to be inconsistent with the documentation; which suggests that one or the other maybe ought to be changed.

    The documentation to HardwareObjectMixin.get_state gives the wrong type for the result. This is my fault, a hangover from an old proposal, AFAIAC the documentation ought to be changed to to match that for HardwareObjectMixin.update_state. But since neither documentation matches the current BlissMotor code we had better make a single, concerted set of changes.

    The documentation for BlissMotor.update_state says that the input must be a HardwareObjectState (as indeed required by the superclass function that is being overridden). But the code shows that the input can in practice also be a boolean or a state returned from the BlissMotor

    specific_state is sometimes set to a HardwareObjectState, not a BlissMotorState (line 109). If we want to do that it really ought to be reflected in th documentation, also in the superclass. Also it should be clarified whether the specific state for a given class should return the BlissMotorState or the HardwareObjectState for READY and UNKNOWN, a mixture, or any state with the right name.

If we want to keep to the superclass documentation, we could rewrite the functions to suit. Most obviously we could connect the Motor to a _update_state function, that in turn called update_state with a parameter of the correct type. Alternatively we would decide that the superclass documentation does not reflect what we want, and rewrite the superclasses so that they show the desired implementation.
"
Thanks a lot

beteva:
@rfogh, I am currently working on BlissMotor in order to change the update_state to a local _updte_state method, which only translates the BLISS states to HardwareObjectStates The PR will follow today.

rhfogh:
@beteva Sounds good. Pity I had not noticed your comment it two minutes earlier.

marcus-oscarsson:
@beteva great ! and thanks @natxo14 for the very detailed and thorough feedback :)

marcus-oscarsson:
@beteva, done right ?

[...]

PR #540, opened 2020-04-17
BlissMotor

beteva:
Changed update_state to _update_state. Implemented get_state and get_specific_state.
Closes pull requests #533 and #537

### DIFF
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -92,14 +92,14 @@ class BlissMotor(AbstractMotor):
         self.update_state(self.get_state())
 
     def _state2enum(self, state):
-        """Translate the state to HardwareObjectState ans BlissMotorStates
+        """Translate the state to HardwareObjectState and BlissMotorStates
         Args:
            state (string): state
         Returns:
            (tuple): (HardwareObjectState, BlissMotorStates)
         """
         try:
-            _specific_state = BlissMotorStates.__members__[state]
+            _specific_state = BlissMotorStates[state]
         except (AttributeError, KeyError):
             _specific_state = BlissMotorStates.UNKNOWN
 
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -45,7 +45,7 @@ class BlissMotorStates(enum.Enum):
     LIMNEG  : 'Hardware low limit active'
     HOME    : 'Home signal active'
     OFF     : 'Axis power is off'
-    DISABLED: 'Axis cannot move (must be enabled - not ready ?)' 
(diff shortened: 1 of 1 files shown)


HardwareObjects/BlissMotor.py
```diff
@@ -77,52 +78,72 @@ class BlissMotor(AbstractMotor):
...
+
+    def _state2enum(self, state):
+        """Translate the state to HardwareObjectState ans BlissMotorStates
+        Args:
+           state (string): state
+        Returns:
+           (tuple): (HardwareObjectState, BlissMotorStates)
+        """
+        try:
+            _specific_state = BlissMotorStates.__members__[state]
```
rhfogh, line 102: I think you could also use 
_specific_state = BlissMotorStates[state]
but  this will work as well

HardwareObjects/BlissMotor.py
```diff
@@ -77,52 +78,72 @@ class BlissMotor(AbstractMotor):
...
+        self.connect(self.motor_obj, "move_done", self._update_state)
 
         # init state to match motor's one
-        self.update_state(self.motor_obj.state)
-        
-    def update_state(self, state=None):
+        self.update_state(self.get_state())
+
+    def _state2enum(self, state):
+        """Translate the state to HardwareObjectState ans BlissMotorStates
```
rhfogh, line 95: ans -> and

marcus-oscarsson:
Looks good to me, well done @beteva !

marcus-oscarsson:
> This looks good - very quick work too.
> 
> Since this is obviously a good way to handle specific states, maybe we should modify the specific_state part of AbstractActuator to fit? I am thinking about
> 
> 1. changing the return type of AbstractActuator.get_specific_state to, effectively, 'Any' (unless you can think of a more restrictive type).

I guess specific state could be of "any" type but mapping it to a enum would probably not be too complicated in most cases. So enum would be fine in my opinion 

> 2. removing AbstractActuator.update_specific_state and the specificStateChanged signal. I put them in AbstractActuator to treat specific states the same way as normal states, but if there is no need for the extra function and signal so much the better.

I guess it could be be interesting in some cases but the again, if stateChanged is emitted one can assume that the specific state is changed and get the specific state through a function call. The stateChanged signal should always be emitted on a specific state change unless specifically ignored by the specific implementation.

> 
> Please merge this at your convenience, and let me know afterwards what you think should be done about the AbstractActuator (and who should do it).

So we can in my opinion leave it as it is for the moment as we are focusing in the common API and wait with the specific states (especially since there is a way to handle the specific states)

marcus-oscarsson:
Merging this !, @rhfogh please let me know if you would like me to create and issue with the discussion above for the wrap up ?

rhfogh:
@marcus-oscarsson 
Yes. please create an issue about this.

PR #543, opened 2020-04-20
Improved generic tests v6

rhfogh:
This PR is up to date with master, and is a replacement for #536 

Is there a way we can get this one merged relaltively soon, so I do not have to keep making new PRs to catch up with modifications to master (given that rebasing is an unpassable labyrinth)?


IvarsKarpics:
Sorry I do not get this... why there was Imroved generic tests v1..4? Why you are closing PRs and opening new ones? Now I lost completely track of whats been added, removed, changed...

rhfogh:
Sorry about this. The problem comes because I try to rebase/merge each new version so that it is up-to-date with the current master (which keeps changing).  Every time I tried to add new commits on top of the existing PRs, I got around 30 new commits in the list, the same commits appearing more than once with different hashes, ... till Magnus had to ask me to make a new clean PR. Trying to rebase on the newest master kept getting me into complex merges where I kept having to do the same manual merges on the same files again and again. Making a new PR was the only way I could make something that was reasonably simple by itself.

I think this is what can easily happen when you keep adding new PRs to master that modify the same files (AbstractMotor, MotorMockup) that are in older PRs. Of course we should ideally not start from here, as the saying goes. But now we are here, what do you think is the best way forward?

marcus-oscarsson:
Looks fine to me, thanks !

marcus-oscarsson:
The easiest would be git fetch --all and then git rebase origin/master then you should get rid of all those additional commits.

marcus-oscarsson:
I'm ready to merge this

rhfogh:
@marcus-oscarsson 
Great if someone could merge it.

As for 'git fetch' etc., that is my standard procedure, so I shall continue to do that. With this PR, though, I keep having to merge the same files by hand, several times, again and again, as part of every rebase I do, with a high risk of getting the final result mixed up. I suspect I may have to take refuge in merging again in the future, if we cannot avoid these changes-upon-changes-upon-changes situations.

rhfogh:
Thanks!!!!!

PR #544, opened 2020-04-20
Fixed small bug in queue_entry

marcus-oscarsson:
Hi,

Found this small bug when I did some tests

Cheers,
Marcus


IvarsKarpics:
Currently sample_view.shapes is not implemented in the QtGraphicsManager. Could we wait a bit with this PR?

marcus-oscarsson:
Hm, Its a bit complicated since a change made a few PR's back changed from passing shapes to passing sample_view, and this unfortunately brakes things for us as its two separate objects.So we needs this change for things to work. I thought that there was a temporary solution put in place to point .shapes to self by @rhfogh ? I would maybe suggest to put that back and then try to find a common way to do this ?

marcus-oscarsson:
Or if you have another idea to solve the problem :)

marcus-oscarsson:
@IvarsKarpics I was going to start to look at the SampleView would you like to discuss it before i start ?

marcus-oscarsson:
Superseded by #546

PR #545, opened 2020-04-21
MicrodiffAperture inherit from ExporterNState instead of ExporterMotor.

beteva:
Change the MicrodiffAperture to inherit from ExporterNstate rather than ExporterMotor.
I want to ask @all if aperture in/out action should in this class, or treated by a separate class.
Being a separate class is easier as it will be a simple ExporterNState, with the standard set/get_value methods. If we put it here, we would need an extra set/get_in.

### DIFF
--- a/HardwareObjects/MicrodiffAperture.py
+++ b/HardwareObjects/MicrodiffAperture.py
@@ -65,6 +65,16 @@ class MicrodiffAperture(ExporterNState):
         if self.inout_obj:
             self._initialise_inout()
 
+    def _set_value(self, value):
+        """Set device to value
+        Args:
+            value (str, int, float or enum): Value to be set.
+        """
+        if value.name in ("IN", "OUT"):
+            self.inout_obj.set_value(value, timeout=60)
+        else:
+            super(MicrodiffAperture, self)._set_value(value)
+
     def _initialise_inout(self):
         """Add IN and OUT to the values Enum"""
         values_dict = {item.name: item.value for item in self.inout_obj.VALUES}
--- a/HardwareObjects/MicrodiffAperture.py
+++ b/HardwareObjects/MicrodiffAperture.py
@@ -28,7 +28,7 @@ Example xml file:
   <state_channel_name>State</state_channel_name>
   <-- either only factor -->
   <factor>(0.15, 0.3, 0.63, 0.9, 0.96)</factor>
-  <!-- or complete, corresponding to label: (index, size[mm], factor) -->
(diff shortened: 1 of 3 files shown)


marcus-oscarsson:
I personally think it sounds like a good idea

HardwareObjects/MicrodiffAperture.py
```diff
@@ -17,85 +17,111 @@
...
+from HardwareRepository.HardwareObjects.ExporterNState import ExporterNState
+
+__copyright__ = """ Copyright © 2020 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+class MicrodiffAperture(ExporterNState):
     """MicrodiffAperture class"""
 
+    unit = "mm"
```
IvarsKarpics, line 47: Shoud we synchronize units among all hardware objects that are related to beam (slits, aperture, beam). Should it be mm or microns?

beteva: I am not sure it is worth the effort. We have unit and maybe we should use it more often.

HardwareObjects/MicrodiffAperture.py
```diff
@@ -17,85 +17,111 @@
...
+
+Example xml file:
+<object class="MicrodiffAperture">
+  <username>aperture</username>
+  <exporter_address>wid30bmd2s:9001</exporter_address>
+  <value_channel_name>CurrentApertureDiameterIndex</value_channel_name>
+  <state_channel_name>State</state_channel_name>
+  <-- either only factor -->
+  <factor>(0.15, 0.3, 0.63, 0.9, 0.96)</factor>
+  <!-- or complete, corresponding to label: (index, size[mm], factor) -->
```
IvarsKarpics, line 31: I assume these are microns?

beteva: You are right, these are microns

beteva: Yes, repared

HardwareObjects/MicrodiffAperture.py
beteva: Changed unit to microns (um).

marcus-oscarsson:
@IvarsKarpics Are there still changes pending or is this done ?

IvarsKarpics:
We discussed the question regarding the units (to be reviewed later) during the last web meeting. So PR is fine for me.

marcus-oscarsson:
Thanks :)

PR #546, opened 2020-04-21
[CODE CAMP] AbstractSampleView

marcus-oscarsson:
Hi,

Updated AbstractSampleView and related classes, mainly integrates the Shape handling into SampleVIew. This makes the SampleView handling more similair between the two version, the Shapes object was removed.

This effectively makes #544 unnecessary 

All the best,
Marcus


HardwareObjects/SampleView.py
```diff
@@ -29,11 +33,511 @@
...
+
+        if shape:
+            shape.de_select()
+
+    def is_selected(self, sid):
+        """
+        Check if Shape with <sid> is selected.
+
+        Returns:
+            (Boolean) True if Shape with <sid> is selected False otherwise
```
rhfogh, line 169: If there is no shape with that sid it will currently return None. Is taht OK? Should hte documentation say so?

marcus-oscarsson: Probably better if it returns False

marcus-oscarsson:
@all, @IvarsKarpics It would be great it we could get this merged our at least a comment if there is something fundamentally wrong. Since this solves and issue described in #544 that prevents MXCuBE3 from running properly

marcus-oscarsson:
Thanks :)

PR #548, opened 2020-04-22
AbstractShutter

### LINKED ISSUE #374: Incompatible shutter statees
The states for different kinds of shutters are incompatible, some use 'OPEN', some use 'opened', and the code that checks is not handling this confusion properly. 
How should this be fixed? Which is the standard? Who is empowered to modify all code to fit?
In the interim I shallp ut in any old hack to get my tests to run

This give me the following error when trying to run in mock mode:

2019-04-24 13:22:01,148 |GUI  |WARNING| Data collect is disabled
2019-04-24 13:22:01,148 |GUI  |WARNING|   - Safety shutter is closed (Open the safety shutter to enable collections)

Looking at the states, you have :

AbstractNSTate:@
class ShutterState(IntEnum):
    """
    Defines the valid Shutter states
    """
    UNKOWN = 0
    CLOSED = 1
    OPEN = 2
    MOVING = 3
    AUTOMATIC = 4
    DISABLED = 5
    FAULT = -1
    ERROR = -2

class EMBLSafetyShutter(Device):
    """
    EMBLSafetyShutter defines interface to DESY ics
    """

    shutter_state_list = {
        3: "unknown",
        1: "closed",
        0: "opened",
        9: "moving",
        17: "automatic",
        23: "fault",
        46: "disabled",
        -1: "error",
    }

class MDFastShutter(Device):
    """
    MD Fast shutter
    """
    shutterState = {3: "unknown", 1: "closed", 0: "opened", 46: "disabled"}

beteva:
First attempt to resolve #374.
The class defines BaseValueEnum with 3 "contractual" values - OPEN, CLOSE and UNKNOWN and a specific states Enum - ShutterStates to convert shutter states to BaseHardwareObject states.
The class has only the initialisation methods although this is not really need.


beteva:
Please fill free to comment not only on the implementation, but also on the need of this a class.

rhfogh:
@beteva 
This looks promising. Just one question (not a suggestion, I do not have an answer):

How do we plan to use BaseValueEnum? At the moment it does not seem to be accessed anywhere, unlike the BaseValueEnum in AbstractNState which is 'added to' the values given in the configuration. Just having it there for documentation would be good in itself; if we want we can also write tests to ensure that the classes have value enums that contain the values given in the BaseValueEnum. That would actually be enough.

Can we do better, though? If we 'add' BaseValueEnum automatically like we do in AbstractNState, then we would need to make sure that the configurable values were consistent with the fixed ones. Perhaps we could make BaseValueEnum a class attribute; then we could have an initialise_values function that accessed self.BaseValueEnum, and just override cls.BaseValueEnum in subclasses. It would all depend on whether it is only the names of the enums or also the values that can be the same in various subclases.

Just an idea, over to you.

beteva:
> How do we plan to use BaseValueEnum? At the moment it does not seem to be accessed anywhere, unlike the BaseValueEnum in AbstractNState which is 'added to' the values given in the configuration. Just having it there for documentation would be good in itself; if we want we can also write tests to ensure that the classes have value enums that contain the values given in the BaseValueEnum. That would actually be enough.

@rhfogh, you are right, we have to do something about BaseValueEnum in this case. What I suggest as a solution is to change a bit the update_values in AbstractNState, which will make that the BaseValueEnum is taken into account in the case of AbstractShutter or any other class, that may want to replace the simple BaseValueEnum by somthing else can do it. That implies, though, that we always check that there is UNKNOWN in VALUES. I'll make a PR soon.

rhfogh:
@beteva 
That looks excellent

HardwareObjects/BlissShutter.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+@unique
+class ShutterStates(Enum):
+    """Shutter states definitions."""
+
+    OPEN = HardwareObjectState.READY, "OPEN"
+    CLOSED = HardwareObjectState.READY, "CLOSED"
+    MOVING = HardwareObjectState.BUSY, "MOVING"
+    DISABLED = HardwareObjectState.WARNING, "DISABLE"
```
rhfogh, line 50: Is it "DISABLE" rather than "DISABLED"?

beteva: Yes, this is a typo. It is  "DISABLED"

marcus-oscarsson:
Looks good to me too :)

PR #551, opened 2020-04-27
Abstract nstate fix

beteva:
Bug fix.


HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -58,12 +59,12 @@ def init(self):
         self.default_value = self.getProperty("default_value")
         if self.default_value is not None:
             self.update_value(self.default_value)
-        # NB We cannot set limits to (default_value, default_value) for read_only
-        # - the value might be variable even if unsettable.
-        # If this behaviour is desired e.g. for AbctractEnergy it must ba coded locally
         limits = self.getProperty("default_limits")
         if limits:
-            self.update_limits(tuple(eval(limits)))
+            try:
```
rhfogh, line 64: @beteva 
I know we have already looked at it, but, for my education,  why is it better to avoid 'update_limits'? The update_limits functions we have just set _nomal_limits and emit "limitsChnaged". Is it harmful to emit the signal at this point? Or is it because update_limits is not implemented in all subclasses? In the latter case, could you not assume that unless update_limits is implemented, it would be an error to have a default_limits porperty in the first place?

beteva: @rhfogh, as you point out, sometimes (e.g. AbstractNState) raises NotImplementedError for set and update limits. I would like to avoid  the NotImplementedError, but still have the _nominal_limits set, as they may be used.

rhfogh:
@beteva 
That makes sense. I wonder if we need to revisit update_limits at some point (see below), but for now I agree with you.

The reason for revisiting would be that I can see update_limits should be disabled where there are no meaningful limits in the class, or where the limits come from the hardware and are not settable. But if the limits can be set by configuration then they would be in some way 'user settable', no? But anyway, that will be for another time, if ever.

beteva:
@rhfogh, you have closed, but not merged this PR. Is it for purpose?

rhfogh:
@beteva 
Sorry, that was a complete mistake. I thought I was closing the discussoin thread, not teh PR.

PR #552, opened 2020-04-27
Bliss classes

beteva:
Change ESRF/Transmission to use AbstractTransmission.
Added warning for few deprecated classes.
Change MicrodiffLightBeamstop to use ExporterNState.

### DIFF
--- a/HardwareObjects/ESRF/Transmission.py
+++ b/HardwareObjects/ESRF/Transmission.py
@@ -19,93 +19,40 @@
 #  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
 """Transmission with bliss """
 
-from HardwareRepository.BaseHardwareObjects import HardwareObject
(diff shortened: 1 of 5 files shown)


HardwareObjects/ESRF/Transmission.py
```diff
@@ -19,93 +19,40 @@
...
-    def __init__(self, name):
-        HardwareObject.__init__(self, name)
+    unit = "%"
 
-        self.__transmission = None
-        self.labels = []
-        self.indexes = []
-        self.attno = 0
+    def __init__(self, name):
+        AbstractTransmission.__init__(self, name)
```
rhfogh, line 36: Might it be better style tu use 'super'? It makes no difference now, but if we ever change the inheritance hierarchy that is one thing less to modify.

beteva: Thanks, that is what I was intending to do for all my classes.

marcus-oscarsson:
:+1: :100: :)

PR #553, opened 2020-04-27
[CODE CAMP] AbstractMachineInfo issue #448 (Review 2)

laispc:
Hi,
This reviews the PR #518, implementing improvements discussed on it.

- [x] Machine current attribute
- [x] Message attribute
- [x] Lifetime attribute
- [x] Topup remaining attribute
- [x] Adapt mockup class (MachInfoMockup)
- [x] Review docstrings
- [x] Remove old MachineInfoMockup
- [x] Rename: MachInfoMockup -> MachineInfoMockup
- [x] Write test (test_mach_info.py)

Again, feedback is very welcome!
Thanks,
Laís


laispc:
Question:

- Beam size and transmission show be in this class? If so, I will add them following the same logic as current, message and so on.

laispc:
I made some changes in MachInfoMockup, such as sub-classing the new abstract class and using BaseHardwareObjects.update_values. Let me know if this is going in the right way. :)

marcus-oscarsson:
Thanks @laispc looks good, beam size and transmission does not belong to machine info :)

HardwareObjects/abstract/AbstractMachineInfo.py
```diff
@@ -0,0 +1,81 @@
...
+    @abc.abstractmethod
+    def get_topup_remaining(self):
+        """Read top up remaining.
+        Returns:
+            value: Top up remaining.
+        """
+        return None
+
+    @abc.abstractmethod
+    def get_value(self):
```
rhfogh, line 76: Should we maybe rename this get_info?
get_value is OK semantically (the value of the info is information) but since this is not an AbstractActuator maybe we should have a different name to differentiate it.

Also, should the documentation say that this should return a *copy* of the dict - we do not want to expose the internal dict.

HardwareObjects/abstract/AbstractMachineInfo.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @abc.abstractmethod
+    def get_message(self):
+        """Read message.
+        Returns:
+            value: Message.
+        """
+        return None
+
+    @abc.abstractmethod
```
beteva, line 59: Could it be not abstract, please.

HardwareObjects/abstract/AbstractMachineInfo.py
```diff
@@ -0,0 +1,81 @@
...
+    def get_lifetime(self):
+        """Read life time.
+        Returns:
+            value: Life time.
+        """
+        return None
+
+    @abc.abstractmethod
```
beteva, line 67: Could it be not abstract, please.

HardwareObjects/abstract/AbstractMachineInfo.py
```diff
@@ -0,0 +1,81 @@
...
+    def get_current(self):
+        """Read current.
+        Returns:
+            value: Current.
+        """
+        return None
+
+    @abc.abstractmethod
```
beteva, line 51: Could it be not abstract, please.

beteva:
Great work, thanks.

laispc:
> Thanks @laispc looks good, beam size and transmission does not belong to machine info :)

Ok, thanks!

HardwareObjects/abstract/AbstractMachineInfo.py
laispc: Nice! I agree, @rhfogh . Thanks for the review. I first thought that all abstract classes should have a get_value, but now I see it may not apply to some cases like this one. :)

I found an example on AbstractBeam:

https://github.com/mxcube/HardwareRepository/blob/master/HardwareObjects/abstract/AbstractBeam.py#L176 

I will modify the method here following the example above. What about renaming to 'get_info_dict'? What do you think?

HardwareObjects/abstract/AbstractMachineInfo.py
laispc: Ok, I agree. I will change it. :)

HardwareObjects/abstract/AbstractMachineInfo.py
rhfogh: @laispc
get_info_dict sounds exactly right.

laispc: Ok :)

HardwareObjects/abstract/AbstractMachineInfo.py
laispc: Ok :)

laispc:
> Great work, thanks.

Thanks for the review, @beteva !

laispc:
I just finished the proposed changes and everything is now uploaded here.

If there is further suggestions, please, let me know. Thanks! :)

HardwareObjects/mockup/MachineInfoMockup.py
```diff
@@ -1,189 +1,131 @@
...
-        Updates storage disc information, detects if intensity
-        and storage space is in limits, forms a value list
-        and value in range list, both emited by qt as lists
-        """
-        self.emit("valuesChanged", self.values_ordered_dict)
+        """Override method."""
+        super().update_values()
+        self.t0 = time.time()
+
+        while True:
```
rhfogh, line 70: Maybe it is just me, but how would you ever get out of this loop?

marcus-oscarsson: So I guess the intention is to run forever to simulate the machine current changing ...

laispc:
@rhfogh , @marcus-oscarsson the changes are done. I reverted this part and '_update_me()' is back. :)

marcus-oscarsson:
Its been a while now :), sorry for not checking this one sooner. Could you quickly check what the conflict is ? Then we can merge this one as well

laispc:
@marcus-oscarsson , ok, I'll check. Thanks for the feedback!

laispc:
Hi! I rebased this branch to the current master. So merge conflict seems to be okay now.

About Travis, does it have to pass to all Python versions? Currently, I am developing with 3.6, all tests pass locally and here with Travis. But for other versions (e.g. 2.7) it fails for Travis.

I started to build a 2.7 conda env for local tests, but package versions are conflicting. So now I wonder if tests have to pass to all Py versions. Thanks!

marcus-oscarsson:
Thanks alot, we have an issue with the tests for Python 3.7 and 3.8 due to some update in the build environment. The tests for 2.7 and 3.6 should pass though :)

marcus-oscarsson:
I had a quick look at the test out put and this seems to be the problem for Python 2.7:

```
ERROR    HWR:HardwareObjectFileParser.py:140 Cannot instanciate class MachineInfoMockup
  File "/home/travis/build/mxcube/HardwareRepository/HardwareObjectFileParser.py", line 137, in instanciate_class
    new_instance = class_obj(object_name)
TypeError: Can't instantiate abstract class MachineInfoMockup with abstract methods get_mach_info_dict
ERROR    HWR:HardwareRepository.py:501 Failed to load Hardware object /mach-info-mockup
```
It looks like there is an issue with the definition of an abstract method: `(TypeError: Can't instantiate abstract class MachineInfoMockup with abstract methods get_mach_info_dict)`

rhfogh:
@marcus-oscarsson 
Isn't that just abstract methods working as designed?
The superclass has an abstract method, which means that the subclass must override it. And in this case the subclase doesn't?

marcus-oscarsson:
@rhfogh, @laispc I suspect that might be the problem as well, but I did not double check

HardwareObjects/abstract/AbstractMachineInfo.py
```diff
@@ -0,0 +1,82 @@
...
+    def get_mach_info_dict(self):
+        """Read machine info dictionary.
+        Returns:
+            (dict): Copy of mach_info_dict.
+        """
+        return self._mach_info_dict.copy()
```
beteva, line 82: @all, do you think this method should be implemented as abstract? At least at ESRF we would need several calls to the hardware to fill such a dictionary. Moreover, is there s fixed keys list for the dictionary, or it could be anything? Maybe it would be easier to remove the abstract decorator.

laispc:
>Isn't that just abstract methods working as designed?
The superclass has an abstract method, which means that the subclass must override it. And in this case the subclase doesn't?

@rhfogh , exactly! Thanks for the comment. The confusion lies on this point: Python 3 should be breaking too because there is an abstract method not implemented. So this indicates that Python 3 is, for some reason, not considering MachineInfoMockup an Abtract class. More details in the following lines.

> It looks like there is an issue with the definition of an abstract method: `(TypeError: Can't instantiate abstract class MachineInfoMockup with abstract methods get_mach_info_dict)`

@marcus-oscarsson , thanks, you are right. My bad! I misunderstood this line when I first saw it due to the point I commented above.

>@ALL, do you think this method should be implemented as abstract? At least at ESRF we would need several calls to the hardware to fill such a dictionary. Moreover, is there s fixed keys list for the dictionary, or it could be anything? Maybe it would be easier to remove the abstract decorator.

@beteva I am ok with both keeping or removing. Maybe it would be good to remove it to make it more flexible. I believe that the dict keys used on MachineInfoMockup, for example, are sugested, but not mandatory, and others can be included if needed.

laispc:
So, basically there is 2 topics now:
* Abstract class implementation/syntax differs in Python 2 and 3.
* Should `get_mach_info_dict` remain abstract or not?

To address the 1st topic, I am keeping `get_mach_info_dict` as an abstract method. As @rhfogh already said, both Python 2 and 3 tests should be breaking because this method is not implemented in the mockup class.

Looking on Python docs, the syntax for Abstract Classes indeed changed in Python 3! So if we want to keep both Python 2 and 3 compatible, something should be adapted. I just sent a commit with a suggestion for it (explanation is in the commit itself).
Now both tests are breaking as expected. lol
Please, see if it looks ok or if we should go for another idea.
After this, we can implement this missing abstract method in the mockup OR remove the abstract decorator.

HardwareObjects/mockup/MachineInfoMockup.py
laispc: Just annotating what was done: methos was renamed back to "_update_me()".

HardwareObjects/abstract/AbstractMachineInfo.py
laispc: @beteva I am ok with both keeping or removing. Maybe it would be good to remove it to make it more flexible. I believe that the dict keys used on MachineInfoMockup, for example, are sugested, but not mandatory, and others can be included if needed.

marcus-oscarsson:
We have so far been using ` __metaclass__ = abc.ABCMeta`, which is strictly Python 2.7, and not inheriting  the ABC base class. This does not give us the full benefit of abstract base classes in Python3. Like you point out @laispc we need to make a fix like the one you proposed to do that. 

We have had a brief discussion about it sometime a go and I think we've said that the current solution was acceptable as the intention of whats abstract is clearly stated. But I don't think we were aware that such a nice "one liner" could solve the problem, it also looks like the [six](https://six.readthedocs.io/#six.with_metaclass) package has a solution for it.

It's of course better if we could get all the benefits of the abstract classes as well, so Id suggest that we try the solution proposed by @laispc but in a separate PR and in the meantime we try to get the tests running :)

What do you think ?

laispc:
@marcus-oscarsson thanks for the details!

> But I don't think we were aware that such a nice "one liner" could solve the problem, it also looks like the six package has a solution for it.
Exactly, six can be a good alternative too.

>It's of course better if we could get all the benefits of the abstract classes as well, so Id suggest that we try the solution proposed by @laispc but in a separate PR and in the meantime we try to get the tests running :)
Nice, I agree to treat this in a separated issue. I am adding an issue based on this PR just to annotate this idea for future work, ok?
I am going to remove the one-line fix for now, and remove the abstract method on `get_mach_info_dict` as suggested by @beteva so we get the tests running.

beteva (approved):
Thanks @laispc for sorting out the pytho2-python3 abstract implementation differences.

laispc:
Done!

- Issue created
- One-line fix reverted
- Abstract decorator removed
- Tests passing for Py 2.7 and 3.6

marcus-oscarsson:
Hi, 

Thanks for the nice work ! :), there is a conflict due to a recent merge could you have look at that ?

laispc:
@marcus-oscarsson , rebase done, conflicts solved.

marcus-oscarsson:
Wonderful :)

PR #554, opened 2020-04-28
AbstractShutter remove specific states

beteva:
Remove ShutterStates. Upgrade ShutterMockup to use AbstractShutter.

### DIFF
--- a/HardwareObjects/mockup/ShutterMockup.py
+++ b/HardwareObjects/mockup/ShutterMockup.py
@@ -17,28 +17,46 @@
 #  You should have received a copy of the GNU Lesser General Public License
 #  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
 
-from HardwareRepository.HardwareObjects.abstract. AbstractNState import AbstractNState
+""" Mockup shutter implementation"""
+
+from enum import Enum, unique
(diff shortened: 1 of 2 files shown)


HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -17,28 +17,46 @@
...
     ShutterMockup for simulating a simple open/close shutter.
     """
 
+    SPECIFIC_STATES = ShutterStates
+
     def init(self):
+        """Initialisation"""
         super(ShutterMockup, self).init()
-        self.update_value(self.VALUES.CLOSED)
+        self.update_value(self.VALUES.CLOSE)
```
rhfogh, line 49: @beteva You are *sure* that this should be 'CLOSE', and not 'CLOSED'??
'CLOSE' is either a verb ('fermer') which is strange for a value (not an action), or it is an adjective meaning 'not far away ('proche'). 'CLOSED' would be the obvious linguistical choice

beteva: @rhfogh, I was thinking as a verb, but you are right, it should be a noun. I will change it.

IvarsKarpics:
Are we happy with this PR? I am fine with it.

marcus-oscarsson:
Agree :)

beteva:
Done 5 days ago.

PR #556, opened 2020-04-29
AbstractResolution class.

### LINKED ISSUE #450: Create AbstractResolution
Create and AbstractResolution derived from AbstractMotor

beteva:
Defines AbstractResolution as AbstractMotor. Implements distance_to_resolution and resolution_to_distance.
Resolves #450

### DIFF
--- a/HardwareObjects/abstract/AbstractResolution.py
+++ b/HardwareObjects/abstract/AbstractResolution.py
@@ -1,0 +1,218 @@
+# -*- coding: utf-8 -*-
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU General Lesser Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Resolution abstract implementation.
+Overloaded methods: get_state, get_value, get_limits.
+Implemented methods: _set_value, distance_to_resolution, resolution_to_distance.
+Emited signals: valueChanged.
+Hardware object used: energy and detecor.
+The detector object can be defined in the configuration file. If not, the
+one set from the beamline configuration is used.
+"""
+
+import abc
+import logging
+from numpy import arcsin, arctan, sin, sqrt, tan
+from HardwareRepository import HardwareRepository as HWR
+from HardwareRepository.HardwareObjects.abstract.AbstractMotor import AbstractMotor
+
+__copyright__ = """ Copyright © 2010-2020 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+class AbstractResolution(AbstractMotor):
+    """Abstract Resolution class"""
+
+    __metaclass__ = abc.ABCMeta
+    unit = "Å"
+
+    def __init__(self, name):
+        super(AbstractResolution, self).__init__(name)
+        self._hwr_detector = None
+        self._hwr_energy = None
+        self.det_metadata = {}
+
+    def init(self):
+        """Initialisation"""
+        super(AbstractResolution, self).init()
+        self._hwr_detector = self.getObjectByRole("detector") or HWR.beamline.detector
+        self._hwr_energy = HWR.beamline.energy
+        self.det_metadata = self._hwr_detector.get_metadata()
+
+        self.connect(self._hwr_detector.distance, "stateChanged", self.update_state)
+        self.connect(self._hwr_detector.distance, "valueChanged", self.update_distance)
+        self.connect(self._hwr_energy, "valueChanged", self.update_energy)
+        self.connect(self._hwr_energy, "stateChanged", self.update_state)
+
+    def get_state(self):
+        """Get the state of the distance motor.
+        Returns:
+            (enum 'MotorStates'): Motor state.
+        """
+        return self._hwr_detector.distance.get_state()
+
+    def get_value(self):
+        """Read the value.
+        Returns:
+            (float): value.
+        """
+        if self._nominal_value is None:
+            _distance = self._hwr_detector.distance.get_value()
+            self._nominal_value = self.distance_to_resolution(_distance)
+        return self._nominal_value
+
+    def get_limits(self):
+        """Return resolution low and high limits.
+        Returns:
+            (tuple): two floats tuple (low limit, high limit).
+        """
+        _low, _high = self._hwr_detector.distance.get_limits()
+
+        self._limits = (
+            self.distance_to_resolution(_low),
+            self.distance_to_resolution(_high),
+        )
+        return self._limits
+
+    def set_limits(self, limits):
+        """Resolution limits are not settable.
+           Set the detector distance limits instead
+        Raises:
+            NotImplementedError
+        """
+        raise NotImplementedError
+
+    def _set_value(self, value):
+        """Move resolution to value.
+        Args:
+            value (float): target value [Å]
+        """
+        distance = self.resolution_to_distance(value)
+        msg = "Move resolution to {} ({} mm)".format(value, distance)
+        logging.getLogger().info(msg)
+        self._hwr_detector.distance.set_value(distance)
+
+    def _calculate_resolution(self, radius, distance):
+        """Calculate the resolution as function of the detector radius and the distance.
+        Args:
+            radius (float): Detector radius [mm]
+            distance (float): Distance from the sample to the detector [mm]
+        Returns:
+            (float): Resolution [Å]
+        """
+        _wavelength = self._hwr_energy.get_wavelength()
+        try:
+            ttheta = arctan(radius / distance)
+            if ttheta:
+                return _wavelength / (2 * sin(ttheta / 2))
+        except (TypeError, ZeroDivisionError):
+            logging.getLogger().exception("Error while calculating resolution")
+        return None
+
+    def distance_to_resolution(self, distance=None):
+        """Convert distance to resolution.
+        Args:
+            distance (float): Distance [mm].
+        Returns:
+            (float): Resolution [Å].
+        """
+        distance = distance or self._hwr_detector.distance.get_value()
+
+        return self._calculate_resolution(
+            self._hwr_detector.get_radius(distance), distance
+        )
+
+    def resolution_to_distance(self, resolution=None):
+        """Convert resolution to distance.
+        Args:
+            resolution(float): Resolution [Å].
+        Returns:
+            (float): distance [mm].
+        """
+        resolution = resolution or self._nominal_value
+        _wavelength = self._hwr_detector.get_wavelength()
+
+        try:
+            ttheta = 2 * arcsin(_wavelength / (2 * resolution))
+
+            dist_1 = self.det_metadata["bx"] / (tan(ttheta) - self.det_metadata["ax"])
+            dist_2 = self.det_metadata["by"] / (tan(ttheta) - self.det_metadata["ay"])
+            dist_3 = (self.det_metadata["width"] - self.det_metadata["bx"]) / (
+                tan(ttheta) + self.det_metadata["ax"]
+            )
+            dist_4 = (self.det_metadata["height"] - self.det_metadata["by"]) / (
+                tan(ttheta) + self.det_metadata["ay"]
+            )
+
+            return min(dist_1, dist_2, dist_3, dist_4)
+        except (KeyError, ZeroDivisionError):
+            return None
+
+    def get_value_at_corner(self):
+        """Get the resolution at the corners of the detector.
+        Returns:
+            (float): Resolution [Å]
+        """
+        _distance = self._hwr_detector.distance.get_value()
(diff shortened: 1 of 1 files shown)


HardwareObjects/abstract/AbstractResolution.py
```diff
@@ -0,0 +1,218 @@
...
+        self.connect(self._hwr_energy, "stateChanged", self.update_state)
+
+    def get_state(self):
+        """Get the state of the distance motor.
+        Returns:
+            (enum 'MotorStates'): Motor state.
+        """
+        return self._hwr_detector.distance.get_state()
+
+    def get_value(self):
```
rhfogh, line 71: This works. But what is the reason for calculating it off _nominal_value, instead of going directly to distance.get_value() every time? Working off a stored value does add a (maybe small) risk of numbers going out of sync.Is the call to distance.get_value slow?

HardwareObjects/abstract/AbstractResolution.py
```diff
@@ -0,0 +1,218 @@
...
+        Returns:
+            (float): Resolution [Å].
+        """
+        distance = distance or self._hwr_detector.distance.get_value()
+
+        return self._calculate_resolution(
+            self._hwr_detector.get_radius(distance), distance
+        )
+
+    def resolution_to_distance(self, resolution=None):
```
rhfogh, line 142: Would it not be better to move resolution_to_distance and  get_value_at_corner (in whole or in part) to the detector ? Parameters like ax,ay,bx,by are 1) more logical as part of the detector assembly, 2) not used weerywhere. This would allow us to reduce (or even remove) get_metadata from Resolutoin, and put it on teh detector instead.

beteva: The reason I prefer these methods to belong to the resolution class is that I  did not want to make the detector structure too heavy. The detector does not need to depend on energy/wavelength (which is necessary when calculating the resolution) and has to handle already quite a lot of other methods. This is simply question of taste, no technical obstacle. If you think that the detector metadata other than width and height are not used everywhere, than I think we should remove the calculation of the resolution at corners from the abstract class and put it in a specific implementation.

HardwareObjects/abstract/AbstractResolution.py
```diff
@@ -0,0 +1,218 @@
...
+    def __init__(self, name):
+        super(AbstractResolution, self).__init__(name)
+        self._hwr_detector = None
+        self._hwr_energy = None
+        self.det_metadata = {}
+
+    def init(self):
+        """Initialisation"""
+        super(AbstractResolution, self).init()
+        self._hwr_detector = self.getObjectByRole("detector") or HWR.beamline.detector
```
IvarsKarpics, line 55: Hmm, do we need this? Does it implies possibility to use several detectors. Other question: why not use in the code directly HWR.beamline.energy? For me it is confusing if there is various names to reference hardware objects.

HardwareObjects/abstract/AbstractResolution.py
```diff
@@ -0,0 +1,218 @@
...
+    """Abstract Resolution class"""
+
+    __metaclass__ = abc.ABCMeta
+    unit = "Å"
+
+    def __init__(self, name):
+        super(AbstractResolution, self).__init__(name)
+        self._hwr_detector = None
+        self._hwr_energy = None
+        self.det_metadata = {}
```
IvarsKarpics, line 50: maybe also self._det_metadata. But is it not available in the abstract detector? there could be a method detector.get_metadata()?

HardwareObjects/abstract/AbstractResolution.py
rhfogh: See PR #557 for a proposed fix

HardwareObjects/abstract/AbstractResolution.py
rhfogh: My PR #557 gets rid of the need for det_metadata altgether

IvarsKarpics: Ahh, ok @rhfogh .

HardwareObjects/abstract/AbstractResolution.py
marcus-oscarsson: Correct me if I'am wrong, but the way that resolution is currently implemented prevents us from using if with several detectors in a practical way without having to define "resolution detector 1" , "resolution detector 2", ... 

One way to make it more scalable would be to attach resolution to the detector, although I think its fine as it is for the moment.

I agree that getting both detector from config or .beamline seems unnecessary

HardwareObjects/abstract/AbstractResolution.py
marcus-oscarsson: I think its more of a problem in theory, the most important "event" is the last movement and state change which moves the detector to its final position and changes the state from BUSY to READY.  This should be fine if we assume that the underlying motor/controller sends a position for its final move (which I think we have to relay on ...)

It could become a real problem if get_value is used in a polling loop, which I hope its not :) ?

marcus-oscarsson:
#557 was merged into this branch ...

HardwareObjects/abstract/AbstractResolution.py
beteva: @IvarsKarpics, @marcus-oscarsson, the fact o get the detector from beamline is to make it easier to configure. Getting the detectorobject from the configuration makes it possible to choose which detector you want to use and so gives the possibility to use several detectors. All specific to the detector parameters are implemented in the detector object itself, the resolution only adds the energy/wavelength and the calculations. I would not add the resolution as part of the detector, but as I said to @rhfogh, it is question of taste. My opinion is not to put a tot of things in one hardware object.

HardwareObjects/abstract/AbstractResolution.py
rhfogh: @marcus-oscarsson 
I do not think get_value is so used.

It is just a matter of clarity. When I tried to figure out this, it was a mess of whether the authoritative value was nominal_value or actually got from the hardware, and if it is from nominal_value how sure are you that the value is always updated in time, what could go wrong, what signals need to be sent by what objects, what do you need to do to implement something new...
If get_value goes directly to the hardware you know at least that the value you get is current. So unless there is a reason to the econtrary...

HardwareObjects/abstract/AbstractResolution.py
rhfogh: @IvarsKarpics 
I definitely agree that energy should be accessed directly as HWR.beamline.energy., There will be no need to have multiple energies, so keep the code simple.

For the detector there will be in some cases a need for multiple deetectors (@MartinSavko ).But I think the code we have will not realy support multiple simultaneous detectors anyway. Might it not make more sense to do it the simple way for now (HWR.beamline.detector) and then refactor once someone produces an implementation of muti[ple detectors that is known to work?

beteva: @all, please have a look at the init an in particular at the line, which has been ther from the vry beginning

> self._hwr_energy = HWR.beamline.energy

@IvarsKarpics, The problem I see is that we cannot use several detectors simultaneously, but this is the case for everything. Is this what you mean?
What is possible, though is having several detector/resolution couple and choose the one, needed for a particular acquisition, although, as @marcus-oscarsson pointed out, we need to configure detector1/resolution1, detector2/resolution2 .... The current code of both AbstractDetector and AbstractResolution allow this.

HardwareObjects/abstract/AbstractResolution.py
marcus-oscarsson: I would personally ask the "hardware" directly and rely on the underlying libraries take care of caching if necessary. nominal_value is maybe needed for convenience or for some other reasons. It looks to me that its in this case is done to avoid a calculation that is not needed ?

HardwareObjects/abstract/AbstractResolution.py
marcus-oscarsson: I also think that the support for multiple detectors at once needs some more though and we should not try to wiggle it in. Its true that we have rudimentary support for multiple detectors if we keep it configurable as an option. 

I think we should keep the detector configurable because it allows for multiple detectors even if the support we have today is limited it would work, while removing the configuration option would force us to implement one resolution class for each detector (which is even worse).

marcus-oscarsson: Oh, and I think @IvarsKarpics you meant `HWR.beamline.detector` and not `.energy` right ?

rhfogh: @beteva 
The point is not whether self._hwr_energy is a diferent object from HWR.beamline.energy. In practice it is not, that is clear. The problem is that it is an alternative way of writing it. Looking at this code from teh outside you do not know it is the same object, unless you check through the code to make sure; you do not know if someone, from somewhere, might be changing the local attribute. If you ever do reload of the energy object you need to worry whether there is an object, somewhere, that still retains the old copy.  You might even spend time looking for reasons why this is done differently, or be reluctant to refactor, in case there is a reason for the different implementation somewhere. 

I do not want to be fanatic about this; if there is some reason for doing it this way, by all means. But I think that saving a couple of operations is not worth the additional complexity.

rhfogh:
@All

I had an idea: How about making Resolution contained within Detector, so you accessed it as 
HWR.beamline.detector.resolution?

 - You would keep Resolution._hwr_detector and make sure it always pointed to the detector that contained it
 - The code of Detector and Resolution would be completely separate, exactly as it is now, in fact.
 - There would be no problem with tracking which detector matched a Resolution object.
  - Changing or reloading either detector or Resolution would still guarantee that the mapping between detector and resolution was maintained.
 - If you really had several active detectors, you would anyway have to know which detector to use wwhen calculating the resolution

Worth considering?

marcus-oscarsson:
> @ALL
> 
> I had an idea: How about making Resolution contained within Detector, so you accessed it as
> HWR.beamline.detector.resolution?
> 
> * You would keep Resolution._hwr_detector and make sure it always pointed to the detector that contained it
> * The code of Detector and Resolution would be completely separate, exactly as it is now, in fact.
> * There would be no problem with tracking which detector matched a Resolution object.
> * Changing or reloading either detector or Resolution would still guarantee that the mapping between detector and resolution was maintained.
> * If you really had several active detectors, you would anyway have to know which detector to use when calculating the resolution
> 
> Worth considering?

Hm, that's what I just suggested in our discussion above. So yes I think its worth considering but I also think we need to look at why we would do that (to get multiple detector support) and if its worth to do that now. Basically what I just said above :)

HardwareObjects/abstract/AbstractResolution.py
marcus-oscarsson: aha, I did not even realize that it was confusing :) but its true that one could look at it that way. That was enlightening (I think Buddha just slapped me in the face :) )

Jokes a side, I don't think its a big problem and the number of extra operations would be +1 (an assignment) which I think we can afford :). But I see your point, for me its ok as it is but alright if you like to change.

rhfogh:
@marcus-oscarsson 
So you said, yes. I had not read it properly (Sorry). But I not think it would be a problem, really. 

First, most beamlines would have only one detector, certainly at a a time. Second, I think we could use the exact same Resolution class pretty much everywhere, we would just have to instantiate the same class twice, possibly with dfferent config.As it is currently implemented, AbstractResolution will work directly with pretty much any detector without code changes. The only difference is that it would call get_radius() etc. on different detector objects.

The only need I can see for having different Resolutoin classes (as opposed to dfferent objects of the same class) is for detectors with very different geometry - curved or half-cylindrical surfaces, detectors that are offset or have settable 2*theta motors, where the concept of 'radius' does not make sense. But in those cases we would need a differet Resolution class no matter how we configured hte beamline..

beteva:
@ALL, energy, detector distance and resolution are interconnected like this - :
- when energy changes value, resolution should change, but not detector distance.
- when detector distance changes value, resolution should change, but not energy.
- when resolution changes value, detector distance should, but not energy.
- when any of the three changes state, the other two should change state as well.
- in the AbstractDetector, the detector distance serves to calculate parameters, and to give acess via HWR.beamline.detector,  but when it comes to setting the value or handling the state, it is independent.

Considering this, I thought it might be easier and cleaner to not make the detector class heavier by adding the resolution to it.
There is an AbstractResolution, not just Resolution, in order to standardise some of the methods.

rhfogh:
@beteva 
I completely agree with you on this. It is better to have Detector and Resolution as two separate objects, where only the resolution knows about the wavelength, and only the detector nows about width, height, and pixel count.
The question is how to connect those separate objects into the beamlne. And here I wonder whether it would not be the simplest to put the resolution object inside the detector object, like we put hte phimotor object inside the diffractometer object.

marcus-oscarsson:
@beteva, @rhfogh,

It would probably be more elegant to have resolution as a member of detector but I think it could wait until we have thought a bit more on the multiple detector scenario. There is the way I see it no immediate advantage of changing it right now, right ?

rhfogh:
@marcus-oscarsson 
It is an easy enough code change to make even if we make it later. The only questoin is what we do with the _hwr_detector attribute - changes in the config files seem to be harder to get through than code changes. But if we are not sure how to change it I guess we might as well keep it.

marcus-oscarsson:
Sure, I think we need to keep it configurable for the time being if we don't want to move resolution to detector right away. Ill let @beteva have the final word on this :)

MartinSavko:
Hi,

I think it would be most fruitful to consider the concept of Resolution from the point of view of the experiment and a user. 

I am imagining a user defining an experiment or  writing a diffraction plan into ISPyB -- it is only one resolution that he is interested in -- the minimum resolution of the structure that is hoped for or required. It would be unreasonable to expect him being aware of a  particular geometry of detectors in a multi detector setup (this can get complicated further as for a detector positioned of beam we might want to consider both minimum and maximum resolution of bragg spots it can measure). 

Looking at it from that perspective, especially in the multi detector scenario, it seems the most appropriate to hide the complexity of relative configuration of different detectors inside a particular subclass inheriting from AbstractResolution. 

Martin

beteva:
@IvarsKarpics, can we merge this PR ?

IvarsKarpics:
Yes!

PR #561, opened 2020-05-15
Fixed test vulnerability to roundoff error

rhfogh:
Updated tests to avoid test errors in case of roundoff error when setting floats.


marcus-oscarsson:
Thanks :)

PR #562, opened 2020-05-15
Abstract detector and Resolution fixes

marcus-oscarsson:
Hi,

Two minor fixes to AbstractDetector and Resolution.

All the best,
Marcus

### DIFF
--- a/HardwareObjects/abstract/AbstractDetector.py
+++ b/HardwareObjects/abstract/AbstractDetector.py
@@ -70,7 +70,7 @@ class AbstractDetector(HardwareObject):
         """Initialise some common paramerters"""
 
         try:
-            self._metadata = self["beam"]
+            self._metadata = dict(self["beam"].getProperties())
         except KeyError:
             pass
         
--- a/HardwareObjects/Resolution.py
+++ b/HardwareObjects/Resolution.py
@@ -77,9 +77,8 @@ class Resolution(AbstractMotor):
         Returns:
             (float): value.
         """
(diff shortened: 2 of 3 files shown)


HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -70,7 +70,7 @@ def init(self):
         """Initialise some common paramerters"""
 
         try:
-            self._metadata = self["beam"]
+            self._metadata = dict(self["beam"].getProperties())
```
beteva, line 73: This for some reason did not work the last time when I've tested (before the confinement). That's why I've used self["beam"] instead.

marcus-oscarsson: I see, it seems to work fine now :)

rhfogh: Could it be that getProperties osmetimes returns None, if the properties are not set?

marcus-oscarsson: In that case self["beam"] would give AttributeError

marcus-oscarsson: Actually it gives KeyError, the current code is fine :)

marcus-oscarsson:
So it seems that I can merge it then :)

marcus-oscarsson:
Let me know if I commit a terrible crime as I merge my own PR, but it seems like its approved ? :)

PR #564, opened 2020-05-18
Abstract resolution

beteva:
Resolution inherits from AbstractResolution and nor AbstractMotor any  more.


marcus-oscarsson:
Solved conflict

PR #566, opened 2020-05-18
Added a preliminary version of "DataPublishser"

marcus-oscarsson:
The DataPublisher contain DataSources(s) that publishes data to a listening client. The data is published through Redis and can thus be used in a distributed system. So the results can be generated on a computer different from the computer hosting HWR. It further provides a convenient
way to get the data back if the client/server crashes or disconnects while scanning

The UI needs to listen to the start, data, and end events and provide a component that can display the data.

This can perhaps be extended to handle any underlying message queue system but I waned to avoid the complexity for now. What we needed was a central place through which we send plot data or other data to the client and a convenient way to access any "unsaved" data in case of a problem.

Marcus

### DIFF
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -538,6 +538,22 @@ class Beamline(ConfiguredObject):
 
     __content_roles.append("mock_procedure")
 
+    @property
+    def scan_mockup(self):
+        """
+        """
+        return self._objects.get("scan_mockup")
+
+    __content_roles.append("scan_mockup")
+
+    @property
+    def data_publisher(self):
+        """
+        """
+        return self._objects.get("data_publisher")
+
+    __content_roles.append("data_publisher")
+
     # NB this is just an example of a globally shared procedure description
     @property
     def manual_centring(self):
--- a/HardwareObjects/DataPublisher.py
+++ b/HardwareObjects/DataPublisher.py
@@ -1,0 +1,370 @@
+#
+#  Project: MXCuBE
+#  https://github.com/mxcube.
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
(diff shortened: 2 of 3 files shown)


HardwareObjects/DataPublisher.py
```diff
@@ -0,0 +1,370 @@
...
+
+from enum import Enum, unique
+
+from HardwareRepository.BaseHardwareObjects import HardwareObject
+
+
+@unique
+class PlotType(Enum):
+    """
+    Defines defualt plot
```
rhfogh, line 32: typo

rhfogh:
Looks interesting and useful. Is this MXCuBE3 only, though, or how would it fit into the Qt version? Can it be called from anything except UI code? What would be expected to happen when data are pulished this way, and what would it require of teh UI?

HardwareObjects/Beamline.py
```diff
@@ -538,6 +538,22 @@ def mock_procedure(self):
 
     __content_roles.append("mock_procedure")
 
+    @property
```
rhfogh, line 541: I am a little dubious about adding mock objects (like scan_mockup) in the Beamine HO. What would be the case for having it availabe in a non-mock envronment? OK, there is mlock_procedure, but I had assumed that this was only a temorary placeolder to show how procedure would fit in. What is tha status of scan_mockup?

HardwareObjects/DataPublisher.py
marcus-oscarsson: Oh, yes i will fix that :)

HardwareObjects/Beamline.py
marcus-oscarsson: Sure you are right, It was just for my tests. I shall remove it

marcus-oscarsson:
It should work for both  Qt and Web one just needs to listen to the right signals from the DataPublisher. Its up to the client to deal with the data so far its only 1D and possibly 2D data for plotting but it could be basically any data, and exactly what the client does with this data depends on the client.

You get a start, data and stop events from the DataPublisher containing the source and the data. You can also interact directly with the DataPublisher as it is a HardwareObject. To get all sources for instace,

You can of course interact with the DataPublisher from both client and backend but the data flow is meant to go from backend to client (obviously)

rhfogh:
Sounds cool

marcus-oscarsson:
Thanks !

IvarsKarpics:
Hi @marcus-oscarsson 

Thanks for the work to introduce DataPublisher. This was missing in mxcube.
I will have a deeper look on how this could be implemented in qt version.

marcus-oscarsson:
Its a first idea, it can surely evolve I'm happy for any input on this. I'm going to use it for plots in MXCuBE3.

marcus-oscarsson:
Could this be merged while its being evaluated ?

marcus-oscarsson:
Merging this as it's accepted and older than 15 days.

marcus-oscarsson:
Merging this as it is accepted and older than 21 days

PR #567, opened 2020-05-18
Decoding images for bayer_rg16 video mode

natxo14:
Adding image decoding for video mode 'bayer_rg16' with QtTangoLimaVideo

### DIFF
--- a/HardwareObjects/abstract/AbstractVideoDevice.py
+++ b/HardwareObjects/abstract/AbstractVideoDevice.py
@@ -136,7 +136,7 @@ class AbstractVideoDevice(Device):
 
         # Apply defaults if necessary
         if self.cam_encoding is None:
-            self.cam_encoding = self.default_cam_encoding
+            self.cam_encoding = AbstractVideoDevice.default_cam_encoding
 
         if self.poll_interval is None:
             self.poll_interval = self.default_poll_interval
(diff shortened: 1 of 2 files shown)


HardwareObjects/abstract/AbstractVideoDevice.py
```diff
@@ -251,6 +252,16 @@ def yuv_2_rgb(self, raw_buffer):
         image.resize(raw_dims[1], raw_dims[0], 2)
         return cv2.cvtColor(image, cv2.COLOR_YUV2RGB_UYVY)
 
+    def bayer_rg16_2_rgb(self, raw_buffer):
+        image = np.fromstring(raw_buffer, dtype=np.uint16)
+        raw_dims = self.get_raw_image_size()
+        image.resize(raw_dims[1], raw_dims[0])
+        out_buffer =  cv2.cvtColor(image, cv2.COLOR_BayerRG2BGR)
+        if out_buffer.ndim == 3 and out_buffer.itemsize > 1:
```
marcus-oscarsson, line 260: Well done !

HardwareObjects/abstract/AbstractVideoDevice.py
```diff
@@ -191,11 +191,12 @@ def get_new_image(self):
         raw_buffer, width, height = self.get_image()
 
         if raw_buffer is not None and raw_buffer.any():
-            if self.cam_type == "basler":
+            if self.cam_type == "basler" or self.cam_type == "bayer":
```
marcus-oscarsson, line 194: I know that this check was here from the start but would it not be better to check set encoder to None or something if we dont want to run this. so the check would be if 

```
if self.decoder:
    self.decoder(...)
else:
   bla bla
```

Otherwise great !

natxo14:
Corrections done:
Added self.decoder = None in init()
Used as condition in get_new_image()

marcus-oscarsson:
Wonderful, thanks !

PR #570, opened 2020-05-20
Inherit from AbstractActuator.

beteva:


rhfogh:
Would it be possible to inherit from AbstractFlux?
That gives the added function get_dose_rate, which I wsa planning to rely on for calculationg transmissions.

beteva:
No problem @rhfogh. I just did not know AbstractFlux was OK. Just out of curiosity, where are the dose numbers come from?

rhfogh:
@beteva 
Good point. There ought to have been a reference to give their pedigree.
As it happpens I took them from EMBLFlux, so I guess they are backed by the authority of Gleb Bourenkov, but at some point I need to find something more explicit.

rhfogh:
@IvarsKarpics 
Actually, Ivars, do you have a reference to back up those dose figures that used to be in EMBLFlux?

bourenkov:
The interpolation table dose_rate_per_photon_per_mmsq was created using "Absorbed dose calculations for macromolecular crystals: improvements to RADDOSE"
Paithankar, K.S., Owen, R.L and Garman, E.F J. Syn. Rad. (2009), 16, 152-162,
for some sensible crystal  composition - I no longer remember which one exactly. Can easily be re-done if necessary. 

To me the table is abstract enough to be here. 

But except for P14 under certain conditions I would not know an instrument where the AbstractFlux.get_dose_rate() would be valid. Likewise, the AbstractBeam.evaluate_beam_info() method.

Also - why energy is an argument but beam.get_beam_size() is mandatory?

beteva:
Thanks @bourenkov for the explanation.
As for the rest, this PR is only flux calculations for ID30B at ESRF. We need that somebody (probably @IvarsKarpics  or you) have a closer look at AbstractFlux.

rhfogh:
@bourenkov 
True, the dose rate function assumes that the flux density can be calculated as the flux value divided by the rectangular beam area. And also that this average flux density is what actully bathes the crystal, which might not be the case in general. Either of which is probably true only for a well-adjusted top-hat beam. But then in more complex cases you would probably need to consider both the actual flux profile and the size and shape of the crystal to get a precise result. Maybe there are better, but still practicable ways of doing this. But would it not be useful to have a formula like this at least as an approximation, instead of simply having no information available at all?

Yes, @IvarsKarpics, your input would be very ueful.

rhfogh:
I should add our use case:
We are trying to have our workflow set a recommended transmission, which requires calculating the dose rate. If there is no MXCuBE function for calculating dose rate we would simply put the same function as part of the GPhL workflow, on the basis that an approximate calculation can still give better transmission settings than leaving it to the user to set by hand. But it is likely that having the function as part of the main MXCuBE might lead to more reliable etimates, at least on some beamlines.

beteva:
Hi @all, I will open an issue on the AbstractFlux and will close this PR

PR #575, opened 2020-06-25
Add ESRFPhotonFlux

beteva:


HardwareObjects/ESRF/ESRFPhotonFlux.py
```diff
@@ -0,0 +1,87 @@
...
+        self._counter = None
+        self._flux_calc = None
+        self._aperture = None
+
+    def init(self):
+        """Initialisation"""
+        super(ESRFPhotonFlux, self).init()
+        controller = self.getObjectByRole("controller")
+        # ultimately it should be HWR.beamline.diffractometer.aperture
+        self._aperture = self.getObjectByRole("aperture")
```
IvarsKarpics, line 49: why not get aperture as HWR.beam.aperture?

beteva: Because it might not be present. I want to test what happens when HWR.beamline.diffractometer.aperture used and no aperture present before.

PR #576, opened 2020-07-10
Features added and bugs fixed while taking first steps towards MXCuBE at P11

jan-meyer:


Command/Tango.py
```diff
@@ -110,7 +113,11 @@ def __init__(self, event):
 class TangoChannel(ChannelObject):
     _tangoEventsQueue = Queue()
     _eventReceivers = {}
-    _tangoEventsProcessingTimer = gevent.get_hub().loop.async_()
+
```
marcus-oscarsson, line 116: Great !

HardwareObjects/TangoMotor.py
```diff
@@ -0,0 +1,181 @@
+#
```
marcus-oscarsson, line 1: Very nice :)

HardwareObjects/TangoMotor.py
```diff
@@ -0,0 +1,181 @@
...
+        if self.cmd_stop is None:
+             self.cmd_stop = self.add_command(
+                  {
+                      "type": "tango",
+                      "name": "stopAxis",
+                      "tangoname": self.tangoname,
+                      "polling": self.polling,
+                  }, "Stop",)
+
+    def connectNotify(self, signal):
```
marcus-oscarsson, line 104: Hm, watch out is this not camel case yet ?

marcus-oscarsson:
Thanks for all this !

marcus-oscarsson:
One of the tests are failing, have a closer look at the travis logs

_______________________ TestShutter.test_limits_setting ________________________
self = <HardwareRepository.test.pytest.test_shutter.TestShutter object at 0x7fa8e9f3f908>
test_object = <ShutterMockup.ShutterMockup object at 0x7fa8e9f4ea20>
    def test_limits_setting(self, test_object):
        """Test that set_limits and update_limits are diabled
        NB override ,ocally if you have an NState with limits"""
        limits = test_object.get_limits()
        with pytest.raises(NotImplementedError):
>           test_object.update_limits(limits)
E           Failed: DID NOT RAISE <class 'NotImplementedError'>
test/pytest/TestAbstractNStateBase.py:57: Failed

marcus-oscarsson:
The changes otherwise looks fine !

HardwareObjects/QtGraphicsLib.py
```diff
@@ -51,6 +51,7 @@
 
 from HardwareRepository.HardwareObjects import queue_model_objects
 from HardwareRepository.ConvertUtils import string_types
+from HardwareRepository.utils.mxcube_logging import log
```
IvarsKarpics, line 54: import is not used.

marcus-oscarsson:
Hi agian @jan-meyer and @vrey01, I guess you already noticed but the tests are failing, do you think you could fix the problem ?

vrey01:
Dear Marcus,

Sorry I did not noticed the error from travis. Still I do not think that it is something that we could have generated.  We did not change anything in ShutterMockup.

From Travis:



I have studied the issue. It seems to me that the error comes from a series of commits, in particular “d7f4c0” from rfogh in 2020-04-13 where ShutterMockup inherits now from both AbstractActuator and AbstractNState.    This is strange as AbstractNState already inherits from AbstractActuator.  That multiple inheritance makes ShutterMockup to cancel all method overwriting done by AbstractNState on AbstractActuator methods.  And in particular that of update_limits() that is producing the error during the tests. 

Shouldn’t ShutterMockup derive only from AbstractShutter??

I can modify ShutterMockup myself but I wonder if there was a reason for that in the first place. Can you comment on this Rasmus?

Best,


Bixente
______________________

Bixente Rey Bakaikoa
vrey01@gmail.com
+34 696.613.500

marcus-oscarsson:
@jan-meyer @vrey01: I merged your other PR which resulted in a conflict in Poller.py. I should maybe have merged them the other way around but the tests are still failing. The tests are not failing on master so there is something that goes wrong in this PR. Could you try to rebase this on master maybe that will solve both problems ?

vrey01:
Dear Marcus,

Independently of the pull request from DESY P11 I have looked at the issue with ShutterMockup raising NotImplementedError 

For that I have installed mxcube from sources and tried to run in mockup mode.   There is an error from that exact place.  I will try to explain:

   ShutterMockup inherits from both AbstractActuator and from AbstractShutter. 
   AbstractShutter inherits from AbstractNState that inherits from AbstractActuator
 
   NOTE — In my opinion AbstractActuator in the class definition for ShutterMockup should not be there

    I have tried removing it. But the error appears anyway. My analysis is as follows:

   TreeBrick asks Shutter hwo to “re_emit_values”
        re_emit_values
            Shutter re_emit_values is implemented in AbstractNState. It calls update_value() and then calls re_emit_values() of the superclass
            superclass for AbstractNState is AbstractActuator
            AbstractActuator.re_emit_values() calls update_values AND “update_limits”
            update_limits() raises “NotImplementedError”

In my opinion raising “NotImplementedError is wrong in that case. “NotImplementedError” should be used to FORCE developers to implement something in an abstract class.  In this case it is used only for checking a test condition saying something like:  “I want to make sure that somebody does not implement update_limits() for a NState class as this is a feature just for Actuators”. 

What we want in that case is to override a valid implementation of update_limits() in AbstractActuator and replace it with a “do nothing” logic.  I have changed the “raise..” with a “pass” and all works fine. 

Another solution would be to make AbstractActuator and AbstractNState inherits from still another new superclass where update_limits() is not implemented. Then AbstracNState would inherit from that class but not directly form AbstractActuator. But maybe the chain of inheritance would then be too long.

If you agree I will submit that as a different “pull request”.  It will require to change the test_shutter() in tests too, of course.

Bixente
______________________

Bixente Rey Bakaikoa
vrey01@gmail.com
+34 696.613.500

marcus-oscarsson:
Hm, I think that you are right that its unnecessary to inherit AbstractActuator twice. So this should be corrected. The problem that we are seeing seems to have to do with a problem with the test. 

The test if `update_limits` raises `NotImplementedError`, we can probably agree that `update_limits` is a strange method for a shutter but I think the tests needs to be changed to maybe a warning or similar as it seems like this behavior is currently used ?

I would not mind disabling the test or even better log a warning instead of failing the test ?

What do you think ?

vrey01:
Sure. I agree.

Still, regardless of the tests… there is a problem when “running” mxcube coming from that implementation as I described in my previous message.  Raising “NotImplementedError” is not appropriate and is causing problems.

I have gone further in my thinking and analysis.   If I understand the general logic behind.  The base idea is:
   
   * AbstractActuator adds a “set_value()” on top of a base hardware object. The implementation provides validation of the value (limits check)
   * AbstractNState implements a discrete number of values actuator (with “set_value”) but should not check limits

I have submitted a pull request adding a flag “check_limits" in any AbstractActuator. By default it is set to True. AbstractNState sets is to False
I have added logic in that direction in AbstractActuator

I have checked the question of value validation “validate_value" in all classes deriving from AbstractActuator :  AbstractEnergy, AbstractFlux, AbstractMotor and AbstractTransmission. They can all share the same code for “validate_value".  Having a “validate_value” at the AbstractActuator level is consistent 
with the implementation of logic for “limits” (speciality of AbstractActuator).
I have checked other possible classes deriving from AbstractActuator (BlissActuator, MicrodiffActuator) to check that there is no conflict with the proposed changes.

I have tested both with “pytest” and by running “mxcube” with mockup objects.  All runs well.

See if you think that the new “pull request” I just submitted is correct. Then I will see for the pending DESY-P11.  My guess is that the tests then will be fine.

Bixente
______________________

Bixente Rey Bakaikoa
vrey01@gmail.com
+34 696.613.500

HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -60,3 +60,6 @@ def open(self):
 
     def close(self):
         self.set_value(self.VALUES.CLOSED, timeout=None)
+
+    def update_limits(self, limits=None):
```
marcus-oscarsson, line 64: It looks like its this that makes the test fail, I think you can remove this and this will work with the fix we did a while back for the update_value

marcus-oscarsson:
Could you have a look at my last comment, and maybe that could make the tests pass then I'm more than happy to merge !

marcus-oscarsson:
Thanks for the fixes, it looks like the tests pass now, the one for 3.7 fails but that's for another reason. So ill merge this one

PR #582, opened 2020-08-13
HWR API - fixes and updates 

marcus-oscarsson:
Dear all,

As you all know we have been in the process of testing the work done during the last months and we 
have adapted the new API here at ESRF. Here are a few fixes and updates we've done in the process.

The last commit is all the changes made to ESRF specific Hardware Objects, I originally though that I
would submit two independent PR's one with the generic changes and one with the ESRF specific changes,
but I finally decided that it would be easier for us all with one PR.

I will follow up this PR with a lint PR, because I noticed that there are few files that are not linted correctly.

All the best,
Marcus


HardwareObjects/XMLRPCServer.py
```diff
@@ -525,19 +547,19 @@ def flux(self):
         return float(flux)
 
     def set_aperture(self, pos_name, timeout=20):
-        HWR.beamline.beam.aperture.move_to_position(pos_name)
+        HWR.beamline.beam.aperture_hwobj.moveToPosition(pos_name)
```
IvarsKarpics, line 550: Should it stay as snake case?

HardwareObjects/video_utils/websocket-relay.js
```diff
@@ -0,0 +1,89 @@
+// Use the websocket-relay to serve a raw MPEG-TS over WebSockets. You can use
```
IvarsKarpics, line 1: Is this part of mxcube?

HardwareObjects/BlissEnergy.py
```diff
@@ -106,7 +109,8 @@ def set_value(self, value, timeout=0):
             ValueError: Value not valid or attemp to set write only actuator.
```
rhfogh, line 109: I have problems here:
1)  the documentatono still says that you will raise ValueError if you try to write to a read-only actuator.
2) If I set a HardwareObject value to say 10, I expect that the value will henceforth be 10, or that I get an error or otherwise be noticed that something went amiss. I wold not expect that the value remains at 7 and my command is just silently ignored. Likely there is a reason for this change, but I wonder what it is

HardwareObjects/ESRF/ESRFMultiCollect.py
```diff
@@ -24,8 +24,366 @@
...
+#         if self._detector:
+#             self._detector.stop()
+#         else:
+#             self.get_command_object("reset_detector").abort()
+#             self.execute_command("reset_detector")
+
+
 class ESRFMultiCollect(AbstractMultiCollect, HardwareObject):
-    def __init__(self, name):
+    def __init__(self, name  ):
```
rhfogh, line 386: Typo?? (addtional spaces)

HardwareObjects/ESRF/ESRFMultiCollect.py
```diff
@@ -119,10 +478,9 @@ def data_collection_hook(self, data_collect_parameters):
     def data_collection_end_hook(self, data_collect_parameters):
         self._metadataClient.end(data_collect_parameters)
 
-    @task
     def prepare_oscillation(
-            self, start, osc_range, exptime, number_of_images, shutterless, npass, first_frame
-    ):
+        self, start, osc_range, exptime, number_of_images, shutterless, npass, first_frame
+    ):
```
rhfogh, line 483: Extra spaces

HardwareObjects/ESRF/ESRFMultiCollect.py
```diff
@@ -232,21 +591,28 @@ def move_motors(self, motor_position_dict):
             logging.getLogger("HWR").info("Waiting for end of motors motion")
             time.sleep(0.02)
 
-    @task
     def open_safety_shutter(self):
         try:
-            HWR.beamline.safety_shutter.set_value(HWR.beamline.safety_shutter.VALUES["OPEN"], timeout=10)
-        except RuntimeError:
+            HWR.beamline.safety_shutter.set_value(HWR.beamline.safety_shutter.VALUES.OPEN, timeout=10)        
+        except Exception:
```
rhfogh, line 597: Extra Spaces

Also, do ew use Exceptoin or BaseExceptoin normally?

marcus-oscarsson: @rhfogh I made a note about the extra spaces (linting) in the main text of the PR, I will make an additional PR with linting fixes if that's alright ?

HardwareObjects/XMLRPCServer.py
IvarsKarpics: Ah, I see. It looks like an old code with aperture_hwobj. First version was correct.

marcus-oscarsson: Oh, good catch. It should be neither. Its should be set_value with the new API. I guess this is code specifically is not really used and only tested with mockups which does indeed have moveToPosition.

HardwareObjects/ESRF/MD2MultiCollect.py
```diff
@@ -136,10 +131,14 @@ def prepare_acquisition(
         )
 
     def open_fast_shutter(self):
-        self.getObjectByRole("fastshut").actuatorIn()
+        self.getObjectByRole("fastshut").set_value(
```
rhfogh, line 134: We are not gettring these objects through HWR?

HardwareObjects/video_utils/websocket-relay.js
marcus-oscarsson: Hm, it is part of MXCuBE if one wants to stream video over web sockets. I can move it to the MXCuBE3 specific part but then one has to provide a "streamer" path or something which I thought was clumsy. I don't mind either, I tough this was simpler for the ones that wanted to use it.

HardwareObjects/BlissEnergy.py
marcus-oscarsson: Hm I somehow agree, The code that uses this however did not at all handle ValueError so I think this was the easiest fix. I don't think we saw any problem with it returning silently. We can go through the code again and handle the ValueError if its really a problem.

HardwareObjects/FlexHCD.py
```diff
@@ -627,14 +639,12 @@ def is_sequencer_ready(self):
...
-            state = self._execute_cmd_exporter("State", attrubute=True)
+            state = self.swstate_attr.get_value().upper()
         else:
             state = "RUNNING" if self._execute_cmd("robot.isBusy") else "STANDBY"
             if state == "STANDBY" and not self.is_sequencer_ready():
                 state = "RUNNING"
-
-        state = "READY"
+                
         state_converter = {
```
rhfogh, line 648: HardwareObjectStates are upper case, e.g. HardwareObjectState.READY
If it is not too much trouble it would be neat to use the same convention for all sets of states.

HardwareObjects/video_utils/websocket-relay.js
IvarsKarpics: Okay, maybe the HardwareRepository/utils is better place? I would suggest to keep HardwareObjects directory for hardware objects.

HardwareObjects/FlexHCD.py
marcus-oscarsson: It would but its not done for the AbstractSampleChanger yet and its not part of this PR

HardwareObjects/BlissEnergy.py
rhfogh: Arguably it is an error to try to set the energy on a fixed-energy beamline. Well, as long as this is only called from this context or by people who understand hte behaviour, there will be no problem. If some outsider (like me) writes code that  sets the energy it would be a lot easier for him to find his mistake if he got an error, rather than letting his code proceed on the assumption that the eneergy had been changed.

HardwareObjects/FlexHCD.py
rhfogh: Cool.

HardwareObjects/video_utils/websocket-relay.js
marcus-oscarsson: hm, sure. Would it be alright with you if I add the entire video_utils directory to utils as a sub-directory then ?

IvarsKarpics: Fine for me.

HardwareObjects/LimaPilatusDetector.py
```diff
@@ -110,21 +112,23 @@ def init(self):
...
             logging.getLogger("HWR").error(
                 "Could not connect to detector %s" % lima_device
             )
 
     def has_shutterless(self):
         return True
 
-    def wait_ready(self, timeout=10):
-        with gevent.Timeout(timeout, RuntimeError("Detector not ready")):
+    def wait_ready(self):
```
rhfogh, line 123: wait_ready is defiend in a supeclass (in master at least) where is has a second parameter timeout=None
It is not nice to change the seignture. How about
 def wait_ready(self, timeout=3500)

HardwareObjects/MiniDiff.py
```diff
@@ -160,9 +160,14 @@ def init(self):
 
         try:
             phiz_ref = self["centringReferencePosition"].getProperty("phiz")
-        except BaseException:
+        except:
```
rhfogh, line 163: I actuallly quite like just 'except'. BaseException does not add anything except protection from linters - but we ought to agree on a standard (no matter which one). What would be best?

HardwareObjects/SampleView.py
```diff
@@ -501,7 +510,7 @@ def __init__(self, mpos_list, screen_coord):
         self.set_id(Grid.SHAPE_COUNT)
 
     def update_position(self, transform):
-        phi_pos = self.shapes_hw_object.diffractometer.phiMotor.get_value() % 360
+        phi_pos = self.shapes_hw_object._diffractometer.phiMotor.get_value() % 360
```
rhfogh, line 513: Is there no better way of getting teh difractometer (HWR, api, ...)?

HardwareObjects/XMLRPCServer.py
```diff
@@ -469,15 +474,32 @@ def get_cp(self):
 
         return json_cplist
 
+    def _getattr_from_path(self, obj, attr, delim="/"):
```
rhfogh, line 477: This should work. But thePython  built-in 
operator.attrgetter does exactly the same job.

HardwareObjects/base_queue_entry.py
```diff
@@ -20,9 +20,9 @@
 import traceback
 from collections import namedtuple
 
-status_list = ["SUCCESS", "WARNING", "FAILED", "SKIPPED"]
+status_list = ["SUCCESS", "WARNING", "FAILED", "SKIPPED", "RUNNING", "NOT_EXECUTED"]
 QueueEntryStatusType = namedtuple("QueueEntryStatusType", status_list)
-QUEUE_ENTRY_STATUS = QueueEntryStatusType(0, 1, 2, 3)
+QUEUE_ENTRY_STATUS = QueueEntryStatusType(0, 1, 2, 3, 4, 5)
```
rhfogh, line 25: This is nice

HardwareObjects/queue_entry.py
```diff
@@ -116,17 +116,23 @@ def execute(self):
                     "sessionId": HWR.beamline.session.session_id,
                     "experimentType": "OSC",
                 }
-
+                
             sample_model = task_model.get_sample_node()
-            # task_model.get_parent()
-            if sample_model.lims_container_location > -1:
+            task_model.get_parent()
+            if sample_model.lims_container_location != -1:
```
rhfogh, line 122: What should happen if you get -2?

HardwareObjects/queue_model_objects.py
```diff
@@ -384,6 +385,7 @@ def init_from_lims_object(self, lims_sample):
             self.lims_id = lims_sample.sampleId
         else:
             self.lims_id = lims_sample.get("sampleId")
+        print("---------------->", self.lims_id)
```
rhfogh, line 388: Remove?

HardwareObjects/queue_model_objects.py
```diff
@@ -1978,14 +1981,15 @@ def get_path_template(self):
...
     def get_workflow_parameters(self):
-        result = HWR.beamline.gphl_workflow.get_available_workflows().get(
-            self.get_type()
-        )
-        if result is None:
-            raise RuntimeError(
-                "No parameters for unknown workflow %s" % repr(self.get_type())
-            )
-        return result
+        pass
```
rhfogh, line 1984: Why not leave it in? If ou remove the gphl workflow roles from the beamline setup this should never be called (and if it does anyway I should fix it).

HardwareObjects/XMLRPCServer.py
marcus-oscarsson: So I suggest that we keep this as it is, and then update it to set_value when we have some time to test it, if thats fine with you ?

HardwareObjects/BlissEnergy.py
marcus-oscarsson: Hm, I think its good of we can keep the ValueError in general for actuators and I will investigate where we need to change to also have it for BlissEnergy. I suggest that we keep it like this for the moment as we have tested with this. Ill make a comment in the code that we need to take another look at handling the ValueError, and Ill try to test that the coming weeks. How about that ?

HardwareObjects/ESRF/ESRFMultiCollect.py
marcus-oscarsson: Hm, well there is a slight difference between BaseException and Exception and I think we should use Exception if we really cant find any more specific class to use. Its in general of course better to use a more narrow exception clause, but I think everybody is aware about that. 

The problem with using `BaseException`, which we are indeed doing in quite alot of places, is that it also catches KeyboardInterrupt which might an issue.

In python 3 just `except` is as far a I know the same as `except BaseException` so I think it would be better in general to use `except Exception` as long as one really does not want to catch `SystemExit`, `KeyboardInterrupt` or `GeneratorExit.

HardwareObjects/ESRF/MD2MultiCollect.py
marcus-oscarsson: Yes, we are :), thanks

HardwareObjects/MiniDiff.py
marcus-oscarsson: Se my comment above

HardwareObjects/queue_entry.py
marcus-oscarsson: -1 means that there is no location everything else is in theory valid although -2 would be a strange position

HardwareObjects/queue_model_objects.py
marcus-oscarsson: Hm it did actually cause us issues so that was why I commented it out, I don't really remember why at the moment but I think it was called at init somehow. I don't remember if we removed it from the "beamline configuration" or not, I guess we did not. But it should maybe not raise an exception even if its left in ?

marcus-oscarsson:
Thanks @rhfogh and @IvarsKarpics for the reviews so far, I've most of the things that you've mentioned and left a few comments. I also linted the files changed in this PR but I will still make a lint pr after this one. We should maybe use pre-commit hooks to prevent this from happening.

HardwareObjects/BlissEnergy.py
rhfogh: Sounds fine.

HardwareObjects/ESRF/ESRFMultiCollect.py
rhfogh: Sounds right. I shall use Exception myself in the future (unless anyone disagrees).

HardwareObjects/queue_entry.py
rhfogh: OK.

HardwareObjects/queue_model_objects.py
rhfogh: In principle it would be better to make sure GPhL init is not called when the GPhL workflow is not installed, and I thought I had installed a mechanism for that. But it seem simpler to just comment it out for now, as you did. Once we have made a version for MXCuBE3, I shall likely push for putting this back in and fixing the problems in another way, but that is for another time

HardwareObjects/XMLRPCServer.py
IvarsKarpics: Fine with me.

IvarsKarpics:
PR is fine for me. I guess there will be few more api related fixes. @rhfogh  are you fine with the PR?

marcus-oscarsson:
Thanks, he said above that he approves so I guess its fine. We will soon follow up with a few more PR's :)

marcus-oscarsson:
As you both seem to approve and 15 days have passed Ill go a head and merge this !

PR #584, opened 2020-09-02
Vicente dev

vrey01:


HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -47,6 +47,7 @@ def __init__(self, name):
         self._nominal_limits = (None, None)
         self.actuator_name = None
         self.read_only = False
+        self.check_limits = True
```
beteva, line 50: I do not think this check should be in AbstractSActuator. Not all of the object that inherit from it have limits, even worse will be to force the check of the limits. If there is a need to force such a check, it should be in the particular HardwareObject or eventually in AbstractMotor, not in AbstractActuator.

marcus-oscarsson:
Great !, this also removes some duplicated code so its really nice ! Looks good to me !

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -92,17 +93,32 @@ def set_limits(self, limits):
         if self.read_only:
             raise ValueError("Attempt to set limits for read-only Actuator")
 
+        if not self.check_limits: 
+            if limits != (None,None):
+                raise ValueError("Attempt to set limits for Actuator with no limits checking")
+            return
+
```
beteva, line 100: See the previous comment.

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -92,17 +93,32 @@ def set_limits(self, limits):
...
+            return True
+
+        if value is None:
+            return True
+        if math.isnan(value) or math.isinf(value):
+            return False
+        limits = self._nominal_limits
+        if None in limits:
+            return True
+        return limits[0] <= value <= limits[1]
```
beteva, line 121: The same as above.

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -153,6 +169,11 @@ def update_limits(self, limits=None):
         Args:
             limits (tuple): two elements tuple (low limit, high limit).
         """
+        if not self.check_limits:
+            if limits != (None,None):
+                raise ValueError("Attempt to update limits for Actuator with no limits checking")
+            return
+
```
beteva, line 176: The same as above

IvarsKarpics (approved):

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -153,6 +169,11 @@ def update_limits(self, limits=None):
         Args:
             limits (tuple): two elements tuple (low limit, high limit).
         """
+        if not self.check_limits:
+            if limits != (None,None):
```
marcus-oscarsson, line 173: This check is perhaps unnecessary, I guess that any update_limits should raise ValueError if check_limits are false ? 
The flag check_limits should perhaps also be named validate_limits or use_limit_validation or simply use_limits ?

beteva (changes requested):
We had a long discussion at the time when the new structure has been put in place and agreed not to put any mechanism for checking the limits  in the AbstractActuator class. The value could be anything, not just numerical and in sucjh a case you cannot check it against limits, but maybe something else (e.g. a list of existing strings) .That's why also in the documentation it does not set any type.
I would request to not do any of the changes you propose

HardwareObjects/abstract/AbstractNState.py
```diff
@@ -64,24 +65,6 @@ def validate_value(self, value):
...
-
-    def update_limits(self, limits=None):
-        """Check if the limits have changed.
-        Args:
-            limits(tuple): two elements (low limit, high limit) tuple.
-        Raises:
-            NotImplementedError
-        """
-        raise NotImplementedError
-
```
beteva, line 84: This part of the code raises NotImplementedError for purpose, as we have decided that if somebody uses this in an NState class, he should implement as well what set/get limits is.

marcus-oscarsson:
For curiosity, where and how is the shutter used with limits, I guess that's why this problem surfaces ?

vrey01:
I understand your comments Antonia.  There are two points to that though:

Concerning the "validate_value" and checking limits:
-------------------------------------------------------
- the original issue is that there is behaviour implemented in AbstractActuator to manage limits.  In my opinion the limits "implementation" should come with checking them, otherwise... what's the point? All classes need now to implement the same code, leading to duplication. 
I have checked all existing classes (whether abstract or not) and I have seen duplicated code in all of them and I have not seen any exception.  
If a new class wants a different behaviour (with funny limit values that I cannot imagine of now), it is free to override the default validate_value() method.

For the logic of my proposal
-----------------------------
AbstractActuator  adds a "set_value" feature to a basic HardwareObject and implements (at least partly) the handling of limits
AbstractNState is an actuator that manages a discrete number of values. Needs "set_value" but does not need limits

We could define another superclass that will "only" implement the "set_value" feature and nothing about limits. Then AbstractNState and AbstractActuator would derive from that one.  Still that seems to me as cutting down the features in too many classes and subclasses.   And that is the reason for the proposal.  
If we assume that AbstractActuator implements (fully) the features "set_value" and "limits handling", then in the same way as there is a "read_only" feature in AbstractActuator (to disable the set_value), we can have "check_limits" to enable or disable the "limits handling".



For the original problem that motivated my pull request
---------------------------------------------------------
There is a problem because AbstractNState is raising an exception to say that the "update_limits() should not be implemented".   This is bad implementation. That exception is supposed to be raised to say (in an abstract class) that a certain method "must" be implemented and not the opposite. 

The problem shows up as an error with ShutterMockup that derives from AbstractShutter that on its own derives at the same time (specifically) from AbstractActuator and AbstractNState (still this is bad implementation).   When a re_emit_values() is called on the shutter... a cascade of calls to the re_emit_values() of the superclass ends up in calling re_emit_values of the AbstractActuator that calls "update_limits". And that generates the NotImplementedError exception.

vrey01:
For information: The error I describe appears from ShutterMockup when mxcube is installed right from the distribution in github.

marcus-oscarsson:
I guess this is in the Qt version of MXCuBE ?, could you point us to the line which gives you this problem ?

vrey01:
No. It is HardwareRepository shared by both Qt and mxcube3. 
If you use the ShutterMockup object (and I guess any Shutter deriving from AbstractShutter) you may get the problem. 

In Qt appears because the TreeBrick is calling re_emit_values() on the shutter hwo.

marcus-oscarsson:
Hm hm sure I know that its in HardwareRepository :), I just wanted to know from which line in MXCuBE it originated, but I think I understand from about where it comes now.

We dont have the problem in MXCuBE3, its true that I have not tested with ShutterMockup for some time though. But the mockup (simulation) version works with ShutterMockup though so I found it a bit strange that it did not work in MXCuBE2.

Antonia outlined the history of this decision quite well and I think we need a moment to investigate what other consequences this might have, but your reasoning seems sound to me. I also agree that its nice if we can avoid adding another level of inheritance.

vrey01:
Ok.  My guess is that you will get the problem too if you call re_emit_values() in, for example, BlissShutter, BlissNState or in any AbstractNState.

beteva:
@vrey01 I agree that an extra class will be only a complication. We do not want inheriting from too many classes - at the end what we want is a clear code and not a plate of spaghetti. Your logic is also fine with me and I would agree with the changes once we've checked that it does not brake the code we have and have tested for the last 2 months. It is just that the question (an especially the limits) were debated for so long that I have the impression of history repeating. The current code is kind of mutual agreement.
As for what you call "funny limit values that I cannot imagine of now", an example is checking if the value asked is one of the Enum members of an NState hardware object. Now that I think, we could eventually implement this in the AbstractNState class.
I also do agree that ShutterMockup should only inherit from AbstractShutter.

vrey01:
I understand.  Also I can understand all the questioning about this implementation.  I meant to send this "pull request" as a proposal for improvement. I am really sorry that I could not be in the original discussions and then I come with a different proposal.   I do think though that the code would be simpler and better.  If you find any issue during the tests I will be glad to adapt the code.  Same thing about writing code for validate_value in AbstractNState that is a very good idea indeed.

marcus-oscarsson (approved):

marcus-oscarsson:
Looks good to me !

rhfogh:
Probably fine, but could you hold off on merging it till I get the functioning mock implementation and have time for a quick look? I am just home from oliday, so no chance before this.

rhfogh (changes requested):
I do not think this is the best way to go. I cannot give a precise proposal till I have a functioning mock installation, but since we have the meeting tomorrow here are some provisional comments:

- AbstractNState.re_emit_values has a bug - it should not call AbstractActuator.re_emit_values (as it effectively does) but HardwareObject.re_emit_values. I suspect that should fix the original problem.

- validate_value should indeed be implemented in AbstractActuator, as Vicente propposes. As long as validate_value is overridden in AbstractNState, the rules for multiple inheritance mean that the AbstractActuator function will never be called for an AbstractNState subclass.

- Whatever mechanism we use, I think it is important that calling set_limits or update_limits on an AbstractNState subclass *should* raise an error (unless the functions have been overridden below AbstractNState). If you try to set limits on something that does not have limits it should not pass silently. 

- I think we can manage without the check_limits attribute. For AbstractNState classes we can get the right effect simply by overiding the relevant functions in AbstractNState (or one of its subclasses). For other AbstractActuators I think we should always have check_limits as True, so there is no need to use it.

vrey01:
Dear mxcubers.  I was traveling earlier this week and I could not participate at the latest mxcube dev meeting where you were going to discuss about this "pull request".  I was wondering if decisions on this were taken at the meeting. I will be glad to follow work on this to get this "pull request" solved.

rhfogh:
@vrey01 There was discussion about the best way of doing this. The agrement was that for now we would make a minimum bug fix so that PR #576 could run without errors. We would then raise an issue and discuss how ot deal with the refactoring etc. arising from this PR. I'll send you the draft minutes and bugfix proposal by mail.

marcus-oscarsson:
Thanks both, I've submitted a fix in my last PR (#586 ) and opened issue #587 for further discussion

marcus-oscarsson:
So I will close this PR while we are disusing issue #587

PR #585, opened 2020-09-10
Changes in get_state to prioritise the certain states.

beteva:
DISABLED has to be treated in priority.

### DIFF
--- a/HardwareObjects/BlissMotor.py
+++ b/HardwareObjects/BlissMotor.py
@@ -37,15 +37,15 @@ __license__ = "LGPLv3+"
 
 @enum.unique
 class BlissMotorStates(enum.Enum):
-    """"
-    MOVING  : 'Axis is moving'
-    READY   : 'Axis is ready to be moved (not moving ?)'
-    FAULT   : 'Error from controller'
-    LIMPOS  : 'Hardware high limit active'
-    LIMNEG  : 'Hardware low limit active'
-    HOME    : 'Home signal active'
-    OFF     : 'Axis power is off'
-    DISABLED: 'Axis cannot move (must be enabled - not ready ?)'
+    """
+    MOVING = "Axis is moving"
+    READY = "Axis is ready to be moved (not moving ?)"
+    FAULT = "Error from controller"
+    LIMPOS = "Hardware high limit active"
+    LIMNEG = "Hardware low limit active"
+    HOME = "Home signal active"
+    OFF = "Axis power is off"
+    DISABLED = "Axis cannot move"
     """
 
     MOVING = 0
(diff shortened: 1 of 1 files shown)


HardwareObjects/BlissMotor.py
```diff
@@ -121,11 +133,8 @@ def get_specific_state(self):
         """
         state = self.motor_obj.state.current_states_names
         state_list = []
-        state_list.append(self._state2enum(state[0])[1])
-        if len(state) > 1:
-            for _state in state[1:]:
-                state_list.append(self._state2enum(_state)[1])
-
+        for _state in state:
```
rhfogh, line 136: This works. But as a matter of style 1) singular name 'state' for plural object; 2) unnecessary variable. If you feel like it you could do

state_list = = list(self._state2enum(stat)[1] for stat in  self.motor_obj.state.current_states_names)

rhfogh (approved):

IvarsKarpics:
@beteva will you change the code as @rhfogh  is suggesting or should I merge the PR?

HardwareObjects/BlissMotor.py
natxo14: Thanks for your comments Rasmus.

It's true that we can save one variable, but personally I find that the code ( though more pythonic ) is less readable for new code readers.
But anyway, code is working fine in both ways, so review is ok for me.

rhfogh: Fair enough, style is in the eye of the beholder..
I feel a little more strongly about a variable called 'state' that is actually a *list* of states (why not call it 'states')?
But cool either way.

beteva: Indeed it is a list, but most of the time it has only one element. My English is probably not good enough, I meant the name more like a title, not really taking into account if there is one or several states.

rhfogh: It is not an English thing, but a style thing - and as such an arbitrary convention. Yours is as good as mine, and we never agred to a variable name style.
But from my former project I am used to the convention that you use the name to distinguish between collections and individual values, and I think it is a useful guard against mistakes that you can see from the variable what kind of thing it contains. Particularly since we generally do not have reliable type documentation even for parameters, and certainly not for variables. For coding purposes a list of one element is vastly different from just having the same element alone.

I think I shall drop this point, and raise variable naming as an issue.

beteva: @rfiogh, I do not understand how you got to this code. This is an old version, which should not be there. I will have a look and remove obsolete parts.

natxo14 (approved):

PR #586, opened 2020-09-14
Some improvements and bug fixes

marcus-oscarsson:
Dear all,

A few fixes we have made the last weeks, since the last PR, it mainly contains a few fixes to the 
ESRF specific parts but also some general improvements. There is also the re_emit_values fix
for AbstractNState that was discussed during last developers meeting and which caused @vrey01 
and @jan-meyer some issues.

All the best,
Marcus


IvarsKarpics (approved):

HardwareObjects/ESRF/BeamCmds.py
```diff
@@ -117,15 +117,15 @@ def __call__(self, *args, **kwargs):
             value = self._hwobj.VALUES.OUT
         else:
             value = self._hwobj.VALUES.IN
-        cmd = getattr(self._hwobj, "set_value")(value)
+        cmd = getattr(self._hwobj, "set_value")(value, timeout=None)
```
rhfogh, line 120: Are you sure this works?
It looks (to me) like it should evaluate to

cmd = self._hwobj.set_value(value, timeout=None) # which should return None

So the next line would evaluat to 

self._cmd_execution = gevent.spawn(None)

Or have I missed something?

HardwareObjects/ESRF/BeamCmds.py
```diff
@@ -117,15 +117,15 @@ def __call__(self, *args, **kwargs):
...
         self._cmd_execution = gevent.spawn(cmd)
         self._cmd_execution.link(self._cmd_done)
 
     def _cmd_done(self, cmd_execution):
         try:
             try:
                 cmd_execution.get()
-                res = getattr(self._hwobj, "get_value")().name.lower()
+                res = getattr(self._hwobj, "get_value")(timeout=None).name.lower()
```
rhfogh, line 128: If self._hwobj is a hwobj this ought to be the same as

self._hwobj.get_value(timepout=None).name.lower()

So what is actually happening here?

HardwareObjects/ESRF/BeamCmds.py
```diff
@@ -144,5 +144,8 @@ def value(self):
         value = "UNKNOWN"
 
         if hasattr(self._hwobj, "get_value"):
-            value = getattr(self._hwobj, "get_value")().name
-        return value
+            value = getattr(self._hwobj, "get_value")()
```
rhfogh, line 147: Like previous comments

HardwareObjects/ESRF/Transmission.py
```diff
@@ -50,6 +50,10 @@ def _set_value(self, value):
         """
         self._transmission.set(value)
 
+        # Busy is set by Abstractactuator, simply set state to ready
```
rhfogh, line 53: AbstractActuator (A not a)

HardwareObjects/FlexHCDMaintenance.py
```diff
@@ -145,7 +154,7 @@ def send_command(self, cmdname, args=None):
         if cmdname in ["reset_sample_number"]:
             self._do_reset_sample_number()
         if cmdname == "change_gripper":
-            self._do_change_gripper()
+            self._do_change_gripper(int(args))
```
rhfogh, line 157: args (plural!) would normally be assumed to be a list. Is it a string? And how do we know it is convertible to int?

HardwareObjects/XMLRPCServer.py
```diff
@@ -428,7 +336,7 @@ def is_queue_executing(self, node_id=None):
     def queue_status(self):
         pass
 
-    def shape_history_get_grid(self):
+    def shape_history_get_grid(self, sid):
```
rhfogh, line 339: Documentaiont for 'sid' parameter?

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -126,7 +126,6 @@ def set_value(self, value, timeout=0):
         if self.read_only:
             raise ValueError("Attempt to set value for read-only Actuator")
         if self.validate_value(value):
-            self.update_state(self.STATES.BUSY)
```
rhfogh, line 129: This needs consderation. Which function and class is responsible for setting state to BUSY, and where is it documented - if we canot assume that the abstract classes are correct? I think this change will (partially) break ActuatorMockup - in that class _set_value sets state to READY at the end, but nothing ever sets state to BUSY.

No objection to having the state setting done in _set_value if you think that is better for some reason (maybe that is what we did or should have done all along) but I think we need to have it clear what must happen where.

rhfogh (changes requested):

HardwareObjects/abstract/AbstractActuator.py
marcus-oscarsson: Hm, I believe was me who added this line in my last PR due to a delay in certain events for diffractometer motors. The event were delayed and the movement short so that the motor finished moving before the busy event arrived. We then realized that it was better to not have this here so we remove this line.

HardwareObjects/XMLRPCServer.py
marcus-oscarsson: Its shape id, sid for short, sorry I did not add a documentation this time

HardwareObjects/FlexHCDMaintenance.py
marcus-oscarsson: We know that its convertible to int because gripper in the context of the FlexHCD is passed as a string and in this case args is a single string.The code was written like this and it was simply addition to be able to handle several different grippers. I'm sure it can be improved.

HardwareObjects/ESRF/BeamCmds.py
marcus-oscarsson: Hm, I think this is a git mishap Ill take a closer look too see what happened :)

HardwareObjects/abstract/AbstractActuator.py
rhfogh: OK. 

Can we sum it up by saying that for real motors the state is set by the hardware? While for mokcups the mockup implementation must set BUSY?

HardwareObjects/FlexHCDMaintenance.py
rhfogh: It is basically a naming issue. If you haven avariable called 'args' you take it for granted that it is a list - and you likely code on that assumption. Better to have name and use consistent.

HardwareObjects/ESRF/BeamCmds.py
marcus-oscarsson: So, we managed to fix this little problem when merging our different pieces of code. The BeamCmds object will likely be reworked a bit more soon but this is whats used today

HardwareObjects/abstract/AbstractActuator.py
marcus-oscarsson: Sure any "soft" motor would need to call `self.update_state(self.STATES.BUSY)` in someway and I guess it in most cases dont hurt to have it in implementation on top of real hardware. That hardware would/should in anyways handle the READY state

I'd be happy to put it back in again, but I wanted to remove it as it was me who added it and it was to resolve a issue we had a with a specific motor that I though was general enough to be in AbstractActuator. I than had another think about thought it was better to remove this logic from AbstractActuator and push it down to in this case ExporterMotor and perhaps move it up to AbstractMotor if it turns out to be a general problem.

But its otherwise fine as it is now

HardwareObjects/FlexHCDMaintenance.py
marcus-oscarsson: Sure I agree with your argument on the naming, we could maybe leave as it is right now and ill have a look at this part of the code at a later time. If thats not too painful for your eyes ...

marcus-oscarsson:
Okey we fixed our little git mishap and also added a few more fixes.

@rhfogh I'am sorry for the variable naming, we simply did what we needed to add a function to some existing code so I think it would be good if could that one go for the moment until we have another look at that code in general.

HardwareObjects/ESRF/BeamCmds.py
rhfogh: Cool.

For my own education, could you tell me what kind of object self._hwobj is? 'Cause if it is a HardwareObject I still think that cmd evaluates to None. The code could work even so, with the setting done in the 'cmd =' line and the spawned process being a no-op. If this is good enough for now that is fine with me, but I just want to make sure there is not something there that I have misunderstood

HardwareObjects/abstract/AbstractActuator.py
rhfogh: @marcus-oscarsson 
This makes sense. I think you are right in removing it and not putting it back.  I just was not sure which layer was responsible for setting BUSY an assumed that it was this function (since the code was there).

We just need to remember to go through the mock motors and make sure they set BUSY as they should.

HardwareObjects/FlexHCDMaintenance.py
rhfogh: Cool. I care about the principle of it, but not that much about a small individual instance.

HardwareObjects/ESRF/BeamCmds.py
marcus-oscarsson: Well this code is now valid any longer but there is something similar in another place and in that case its actually a function

rhfogh (approved):
All issues sorted now.

marcus-oscarsson:
Okey so Ill merge then :)

PR #588, opened 2020-09-14
Rework of re_emit_values

IvarsKarpics:
PR is linked to the https://github.com/mxcube/mxcube/pull/402

Some fixes to run qt version of mxcube
* added re_emit_values method to the beamline to be called after loading the gui.
* use re_emit_values method just to re emit actual values. In the case of mxcube qt we would like to re emit values after the gui is loaded. So there should not be a logic in the re_emit_values method that checks if the value is changed. Gui needs these values.
* Fixed AbstractCollect


HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -133,17 +133,15 @@ def validate_value(self, value):
             return True
         return limits[0] <= value <= limits[1]
 
-    def update_value(self, value=None):
+    def get_undulator_gaps(self):
+        return ()
+
+    def re_emit_values(self):
```
rhfogh, line 139: This looks correct and matches the pre-change code. Still, are we sure we do not need an update function for the wavelength, and that hte valueChanged signal emitte from AbstractAcrtuator will do everything we need?

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -467,34 +471,39 @@ def _update_data_collection_in_lims(self, cp):
...
             i = 1
             for jj in self.bl_config.undulators:
                 key = jj.type
                 if key in und:
                     cp["undulatorGap%d" % (i)] = und[key]
                     i += 1
 
-            cp.dangerously_set("resolutionAtCorner", self.get_resolution_at_corner())
+            #TODO check if this is used
+            cp.dangerously_set("resolutionAtCorner", HWR.beamline.resolution.get_value())
```
rhfogh, line 494: This does not look right. resolution.get_va;lues() should surely return resolutoin at edge, not ncorner?

rhfogh (approved):
Looks good

HardwareObjects/abstract/AbstractCollect.py
IvarsKarpics: Should I add method get_resolution_at_corner to the AbstractResolution?

HardwareObjects/abstract/AbstractEnergy.py
IvarsKarpics: Something like this?: 


```
AbstractActuator.re_emit_values(self)
self.emit("energyChanged", (self.get_value(), self._wavelength_value))
```

rhfogh: Looks good

HardwareObjects/abstract/AbstractCollect.py
rhfogh: @IvarsKarpics 
I do not think we need 'resolution_at_corner' at all. AbstractResolution already has a get_value_at_corner function - which calculates the value on the fly - and presumably that should be enough?

HardwareObjects/ALBA/ALBABeamInfo.py
```diff
@@ -149,19 +149,19 @@ def get_beam_shape(self):
 
     def beam_width_changed(self, value):
         self.beam_info_dict["size_x"] = value
-        self.emit_beam_info_changed()
+        self.re_emit_valuesd()
```
marcus-oscarsson, line 152: Should there really be a `d` at the end ? `re_emit_valuesd`

HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -163,7 +164,10 @@ def update_limits(self, limits=None):
...
-        self.update_value(self.get_value())
-        self.update_limits(self.get_limits())
-        super(AbstractActuator, self).re_emit_values()
+        """re emits values for all internal attributes"""
+        self.emit("valueChanged", (self.get_value(),))
+        self.emit("limitsChanged", (self.get_limits(),))
+        self.emit("stateChanged", (self.get_state(),))
+        #self.update_value(self.get_value())
+        #self.update_limits(self.get_limits())
+        #super(AbstractActuator, self).re_emit_values()
```
beteva, line 173: Could you, please, explain why do you choose the signals rather that the upadte_* methods. Especially why do you force the emitting of the signals, while in the update_* methods we have checks if the signals have to be emitted.

HardwareObjects/abstract/AbstractEnergy.py
beteva: In the AbstractEnergy the energyChanges was kept only meanwhile. If we follow our rules, there should be wavelengthChanged and valueChanged (for the energy). I still do not understand why not using upfdate_value, though

HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -133,17 +133,15 @@ def validate_value(self, value):
...
-            value = self.get_value()
-        self._nominal_value = value
-
         if not self._wavelength_value:
-            self._wavelength_value = self._calculate_wavelength(value)
-        self.emit("energyChanged", (value, self._wavelength_value))
-        self.emit("valueChanged", (value,))
+            self._wavelength_value = self._calculate_wavelength(self.get_value())
+        self.emit("energyChanged", (self.get_value(), self._wavelength_value))
+        self.emit("valueChanged", (self.get_value(),))
```
beteva, line 147: I would not force the reading, if we do have a value.

HardwareObjects/abstract/AbstractFlux.py
```diff
@@ -64,10 +64,6 @@ def init(self):
         super(AbstractFlux, self).init()
         self.read_only = self.getProperty("read_only") or True
 
-    def _set_value(self, value):
-        """Local setter function - not implemented for read_only clases"""
-        raise NotImplementedError
-
```
beteva, line 70: Why this did has been removed?

beteva (commented):
Why is the Exporter part of this PR?

HardwareObjects/abstract/AbstractEnergy.py
IvarsKarpics: Ok, I will change signals to:
valueChange, self.get_value() and wavelenghtChanged, self.get_wavelength()

HardwareObjects/abstract/AbstractCollect.py
IvarsKarpics: Ok

HardwareObjects/ALBA/ALBABeamInfo.py
IvarsKarpics: ups, typo

HardwareObjects/abstract/AbstractActuator.py
IvarsKarpics: I did try to explain this in our last dev meeting. The issue is with loading qt gui (I do not know how it is done in web gui). At the beginning the beamline object with all hardware objects is initialized. During the initialization channels are connected and signals emitted. After the initialization gui is assembled and loaded and bricks do not get actual values from hardware objects because connect is after the values have been emitted.
I create a method in the beamline object to call re_emit_values of all hardware objects and I call this method after the gui is loaded. So in this case update_value and update_limits within re_emit_values will not work because values have not been changed. I would like to avoid calling from bricks something like `self.value_changed(HWR.beamline.some_hwobj.get_value())`.
Maybe to avoid creating two method we could add argument force to re_emit_values?
Something like:
```
def re_emit_values(self, force):
    self.update_value(self.get_value(), force)
```

IvarsKarpics:
Hi @beteva : I added accidentally exporter file in the first commit. Now they are removed

HardwareObjects/abstract/AbstractFlux.py
IvarsKarpics: reverted

HardwareObjects/Beamline.py
```diff
@@ -694,3 +694,20 @@ def get_default_path_template(self):
             pass
 
         return path_template
+
+    def get_default_characterisation_parameters(self):
+        return self.characterisation.get_default_characterisation_parameters()
+
+    def re_emit_values(self):
```
marcus-oscarsson, line 701: Hm, do we agree that this should really only be called in certain very specific cases for instance right after the HWR is initialized or when a UI "reconnects"

marcus-oscarsson: and that this is never done by anything inside of HWR unless explicitly asked for somehow ?

HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -5,21 +5,20 @@
 #  This file is part of MXCuBE software.
 #
 #  MXCuBE is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU General Public License as published by
+#  it under the terms of the GNU Lesser General Public License as published by
```
marcus-oscarsson, line 8: This file also seems to contain some things that are not directly related with re_emit_values, I'am fine with these but is it intentional ?

marcus-oscarsson:
Thanks, I'm looks food in general. I just want to understand the intention with re_emit_values on the Beamline object. So this is only called from the UI ?

HardwareObjects/abstract/AbstractActuator.py
beteva: This could be an idea. Or we make  explicit ui_(re)emit_all for the UI, which forces the reading.

IvarsKarpics:
Hi @marcus-oscarsson  and @beteva 

Yes, this method currently is called just from qt gui after HWR init. Should we discuss this PR during the code camp?

marcus-oscarsson:
Hm, I'm fine with the function/functionality if you think that its needed, as long as its called explicitly from the UI. We can definitely discuss it as well. Just let me know what you prefer.

marcus-oscarsson:
Oh by the way, could you try to trigger another build, all the tests passes here but it seems like there is/was some issue with Travis the last time we merged.

marcus-oscarsson:
@IvarsKarpics and if its possible maybe you could add me as admin to the Travis project, I think its on your travis profile ?

rhfogh:
@marcus-oscarsson @IvarsKarpics @beteva 
Looking at the code, this function is called from the HO layer in some places.
Maybe ew should rename it to ui_re_emit_values, as suggested above, and go through and replace these calls with plaiin update_values cals whenever possible

IvarsKarpics:
rename re_emit_values to force_signal_emit

IvarsKarpics:
Dear all,

I kept the method re_emit_values with its functionality (emit signals if value, state or limits has been changed) and added a new method force_emit_signals that will emit all signals and will be called from gui.
Are you @marcus-oscarsson , @rhfogh , @beteva  and others ok with this? Can we merge the PR?

rhfogh:
What is the use case for re_emit_values, and the dfference with force_emit_values?

beteva:
As I understood, the force_emit_signals is reading explicitly the value, state and limits and sends the three attached to that signals, while the re_emit_values updates all three in one go,  but only emits the relevant signals if change. I guess the use case would be to do all in one, instead of using three update_*, not sure this is needed, though.

IvarsKarpics:
Exactly  @beteva . Thanks for the explanation!

rhfogh:
When would you actually need re_emit_signals (except maybe foir beam_info)?

IvarsKarpics:
Thanks @rhfogh

HardwareObjects/ALBA/ALBABeamInfo.py
```diff
@@ -149,19 +149,19 @@ def get_beam_shape(self):
...
 
     def beam_height_changed(self, value):
         self.beam_info_dict["size_y"] = value
-        self.emit_beam_info_changed()
+        self.re_emit_valuesd()
```
marcus-oscarsson: I think there still a typo here ? an extra "d" at the end

PR #590, opened 2020-09-18
Better handling of the Pilatus energy threshold.

beteva:
Fix a bug in the set_energy_threshold.


rhfogh (approved):
Hardly an area I am on top of, but FWIW it looks fine.

beteva:
Replaced by #591

PR #591, opened 2020-09-21
Better handling of the Pilatus energy threshold.

beteva:
Fix a bug in the set_energy_threshold.


marcus-oscarsson:
Merging this as previous PR was approved and this is only a resolution of a merge conflict

PR #592, opened 2020-09-30
Esrf hutchtrigger

beteva:
Triggering actions to execute when enter/exit hutch. ESRF specific.

### DIFF
--- a/HardwareObjects/ESRF/BeamCmds.py
+++ b/HardwareObjects/ESRF/BeamCmds.py
@@ -73,10 +73,6 @@ class ControllerCommand(BaseBeamlineAction):
         self.type = PROCEDURE_COMMAND_T
         self.argument_type = ARGUMENT_TYPE_LIST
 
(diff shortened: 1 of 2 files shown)


IvarsKarpics (approved):

rhfogh (approved):
Looks fine to me - though I cannot really judge it properly

marcus-oscarsson:
It looks like we have a test for python 2.7 that fails and this seems to be due to a previous PR (#590)  @beteva maybe you can have a look, its test/pytest/test_beam.py that fails.

marcus-oscarsson:
Ill merge this PR since its a previous PR that makes the test fail

beteva:
Sorry, no idea why the test are making core dump for python 3.7 Did anybody change the environment?

marcus-oscarsson:
Hm, I dont think so. I guess we just have to run it on one of our machines. We can try that next week if you like

PR #594, opened 2020-10-13
Converted hasObject uses to has_object

rhfogh:
Changed uses of hasObject to has_object


IvarsKarpics (approved):

rhfogh:
I notice
1) lots of camelCase functions in (just as as start)SpecScan and MDClient. I did not embark on changing htem. Presumably there is a reason?

2) isConnected still camelCase in a lot of places. Do some ot the underlying libraries use camelCase?

3) Python-3-only constructs in some cases (f"" f-strings, and super()). As long as my own code does not break are these still worth fixing?

rhfogh:
Doing some global searches, I can find quite a few camelCase functions that start with 'is', 'get', or 'set'. Would it be sensible for me to try to change some of them? Or is all that can reasonably/safely be done already done?

beteva (approved):

beteva:
>     1. lots of camelCase functions in (just as as start)SpecScan and MDClient. I did not embark on changing tem. Presumably there is a reason?
I do not think MDClient is used anywhere.
>     2. Python-3-only constructs in some cases (f"" f-strings, and super()). As long as my own code does not break are these still worth fixing?
I guess it might be me in some of the ESRF HO. I would say, leave it as it is, if it is in a specific synchrotron HO.

Issue #587, opened 2020-09-14
Limit handling for AbstractNState


marcus-oscarsson:
This topic was already discussed in detail during the code camp in [April](https://mxcube.github.io/mxcube/doc/developers_meetings/MinutesGrenoble20200402Virtual.pdf) and the current solution was put in place.

The principle question discussed where whether "n-state" objects have limits and how to handle them, resulting in the decision that they might in some rare cases have limits but there is no default implementation for handling those provided by AbstractNState. 

A bug in re_emit_values of AbstractNState incorrectly calling re_emit_values of AbstractActatutor (that does have limit handling) raised an exception.

The bug is fixed by #586 but this issue is created to discuss any remaining questions and perhaps draft another solution if necessary.

rhfogh:
In line with Vicente's original PR, I think that we should implement validate_value in AbstractActuator, using (I'd  say) a copy of AbstractMotor.validate_value. The alternative implementation in AbstractActuator will overide the implementaiton in teh superclass, and should give the fight behaviour everywhere. If we want diferent behavious in special cases we can always override.

The reasons are those that VIcente mentioned: avoiding duplication and having the validation implkementatoin close to where limits are introduced

PR #596, opened 2020-10-14
validate_value in AbstractActuator

beteva:
Solves issue #587. Implement validate_value in AbstractActuator to avoid duplication in Abstractmotor, AbstractEnergy and AbstractTrensmission.


HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -100,11 +100,17 @@ def set_limits(self, limits):
     def validate_value(self, value):
         """Check if the value is within limits.
         Args:
-            value: value
+            value(numerical): value
         Returns:
             (bool): True if within the limits
         """
-        return True
+        if value is None:
```
IvarsKarpics, line 107: Maybe I missed something. But why if value is None it is validated as True. I thought value should always be not None and None means it is not initialized.

beteva: Honestly I do not remember. Has been added by @rhfogh during the April Code Camp.

rhfogh: They are different semantics, mandatory versus de facto optional. The question is if it is a problem during or after initialisation that you initialise to an invalid value. It would be nicer to have None as invalid, I agre, so we could try to make the change and see if that gets us into problems.

rhfogh (approved):

HardwareObjects/abstract/AbstractActuator.py
marcus-oscarsson: Hm, sounds a bit convoluted, so what you say is that the value is "valid" if the value is None and the HO is not yet initialized ?

rhfogh: Depends what you want it to be, I'd say. It would be best to say that None was invalid, in most cases. The only problem is if that gets us errors or funny messges during initialisation.

beteva: > The only problem is if that gets us errors or funny message during initialisation.

I would guess that was the problem, as the two lines were added much later that the initial implementation of the validate_value. I'd rather prefer that the value is by default valid, even if the HO is not yet initialised.

marcus-oscarsson:
Looks good to me :+1: , there is a small conflict to look at, I think its better that you do that @beteva :)

beteva:
Hi @all, can we conclude on this - should the validate _value return True by default or invalidate None value? My vote is for True by default.

IvarsKarpics:
But how we would define an uninitialized value? GUI is based on the principle that a value None is not valid and will not be displayed (string formatting will not work on None).

beteva:
@IvarsKarpics, probably value will only be None during the initialisation, so nothing to be displayed by the GUI. The question is, if the value is None, should we return True or raise a InvalidValue exception.

IvarsKarpics:
In this case I assume we have to return True, because we consider value None as a "predefined" value.

marcus-oscarsson:
@beteva I guess you noticed already, but there is a conflict again ?

HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -117,34 +120,17 @@ def _calculate_energy(self, wavelength=None):
...
             value: Not used, but kept in the method signature.
         """
-        AbstractActuator.force_emit_signals(self)
-        self.emit("wavelengthChanged", self.get_wavelength())
+
+        if value is None:
+            value = self.get_value()
+        self._nominal_value = value
+
+        if not self._wavelength_value:
```
rhfogh, line 133: That does not look right. 
1) I do not think you should store the wavelength|_value at all. Admittedly it is in the current code, but it gives addtional problems with syncronisation and updatting. Much better, surely, to just calculate it as needed. 
2) If you do store it, you should recalculate it *every* time the energy changes (that is one reason it is better not to store it); this code looks like it only recalculates wavelength it was not set before., which should lead to inconsistencies

beteva: Storing the  wavelength value has been introduced, as we can ask either to change the energy, or the wavelength. 
I'll do the re-calculation of the wavelength in update_value.

rhfogh: I still do not understand why this is necessary. The wavelength is a direct function fo the energy; if you need the current wavelength for comparison or any other reason you can always, simply and quickly, calculate it from the energy. By not storing the eavelength we are guaranteeing, without any extra work, that the two values are always in sync. If we store it we introduce a potential weakness that we then need to work to overcome. What is the advantage, again?

HardwareObjects/abstract/AbstractActuator.py
beteva: @IvarsKarpics, is this code OK for you and if yes, could you, please merge this pull request. Thank you

IvarsKarpics:
Hi @beteva  and @marcus-oscarsson 

I am fine with the PR. How about @rhfogh question regarding the wavelength? I do not remember if there was an agreement.

HardwareObjects/abstract/AbstractEnergy.py
beteva: I will do as you suggest @rhfogh, but we have to be sure, that there is always an energy object with get_value. That excludes a wavelength object only (if there is a case anywhere). This is not a problem for us, though.

beteva:
Hi @all. This PR is ready to be merged.

IvarsKarpics (approved):

rhfogh (approved):

PR #597, opened 2020-10-15
Update guidelines

rhfogh:
Expanding the guidelines a bit.

NB I would dearly love an agree whether e.g. 'is_connected' should either  be always  a property or always a function. At present it could be both, and I cannot see a good way to distinguish the two, for booleans. OBJ.connected (attribute) is a bit cryptic, but OBJ.get_is_connected is clunky. Any ideas?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,12 +110,20 @@ Imports that are incompatable between Python 2x and 3x should be handled with:
 #### Naming convention
 
 ##### Functions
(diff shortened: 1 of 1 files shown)


IvarsKarpics (approved):

marcus-oscarsson:
I guess, if we go with the reasoning that we had during our meeting,  it should be a property if its simply a boolean indicating whether the HO is connected or not or ?

PR #598, opened 2020-10-15
Get xml filename

IvarsKarpics:
Added method to the HardwareObjectNode to get configuration xml filename


marcus-oscarsson:
Hi,

I investigated a little bit why travis fails and I noticed that we get a segfault for certain imports in the mockup folder. I dont know if its PyTests "test runner" that  messes something up or if its something else. Its definitely related to a dependency update because I ran all the tests without problems in an older Py3 environment but they fail when I install from fresh.

Ill try to narrow down the import that causes this issue, do you have some other clues/ideas ?

PR #599, opened 2020-10-16
Fix travis

IvarsKarpics:
* Moved test config to mockup/test.
* Trying to fix the seg fault


marcus-oscarsson:
Hi,

I investigated a little bit why travis fails and I noticed that we get a segfault for certain imports in the mockup folder. I dont know if its PyTests "test runner" that messes something up or if its something else. Its definitely related to a dependency update because I ran all the tests without problems in an older Py3 environment but they fail when I install from fresh.

Ill try to narrow down the import that causes this issue, do you have some other clues/ideas ?

IvarsKarpics:
Hi @marcus-oscarsson ,

I found out that the tests started to fail when I updated re_emit_values branch for PR #588 . I am looking at this issue as well.

marcus-oscarsson:
Hm, ohh so no they are all failing :/. Travis is sometimes a bit strange because I checked the 2.x tests before and they passed. I looks like some update is a bit slow

marcus-oscarsson:
So there seemed to be an issue with the latest version of greenlet when used with certain versions of gevent. So I've pinned greenlet to 0.4.16, which seems to solve the core dump problem for python 3.

marcus-oscarsson:
@IvarsKarpics Could you have a look at the conflict in the meantime ?

marcus-oscarsson:
So it looks like that works better :)

marcus-oscarsson:
I guess this is not WIP anymore than ?

beteva:
Well done, thanks. Only test_beam missing ✴️

IvarsKarpics:
Thanks @marcus-oscarsson  for fixing the bug. Somehow the branch is not in sync with master, I will update it, revert the test_beam and then it will be ready to merge.

IvarsKarpics:
Ready to be merged.

PR #603, opened 2020-12-16
Detector radius fix

beteva:
Fix the detector radius to be the min(width/2, height/2) if no distance argument in get_radius. The  get_radius with no argument is used to calculate resolution to distance.
Raise RuntimeError if no distance for outer radius -  cannot calculate the radius if both distance argument and distance motor are None.
Fixed bug in force_emit_signals.
Run black. Add documentation.

### DIFF
--- a/HardwareObjects/abstract/AbstractDetector.py
+++ b/HardwareObjects/abstract/AbstractDetector.py
@@ -20,8 +20,24 @@
 
 """
 Detector API
-
-emits signals:
+Abstract methods:
+    prepare_acquisition
+    start_acquisition
+    has_shutterless
+Overloaded methods:
+    force_emit_signals
+Implemented methods:
+    stop_acquisition
+    get_pixel_size, get_width, get_heigth, get_metadata
+    get_radius, get_outer_radius
+    get_beam_position
+    get_roi_mode, set_roi_mode, get_roi_mode_name, get_roi_modes
+    get_exposure_time_limits
+    get_threshold_energy, set_threshold_energy
+    get_binning_mode, set_binning_mode
+Implemented propertries:
+    distance
+Emited signals:
     detectorRoiModeChanged
     temperatureChanged
     humidityChanged
@@ -29,6 +45,7 @@ emits signals:
     frameRateChanged
     stateChanged
     specificStateChanged
+Hardware objects used: energy
 """
 
 import abc
@@ -75,10 +92,7 @@ class AbstractDetector(HardwareObject):
         except KeyError:
             pass
 
-        try:
-            self._distance_motor_hwobj = self.get_object_by_role("detector_distance")
-        except KeyError:
-            pass
+        self._distance_motor_hwobj = self.get_object_by_role("detector_distance")
 
         self._roi_modes_list = ast.literal_eval(self.get_property("roiModes", "()"))
 
@@ -87,13 +101,13 @@ class AbstractDetector(HardwareObject):
         self._height = self.get_property("height")
 
     def force_emit_signals(self):
+        """Emit all hardware object signals.
+        """
         self.emit("detectorRoiModeChanged", (self._roi_mode,))
         self.emit("temperatureChanged", (self._temperature, True))
         self.emit("humidityChanged", (self._humidity, True))
         self.emit("expTimeLimitsChanged", (self._exposure_time_limits,))
-        self.emit(
-            "frameRateChanged", self._actual_frame_rate,
-        )
+        self.emit("frameRateChanged", (self._actual_frame_rate,))
         self.emit("stateChanged", (self._state,))
         self.emit("specificStateChanged", (self._specific_state,))
 
@@ -120,25 +134,23 @@ class AbstractDetector(HardwareObject):
         """
 
     def last_image_saved(self):
-        """
+        """Get the path to the last saved image
         Returns:
-            str: path to last image
+            (str): full path.
         """
         return None
 
     @abc.abstractmethod
     def start_acquisition(self):
-        """
-        Starts acquisition
+        """Start the acquisition.
         """
 
     def stop_acquisition(self):
-        """
-        Stops acquisition
+        """Stop the acquisition.
         """
 
     def get_roi_mode(self):
-        """
+        """Read the current ROI mode.
         Returns:
             (str): current ROI mode
         """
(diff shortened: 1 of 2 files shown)


HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -120,25 +134,23 @@ def prepare_acquisition(self, *args, **kwargs):
...
 
     def stop_acquisition(self):
-        """
-        Stops acquisition
+        """Stop the acquisition.
         """
 
     def get_roi_mode(self):
-        """
+        """Read the current ROI mode.
```
rhfogh, line 153: Why Read and not Get?

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -208,16 +220,16 @@ def get_beam_position(self, distance=None, wavelength=None):
         # is required, e.g. using other detector positioning motors and
         # wavelength
 
-        distance = distance or self._distance_motor_hwobj.get_value()
-        wavelength = wavelength or HWR.beamline.energy.get_wavelength()
-        metadata = self.get_metadata()
-
         try:
+            distance = distance or self._distance_motor_hwobj.get_value()
```
rhfogh, line 224: Just out of curiosity: When could you get the AttributeError?

beteva: If self._distance_motor_hwobj is None. Nothing prevents us of the motor not being initialised.

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -229,23 +241,28 @@ def get_radius(self, distance=None):
...
     def get_outer_radius(self, distance=None):
         """Get distance from beam_position to the furthest point on the detector.
         Args:
             distance (float): Distance [mm]
         Returns:
-            (float): Detector router adius [mm]
+            (float): Detector outer adius [mm]
         """
-        distance = distance or self._distance_motor_hwobj.get_value()
+        try:
```
rhfogh, line 262: Again, when would you get the AttrributeError?

HardwareObjects/abstract/AbstractDetector.py
beteva: Yes, you are right, should be Get.

HardwareObjects/abstract/AbstractDetector.py
beteva: Same as above, If self._distance_motor_hwobj is None.

HardwareObjects/abstract/AbstractDetector.py
rhfogh: Cool.

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -229,23 +241,28 @@ def get_radius(self, distance=None):
         Returns:
             (float): Detector radius [mm]
         """
-        distance = distance or self._distance_motor_hwobj.get_value()
+        if distance is None:
```
rhfogh, line 244: Why would you do this? The point of using the distance is to get the beam centre, which allows you to get the actual radius. The previous behaviour used the current distance, which would get you the resolution matching the current settings. Your new version will give you a nominal radius which uses less information and does not match current conditions. 

When might _distance_motor_hwobj be None. During initialisation? Are we talking about beamlines with unmovable detectors? Or can you not assume that during actual operation you will have a distance motor?

HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -229,23 +241,28 @@ def get_radius(self, distance=None):
...
         """Get distance from beam_position to the furthest point on the detector.
         Args:
             distance (float): Distance [mm]
         Returns:
-            (float): Detector router adius [mm]
+            (float): Detector outer adius [mm]
         """
-        distance = distance or self._distance_motor_hwobj.get_value()
+        try:
+            distance = distance or self._distance_motor_hwobj.get_value()
```
rhfogh, line 263: You could calculate the outer radius assuming the the beam in in the middle, just li you do for the inner radius. But again i do not understand which actual situation you are allowing for,

HardwareObjects/abstract/AbstractDetector.py
beteva: Changing the code in the get_radius comes from the fact that in AbstractResolution calculating the detector distance from the resolution uses the get_raduis without a parameter (which is normal, as we want to calculate the distance). This is only an approximation. If you have a better idea how to calculate the distance from the resolution, we could leave the code here to be dependant of the distance and change the AbstractResolution.
As for the _distance_motor_hwobj be None, this comes from the initialisation - if there is no property detector_distance, there will be no motor object. I do not know the reason, though. Would think that the detector will be still initialised and used even if not movable. Unfortunately I cannot give any example.
When might _distance_motor_hwobj be None?  - it will be None if not declared in the xml file. As we allow this (I do not know the reason), the case should be treated wherever the motor object is used.

HardwareObjects/abstract/AbstractDetector.py
beteva: I was unsure what to be done for the outer radius if there is no distance. Should it be raising an exception?

HardwareObjects/abstract/AbstractDetector.py
rhfogh: If it is only during the intialisation then we should clearly catch the error, but providing the right value in those circumstances is maybe less important. 

The idea was that if you did not pass in a specific value (including from the resolution) you should use the current detector distance. That seems to me to be the normal case - you want to get the detector radius at the current distance setting. If that sometimes fails we need to catch the failures. But I still think it is better to default to the current distance whenever you can, than to default to a nominal 'beam in detector-centre'.

beteva: You cannot use the current distance, as the detector has not yet moved, when you want to move the resolution and calculate the corresponding distance.

rhfogh: Hmmm, you have a point. I'd still argue that 1) it is better in general if get_radius() works off the current distance, (like e.g. bet_beam_position), 2) that you might actually get better values using the current distance since a) the distance-dependent correction should not be that big, b) assuming that the beam is perfectly centred gives you the (extreme) value of the maximum possible radius, c) resolution changes will often be to reasonably similar values. But since this is a correction, how about doing it in two steps? Keep get_radius working off the current distance and do the following in resolution_to_distance:

distance = self._hwr_detector.get_radius() / (tan(2 * asin(_wavelength / (2 * resolution))))
return (round(self._hwr_detector.get_radius(distance) / (tan(2 * asin(_wavelength / (2 * resolution)))), 2)

HardwareObjects/abstract/AbstractDetector.py
rhfogh: I think it should be treated the same as the inner radius. The distance is only used to calculate a correction, for the distance dependence of the beam centre.

HardwareObjects/abstract/AbstractDetector.py
beteva: @rhfogh, thank you for the solution. I will withdraw my PR and will do another one for AbstractResolution

PR #606, opened 2020-12-18
Missing member in Beamline object

jordiandreu:
The Qt widget create_discrete_widget requires this **run_online_processing** boolean value on its initialization.

### DIFF
--- a/HardwareObjects/Beamline.py
+++ b/HardwareObjects/Beamline.py
@@ -102,7 +102,10 @@ class Beamline(ConfiguredObject):
 
         # bool By default run processing of (certain?)data collections in aprallel?
         self.run_processing_parallel = False
-
+        
+        # bool By default run online processing (characterization/mesh?)
+        self.run_online_processing = True
+        
         # Dictionary-of-dictionaries of default acquisition parameters
         self.default_acquisition_parameters = {}
 


rhfogh (approved):
I think I ran into that one as well. Thanks.

beteva:
@jordiandreu, @rhfogh, is this the same run_online_processing, defined by default as Falsein queue_model_objects.py?

jordiandreu:
> @jordiandreu, @rhfogh, is this the same run_online_processing, defined by default as False in queue_model_objects.py?

I think the one discussed in this PR is a boolean value related to the check box present in different Qt widgets. The one in the queue_model_object is, in fact, a string value which contains the processing name (LineScan, XRayCentring,...) also used in these widgets. So, I guess the initialization of this latter should be an empty string.

marcus-oscarsson:
I see, i'm not sure that run_online_processing nor run_processing_parallel should be members of Beamline. @beteva has a point, its already defined in queue_model_objects so the question is what the difference is supposed to be. The location of the members is maybe less important but more so that we are consistent.  So, how are run_online_processing on Beamline different from the one defined in queue_model_objects ?

marcus-oscarsson:
Oh, sorry I see I understand what you meant now @jordiandreu. The ones in Beamline are meant to used as default (configurable) values, right ?

jordiandreu:
Hi @marcus-oscarsson , in fact, I have no clear idea about the need or usage of these variables but I just noticed that the one defined in Beamline object is a boolean value which is used to set the initial value of a check box in the data collection for widget.
The one in the queue model object is not a boolean value (even it is initialized to False). In fact, this variable takes a string value corresponding to the processing that is associated with the queue entry.

That only my guess and maybe someone else can correct or verify my interpretation is correct? @rhfogh @IvarsKarpics ?

I just added the member to the Beamline object just to correct the exception I got, since the task widget required it to initialize the check box value.

IvarsKarpics:
Hi,

There is small mess with these variables. It should be run_online_processing and run_offline_processing. Currently they are boolean, but during the execution they change data type. For example run_offline_processing (currently run_parallel_processing) turns into string describing workflow type (XrayCentering, Mesh), etc. What we actually need is a more complex data type in the beamline to define all processing methods. For example:
```
run_online_processing = {"Dozor": True, "OtherOnlineProcessingMethod": False}
run_offline_processing = {"EDNAProc": True, "Dials": True, "AutoProc": True}
```
And then display in gui checkboxes based on these dictionaries.
I think we can accept this PR as a bug fix and after all these big changes I can propose a solution to this.

rhfogh:
I agree with Ivars, on general principles, do the bug fix now, then reorganise. 

I have not loked at the details, and anyway this one is clearly in good hands

Just a small idea:
Instead of a boolean that turns into a string, how about having emprty string == False (like Jordi suggested (?)) and any other string== True?
And a question: If these run_processing setting differ between different experiment types, could the value not be taken out of the beamline and moved somewhere else, like the relevant hardware object and theri configuration?

marcus-oscarsson:
Sounds good :)

beteva:
Sound OK for me as well to remove the run_*_processing from beamline configuration. I guess their initialisation/implementation should be foreseen somewhere in the AbstractProcessing HO.

PR #607, opened 2020-12-18
[WIP] Fixes for setup.py in mxcube

jordiandreu:
This PR is related to the PR 417 in mxcube/mxcube, which defines a setup.py files for the project. Nevertheless, some changes need to be adopted or discussed for the HardwareRepository as well.


beteva (changes requested):
@IvarsKarpics, do we still need this to be in hardware repository? Normally there should be no dependencies of any kind of gui.

IvarsKarpics (changes requested):
I would suggest to use try, except to import Qt classes. Over the code there are cv2, PIL and I guess other graphical library used.

HardwareObjects/InstanceServer.py
```diff
@@ -8,7 +8,7 @@
 import socket
```
beteva, line 8: Maybe we could rename at least the file (maybe the class as well) QtInstanceServer as it depends on Qt.

jordiandreu:
Dear @IvarsKarpics, @beteva,
Let me add here a reference to this issue from mxcube/mxcube in which I explain what i understand should be solved: https://github.com/mxcube/mxcube/issues/416

For me the point is that these three modules are importing the _QtImport_ module from mxcube repository so, this is creating a circular dependency between the mxcube/mxcube and mxcube/HardwareRepository.

One option would be to move the _QtImport_ module to the mxcube/HardwareRepository as a helper module for the Qt front end. In addition, we will need to relax the Qt dependency in the HardwareRepository. As @IvarsKarpics suggested, we can use try/except in the import statement. Then, we could add the Qt dependency as **extras_require**s in the setup.py file from the HardwareRepository.

marcus-oscarsson:
@IvarsKarpics @jordiandreu @beteva 

We have discussed already discussed this sometime ago and I think we all agreed that the aim was to eventually move all Qt dependent code to MXCuBE Qt. Its true that both cv and PIL are graphics libraries but they are not used for the frontend the same way that Qt is. We have so far been quite liberate on the dependencies but I think it would be good if we can avoid to use redundant dependencies in general. 

Having said that, I guess we could argue that QtGraphicsManager should be moved to the frontend but I however don't really see any direct benefit of doing that. The easiest solution for now would simply be to move the QtImport module to HardwareRepository but make it an optional dependency (Only installed for the Qt frontend).

jordiandreu:
So, as far as I understand, what @marcus-oscarsson said is in agreement with my proposal.

Let me also add that, some more work is required to properly decouple the Qt from the HardwareRepository. For instance, the AbstractVideoDevice ALSO depends on the Qt but I did not notice before cause it does not depend on the QtImport since it imports the Qt directly. A question arise here, shall we import always Qt object using the QtImport?

In agreement with the simplest solution, maybe we could move all these codes which depend on Qt into a subfolder named **qt** in HardwareRepository?

And last, regarding the naming of the modules, we also have modules which did not really depend on Qt but they have the Qt added to the filename: _QtLimaVideo.py_, _QtTangoLimaVideo.py_,...shall we rename these modules to _LimaVideo.py_ and _TangoLimaVideo.py_?


-

IvarsKarpics:
I would suggest remove Qt and PIL from AbstractVideoDevice and move them to necessary subclasses. There is no need of creating a new qt directory for few hardware objects. It is enough to have a Qt prefix.

marcus-oscarsson:
Hi @jordiandreu and @IvarsKarpics 

- @jordiandreu, yes it was maybe a bit unclear but I agree with keeping Qt as an optional dependency :). 
 
- I agree with @IvarsKarpics that I dont think its necessary with an additional subfolder but a Qt prefix would be enough.

- It would finally be good if the Qt import could be done in the same way everywhere and done only were its needed andnot in the abstract class (assuming that its not needed there).

marcus-oscarsson:
@beteva and @IvarsKarpics is it fine like this ?

beteva:
Agree with @IvarsKarpics (remove any GUI dependence from the abstract class) and @marcus-oscarsson. Once all this done I will accept the PR

rhfogh (approved):
Fine with me.

marcus-oscarsson:
Can this be closed now, or do you want to update it ?

jordiandreu:
It can be close since the setup.py has been already added to the master bracnh.

PR #613, opened 2021-03-15
Package structure

marcus-oscarsson:
Dear all,

I've made a filter branch operation to move most files into a mxcubecore subfolder (the python package). This means that the history is preserved which is not the case with the normal `git mv` command. We can probably not merge this as a normal PR but I will likely need to push this with the -f option to master.

So this PR is just to give you an idea of what the structure looks like and if we need to discuss any changes. Once it's done I'll push this to master.

I'll add a preliminary setup.py in the meantime

Marcus

### DIFF
--- a/test/pytest/conftest.py
+++ b/test/pytest/conftest.py
@@ -7,9 +7,12 @@ from gevent import monkey
 monkey.patch_all(thread=False)
 
 TESTS_DIR = os.path.abspath(os.path.dirname(__file__))
-MXCUBE_CORE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../.."))
+ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../.."))
 
-sys.path.insert(0, MXCUBE_CORE_DIR)
+sys.path.insert(0, ROOT_DIR)
+
+print("DEBUG TESTS")
+print(sys.path)
 
 from mxcubecore import HardwareRepository as HWR
 
@@ -26,9 +29,9 @@ from mxcubecore import HardwareRepository as HWR
 @pytest.fixture(scope="function")
 def beamline():
     hwr_path = "%s%s%s" % (
-        os.path.join(MXCUBE_CORE_DIR, "configuration/mockup"),
+        os.path.join(ROOT_DIR, "mxcubecore/configuration/mockup"),
         ":",
-        os.path.join(MXCUBE_CORE_DIR, "configuration/mockup/test")
+        os.path.join(ROOT_DIR, "mxcubecore/configuration/mockup/test")
     )
     HWR._instance = HWR.beamline = None
     HWR.init_hardware_repository(hwr_path)


jordiandreu:
Hi @marcus-oscarsson , thanks for the work. Indeed, is great to preserve the history at that point.

Here are some comments regarding the package structure:

* I think that all files we would like to distribute within the package, is convenient to place them inside the package folder. For this reason, I would suggest to move the pytest folder at the project level.
* Regarding the xml configuration files, i would suggest to distribute the mockup configuration files at the package level, in this way, i would be easy to create a mockup or example application from them. What is not so clear to me is what would be the best way to share the site xml files for MXCuBE (if we want to do so).
* We need also to update the contributing guidelines (replace HardwareRepository by mxcubecore, for instance)
* Also the `py_file_header.txt` file can be added as text into the contributing guide lines (?).
* We should decide a fisrt version number for this new repository. We could check on the semantic versioning recommendations and agree on a first starting number (https://semver.org/).
* Meanwhile, maybe we could add the WIP keyword to the PR title.

marcus-oscarsson:
Hi,

> Hi @marcus-oscarsson , thanks for the work. Indeed, is great to preserve the history at that point.
> 
> Here are some comments regarding the package structure:
> 
> * I think that all files we would like to distribute within the package, is convenient to place them inside the package folder. For this reason, I would suggest to move the pytest folder at the project level.

I agree,

> * Regarding the xml configuration files, i would suggest to distribute the mockup configuration files at the package level, in this way, i would be easy to create a mockup or example application from them. What is not so clear to me is what would be the best way to share the site xml files for MXCuBE (if we want to do so).

I think they should definitely be shared but I personally dont think its, in general, a good idea to distribute the configuration files with the package. It's definitely worth thinking about how we would like to handle the configuration files. I used the package_data option to include the configuration files in the package at install. But we can discuss the different options and to have something running with the mockup configuration would be needed.

> * We need also to update the contributing guidelines (replace HardwareRepository by mxcubecore, for instance)

We already started to do this, but I think that this work needs to be extended beyond this PR.

> * Also the `py_file_header.txt` file can be added as text into the contributing guide lines (?).

I think it was done like this so that one could configure ones editor to read this file, but I don't mind if its included
in the contributing guide lines instead.

> * We should decide a fisrt version number for this new repository. We could check on the semantic versioning recommendations and agree on a first starting number (https://semver.org/).

We mentioned that at the last developers meeting as well, so it will be one of the topics on the next meeting.

> * Meanwhile, maybe we could add the WIP keyword to the PR title.
> 
Yes of course :)
> Cheers,
> Jordi

rhfogh (approved):
Looks good, but nothing much I could add - hard to judge such an extensve change. Maybe an overview summary at the next developers' meeting?

beteva:
Thanks for the hard work @marcus-oscarsson.
Concerning the configuration files (xml and yml) I think we should exclude the site specific ones from the package and would only put the template ones, needed for the tests. For one the configuration files directory structure may differ from one site to another, for second it will be quite cumbersome to make PRs just for changing the configuration and for third, they can be quite outdated.
One could add in the hardware object itself (e.g. BlissEnergy.py) an example configuration file.

marcus-oscarsson:
> Looks good, but nothing much I could add - hard to judge such an extensve change. Maybe an overview summary at the next developers' meeting?

Yes, we can manage that. I can explain how I did this and what it changes

marcus-oscarsson:
The tests now passes, I will move the configuration files into the package directory and then we can have a discussion on where to
keep the site specfic ones later.

marcus-oscarsson:
So, I finally got the tests to pass :)

The test folder was moved out of the package folder and the configuration folder moved back in. We can as I said earlier then discuss the exact contents of the later. It would be necessary to keep at least the test and mockup configuration files ...

Please have a last look, let me know when you are fine with this and Ill push this to master

Marcus

IvarsKarpics:
Hi @marcus-oscarsson ,

I see that the branch is out-of-date and tests are failing (import error in python 2.7). After fixing these two issues I will merge this PR.

marcus-oscarsson:
@IvarsKarpics hm the tests were actually passing just a few days ago and they pass for both Python 2.7 and Python 3 on my setup. I will have a closer look to get the Travis one to pass again.

I don't think we can merge this PR like this, I think we might need to force push it.

Marcus

IvarsKarpics:
Ah, ok. Then I would suggest to force push these changes and open an another one for fixing travis tests.

PR #619, opened 2021-05-27
Fixed bug in QtVideoMockup

IvarsKarpics:
* added missing import
* added sample.jpg


jordiandreu (approved):
Fine for me.
Maybe in the future we can adopt the pkg_resources module to load static files from within the python module.

rhfogh (approved):
Looks fine to me

PR #620, opened 2021-06-02
[WIP] Abstract procedure

IvarsKarpics:
* Added variables offline_processing_methods and online_processing_methods to the Beamline. Both are lists of strings and are populated based on beamline_config.yml. Processing methods are are listed in the processing widget. For example: offline_processing_methods  = ["EDNAProc", "AutoProc"] will result in displaying "Run offline processing (EDNAProc, AutoProc)". If not method is defined then the checkbox is unchecked and disabled.
* Renamed run_processing_parallel to run_offline_processing. True will check the checkbox "Run offline processing" in processing_widget.
* Cleaned up AbstractCollect: removed not used imports, inherit class from AbstractProcedure, implemented _pre_execute, _execute and post_execute. Currently methods are empty and in the next PR will be reviewed.
* AbstractOnlineProcessing is inherited now from AbstractProcedure. Removed not used code.
* Cleaned AbstractProcedure. For sure it is not the final version, but at least to start with.
* CollectMockup reviewed and cleaned.
* queue_entry is modified based on signals emitted by CollectMockup, OnlineProcessingMockup (both based on AbstractProcedure)

Now mockup data collection is running and there are online processing results displayed.


rhfogh (approved):
This is too big to look at in detail - we shall have to wait till it is in, try it out, and then propose any changes. But it looks good and clean. Onward it is!

marcus-oscarsson:
Hm, we are still using the old fashioned AbstractMultiCollect so we can use this a it is with the changes to queue_entry, could you make a separate queue_entry for your tests. We need to be able to use AbstractMultiCollect for a little longer.

mxcubecore/HardwareObjects/queue_entry.py
```diff
@@ -595,62 +596,52 @@ def __getstate__(self):
...
+        HWR.beamline.collect.start(self.get_data_model())
+
+        HWR.beamline.collect.wait()
+        HWR.beamline.online_processing.wait(timeout=10)
 
     def pre_execute(self):
         BaseQueueEntry.pre_execute(self)
 
         qc = self.get_queue_controller()
-
```
marcus-oscarsson, line 625: So these changes would be problematic for us, as we are still not using AbstractProcedure for the collect

marcus-oscarsson (changes requested):

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -91,6 +91,9 @@ def __init__(self, name):
         # List[str] of available methods
         self.available_methods = []
 
+        # List[str] of available methods
+        self.available_tasks = []
+
```
beteva, line 96: Could you, please, change the comment and explain the difference between tasks and merhods

mxcubecore/HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -34,6 +34,7 @@
 
 
 __credits__ = ["MXCuBE collaboration"]
+__license__ = "LGPLv3+"
 
 
 # Temporary definition should use common denfinition from
```
beteva, line 40: Maybe it is a good time to change the states implementation, as specified in the comment.

rhfogh (approved):
There is too much here for me to judge the details, but it looks about right, and we shall sort out the final details ater in any case.

mxcubecore/HardwareObjects/abstract/AbstractProcedure.py
```diff
@@ -49,7 +50,7 @@ class ProcedureState(IntEnum):
     READY = 3
 
 
-class AbstractProcedure(ConfiguredObject):
+class AbstractProcedure(HardwareObject):
```
rhfogh, line 53: Any paeticular reason to stay with xml configuration instead of moving to yaml configuration? From the current class names I think a yaml-configured object should have been HardwareObjectYaml, so there is another alternative to ConfiguredObject

rhfogh:
Is this PR still relevant, or has it been superseded by events?

marcus-oscarsson:
Indeed, I think we can consider this superseded

PR #621, opened 2021-06-02
Add setup.py file

jordiandreu:
I have added a setup.py file for the project in order to be installed using pip. The setup.py provides:

* Installation using pip command.
* It removes the need to manually modify the PYTHONPATH environment variable.
* It install dependencies according to the python version required, using the requirement files already defined.
* It defined the option to also intall de dependencies related to the GPhL workflows.

Out of the scope:
* It does not add the MANIFEST.in file required to generate source distributions.
* Some metadata still need to be defined (authors, emails, version,...)
* It does not include the configuration of any versioning tool.


setup.py
```diff
@@ -0,0 +1,63 @@
...
+        'console_scripts': console_scripts,
+        'gui_scripts': gui_scripts
+    }
+
+setup(
+    name='mxcubecore',
+    # version='0.1.0',
+    # author='The MXCuBE developers',
+    # author_email='mxcube@mxcube.org',
+    description='MXCuBE core libraries',
```
IvarsKarpics, line 37: Should it be MXCuBE core library?

IvarsKarpics (approved):

setup.py
jordiandreu: Fine for me.

marcus-oscarsson:
Looks good to me, thanks !

PR #622, opened 2021-06-02
Added a few missing mock objects mainly used by mxcube-web

marcus-oscarsson:
There were a few changes that for some reason was lost on a previous merge, mainly a few mock objects that are used by mxcube web.

 - Added missing mock-up objects used by mxcube-web
 - Minor update to CommandContainer for allowing beamline actions with input forms based on JSON schema


mxcubecore/HardwareObjects/mockup/AttenuatorsMockup.py
```diff
@@ -0,0 +1,24 @@
+from mxcubecore.BaseHardwareObjects import Device
+
+
+class AttenuatorsMockup(Device):
```
IvarsKarpics, line 4: Do we need this class and file? There is AbstractTransmission class that inherits AbstractTransmission. Here is an old method getAttState and signal attFactorChanged.

mxcubecore/HardwareObjects/mockup/MachInfoMockup.py
```diff
@@ -0,0 +1,120 @@
...
+"""
+
+import gevent
+import time
+
+from mxcubecore import HardwareRepository as HWR
+from mxcubecore.BaseHardwareObjects import Equipment
+
+
+class MachInfoMockup(Equipment):
```
IvarsKarpics, line 42: We have MachineInfoMockup. I suggest not to add an another mockup.

IvarsKarpics (changes requested):

mxcubecore/HardwareObjects/mockup/MachInfoMockup.py
marcus-oscarsson: Ah, yes you are right, missed that simply a config file that referred to an old object

mxcubecore/HardwareObjects/mockup/AttenuatorsMockup.py
marcus-oscarsson: Yes of course :)

mxcubecore/HardwareObjects/mockup/MultiCollectMockup.py
```diff
@@ -0,0 +1,313 @@
...
+    AbstractMultiCollect,
+)
+from mxcubecore.TaskUtils import task
+import logging
+import time
+import os
+import gevent
+
+
+class MultiCollectMockup(AbstractMultiCollect, HardwareObject):
```
rhfogh, line 12: Is that still in use? OI thought we were going over to AbstractCollect.

marcus-oscarsson: Its till used for the moment unfortunately, we will eventually start to use the AbstractCollect but its still not the case

mxcubecore/HardwareObjects/mockup/BeamMockup.py
```diff
@@ -98,6 +100,9 @@ def set_beam_position_on_screen(self, beam_x, beam_y):
         self._beam_position_on_screen = (beam_x, beam_y)
         self.emit("beamPosChanged", (self._beam_position_on_screen,))
 
+    def get_value(self):
```
rhfogh, line 103: This really ocnfuses me. As I rmember, the 'value' attriute was only used when it was clear what it wa teh value of. What is the 'value' of a beam? Also, how come get_value returns a list of teh values in the beam_info_dict, when set_value seems to set the aperture diameter size?

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
```diff
@@ -0,0 +1,130 @@
...
+from enum import Enum
+from gevent import Timeout, sleep
+from mxcubecore.HardwareObjects.abstract.AbstractNState import AbstractNState
+from mxcubecore.Command.Exporter import Exporter
+from mxcubecore.Command.exporter.ExporterStates import ExporterStates
+
+__copyright__ = """ Copyright © 2020 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+class ExporterNStateMockup(AbstractNState):
```
rhfogh, line 40: I am sure it makes sense, but could someone explain this to me? 

Why do you call a function called '_value2state' with a state input parameter?
Why do you maintain a separate _mock_state, when the abstract classes above alrady have _state and _specific_state, and (same thing) why overide get_state?
Why do you need hwstate instead of specific_state?

mxcubecore/HardwareObjects/mockup/BeamMockup.py
marcus-oscarsson: This is code that is currently in use for the mxcube web mock-ups and that was for some reason not merged a while back, its hence a bit strange bit it makes things work as they were.

I think the idea of the beam object was exactly that we could do set and get value and that the underlying mechanism that is used to actually define the beam was used to do the right thing. There were some more thought work to be done to make that work with all devices used so I think it was left like this.

Its a bit strange perhaps, but get_value returns the beam size x and y in microns which I don't think is so unexpected for a get_value function of a Beam object to return, even if could perhaps return some more things.

Would you rather that its get size_in_microns or something  ?

The set in this case only takes one value as the beam shape is assumed to be the same in both dimensions or simple just selected by the horizontal size (for the cases were there is no aperture).

I can rework this a bit, but I think it require some more general thought before it can be an appropriate abstraction

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
marcus-oscarsson: I think @beteva could give you some better answers, so she has to correct me if I'm wrong.

 '_value2state' is used to convert a diffractomter specific state to one of the states defined by HardwareObjectState

I think that '_mock_state' and 'mock_value' are used to "emulate" the values currently held by the diffractometer
and that's why get_state is also overridden. 

I think get_hwstate is there for completeness as its defined in ExporterNState and it gets the state directly from the hardware.

mxcubecore/HardwareObjects/mockup/BeamMockup.py
rhfogh: I checked in the mock branch you are merging from - as well as origin/master - and near as I can see, AbstractBeam does not have get_value or set_value. It has, instead, get_beam_size, get_beam_shape, get_beam_info_dict, and set_beam_size_shape. This makes much more sense to me, and, anyway, BeamMockup ought to have the same functions as AbstractBeam. Have I got the wrong AbstractBeam implementation somehow?

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
rhfogh: It may be just a matter of documentation, but it does rather confuse me when both parameter names and documentation in _update_state and _value2state do not make a clear distinction between what is a state and what is a value. Can you read through those two functions and get a clear idea of what the parameter types for input and values set are?

mxcubecore/HardwareObjects/mockup/MultiCollectMockup.py
rhfogh: OK

mxcubecore/HardwareObjects/mockup/BeamMockup.py
marcus-oscarsson: Well we are using this code today

https://github.com/mxcube/mxcubecore/blob/master/mxcubecore/HardwareObjects/ESRF/ESRFBeam.py

Were we do have set and get value, as I said previously, I think the intention was to create a more general abstraction with get and set value but it was never agreed on and hence never done. We are relying on the set_value and get_value to delegate the "beam shaping" to the right device. I can create a ESRFBeamMockup if that makes more sense or add some methods to AbstractBeam ?

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
marcus-oscarsson: Hm, I think you and @beteva had a very similair discussion for:

https://github.com/mxcube/mxcubecore/blob/master/mxcubecore/HardwareObjects/ExporterNState.py

I don't know what you said then, but the same would apply I guess. Otherwise maybe @beteva can expand on the meaning of the argument names ?

mxcubecore/HardwareObjects/mockup/BeamMockup.py
beteva: I recall discussing AbstractBeam in march 2020. At that time it was said that it will have get_value to replace evaluate_beam_info and set_value to replace set_beam_size_shape. As we needed it an to not brake the rest of the world, I've implemented this in our local version - ESRFBeam. For me this comes under the "Clean Abstract Classes" task.

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
beteva: @marcus-oscarsson is right, _value2state is used to convert the exporter state (which is a string) to HardwareObjectState, which is an enum. It is maybe not well documented in the Mockup, but in the ExporterNState class  the _value2state is documented as:
"""Convert string state to HardwareObjectState enum value
        Args:
            state (str): the state
        Returns:
            (enum 'HardwareObjectState'): state
        """
As from get_hwstate - the microdiff has two different states - the application global state (which is reported by get_state) and an additional hardware state, but only for more recent versions of the MD2 software. For some actuators, the global (software) state does not change, while they are moving, but the hardware state does, so it is handy to use it, but as I said, it is not always available.

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
```diff
@@ -0,0 +1,130 @@
...
+
+            return self.update_state(state)
+
+    def _value2state(self, state):
+        """Convert string state to HardwareObjectState enum value
+        Args:
+            state (str): the state
+        Returns:
+            (enum 'HardwareObjectState'): state
+        """
```
beteva, line 82: Actually we do have the right documentation. Maybe it is not clear enough

rhfogh: It may be be because I am unattached to a synchrotron.  But I see an object that has something called state and something called value, and yet uses 'state' indiscriminately' for both. OK, one is a hardware-state, and one is an object-state, but not being intimate with the details I would be unable to code to this, since I would not know whether to pass the result of get_state() or get_value() to any given function. Similarly I am unable to tell, looking at the code, whether it is correct or not. Not that I do not trust you to get it right, but in order to check it I would have to check the data flow for the times when these functions are called. Would there not be a choice of variable and function names that distinguished the two, so you can see which is which?

mxcubecore/HardwareObjects/mockup/BeamMockup.py
IvarsKarpics: Hi all,

To be honest i do not remember what we decided regarding the get_value, set_value of AbstractBeam. I am ok to add these methods to avoid collapsing all other world :) I have just two comments:
- Please add set_value, get_value to the AbstractBeam, so it is clear what those methods should do.
- In your case the return value of get_value is a list of dict values. This is very unclear. Maybe it is easier to return the

```
self._beam_info_dict = {
                    "size_x": self._beam_width,
                    "size_y": self._beam_height,
                    "shape": self._beam_shape,
                    "label": self._beam_label,
}
```

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
beteva: OK I should have named the method str2state to be more clear and I am going to do so.  It is an internal method and all you need to be sure is that get_state gives you a HardwareObjectState.

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
beteva: OK I should have named the method str2state to be more clear and I am going to do so. It is an internal method and all you need to be sure is that get_state gives you a HardwareObjectState.

rhfogh: OK, with just that modification I can understand what is gong on, including why you need _mock_state etc. The distinction is between string values and enum objects, not (as I thought) between the state of the hardware (value) and state of the object, or state and specific state. Sorry it took so long.

mxcubecore/HardwareObjects/mockup/BeamMockup.py
rhfogh: Looking at the code in AbstractBeam I can see this is a quite complex case. The simplest would be to have just 'beam_size', 'beam_shape', and 'beam_label' as separate attributes, with no 'value', but maybe things do not work that way. If we do have to have a value, I would vote that it should be the beam_info_dict. Either way I think it is mandatory that 1) the data type passed to set_value and the data type returned from get_value must be identical; 2) that if we pass mutable objects like dictionaries, they must be copied both on entry and on exit to avoid inadvertently modifying the internals of the object.

Could you help me with a use case? Suppose I find out the crystal dimensions and want to set the beam to match. Also I want to do this in a way that works on multipe beamlines. Could it work to simply set the beam_size? Do I need to say whether I want elliptical or rectangular beam? Do all beamlines support both? Looking at the get_available_size implementation I see I can get either a list of apertures, or a list of predefined beam positions or the high and low limits for x and y slits, depending on the beamline. How am I supposed to figure out what beam sizes are available and what to set, in a site-independent manner?

mxcubecore/HardwareObjects/mockup/ExporterNStateMockup.py
beteva: Now I understand the confusion. The change has been made.

mxcubecore/HardwareObjects/ExporterNState.py
```diff
@@ -114,11 +114,11 @@ def _update_state(self, state=None):
         if not state:
             state = self.get_state()
         else:
-            state = self._value2state(state)
+            state = self._str2state(state)
```
IvarsKarpics, line 117: could it be better called _convert_str_to_state? Numbers in the method name are not so common and it is recommended to be a verb.

rhfogh: I would be happy enough with _str2state (short and unambiguous), but there is nothing wrong with _convert_str_to_state either.

marcus-oscarsson: Same for me either is fine

mxcubecore/HardwareObjects/mockup/BeamMockup.py
marcus-oscarsson: As I mentioned earlier, this was all part of a rather complex discussion that was postponed and then never concluded.

There is still some work to be done to create something that is more consistent but what we have to day is still somewhat incomplete. There might in theory be a case where you need to specify the shape of a certain beam size even if that its, to my knowledge highly unlikely. We currently don't use multiple shapes for the same size. The available sizes are retrieved by get_available_size (which regrettably is incorrectly named).

We cant really change this behavior based on a mockup without having done the underlying changes to the hardware objects and thus test those. I would suggest that we conclude our discussion around AbstractBeam, it seems like the question of set_value and get_value still remain. Would it in the meantime be possible to merge these changes so that we can continue with the rest of the work ?

I created https://github.com/mxcube/mxcubecore/discussions/624 so that the discussion can continue there

IvarsKarpics: I would suggest to merge this PR as it is and discuss AbstractBeam in another issue, PR.

marcus-oscarsson: Okey, I just created #624 for that

marcus-oscarsson: Ok, So Ill merge this as the pending request for change is from you @IvarsKarpics.

marcus-oscarsson:
Thanks for having a look, so Ill merge this as the pending request for change is from you @IvarsKarpics.

marcus-oscarsson:
The discussion will continue in #624

PR #627, opened 2021-06-14
[Bug] Reverted CollectMockup to working state

IvarsKarpics:
* Reverted abstract_collect and CollectMockup to working version.
* Replaced parallel_processing with online_processing

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractCollect.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractCollect.py
@@ -1,21 +1,21 @@
 #
 #  Project: MXCuBE
-#  https://github.com/mxcube
+#  https://github.com/mxcube.
 #
 #  This file is part of MXCuBE software.
 #
 #  MXCuBE is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU Lesser General Public License as published by
+#  it under the terms of the GNU General Public License as published by
 #  the Free Software Foundation, either version 3 of the License, or
 #  (at your option) any later version.
 #
 #  MXCuBE is distributed in the hope that it will be useful,
 #  but WITHOUT ANY WARRANTY; without even the implied warranty of
 #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU Lesser General Public License for more details.
+#  GNU General Public License for more details.
 #
-#  You should have received a copy of the GNU Lesser General Public License
-#  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
+#  You should have received a copy of the GNU General Public License
+#  along with MXCuBE.  If not, see <http://www.gnu.org/licenses/>.
 
 """
 AbstractMulticollect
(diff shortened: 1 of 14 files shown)


mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -1,48 +1,42 @@
...
-#  it under the terms of the GNU Lesser General Public License as published by
+#  it under the terms of the GNU General Public License as published by
 #  the Free Software Foundation, either version 3 of the License, or
 #  (at your option) any later version.
 #
 #  MXCuBE is distributed in the hope that it will be useful,
 #  but WITHOUT ANY WARRANTY; without even the implied warranty of
 #  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU Lesser General Public License for more details.
+#  GNU General Public License for more details.
```
rhfogh, line 15: ALARM!!!! We are releasing und LGPL and not GPL. GPL would mean aht teh Gobal Phasing collaboration became impossible

rhfogh (changes requested):
The licensing needs to be LGPL.
For the rest, I assume that the changes to AbstractCollect are temporary/WIP? When we finally setle I shall surely have some things to put in (currently from the SOLEIL close-to-master branch). But assuming that this is still in flux I see no reason to go through teh details.

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
IvarsKarpics: done

IvarsKarpics:
Hi @rhfogh 

The changes in AbstractCollect are not temporary. They are bugfixes. Currently the AbstractCollect is not compatible with queue_entry and this PR restores it to a working state.

rhfogh:
@IvarsKarpics True, they are bug fixes. But one way or the other this is not the end of looking at AbstractCollect, I guess, so there is no need for putting a magnifying glass on the code. I may have something for later, when I merge in my WF version from Soleil PX2, but meanwhile I am happy with this.

marcus-oscarsson:
> @IvarsKarpics True, they are bug fixes. But one way or the other this is not the end of looking at AbstractCollect, I guess, so there is no need for putting a magnifying glass on the code. I may have something for later, when I merge in my WF version from Soleil PX2, but meanwhile I am happy with this.

Agree

marcus-oscarsson:
@rhfogh you are still requesting for changes, is that really case or can I merge ?

marcus-oscarsson (approved):

rhfogh:
My only change request is the licensing, and that is fixed. Please go ahead.

Is there something I could have done to make the situation clearer?

IvarsKarpics:
AbstractCollect in detail will be discussed in #620 . I think we are ok with the PR. Can someone merge it?

marcus-oscarsson:
> My only change request is the licensing, and that is fixed. Please go ahead.
> 
> Is there something I could have done to make the situation clearer?

Ah no that is fine but you still have the PR marked as "change requested" so I have to force merge. Just wanted to check so that I did not make mistake. Dont forget to approve the PR after you have requested changes.

PR #629, opened 2021-06-15
Add hierarchy scheme.

beteva:


rhfogh (approved):
It is very good to have this.

I can pick a couple of nits, though.

The link from Beamline to HardwareObject is not a subclass relationship like the others but a containment link, Maybe a different type of arrow (subclass is generally colored, open-headed arrows).

Typos (updatr_state, AbstracResolution)

I am not quit e clear what it means that valueChanged is written in both AbstractActuator and its subclass AbstractMotor

Would it help to have this a a pptx file (so others can edit if there are minor changes) or are you happy to have it so only can edit it?

I did NOT choose to 'request changes', so you are welcome to merge as it stands.

beteva:
> I am not quite clear what it means that valueChanged is written in both AbstractActuator and its subclass AbstractMotor

The method, which sends the signal (update_value) is overloaded in AbstractMotor. That's why it is the AbstractMotor and not AbstractActuator that sends the signal in this case.

rhfogh:
@beteva Makes sense..

marcus-oscarsson:
I also think this is a great start ! It would be really nice though if we can use some UML modeling tool, maybe a web based one so that its easy to collaborate and share the schema. Maybe someone already knows a good tool for this, the best thing would be one that can parse python code and generate (at least to get started).

Issue #628, opened 2021-06-15
Add sample_temperature to beamline roles


beteva:
Add sample_temperature to the roles of Beamline.py. We can access the temperature (of the cryo stream for time being) through HWR. This is also to prepare eventual control of the temperature (ramping, set ...)

rhfogh:
Sounds reasonable.

IvarsKarpics:
Is it a variable representing temperature at the sample position or it is a device? Would it make a sense to add a new role called cryo (AbstractCryo with variable temperature) or sample_environment?

beteva:
It can be just a temperature, but it may as well represent a device. Having AbstractCryo in not contradictory to what I suggest, it is another aspect of it. The sample_temperature is just the way how you want to access it. Right now we access it via the diffractometer, but it is not quite right.  cryo is not a good name, because it may not be just cooling, but also heating.
I quite like sample_environment.

beteva:
The idea behind the sample environment is to be able to access directly (and not through diffractometer) any sample environment device.
For now we only have reading of the temperature of the cryo stream. But in a near future, we will need changing this temperature (immediately or by ramping) and I could think of other parameters as pressure. For this I am not quite sure if the simple get_value/set_value mechanism is enough. If we do have only one parameter (e.g. only temperature) we could read it with get_value. Bit a simple set_value needs a prerequisite that it sets the value immediately. Than we may add other custom (not abstract) commands as ramp, start/stop ... Things get more complicated if we have more than one sample environment parameter, but this is not the case for now. For this I thought that AbstractSampleEnvironment is maybe not needed, as we cannot really specify generic methods, other than already available get/set value. if we do go for AbstractSampleEnvironment methods like get_units/set_units can be useful.

rhfogh:
If there is a limited and predictable set of parameters (temperature, pressure, humidity???) I would vote in favour of having separate objects rather than a generic 'SampleEnvironment' - as you  say "we cannot really specify generic methods". If there may be many environment parameters with generalisable methods or you may need complex multi-parameters time courses, then maybe an 'AbstractSampleEnvironment'. But since this is far from my expertise I shall be happy with whatever you guys think best.

beteva:
[...]

PR #633, opened 2021-07-07
Add sample_environment to the beamline properties.

beteva:
Resolves issue #628.


rhfogh:
This looks fine - but would it not be better to add the AbstractSampleEnvironment at the same time, so we can see how it will be - and so the type given nin this change matches something existing? I am also curious whether this will contain only the temperature or eventually?) also other attributes.

IvarsKarpics (approved):

marcus-oscarsson:
I also think its a good idea to draft/discuss what `AbstractSampleEnvironment` would include so that everybody can get an idea of what it is.

beteva:
The idea behind the sample environment is to be able to access directly (and not through diffractometer) any sample environment device.
For now we only have reading of the temperature of the cryo stream. But in a near future, we will need changing this temperature (immediately or by ramping) and I could think of other parameters as pressure. For this I am not quite sure if the simple get_value/set_value mechanism is enough. If we do have only one parameter (e.g. only temperature) we could read it with get_value. Bit a simple set_value needs a prerequisite that it sets the value immediately. Than we may add other custom (not abstract) commands as ramp, start/stop ... Things get more complicated if we have more than one sample environment parameter, but this is not the case for now. For this I thought that AbstractSampleEnvironment is maybe not needed, as we cannot really specify generic methods, other than already available get/set value. if we do go for  AbstractSampleEnvironment methods like get_units/set_units can be useful.

marcus-oscarsson:
I think that it might be better to add the access to the `cryo` and `humidifier` directly onto the beamline object for the time being. Do we really need to encapsulate them in a `AbstractSampleEnvironment`.

Or do you foresee some functionality to over-arches everything in the "sample enviroment" ?

beteva:
I do not think we need AbstractSampleEnvironment. Please see  issue #628.

PR #639, opened 2021-08-16
Fixes after testing mxcubeqt with python3/qt5

IvarsKarpics:
Here are some changes, fixes after testing the latest version of mxcubecore with python3/qt5 on the beamline p13

- Beamline.py: 
1. added boolean run_offline_processing to indicate if by default the offline autoprocessing will be launched. 
2. added list variable offline_processing_methods and online_processing_methods to indicate autoprocessing pipelines (EDNAProc, Dials, etc). These names are displayed in the processing widget.
3. added beamstop property. Several methods require beamstop parameters and it is convenient to access it via beamline object (and not via diffractometer hwobj)
- DozorOnlineProcessing.py: fixed method name
- ExporterMotor: renamed self.motor_position to self.chan_motor_position and self.motor_state to self.chan_motor_state. I would encourige people to use prefix chan_ and cmd_ to indicate if variable is channel or method. Otherwise it looks like variable describing position and state.
- ExporterNState: changed value validation. To be discussed.
- GenericDiffractometer: added force_emit_signals method
- QtGraphicsLib: fixed bug if qt5 is used
- VimbaVideo.py: rewrote hwobj to be compatible with the latest pymba library
- AbstractBeam: init beam size defined by the beam definer
- AbstractCollect, AbstractDetector: cleaned up the binning_mode (stored in ispyb)
- AbstractEnergy: added get_undulators_gap method
- AbstractShutter: added is_open, is_closed, open, close methods
- queue_model_objects: get container (puck) code and add it to the display name

I will create a corresponding PR in mxcubeqt after merging this PR.


mxcubecore/HardwareObjects/DozorOnlineProcessing.py
```diff
@@ -53,8 +53,9 @@ def create_processing_input_file(self, processing_input_filename):
         input_file.setReversing_rotation(
             XSDataBoolean(self.params_dict["reversing_rotation"])
         )
-        input_file.setPixelMin(XSDataInteger(HWR.beamline.detector.get_pixel_min()))
-        input_file.setPixelMax(XSDataInteger(HWR.beamline.detector.get_pixel_max()))
+        pixel_size = HWR.beamline.detector.get_pixel_size()
```
rhfogh, line 56: Probably fine - I do not understand Dozor. But it makes me wonder a little, since AbstractDetector.get_pixel_size is supposed to return a pair of floats (pixel size in mm), whereas setPixelMin and setPixelMax seem to take some kind of integer sa input. Are we OK here?

rhfogh (approved):
Looks good. Just one minor niggle (see relevant comment)

mxcubecore/HardwareObjects/DozorOnlineProcessing.py
IvarsKarpics: Yes, your are right. get_pixel_size returns a list and setPixelMin, setPixelMax requires an integer as an input. The code is ok, ... or I do not see the issue?

rhfogh: One of us is missing something. The pixel size in mm is two floats between 0 and 1. I cannot see how you can meaningfully transfer that to two integers pixelMin and pixelMax. If you were talking about the *number* of pixels I could just about see it - but (by the definition of get_pixel_size) you are not.

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -205,6 +209,18 @@ def beam(self):
...
+        """Beamstop Hardware object
+
+        Returns:
+            Optional[AbstractActuator]:
+        """
+        return self._objects.get("beamstop")
+
+    __content_roles.append("beamstop")
+
+
```
beteva, line 223: We should revise the roles in the Beamline object before including this one together with other 19 files in a single PR. Please remove this part from the PR

mxcubecore/HardwareObjects/ExporterMotor.py
```diff
@@ -50,8 +50,8 @@ def __init__(self, name):
         self._motor_state_suffix = None
         self._exporter = None
         self._exporter_address = None
-        self.motor_position = None
-        self.motor_state = None
+        self.chan_motor_position = None
+        self.chan_motor_state = None
```
beteva, line 54: Could you, please, tell us the reason to change the names of the variables?

mxcubecore/HardwareObjects/ExporterNState.py
```diff
@@ -104,6 +104,11 @@ def _wait_ready(self, timeout=None):
             while not self.get_state() == self.STATES.READY:
                 sleep(0.5)
 
+    def update_value(self, value=None):
+        if value is None:
+            value = self.value_channel.get_value()
+        AbstractNState.update_value(self, self.value_to_enum(value))
+
```
beteva, line 111: Sorry, I do not get the difference between the AbstractActuator update_value, which handles value=None ant this implementation. Do you want to force the reading anyhow?

mxcubecore/HardwareObjects/ExporterNState.py
```diff
@@ -142,6 +147,12 @@ def abort(self):
         if self.get_state() != self.STATES.UNKNOWN:
             self._exporter.execute("abort")
 
+    def validate_value(self, value):
+        if self.value_to_enum(value) in (None, self.VALUES.UNKNOWN):
+            return False
+        else:
+            return True
+
```
beteva, line 155: Could you, please, explain why the AbstractNState validate_value is not good? In your case you change the type of the value parameter, which makes it inconsistent with the AbstractNState.

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -494,7 +497,7 @@ def get_undulators_gaps(self):
         """
         Descript. :
         """
-        return {}
+        return HWR.beamline.energy.get_undulator_gaps()
```
beteva, line 500: Is it really under energy HO?

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
```diff
@@ -45,6 +45,7 @@ class AbstractEnergy(AbstractActuator):
     def __init__(self, name):
         AbstractActuator.__init__(self, name)
         self._wavelength_limits = (None, None)
+        self.undulator_gaps = ()
```
beteva, line 48: I do not think the place is here. Again before doing such changes, a consensus should be achieved. Please, remove this part from your PR.

beteva (changes requested):

mxcubecore/HardwareObjects/ExporterMotor.py
IvarsKarpics: Hi @beteva 

See my first comment for reason. I do not think self.motor_position describes a channel and a name self.chan_motor_position is more descriptive.

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
IvarsKarpics: Hmm, which is then a better place? Machine info?

mxcubecore/HardwareObjects/ExporterMotor.py
beteva: Sorry @IvarsKarpics, I've missed you first comment.

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
beteva: I do not thing you need any specific place, as you may have a bending magnet, which has no gaps at all. Maybe I did not look carefully enough, but why do you need a special placeholder and not use the motors instead in your specific hardware object.

mxcubecore/HardwareObjects/ExporterNState.py
IvarsKarpics: For some reason I implemented validate_value and update_value to handle Zoom. I will check tomorrow at the beamline if it works without re implementing these methods.

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
IvarsKarpics: The only place where those gaps are used is the ISPyBClient.py. They are stored and displayed in the ispyb gui. I think some time ago we discussed this and I do not remember why they got removed from AbstractEnergy.

mxcubecore/HardwareObjects/Beamline.py
IvarsKarpics: ok

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
beteva: If this is of any help, we use MicrodiffZoom, which inherits from ExporterNState. It even check how many zoom levels are configured (by default 10 with the Prosilica and 6 for the B-zoom).

mxcubecore/HardwareObjects/DozorOnlineProcessing.py
IvarsKarpics: Hmm, indeed something went wrong here. I have not used this hwobj for a while and have to double check what this pixelMin and pixelMax represents. For sure it is not the pixel size. Can we finish this PR and I will open a new one?

mxcubecore/HardwareObjects/ExporterNState.py
IvarsKarpics: Hi @beteva  I reverted AbstractActuator and ExporterNState. For a quick look the MicrodiffZoom worked well and is sufficient.

mxcubecore/HardwareObjects/DozorOnlineProcessing.py
rhfogh: Sure, go ahead.

mxcubecore/HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, name):
 
         self._pixel_size = (None, None)
 
-        self._binning_mode = 0
+        self._binning_mode = "Unbinned"
```
beteva, line 77: @IvarsKarpics, the get/set binning mode to take/return integer. Initialising it to a string is inconsistent.

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
beteva: @IvarsKarpics, historically the undulator gaps were accessed in the collect hardware object. I agree, this is probably not the best place, but AbstractEnergy is not either. Could you, please, remove them from AbstractEnergy and we could decide where to implement them in another PR.

mxcubecore/HardwareObjects/abstract/AbstractShutter.py
```diff
@@ -50,3 +50,15 @@ def __init__(self, name):
     def init(self):
         """Initilise the predefined values"""
         AbstractNState.init(self)
+
+    def is_open(self):
+        return self.get_value() == self.VALUES.OPEN
+
+    def is_closed(self):
+        return self.get_value() == self.VALUES.CLOSED
```
beteva, line 58: Would it be better to set this two as a property?

beteva (changes requested):

mxcubecore/HardwareObjects/abstract/AbstractDetector.py
IvarsKarpics: Hmm, maybe someone can explain me what the detectorMode in ispyb data model means? The variable is a varchar and is used in the findDetector method: https://github.com/ispyb/ISPyB/blob/66771e4bb253ad802209219ba65ec2155cc4ac77/ispyb-ejb/src/main/java/ispyb/server/mx/services/collections/Detector3ServiceBean.java#L252 that is called via webservice.

mxcubecore/HardwareObjects/abstract/AbstractEnergy.py
IvarsKarpics: Well, I am not fully ok, but keep going I will remove the code.

mxcubecore/HardwareObjects/abstract/AbstractShutter.py
IvarsKarpics: ok

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -100,12 +100,16 @@ def __init__(self, name):
...
+        # bool By default run processing of (certain?)data collections?
+        self.run_offline_processing = False
         
         # bool By default run online processing (characterization/mesh?)
-        self.run_online_processing = True
+        self.run_online_processing = False
         
+        self.offline_processing_methods = []
+
+        self.online_processing_methods = []
```
marcus-oscarsson, line 111: What will this list contain ?

mxcubecore/HardwareObjects/ExporterMotor.py
```diff
@@ -50,8 +50,8 @@ def __init__(self, name):
         self._motor_state_suffix = None
         self._exporter = None
         self._exporter_address = None
-        self.motor_position = None
-        self.motor_state = None
+        self.chan_motor_position = None
```
marcus-oscarsson, line 53: IMO  `motor_position_chan` is more readable in that case (or even channel)

mxcubecore/HardwareObjects/QueueManager.py
```diff
@@ -222,14 +222,16 @@ def stop(self):
...
             for qe in self._current_queue_entries:
                 try:
-                    qe.QUEUE_ENTRY_STATUS.FAILED
+                    qe.status = QUEUE_ENTRY_STATUS.FAILED
                     self.emit("queue_entry_execute_finished", (qe, "Aborted"))
                     qe.stop()
                     qe.post_execute()
                 except base_queue_entry.QueueAbortedException:
-                    pass
-                except Exception:
```
marcus-oscarsson, line 231: Oh, yes good :)

mxcubecore/HardwareObjects/abstract/AbstractBeam.py
```diff
@@ -61,6 +61,7 @@ def __init__(self, name):
         self._beam_size_dict = {
             "aperture": [sys.float_info.max, sys.float_info.max],
             "slits": [sys.float_info.max, sys.float_info.max],
+            "definer": [sys.float_info.max, sys.float_info.max]
```
marcus-oscarsson, line 64: Whats the definer ?

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -110,6 +111,7 @@ def init(self):
             detector_model=HWR.beamline.detector.get_property("model"),
             detector_px=HWR.beamline.detector.get_property("px"),
             detector_py=HWR.beamline.detector.get_property("py"),
+            detector_binning_mode=HWR.beamline.detector.get_binning_mode(),
```
marcus-oscarsson, line 114: Great :)

mxcubecore/HardwareObjects/abstract/AbstractDetector.py
marcus-oscarsson: Hm, I guess we can always find a way to deal with how ISPyB wants the data its maybe more important that we are consistent with our internal `AbstractDetector`. It looks like `binning_mode` is indeed a integer in [`AbstractDetector`](
https://github.com/mxcube/mxcubecore/blob/44b8f4886faa0a0c8a14a258437ed25c4aa48dec/mxcubecore/HardwareObjects/abstract/AbstractDetector.py#L77)

mxcubecore/HardwareObjects/abstract/AbstractShutter.py
```diff
@@ -50,3 +50,17 @@ def __init__(self, name):
     def init(self):
         """Initilise the predefined values"""
         AbstractNState.init(self)
+
+    @property
+    def is_open(self):
+        return self.get_value() == self.VALUES.OPEN
+ 
+    @property
+    def is_closed(self):
```
marcus-oscarsson, line 59: Do we need both is_open and is_closed ? , I guess the confusion it causes really minimal but I will for sure always need to double check when I see whichever is the least used ...

mxcubecore/HardwareObjects/Beamline.py
IvarsKarpics: list of strings. See my first comment as a descriptions of PR

mxcubecore/HardwareObjects/ExporterMotor.py
IvarsKarpics: Well, its a question of taste. I find it convenient when all channels have prefix chan and commands cmd. When there are many of them (like in diffractometer) then it is easier to read the code.

mxcubecore/HardwareObjects/abstract/AbstractBeam.py
IvarsKarpics: anything else that can change beam size and shape: crls, beamfocusing motors, etc.

marcus-oscarsson: Oh, I see interesting

mxcubecore/HardwareObjects/ExporterMotor.py
marcus-oscarsson: I agree that its good to have that information in the name. As I remember it we have previously discussed this when it comes to types and that we would add a suffix i.e `list`, `dict`when it would help, for instance `something_list`. I would consider that the same reasoning should apply here. So it should then in my opinion be `motor_position_channel` or `motor_position_chan` if we would like to remain consistent.

beteva: :+1:

IvarsKarpics: ok, then lets have motor_position_chan

mxcubecore/HardwareObjects/abstract/AbstractOnlineProcessing.py
```diff
@@ -712,9 +708,9 @@ def store_processing_results(self, status):
                     str(self.params_dict["reversing_rotation"]),
                     HWR.beamline.detector.get_pixel_min(),
                     HWR.beamline.detector.get_pixel_max(),
-                    self.beamstop_hwobj.get_size(),
-                    self.beamstop_hwobj.get_distance(),
-                    self.beamstop_hwobj.get_direction(),
+                    HWR.beamline.beamstop.get_size(),
+                    HWR.beamline.beamstop.get_distance(),
+                    HWR.beamline.beamstop.get_direction(),
```
beteva, line 713: Still problem with this one. Coulod you, please, remove it from this PR. We have not decided yet if the beamstop should be part of the beamline or the diffractometer, or anything else.

mxcubecore/HardwareObjects/abstract/AbstractShutter.py
beteva: I guess is_open is enough. Closed may have more that one meaning (really closed, closed because of safety reasons...)

mxcubecore/HardwareObjects/abstract/AbstractOnlineProcessing.py
IvarsKarpics: ok. I removed the beamstop. This method stores results in csv file and at some point we needed to store beamstop parameters. I think now it is not the case. We should definitely review beamline object.

mxcubecore/HardwareObjects/abstract/AbstractShutter.py
IvarsKarpics: ok, I removed the is_closed method

mxcubecore/HardwareObjects/abstract/AbstractDetector.py
IvarsKarpics: For some reason this code is commented out
https://github.com/mxcube/mxcubecore/blob/b894885ee65e902769cd73ccf392c8540fffdcef/mxcubecore/HardwareObjects/ISPyBClient.py#L2193-L2205
I assume this binning_mode is never changed in mxcube. I guess if we keep it as None it will be not used as selection criteria in the ispyb web service.

marcus-oscarsson: We are in fact changing binning_mode from MXCuBE. Its not used very often at the moment, but will be used more often in the future. I cant really say way this code is commented out though (maybe due to some table update ?)

IvarsKarpics: fine. I kept the set_ and get_ methods and set the variable to None. Lets clear this in a separate PR

mxcubecore/HardwareObjects/abstract/AbstractOnlineProcessing.py
```diff
@@ -286,12 +285,12 @@ def prepare_processing(self):
...
         try:
             gevent.spawn(
-                self.save_snapshot_task, os.path.join(archive_directory, "snapshot.png")
+                self.save_snapshot_task, self.params_dict["snapshot_path"]
             )
         except Exception:
             logging.getLogger("GUI").exception(
                 "Online processing: Could not save snapshot: %s"
-                % os.path.join(archive_directory, "snapshot.png")
+                % self.params_dict["snapshot_path"]
```
beteva, line 293: I like the idea of snapshot_path. Question: here you only have path, while in the previous code it was the filename. Is this for purpose.

beteva (approved):

beteva:
Side question, do we still use logging.getLogger("GUI") and what would be the difference with logging.getLogger("user_level_log")

mxcubecore/HardwareObjects/abstract/AbstractOnlineProcessing.py
IvarsKarpics: It is just a code clean up to avoid confusion. In line 191 snapshot_path is initialized to `os.path.join(archive_directory, "snapshot.png"`

IvarsKarpics:
I personally use logging.getLogger("GUI"), because it similar to HWR and in it easier to read the log file. There is no difference between GUI and user_level_log

mxcubecore/HardwareObjects/abstract/AbstractOnlineProcessing.py
beteva: Than it is not path, but file. The label is misleading

PR #642, opened 2021-08-31
Fix package

jordiandreu:
Add missing package files and review requirement files. It fixes the installation of the package using pip in a clean environment.


rhfogh (approved):
Not very knowledgeable, but looks fine to me.

marcus-oscarsson (approved):

PR #651, opened 2021-10-15
[WIP] Adapted to new mxcubecore API

marcus-oscarsson:
Mainly adaptations to the new API resulting from the refactoring process and change to mxcubecore.

### DIFF
--- a/mxcubecore/Command/Mockup.py
+++ b/mxcubecore/Command/Mockup.py
@@ -20,7 +20,7 @@
 
 import time
 import logging
-import Queue
+import queue
 import weakref
 
 from mxcubecore.CommandContainer import CommandObject, ChannelObject
--- a/mxcubecore/HardwareObjects/ESRF/id29_calc_gaps.py
+++ b/mxcubecore/HardwareObjects/ESRF/id29_calc_gaps.py
@@ -1,7 +1,6 @@
 import logging
 import math
 import sys
-from conversion import HC_OVER_E
 
 
 class CalculateGaps:
@@ -80,7 +79,8 @@ class CalculateGaps:
         const = 13.056 * arr[1] * 100 / pow(contst_en, 2)
         k2 = (math.pi / arr[1]) / 1000
         # Transform energy in wavelength
-        energy = HC_OVER_E / energy
+        h_over_e = 12.3984
+        energy = h_over_e / energy
         target = energy / const
         nsols = 0
 
--- a/mxcubecore/HardwareObjects/LimaEigerDetector.py
+++ b/mxcubecore/HardwareObjects/LimaEigerDetector.py
@@ -1,30 +1,8 @@
-# encoding: utf-8
-#
-#  Project: MXCuBE
-#  https://github.com/mxcube
-#
-#  This file is part of MXCuBE software.
-#
-#  MXCuBE is free software: you can redistribute it and/or modify
-#  it under the terms of the GNU Lesser General Public License as published by
-#  the Free Software Foundation, either version 3 of the License, or
-#  (at your option) any later version.
-#
-#  MXCuBE is distributed in the hope that it will be useful,
-#  but WITHOUT ANY WARRANTY; without even the implied warranty of
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU Lesser General Public License for more details.
-#
-#  You should have received a copy of the GNU General Lesser Public License
-#  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
-
-"""LimaEigerDetector Class
-Lima Tango Device Server implementation of the Dectris Eiger2 Detector.
-"""
+import gevent
+import time
 import os
 import math
 import logging
-from gevent import Timeout, sleep
 
 from mxcubecore.TaskUtils import task
 from mxcubecore import HardwareRepository as HWR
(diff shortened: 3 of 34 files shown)


mxcubecore/Command/Mockup.py
```diff
@@ -20,7 +20,7 @@
 
 import time
 import logging
-import Queue
+import queue
```
IvarsKarpics, line 23: This is python3x import and will break in python2. Should we add code to handle both python2 and 3 or from now we just support python3?

mxcubecore/HardwareObjects/QueueManager.py
```diff
@@ -143,6 +152,10 @@ def __execute_task(self):
         finally:
             self._running = False
             self.emit("queue_execution_finished", (None,))
+
+            if self._collect_hwobj:
```
IvarsKarpics, line 156: This looks very strange:
1. queue_finished_cleanup do not exist in AbstractCollect and will break the queue execution.
2. QueueManager should be free from hardware objects and handles just queue entries defined in queue_entry.py
This could be done in the level of AbstractMulticollect or AbstractCollect

mxcubecore/HardwareObjects/QueueManager.py
```diff
@@ -151,7 +164,7 @@ def __execute_entry(self, entry):
 
         status = "Successful"
         # self.emit('centringAllowed', (False, ))
-        self.emit("queue_entry_execute_started", (entry))
+        self.emit("queue_entry_execute_started", (entry, status))
```
IvarsKarpics, line 167: Signal requires one argument:

https://github.com/mxcube/mxcubeqt/blob/5c106e662b1dc3c753de39fbd5c43886777bc7c5/mxcubeqt/bricks/tree_brick.py#L519

Please revert it

marcus-oscarsson: Oh, have to double check this signature, it might be that we are using the status for something but then again it seems a bit redundant for ..._execute_started. We can probably do without it

mxcubecore/Command/Mockup.py
marcus-oscarsson: The Poller and Commands/* implementation probably also needs a review in general. It works fine but there also seems to be quite a few outdated bits.

mxcubecore/HardwareObjects/ESRF/id29_calc_gaps.py
```diff
@@ -80,7 +79,8 @@ def _calc_gap(self, energy, arr):
         const = 13.056 * arr[1] * 100 / pow(contst_en, 2)
         k2 = (math.pi / arr[1]) / 1000
         # Transform energy in wavelength
-        energy = HC_OVER_E / energy
+        h_over_e = 12.3984
```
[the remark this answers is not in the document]
beteva: This should not even be here. I will remove it.

mxcubecore/HardwareObjects/LimaEigerDetector.py
```diff
@@ -1,53 +1,30 @@
...
-#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#  GNU Lesser General Public License for more details.
-#
-#  You should have received a copy of the GNU General Lesser Public License
-#  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
-
-"""LimaEigerDetector Class
-Lima Tango Device Server implementation of the Dectris Eiger2 Detector.
-"""
+import gevent
```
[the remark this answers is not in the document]
marcus-oscarsson: :), wopps, Il add it back :)

rhfogh:
Marcus,

I had thought that my PR should be merged into this first, any outstanding NBNBs etc. dealt with, and then the result should be merged into master - that is why I made my PR to fix-esrfupdate rather than master. But I would say that you are free to do it in any order you think best, run black, and then merge the result once you are happy with it.

marcus-oscarsson:
I appreciate your NBNB and TODO comments but its also somewhat odd to both have the possibility to review the PR and add comments and make inline suggestions of improvements in the code. I think we should work out a way to handle those. What Ill do for the moment is to make a comment for each of the NBNB's so that we can claraify them

rhfogh:
I see your point - and I would certainly not say that these NBNBs should hold up the merge. But, then, I saw this a bit as a service, having a nice editor and using it to go through and look for (potential) problems in a code body that you yourself said was still settling down. The rest in a certain sense is up to you. These are things that I could not propose fixes for myself, and it would not really make sense for me to go through code systematically and make one mega-issue or lots of little issues out of the warnings I found. If you think this ends up being more trouble than it is worth I shall not do it again. Otherwise we need to agree on how it should be handled.

marcus-oscarsson:
:), sounds good. I think your comments and remarks are good and that they should be addressed and its a **great** service. Thanks alot for taking the time doing it :+1:. I just wanted to clarify that I though that some of the remarks where somewhat out of context, but I agree that they are important :)

IvarsKarpics (approved):

marcus-oscarsson:
:), Thanks I will address the NBNB that was introduced in this PR and then we could maybe come up with a way to signal and handle potential issues with the code that spot while doing other things and to "heavy" to create issues for.

marcus-oscarsson:
Ive addressed the outstanding comments and additional NBNB (that were added in this PR) added by @rhfogh . The remaining NBNB comments needs to be addressed in separate PR. Thanks for the review everyone :)

PR #656, opened 2021-11-05
Rhfogh mxcube3

rhfogh:
Reverted change to AbstractSampleChanger, and various improvements in GPhL-specific code.

NOTE: The change to AbstractSampleChanger is a temporary fix that reverts a commit from 15/10. The modified version did not register a sample mounted from the samplegrid in  app.CURRENTLY_MOUNTED_SAMPLE. Quite likely a more complex change is required to avoid reintroducing the problems that the previous change was supposed to solve.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractSampleChanger.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractSampleChanger.py
@@ -214,7 +214,7 @@ class SampleChanger(Container, Equipment):
     CONTENTS_UPDATED_EVENT = "contentsUpdated"
 
     def __init__(self, type_, scannable, *args, **kwargs):
-        super(SampleChanger, self).__init__(type_, None, type, scannable)
+        super(SampleChanger, self).__init__(type_, None, type_, scannable)
         if len(args) == 0:
             args = (type_,)
         Equipment.__init__(self, *args, **kwargs)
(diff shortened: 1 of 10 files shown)


mxcubecore/HardwareObjects/abstract/AbstractSampleChanger.py
```diff
@@ -801,23 +801,16 @@ def _reset_loaded_sample(self):
...
             else:
-                if self.get_loaded_sample() == sm:
-                    sm._set_loaded(True)
-
-        if previous_loaded != self.get_loaded_sample():
-            self._trigger_loaded_sample_changed_event(sample)
+                if self.get_loaded_sample() != s:
+                    s._set_loaded(True)
 
+        self._trigger_loaded_sample_changed_event(sample)
```
marcus-oscarsson, line 813: I need to verify this because it was working when I tested it. The previous code also had the problem that the
`self._trigger_loaded_sample_changed_event(sample)` was sent even if the sample did not change.

marcus-oscarsson:
Looks good, but I have to double check the "revert" done on AbstractSampleChanger. Did you have the problem in both Qt and Web ?

rhfogh:
It makes sense that you check this - after all we would not want to unfix what you aave fixed. 

I am only testing on the web version at the moment - it is hard enough to keep up with one version rather than tracking two simultaneously.

As I understood it, the problem was that the _trigger_loaded_sample_changed_event was triggered after the the new sample had been set to is_loaded elsewhere, so that previous_loaded == self.get_loaded_sample() was true  before the function is called - even when you are loading a new sample (at least for me). Could it be a matter of mock v. real - or of which sample moune command you are using? I guess that is one potential complication with keeping track of which sample is loaded separeately in the core and the web layer - you need to syncronise the change and keep track of what is set when where.

rhfogh:
State of play 20211122:

The PR contains 
1) GPhL-specific changess
2) A couple of innocuous changes to log messages
3) Two bug fixes to AbstractSampleChanger.

Of the latter, one is a simple typo - I made an incomplete fix earlier, and missed a name change 'type' -> 'type_'. 
The other regards an earlier change of Marcus to AbstractSampleChanger._set_loaded_sample. An earlier fix from Marcus to avoid setting the same sample twice had the effect in my hands to break the transmission of sample-is-mounted information to the UI. Reverting the change fixes my problem, but clearly messes up the original reason for the change.

PR #658, opened 2021-11-30
EMBL config update

IvarsKarpics:


rhfogh (approved):
Presumably this works - anwyay it is impossible for an outsider to review, really. So I approve. A couple of the settings in the GPhL config files I do not recognise, but one would have to double check against the local installation. 

Approved.

beteva (approved):

beteva:
Being just configuration file, I cannot judge either if all is OK.

beteva:
I was wondering if all this should be in mxcubecore. After all these kind of files are not even institute, but beamline specific. In my opinion they should not be part of mxcubecore.

rhfogh:
Some are nice to have - Session, for instance, and it allows you to see what hardware and settings people have.

beteva:
Most of them are also control system dependant. This makes a lot of particularities.

marcus-oscarsson:
We have already discussed the matter of the example configuration files being a part of the repository during our meetings. The conclusion at the time was, if remember it correctly, that they were to remain there.

PR #660, opened 2021-12-01
Patch for issue #895

marcus-oscarsson:
The problem described in #895 is due to a retry mechanism in the client that re-sends a request if there are no response after a certain amount of time. The call to `execute_entry` is in the case of #895 blocking so a retry is made after n minutes, effectively calling the entry twice (or even more depending on the duration of the call)

A parameter `_async` to make the call non-blocking was added, the task is executed in a greenlet if `_async` is passed `True`. The default behavior is to still that `execute_entry ` is blocking, so all logic that need a non blocking behavior need to set  `_async` to `True`

### DIFF
--- a/mxcubecore/HardwareObjects/QueueManager.py
+++ b/mxcubecore/HardwareObjects/QueueManager.py
@@ -235,14 +235,18 @@ class QueueManager(HardwareObject, QueueEntryContainer):
                 except Exception:
                     pass
 
-        self._root_task.kill(block=False)
+        if self._root_task:
+            self._root_task.kill(block=False)
 
+        self._queue_end()
+
+    def _queue_end(self):
         # Reset the pause event, incase we were waiting.
         self.set_pause(False)
-        self.emit("queue_stopped", (None,))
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
Looks good - though I am not able to test at the moment.

BTW, what is our practice about having parameters that start with underscore? What does it mean?

rhfogh (approved):
I did not realise that async was a keyword, sorry. Still, I would vote, gently, for using 'async_' or 'is_async' instead, since '_async' strongly suggests (to me) 'internal variable or attribute'. But that is just one person's opinion, you are free to stick to your own and merge it anyway.

marcus-oscarsson:
Aha, well I'm just used to add underscore to those variables but we can change I actually like is_async or use_async instead, a better more readable option :)

marcus-oscarsson:
How about like this ? :)

rhfogh (approved):
This looks fine (and better suits my tastes - thanks). I assume you made a corresponding change in mxcube3?

marcus-oscarsson:
doing it :)

PR #661, opened 2021-12-17
Define set_value for better precision.

beteva:
The precision depends on the difference between the current resolution and the target value - the smaller the difference, he better the precision.
Move twice to get the closet possible to the requested resolution

### DIFF
--- a/mxcubecore/HardwareObjects/Resolution.py
+++ b/mxcubecore/HardwareObjects/Resolution.py
@@ -48,3 +48,17 @@ class Resolution(AbstractResolution):
     def init(self):
         """Initialisation"""
         super(Resolution, self).init()
+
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
Looks fine

beteva:
Hi @all, can we merge this?

rhfogh:
This still shows "Some checks haven’t completed yet". Should we ignore this? Or is there some problem in the  testing tha needs fixing?

marcus-oscarsson:
I think we should look over our CI and maybe change from travis to something else, its been relay slow lately. I even think they have changed something in their infrastructure so might need to update it to make it work properly. Its alright for the moment that we merge without this being run but we need to take a closer look at the CI in the near future.

PR #663, opened 2022-01-11
[WIP] Abstract diffarctometer

beteva:
AbstractDiffractometer class:

- inherits from HardwareObject and uses wait_ready, abort methods and the state handling.
- defines some methods, properties and the DiffractometerHead, DiffractometerPhase enums.
- defines some scan types
- in the init method defines standard access to different equipment (motors, actuators and complex/compound equipment). The access relies on the roles, which are defined in the configuration file. To be discussed if checks are to be done. This is a typical xml configuration file:
```
<object class = "MicroDiffractometer"
  <username>MD2S</username>
  <exporter_address>id29md2:9001</exporter_address>
  <motors>
    <object role="omega" hwrid="/udiff_omega"/>
    <object role="phiy" href="/udiff_phiy"/>
    <object role="phiz" href="/udiff_phiz"/>
    <object role="sampx" href="/udiff_sampx"/>
    <object role="sampy" href="/udiff_sampy"/>
    <object role="focus" href="/udiff_phix"/>
    <object role="zoom" href="/udiff_zoom_as_motor"/>
    <object role="front_light" href="/udiff_frontlight_intensity"/>
    <object role="back_light" href="/udiff_backlight_intensity"/>
    <!-- <object role="horizontal_alignment" hwrid="/udiff_phiy"/>
    <object role="vertical_alignment" hwrid="/udiff_phiz"/>
    <object role="horizontal_centring" hwrid="/udiff_sampx"/>
    <object role="vertical_centring" hwrid="/udiff_sampy"/> -->
  </motors>
  <actuators>
    <object role="fast_shutter" href="/udiff_fastshut"/>
    <object role="scintillator" href="/udiff_scint"/>
    <object role="diode" href="/udiff_diode"/>
    <object role="fluo_detector" href="/udiff_fluodet"/>
    <object role="cryostream" href="/udiff_cryostream"/>
    <object role="front_light" href="/udiff_frontlight_inout"/>
    <object role="back_light" href="/udiff_backlight_inout"/>
  </actuators>
  <complex_equipment>
    <object role="capillary" href="/udiff_capillary"/>
    <object role="aperture" href="/udiff_aperture"/>
    <object role="zoom" href="/udiff_zoom"/>
  </complex_equipment>
</object>
```

MD2 implementation - overload or fill in the abstract methods.

### DIFF
--- a/mxcubecore/HardwareObjects/MicroDiffractometer.py
+++ b/mxcubecore/HardwareObjects/MicroDiffractometer.py
@@ -176,7 +176,7 @@ class MicroDiffractometer(AbstractDiffractometer):
             (list): list of tuples (motor role, position)
         """
         motors_positions_list = super().get_values_motors(motors_list)
-        if not self.in_kappa_mode():
+        if not self.in_kappa_mode:
             for _mot in motors_positions_list:
                 if _mot[0] in ("kappa", "kappa_phi"):
                     motors_positions_list.remove(_mot)
@@ -225,7 +225,7 @@ class MicroDiffractometer(AbstractDiffractometer):
         Returns:
             (bool): True (parameters within the limits), False otherwise.
         """
-        if self.in_plate_mode():
+        if self.in_plate_mode:
             scan_speed = math.fabs(end - start) / exptime
             llim, hlim = map(
                 float,
@@ -376,7 +376,7 @@ class MicroDiffractometer(AbstractDiffractometer):
             RuntimeError: Timeout waiting for status ready.
         """
 
-        if self.in_plate_mode():
+        if self.in_plate_mode:
             # to see if needed when plates
             return
         scan_params = f"{nb_frames}\t{start:0.4}\t{scan_range:0.4}\t{exptime:0.4}\t"
--- a/mxcubecore/HardwareObjects/MicroDiffractometer.py
+++ b/mxcubecore/HardwareObjects/MicroDiffractometer.py
@@ -61,7 +61,7 @@ Example xml file:
 """
 import math
 from gevent import Timeout, sleep
-from AbstractDiffractometer import (
+from mxcubecore.HardwareObjects.abstract.AbstractDiffractometer import (
     AbstractDiffractometer,
     DiffractometerHead,
     DiffractometerPhase,
@@ -73,7 +73,7 @@ __copyright__ = """ Copyright © 2010-2022 by the MXCuBE collaboration """
 __license__ = "LGPLv3+"
 
 
-class Microdiff(AbstractDiffractometer):
+class MicroDiffractometer(AbstractDiffractometer):
     """Microdiff with Exporter implementation of AbstartDiffractometer"""
 
     def __init__(self, name):
@@ -82,7 +82,7 @@ class Microdiff(AbstractDiffractometer):
 
     def init(self):
         """Initialise the device"""
-        super.init()
+        super().init()
         # Initialise the commands and channels
         exporter_address = self.get_property("exporter_address")
         _host, _port = exporter_address.split(":")
--- a/mxcubecore/HardwareObjects/MicroDiffractometer.py
+++ b/mxcubecore/HardwareObjects/MicroDiffractometer.py
@@ -27,7 +27,7 @@ Example xml file:
   <username>MD2S</username>
   <exporter_address>wid30bmd2s:9001</exporter_address>
   <motors>
-    <device role="phi" hwrid="/udiff_omega"/>
+    <device role="omega" hwrid="/udiff_omega"/>
     <device role="phiy" hwrid="/udiff_phiy"/>
     <device role="phiz" hwrid="/udiff_phiz"/>
     <device role="sampx" hwrid="/udiff_sampx"/>
--- a/mxcubecore/HardwareObjects/MicroDiffractometer.py
+++ b/mxcubecore/HardwareObjects/MicroDiffractometer.py
@@ -20,7 +20,7 @@
 
 """MD2 implementation of the AbstractDiffractometer class.
 Overloads the methods:
-set_values_motors, _get_values_motors, set_phase, get_phase and all the scans.
+set_value_motors, _get_value_motors, set_phase, get_phase and all the scans.
 
 Example xml file:
 <object class = "Microdiff"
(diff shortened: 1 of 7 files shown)


beteva:
Dear @all, this is the cleaned version of an abstract diffractometer, which only handles the equipment and introduces some standardisation to access it. Please feel free to give you comments, ideas, pro and cons.

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -339,12 +341,49 @@ def do_mesh_scan(
...
+                             If timeout = 0: return at once and do not wait
+                             if timeout is None: wait forever (default).
+        Raises:
+            RuntimeError: Timeout waiting for status ready.
         """
         scan_params = f"{pulse_duration:0.7}\t{pulse_period:0.7}\t{nb_pulse}"
         self._exporter.execute("startStillScan", (scan_params,))
         self._wait_ready(timeout)
+
+    def do_characterisation_scan(
```
rhfogh, line 359: Nice!

rhfogh (approved):
Looks fine.

Just out of curiosity - when is 'no import possible'?

beteva:
` Just out of curiosity - when is 'no import possible'?`
Hi @rhfogh,
not very explicit comment. It actually means that it is more handy in the GUI to use HWR.beamline.diffractometer.get_phase_enum than to use an export.

rhfogh:
> ` Just out of curiosity - when is 'no import possible'?` Hi @rhfogh, not very explicit comment. It actually means that it is more handy in the GUI to use HWR.beamline.diffractometer.get_phase_enum than to use an export.

Cool. Thanks.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,317 @@
...
+        self.current_phase = None
+        self.head_type = None
+        self.timeout = 3  # default timeout 3 s
+
+    def init(self):
+        """Initialise actuator_name and username properties.
+        Initialise the equipment, defined in the configuration file
+        """
+        self.username = self.get_property("username") or self.username
+        all_roles = self._on
```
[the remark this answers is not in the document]
beteva: This was a test I was doing. It is gone now. Sorry for the pollution.

rhfogh (commented):
Are you changing the axis names globally? It used to be 'phi', 'kappa', 'kappa_phi', and those names used to be standard so you could use then as explicit string in your code. I very much prefer 'omega', but is the old 'kappa_phi' now called 'phi'?  And are we sure that this change will go rhgough in teh entire code base without problems or complaints?

beteva:
> Are you changing the axis names globally? It used to be 'phi', 'kappa', 'kappa_phi', and those names used to be standard so you could use then as explicit string in your code. I very much prefer 'omega', but is the old 'kappa_phi' now called 'phi'? And are we sure that this change will got through in the entire code base without problems or complaints?

This is an attempt to standardise the roles, which are still explicit strings. I've left kappa_phi, because it is only relevant with minikappa, but omega is, according to our scientist, what our former phi should be. Of course I am expecting complains, that's why I've only renamed phi to omega in few places, which will anyhow break if the new AbstractDifrractometer is used.

I think, as we are anyhow changing then diffractometer API, it is a good opportunity to be more "standard". Should be discussed at the next developer's meeting.

rhfogh:
I strongly support that. And while in an ideal world you would rename 'kappa_phi' to 'phi' it is probably less error-prone to keep the old name.

beteva:
> I strongly support that. 
😃

beteva:
> Some general comments, since this is a WIP:
> 
>     * functions move_motors and get_motor_positions take a list of tuples as input. Could it not be an ordered dictionary instead? Dictionaries are actually always ordered in the newest pythons, I believe. That would be handier for me (though other may have different requirements of course).

This has been just to not change from what has already been with the previous diffractometer implementation. I do not have a strong opinion, we just have to decide what to be.


>     * functions set_actuator and get_actuator_position take a n actual hardware object as input parameter. Would it not be better if they took a role name since then you would not need to pass actual objects around?

Could you, please, put this comment next to the relevant code. Unfortunately I do not get the idea.

>     * Why a function for get_head_type when it just returns an existing attribute? either ername the attribute (starting '_' or skip the function?

This has been implemented, because we need in the concrete implementation to get it from the hardware. Indeed it could be something that can be set in the init method and remove the get_head_type. We anyhow need to restart everything as we physically change head.
I was just wondering, as we often use get/set methods, this is more in the line of the standards, rather than just reading an attribute. I can very well remove the method, if this is what we decide.

rhfogh:
On the head_type, since there is a good reason for having this as a function, maybe we could rename the internal attribute to
self._head_type
instead of
self.head_type
That would keep people from the temptation to interact directly with the attribute.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,311 @@
...
+                return
+            self.wait_ready(timeout)
+
+    def _set_phase(self, phase):
+        """Specific implementation to set the diffractometer to selected phase
+        Args:
+            phase (Enum): DiffractometerPhase value.
+        """
+
+    def get_phase(self):
```
meghdadyazdi, line 251: `get_current_phase` maybe? as we need it [here](https://github.com/mxcube/mxcube3/blob/develop/mxcube3/routes/diffractometer.py#L23) and [here](https://github.com/mxcube/mxcube3/blob/develop/mxcube3/core/components/beamline.py#L358)
We might need `def get_phase_list` for [here](https://github.com/mxcube/mxcube3/blob/develop/mxcube3/routes/diffractometer.py#L39) too but not enum but list of str (?)

beteva: I would rather keep get_phase, as it is consistent with the decision we've made for all the objects that have get/set method.
get_phase_list is replaced by get_phase_enum, as it returns enum and not list any more. It has been changed to enum in order to cope with different types of diffractometers - we guarantee the names to be the same for different diffractometer types.

PR #664, opened 2022-01-14
Removed name override

marcus-oscarsson:
- Overriding name method gives unexpected names for exporter motors

### DIFF
--- a/mxcubecore/HardwareObjects/BlissMotor.py
+++ b/mxcubecore/HardwareObjects/BlissMotor.py
@@ -195,6 +195,3 @@ class BlissMotor(AbstractMotor):
         """Stop the motor movement"""
         self.motor_obj.stop(wait=False)
 
-    def name(self):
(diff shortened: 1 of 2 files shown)


rhfogh:
Looks fine in itself. But 1) might it not be worth it to remove that method also from BlissMotor? 
2) (much more controversial) might this be the right time to refactor, renaming the name() method to get_name in the superclass? Of course that is likely rather more demanding.

rhfogh (approved):
Subject to previous comment

marcus-oscarsson:
Thanks, you are right there should not be any name override for BlisMotor either.

marcus-oscarsson:
For changing from name() to get_name() or maybe even just to the property name ? would be a good idea.  Not part of this PR but we can make another one.

rhfogh:
@marcus-oscarsson I agree - and I still approve of the PR.

marcus-oscarsson:
Created #665

PR #666, opened 2022-01-21
Added development requirements to requirements files

marcus-oscarsson:
Added development requirements to requirements files


marcus-oscarsson:
Testing GitHub action

PR #667, opened 2022-01-21
Handle light

beteva:
Fix set_light_in/out, used by XMLRPC.
Change obsolete call in MiniDiff
Handle property coming from bliss actuator.

### DIFF
--- a/mxcubecore/HardwareObjects/BlissNState.py
+++ b/mxcubecore/HardwareObjects/BlissNState.py
@@ -88,7 +88,10 @@ class BlissNState(AbstractNState):
             if self._prefix:
                 _attr = self._prefix + "_is_in"
                 _cmd = getattr(self._bliss_obj, _attr)
-                _val = _cmd()
+                if isinstance(_cmd, bool):
+                    _val = _cmd
+                else:
+                    _val = _cmd()
             else:
                 _val = self._bliss_obj.state
         # print(f"get_value:  {self.value_to_enum(_val)}, {_val}")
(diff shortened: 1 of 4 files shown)


mxcubecore/HardwareObjects/BlissNState.py
```diff
@@ -88,7 +88,10 @@ def get_value(self):
             if self._prefix:
                 _attr = self._prefix + "_is_in"
                 _cmd = getattr(self._bliss_obj, _attr)
-                _val = _cmd()
+                if isinstance(_cmd, bool):
```
rhfogh, line 91: Slightly horrible - but if it has to be it has to be.

rhfogh (approved):
Not tested, but looks fine.

mxcubecore/HardwareObjects/BlissNState.py
beteva: You are right, but I do not know of other way to distinguish between outcome of a property and execution of a command. The trick with the prefix is because we have few of those actuators and wanted to have a generic method, rather than one per device. If you can think of a more elegant solution, I'll be glad to know it.

beteva: Thanks for having a look

rhfogh: Actually I cannot think of anything else, and I am sure this is the best possible. If you cannot refactor things upstream so it is the same everywhere, and I strongly suspect you cannot  (or at least that it is not worth doing), then this  is be the right solution. I should have sounded less negative in my first comment, sorry.

beteva: Thank you @rhfogh. I do not think it is negative and I am always happy to see people suggesting better coding.

marcus-oscarsson:
:), thanks @beteva and @rhfogh

PR #669, opened 2022-01-25
Update mep01.md

rhfogh:
The new section covers interpretation of version numbers and a tighter specification of what gets bumped when. It follows the rest of the document and PEP440, but adds extra considerations and should be considered as a proposal for discussion

### DIFF
--- a/doc/MEPS/MEP-01/mep01.md
+++ b/doc/MEPS/MEP-01/mep01.md
@@ -60,7 +60,7 @@ Goals & considerations
 
 The following lists the main goals taken into consideration for this proposal:
 
-1. A release schema have to provide a clear versioning of the code at different stages.ç
+1. A release schema have to provide a clear versioning of the code at different stages.
    This will facilitate the identification of the code used at any time.
 
 2. Each release should provide a comprehensive list of changes with all additions and 
@@ -203,21 +203,24 @@ recommended list: `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed` and `Secu
 
 The naming practice follows directly from the use of [PEP440](https://www.python.org/dev/peps/pep-0440/) 
 and the description in the preceding sections, but some interpretation might be helpful: 
-All version numbers are with reference to a final, public release. Thus e.g. versions 1.2.0.dev4, and 
+All version numbers refer to the version of the *next* public release. Thus e.g. versions 1.2.0.dev4, and 
 1.2.0.rc2 are both part of the preparations *preceding* release 1.2.0. Similarly version 1.2.1.dev1 
 is part of the *preparation* of the (patch) release 1.2.1, and thus is derived from release 1.2.0.
 This should make it clear when and how we need to bump versions.
 
 ### Local versions
 
-We may not want to give explicit version numbers to local beamline-specific code, 
-but it would be useful to agree on the correct way to do it, even so. Following
+Local beamline-specific code is essentially a matter for the beamline, and is not
+expected to appear on github. It would still be useful to have a purely optional
+naming convention for beamline-specific branches, for the case when beamlin-specific
+code is shared with others (e.g. workflow developers). Following
 PEP440, local versions must conform to the API of the official release they match, 
-and are indicared by a suffix starting with '+'. I would recommend that we agree
+and are indicated by a suffix starting with '+'. I would recommend that we agree
 on starting the local version identifier with a beamline identifier. Thus version 
-1.2.1+ESRF.3 would be an ESRF-specific version that conformed to the 1.2.1 API except
-for beamline-specific code (which we have agreed is not considered when deriving 
-version numbers).The last field in the identifier (the '3') is strictly speaking 
+1.2.1+ESRF.3 would be an ESRF-specific version that conformed to the 1.0 API and 
+the 1.2 feature set and incorporated patches up to 1.2.1. Beamline-specific code
+coudl be modified at will, since we have agreed that this is not considered when deriving 
+version numbers. The last field in the identifier (the '3') is strictly speaking 
 free text, so beamlines can deal with it as they wish, though they might find it 
 practical to include some kind of consecutive numbering. 
 
@@ -227,12 +230,12 @@ I would propose we should follow the following sequence when we make a release.
 This should pretty much match the diagram above.
 Supposing we want to make a release from version 1.1.0.dev12:
 
-- First push version 1.1.0.dev12 to the release branch under the name 1.1.0rc0
+- First push version 1.1.0.dev12 to the release branch under the name 1.1.0.rc0
 
 - Then bump the version of the development branch to 1.2.0.dev0. 
   Future development will be happening under this tag.
 
-- After finishing with the release candidates (say we are at 1.1.0rc2) 
+- After finishing with the release candidates (say we are at 1.1.0.rc2) 
   we push the candidate to the master branch as 1.1.0, and merge the 
   accumulated changes into the development branch, incrementing the final
   number to (e.g.) 1.2.0.dev1. 
@@ -245,13 +248,13 @@ Supposing we want to make a release from version 1.1.0.dev12:
   
 - PEP440 and the first version of this document are not quite clear (IMHO)
   on how we should handle patch levels. I would propose that increases in patch
-  level are only done by forking directly from a release, e.g. from 1.2.0 to 1.2.1rc0,
-  and further proceding as for other release candidates. This makes the system
-  a lot simpler, and frees us from having to consider in the development branch 
-  whether we are planning to add new features or not.
+  level are only done by forking directly from a release, e.g. from 1.2.0 to 1.2.1.dev0,
+  and that the final 1.2.1.devn would then be merged back to the main branch as 1.2.1.
+  This makes the system a lot simpler, and frees us from having to consider in the 
+  development branch whether we are planning to add new features or not.
   
 I have assumed that we want to start new branches at patch level 0. This seems 
-sensible (to me), but we can opt for starting at .1 if we so pr3efer.
+sensible (to me), but we can opt for starting at .1 if we so prefer.
  
   
 License
--- a/doc/MEPS/MEP-01/mep01.md
+++ b/doc/MEPS/MEP-01/mep01.md
@@ -197,7 +197,63 @@ The changelog file will be updated manually at each merge commit in different si
 All changes in a release will be grouped using the most appropriate category from the 
 recommended list: `Added`, `Changed`, `Deprecated`, `Removed`, `Fixed` and `Security`.
 
+## Version numbers and their interpretation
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks for having a look at this, it seems like we are starting to get somewhere :)

doc/MEPS/MEP-01/mep01.md
```diff
@@ -197,7 +197,63 @@ The changelog file will be updated manually at each merge commit in different si
...
+The naming practice follows directly from the use of [PEP440](https://www.python.org/dev/peps/pep-0440/) 
+and the description in the preceding sections, but some interpretation might be helpful: 
+All version numbers are with reference to a final, public release. Thus e.g. versions 1.2.0.dev4, and 
+1.2.0.rc2 are both part of the preparations *preceding* release 1.2.0. Similarly version 1.2.1.dev1 
+is part of the *preparation* of the (patch) release 1.2.1, and thus is derived from release 1.2.0.
+This should make it clear when and how we need to bump versions.
+
+### Local versions
+
+We may not want to give explicit version numbers to local beamline-specific code,
```
marcus-oscarsson, line 213: I'm not sure that adding these local version identifiers would be of alot benefit ?  The local code will not be local once its merged with the public repository, so I guess it would only have a meaning locally.

What you propose seems reasonable if we agree that we should have explicit version numbers for local code but I think it decides a bit to much about internal working routines.

doc/MEPS/MEP-01/mep01.md
```diff
@@ -197,7 +197,63 @@ The changelog file will be updated manually at each merge commit in different si
...
+- Then bump the version of the development branch to 1.2.0.dev0. 
+  Future development will be happening under this tag.
+
+- After finishing with the release candidates (say we are at 1.1.0rc2) 
+  we push the candidate to the master branch as 1.1.0, and merge the 
+  accumulated changes into the development branch, incrementing the final
+  number to (e.g.) 1.2.0.dev1. 
+  
+- Further development will be happening under 1.2.0.dev, increasing only 
+  the last number (to .dev2, .dev3, ...) no matter how many features are added.
```
marcus-oscarsson, line 241: Seems reasonable, so if I understand you right its then up to us in the community to decide when to bump the version. For instance during one of our monthly meetings so the bump in version will not in this sense strictly follow semantic versioning. 

I would be fine with that, I'm just trying to understand then what "event" would bump the version ?

doc/MEPS/MEP-01/mep01.md
```diff
@@ -197,7 +197,63 @@ The changelog file will be updated manually at each merge commit in different si
...
+  
+- When we decide for changes in the major version we bump the major version at
+  the start of the process (2.0.0.dev0).
+  
+- PEP440 and the first version of this document are not quite clear (IMHO)
+  on how we should handle patch levels. I would propose that increases in patch
+  level are only done by forking directly from a release, e.g. from 1.2.0 to 1.2.1rc0,
+  and further proceding as for other release candidates. This makes the system
+  a lot simpler, and frees us from having to consider in the development branch 
+  whether we are planning to add new features or not.
```
marcus-oscarsson, line 251: Yes, we could try that. It sounds less complicated :+1:

marcus-oscarsson:
The only part I'm not sure about is the section about local versioning. I think the section can be kept but we might perhaps consider to not adopt it right away ?

doc/MEPS/MEP-01/mep01.md
rhfogh: I was not suggesting that we should force people to use this, rather that we should have an agreed definition, that people could use if they wanted. Clearly people can do whatever they want internally, the local versions would not make it to the main repositories, and the local version identifiers only make sense if you are sharing or using the local code outside the beamline. But, then, I am, doing that continuously at the moment, adapting workflow code to local branches. So, I'd say, if you make a version tagged e.g. 1.3.2+ESRF.something you would be saying that this is local ESRF code, conforming to the 1.3 API, derived from 1.3.2, with possibly major ESRF-specific additions. If such things were available it would be useful for me to have them, but it would be entirely up to people if they wanted to produce or share such versions.

doc/MEPS/MEP-01/mep01.md
rhfogh: As I understand it, the version numbers on development code would correspond to the version of the *next* release, so once you made release 1.2.1 that tag would be retired, and the development versions would have to be numbered either 1.2.2, 1.3.0, or 2.0.0 from then on. You would do the bump immediately on release, in anticipation of future changes. If you are making a new major release you would know it, and I think it would make sense to start 3.0.0dev in parallel with 1.3.0dev.  If you just keep developing you might not know whether the next release would be 1.2.2 or 1.3.0. When I wrote the  PR I thought that we would just assume we always added a feature, and then bump to 1.3.0 straight away. Thinking about it, it might be better if we automatically bumped to 1.2.2 on release, and then branched and bumped to 1.3.0 the first time we added a new feature. We would then be free to decide whether we wanted to actually make the 1.2.2 release or close that branch down and go straight to 1.3.0

doc/MEPS/MEP-01/mep01.md
rhfogh: Or maybe not. See comment above.

doc/MEPS/MEP-01/mep01.md
marcus-oscarsson: I see, that's a bit how it interpreted it. I think the discussion is good to have regardless if we adopt a versioning schema for local code or not.

doc/MEPS/MEP-01/mep01.md
marcus-oscarsson: > As I understand it, the version numbers on development code would correspond to the version of the _next_ release, so once you made release 1.2.1 that tag would be retired, and the development versions would have to be numbered either 1.2.2, 1.3.0, or 2.0.0 from then on. 

That makes sense :)

>You would do the bump immediately on release, in anticipation of future changes. If you are making a new major release you would know it, and I think it would make sense to start 3.0.0dev in parallel with 1.3.0dev. If you just keep developing you might not know whether the next release would be 1.2.2 or 1.3.0. When I wrote the PR I thought that we would just assume we always added a feature, and then bump to 1.3.0 straight away.

That's an assumption I did as well, but It seems to me that it will make it quite complicated when one wants to
go back and do fixes for that feature and the work on next minor release already have started. 

>Thinking about it, it might be better if we automatically bumped to 1.2.2 on release, and then branched and bumped to 1.3.0 the first time we added a new feature. 

Or we simply collectively decide when to bump the minor number, for instance for a set of features.

>We would then be free to decide whether we wanted to actually make the 1.2.2 release or close that branch down and go straight to 1.3.0

I see, something In that direction would perhaps be a good idea. Thinking about a bit more and reading your comments. I think that maybe semantic versioning makes things a bit too complicated and that it might be a good idea to consider something lighter (at-least to begin with). Its maybe the patch level that makes things a bit too tedious.

How about something like this: 
  
     - We bump the patch/micro version for each PR
     - We bump the minor version collectively when we decide that its time to do so (for a set of features)
     - We bump the major version on breaking changes

It maybe makes things easier to handle as there are less decisions of the individual developer to take while committing code ?

rhfogh: Your proposal is radically different from mine, I think. First I think mine is simpler:
- We bump the development branch automatically to 1.2.2dev0 when we  release 1.2.1
- We bump the minor version the first time we commit a feature-adding change (e.g. 1.2.2dev5 -> 1.3.0dev0)
- We make a separate branch when we start making breaking changes and bump the major version.
- Otherwise we bump the development suffix at each PR.

As I understand it, this *is* semantic versioning - it is just that it applies only to the releases, not to the development branches. We do not track if each development PR breaks with its predecessor, only if each release does. The question is do we want or need to track semantically distinct versions in unreleased code? Maybe we do, but we have to decide

As I understand your proposal, you want to get rid of the dev extension completely (breaking with PEP440) and it would go something like this:

- We release 1.2.0
- 1.2.1 Patch PR
- 1.2.2 Patch PR
- 1.2.3 Feature-add PR. At this point we are no longer complying with 1.2, but we have not collectively decide to bump the minor version yet
- 1.2.4 Patch PR - still not compliant with 1.2
- 1.3.0 Feature-add PR - this time we decide to bump
- 1.3.1 Patch PR
- 1.4.0 Feature-add PR and we decide to bump again.
- 1.4.1 Patch PR
At this point the last release is 1.2.0, and the tag on the development version is 1.4.1. If we want to make a release now, what do we call it? 1.3.0? 1.4.1?

We can certainly do something like this if we want to track the semantic versions of development code. But if so I think we should seriously consider whether we want to allow feature-adding PRs that do not bump the minor version.

marcus-oscarsson: Hm, I thought what I suggested was simpler because it removed the step of deciding whatever your change was feature adding or not. Thus the procedure to commit code would more or less be identical each time, I somehow perceived this as being seen an obstacle.

What you describe above is otherwise compatible with semantic versioning which was also what we wanted to begin with and I would personally be happy to give it a go

rhfogh: As usual I may have been too combative. We could certainly go for a situation where we did track semantics from one development to the next. Or we could simply bump the minor version after each release, and reserve patch releases to specific patches (as in my first draft). The most important thing for me would be to work through something like the example I gave above and say explicitly what got to be bumped when,according to the proposed schema.

marcus-oscarsson: I agree, that it would be very nice to have a clear example of when to bump version. I think the example you have above is good. We should maybe circulate it among the others or raise it on our next meeting. One of the things that Id think would be nice is to decrease the "cognitive overload" in the everyday work, meaning making the work against the develop branch as simple as possible.

I'm as I mentioned happy with the routine you outline above, id invite the rest to also give their opinion and then maybe to discuss this example on our next meeting ?

marcus-oscarsson: @beteva @hellmig @MartinSavko @meguiraun @meghdadyazdi (and the ones I forgot :) ) : What do you think ?

doc/MEPS/MEP-01/mep01.md
```diff
@@ -197,7 +197,63 @@ The changelog file will be updated manually at each merge commit in different si
...
+  
+I have assumed that we want to start new branches at patch level 0. This seems 
+sensible (to me), but we can opt for starting at .1 if we so pr3efer.
```
marcus-oscarsson, line 254: prefer instead of ` pr3efer`

marcus-oscarsson:
Could you make this PR against the `develop` branch ?

rhfogh:
Replaced by PR to development branch

PR #671, opened 2022-02-20
Rhfogh master

rhfogh:
Important bugfix in QueueManager. execute_entry assumed that queue ended when the function ended, but workflows execute entries from within a running queue. This messed up the UI behaviour for 'RunQueue' v. 'Stop' buttons. The rest is essentially GPhL-specific


rhfogh:
Replaced by PR to develop

PR #676, opened 2022-02-25
Contributing guidelines -  Language and spelling

marcus-oscarsson:


hellmig:
Hi Marcus, hi All,

Since we are an (still growing ..) international collaboration and American English is in my perception more widely used I would suggest to leave the decision whether to use AE or BE for code, comments or documentation in the responsibility of the respective developer.

Best, Michael

marcus-oscarsson:
We had a discussion regarding this during our last meeting and the consensus then was to use British English. Not everybody was present at the meeting so  we can have another round of discussion regarding this if needed.

So please let us know what you think regarding this in this PR :)

rhfogh:
I would move quite strongly for standardising the spelling. Which one we choose is less important (though I have my preference too). If you have both 'centring' and 'centering', 'characterisation' and 'characterization' it is a source of bugs. You never know which spelling you should use, and you risk ending up with parallel functions that do more or less the same thing but have differently spelled names. All hail standardisation (with an 's')!

rhfogh:
Time to decide? I move we accept the PR.

marcus-oscarsson:
Yes very much so

PR #678, opened 2022-02-25
Contributing guidelines - git flow model

marcus-oscarsson:


rhfogh (commented):
Negative review, pending an explanation.

marcus-oscarsson:
I'ts WIP not completely pushed yet.

rhfogh:
Ah sorry - I missed that.

marcus-oscarsson:
OK :)

CONTRIBUTING.md
```diff
@@ -73,6 +76,26 @@ Pull request (PR) is the most convinient way of submitting a new code to the rep
 * The last reviewer to review the PR have the responsibility of merging it.
 * A PR that has no reviewer can be approved and merged by anyone.
 
+### Versioning 
+Versioning is partly automated by GitHub actions and the software called [bump2version](https://github.com/c4urself/bump2version) and based on the gitflow braching model:
+
+- Each new feature is implemented in a `feature branch`, branching from the `develop branch`.
+
+- The merge of a `feature branch` is made via PR to the `develop branch`. The author of 
+  the PR must solve any conflicts with the latest development version before the merge.
```
rhfogh, line 85: Should we re-emphasize here that you should use rebase rather than merge, to keep the log clean? See https://www.endoflineblog.com/gitflow-considered-harmful for a relevant discussion.

rhfogh (approved):

CONTRIBUTING.md
marcus-oscarsson: Sure why not, I'm not a fan of the merge commits myself

CONTRIBUTING.md
```diff
@@ -17,28 +17,31 @@ If the corresponding issue do not exist then:
...
+This means that all pull requests should be made against the [**develop**](https://github.com/mxcube/mxcubecore/tree/develop) branch. The work on the **develop** branch is performed by simply creating a branch for the work to be done and then making a PR according to the description below.
 
-* Fetch all branches and merge upstream to your forked master:
+* To fetch all branches and merge upstream to your forked develop branch:
   ```bash
   git fetch --all
-  git checkout master
-  git merge upstream/master
+  git checkout develop
+  git merge --no-ff upstream/develop
```
marcus-oscarsson, line 36: Added something about using --no-ff no fast forward is this what you meant and is it enough @rhfogh ? 

What do you guys prefer --no-ff or rebase ?

marcus-oscarsson: I finally changed my mind and I propose to use rebase instead of --no-ff

CONTRIBUTING.md
```diff
@@ -33,7 +33,7 @@ This means that all pull requests should be made against the [**develop**](https
   ```bash
   git fetch --all
   git checkout develop
-  git merge --no-ff upstream/develop
+  git rebase upstream/develop
```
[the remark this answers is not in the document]
rhfogh: 1) I agree with favouring rebase. If the merge commit does not add anything, better not have it (so you do not have to wonder).
2) I think it is important enough to be worth saying twice.  

In practice it will sometimes be necessary to merge, this is a recommendationn, not a straitjacket. As I personally use pycharm for my git handling (recommended), I cannot comment on the command wording.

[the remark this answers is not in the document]
rhfogh: Actually, you might be a bit more nuanced - though I do not know if that makes for good guidelines. A bit loosely:

When you are moving commits from develop to feature branches or from mxcube to your local copy, you should definitely use rebase, if possible. That means the commits in the feature branch are all actually relevant for the feature. When you are moving commits from a feature branch or your local copy into develop you should probably merge. Whether to use no-ff or not  I am not sure, maybe not - after all your PR will show which commits belong together. Moving from RC to master should be  a merge, I guess, Moving from RC to develop I am not sure. Maybe rebase - that would give a consistent picture where you rebased when moving data towards less authoritative versions, and merge when you moved towards more authoritative versions.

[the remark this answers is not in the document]
marcus-oscarsson: Hm, I agree that there might be some advantages to being more nuanced. I think that we should avoid the merge commits as far as possible but also that you are right that there are times when they can be meaningful. 

However, for the time being, I would advocate to always use rebase on develop and feature branches as far as possible. Then we will see if that works out as we like and if people are happy with it or want to change something. What do you think ?

I've added some clarification about using rebase instead of merge.

rhfogh:
@pkeller pointed me to an alternative to gitflow that I think we should consider: [oneflow](https://www.endoflineblog.com/implementing-oneflow-on-github-bitbucket-and-gitlab). It is not a big change - in fact it is arguably what we were trying to do already - but has a lovely precise description of what should be done and why, down to recommending exact git commands in each case. I think this would be really helpful to adopt this. I could make another PR, but it makes more sense if we look at this and discuss it.

A lightning summary:
- A master branch that is *only* used to keep the successive releases. 
- A permanent develop branch with a linear history
- Feature, release, and hotfix branches as already discussed. 
which is all pretty much as already agreed.

I would add a couple of rules:
- you never, ever, rebase the master or develop branches. 
- Release branches should be rebased on to develop before being merged into develop (if at all possible).

In terms of oneflow we would choose the variant with two branches, which is mentioned as an alternative in the link. 
The main choice remaining is exactly how to do the merging of feature branches into develop:
- I would recommend "– rebase / merge --no-ff", which is discussed in the link as an alternative (sort of #option 4)
  This always creates a merge commit, but rebases the feature branch on develop first
- Alternatively we could use "– rebase / merge --ff-only"
  This adds the feature branch commits to develop, avoiding any merge commits (when possible)
- Or we could use "– merge --squash", which squashes the feature branch commits into a single commit before merging.

marcus-oscarsson:
I agree that there are improvements to be done in the workflow we are using, our Idea was to use Gitflow as a base and adapt it to our needs. Looking at Oneflow is a good idea and we can keep this as a point for discussion.

I would suggest that we keep what we have discussed so far for the moment and get what we have discussed so far into place inorder to let people get an opinion about it. We have to remember that we should keep this as simple as we can for all developers. 

It would perhaps be easier to discuss/digest your idea with a seperate PR that adds a section to this file outlining the exact steps (git commands) to take at which moment.

PR #679, opened 2022-03-01
Contributing guidelines, Oneflow model

rhfogh:
This is an alternative PR to #678

My apologies for introducing so big changes so late in the process. @pkeller introduced me to the _oneflow_ workflow, and I thought this would be simper to use and better defined than what we had before. It is very similar to what we have already discussed, except for a much simpler versioning system, and precise procedures for the different actions.

The main points are:

- Using the master branch' explicitly for only storing releases, with the rest of the work done in the develop branch.

- Changed merge etc. instructions to provide a simple and linear commit graph.

- Limited the versioning to actual releases, which means getting rid of the dev and rc suffixes and the beamline-specific tags. This means that we only need to bump versions as part of the release process. We lose the ability to tag and refer to specific commits outside of the git repository - but how often would we need to refer specifically to the code in 2.1.0.dev7 or 2.2.0.rc2?

Clearly we can consider these points in isolation - we do not have to adopt them all. 

Changes in CONTRIBUTING.md are taken directly and unchanged from #678. Adoption of this PR would require modifications elsewhere (change references form gitflow to oneflow), but first we need to decide if we actually want this PR.


doc/MEPS/MEP-01/mep01.md
```diff
@@ -83,179 +101,214 @@ goals:
...
-1.2.1+ESRF.3 would be an ESRF-specific version that conformed to the 1.0 API and 
-the 1.2 feature set and incorporated patches up to 1.2.1. Beamline-specific code
-coudl be modified at will, since we have agreed that this is not considered when deriving 
-version numbers. The last field in the identifier (the '3') is strictly speaking 
-free text, so beamlines can deal with it as they wish, though they might find it 
-practical to include some kind of consecutive numbering. 
+After the pull request is accepted and merged in, you should delete the feature branch
 
-### What versions to bump
+There are three ways to do the merges into develop, which presumably have to be set in
```
rhfogh, line 238: If we do adopt (more or less) this PR, this choice is probably the most open question to discuss. All three choices will fit with the rest, so it depends on how we want the commit graph to look.

doc/MEPS/MEP-01/mep01.md
```diff
@@ -83,179 +101,214 @@ goals:
...
+
+Release branches are named as proposed by oneflow. This is not teh only possibility, 
+but is a clear and visible way to signal that this is a release branch. It avoids 
+the 'rc0' etc. suffixes, since we are not versioning the individual commits.
+
+$ git checkout -b release/2.3.0 develop   # If you want a specific commit check out that instead
+
+$ git push origin release/2.3.0
+
+Now edit the release notes, and bump the internal version string. Be careful to note
```
rhfogh, line 269: When to bump the version string (and when to edit the release notes) could be done in different ways. This approach has the advantage that you only need to bump as part of the release process. It would be possible to bump the version immediately *after* making a release, so that the develop branch tag would reflect the *next* release version. The problem with this is that you do not know whether the next release will be breaking or not, or whether you will need to merge in hotfixes, and the name of the next tag obviously depends on that. The laternative version might force you to do much more work and sometimes to change teh internal version string, depending on how the releae process proceeded.

marcus-oscarsson:
It is indeed quite late in the process and this requires that we discuss this from the top. Id suggest that we therefore add this to the agenda for the next meeting so that the idea and the differences from our current suggestion can be presented and then discussed in detail.

Some questions in the meantime:

   -  How is the version string handled, it seems like we have to do both `git tag` and change certain files containing the version string. Which was handled with `bump2version` in our previous example
  
   -  In extension, how does this work with our Github Actions, did you test this with bump2version and the [test action repository](https://github.com/mxcube/test_action) ?

   -  Did you or Peter use this workflow before ?

rhfogh:
Good points. And you are quite right that this would need to be discussed at the next meeting

As for Github Actions, I do not think I am competent. I should be able to contribute on what kind of flow we want to achieve (hopefully ;-) ) but with no experience of setting up or using GitHub commands it would take quite a while before I could come up with anything practical on how to do it, let alone compare with more experienced people.

Neither I nor Peter have experience of using this workflow before. But then, my impression was that Jordi made his proposal from literature references and that nobody has much experience in using GitFlow either. 

We could use bump2version to handle both the internal version string and the git tag. The reason I proposed otherwise was that I had the impression (from the Oneflow blog) that bump2version would tag first and change the version string afterwards, so that you would still need to commit afterwards, and so the version with the internal tag committed would not be the one that was tagged. But maybe this is incorrect, or there is a simple way of handling it. 

I can see two possible advantages in this PR. One is the simpler version system which should be easier to set up and use, and would avoid a number of bumps and edge cases - if we decide this is enough. The other is having a short and precise set of procedures, which should bring clarity if nothing else. That is something that could be provided equally well in the GitFlow system. I apologise if you think it is already there and I have missed it.

marcus-oscarsson:
> Good points. And you are quite right that this would need to be discussed at the next meeting
> 
Okey, so let's try to take a step back and discuss this on our next meeting. The discussion can in the meantime continue here and I will close my other PR.

> As for Github Actions, I do not think I am competent. I should be able to contribute on what kind of flow we want to achieve (hopefully ;-) ) but with no experience of setting up or using GitHub commands it would take quite a while before I could come up with anything practical on how to do it, let alone compare with more experienced people.
> 
I see, we will have to investigate that a bit closer then.

> Neither I nor Peter have experience of using this workflow before. But then, my impression was that Jordi made his proposal from literature references and that nobody has much experience in using GitFlow either.
>
My impression was that Jordi was used to this workflow but it's true that the rest of us were not. I will repeat that I think it's important that we find a way of working that suits our needs and are as little as possible in the way. We just want to commit
code and keep track of versions.  We don't create release very often so we can afford to adapt our method once we have something basic in place.

> We could use bump2version to handle both the internal version string and the git tag. The reason I proposed otherwise was that I had the impression (from the Oneflow blog) that bump2version would tag first and change the version string afterwards, so that you would still need to commit afterwards, and so the version with the internal tag committed would not be the one that was tagged. But maybe this is incorrect, or there is a simple way of handling it.

I think that is correct, the commit would still be needed.

> 
> I can see two possible advantages in this PR. One is the simpler version system which should be easier to set up and use, and would avoid a number of bumps and edge cases - if we decide this is enough. The other is having a short and precise set of procedures, which should bring clarity if nothing else. That is something that could be provided equally well in the GitFlow system. I apologise if you think it is already there and I have missed it.

That is certainly good points, would you mind presenting a summary of this during our next meeting ?

rhfogh:
I can certainly present something.

marcus-oscarsson:
I thought it was unnecessary to have both as its easier to discuss everything in one place.This PR is also based on #678 so it already contains parts of it so I thought it was sort of a duplication to have both. I'd be happy to open the old PR again if its needed.

PR #680, opened 2022-03-23
Exposing type hinted attributes for UI

marcus-oscarsson:
Adds a feature that allows any type hinted member of a HardwareObject to be exported to the UI. 

A function written with type hints for instance;

`    
def my_fancy_function(self, speed: float, num_images:int, exp_time:float, phase:PhaseEnum) -> bool:
        return True
`

can now be exported so that a default UI can be displayed to the user, once validated the data entered
are passed to member in question.

Example of UI for the example below, illustrating the feature.
md_th

This feature is currently used in two places, it replaces the current beamline actions in MXCuBE 3
sot that they can be expressed as objects with type hinted members. It further makes it possible to
create simple maintenance UIs like the example above.

The type annotated beamline actions currently handles return data and errors like the previous
beamline actions. Error handling and return data for the the pure HardwareObject version still needs
to be implemented.

The feature itself is UI agnostic, its up the the UI-Application to display the exported information and
ultimately call ` execute_exported_command` once the input is validated.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -21,10 +21,13 @@
 
 from __future__ import absolute_import
 
+import typing
+import ast
 import enum
 from collections import OrderedDict
 import logging
 from gevent import event, Timeout
+import pydantic
 
 from mxcubecore.dispatcher import dispatcher
 from mxcubecore.CommandContainer import CommandContainer
@@ -680,6 +683,63 @@ class HardwareObject(HardwareObjectNode, HardwareObjectMixin):
         HardwareObjectMixin.__init__(self)
         self.log = logging.getLogger("HWR").getChild(self.__class__.__name__)
         self.user_log = logging.getLogger("user_log_level")
+        self.__exports = {}
+        self.__pydantic_models = {}
+        self._exported_attributes = {}
+        self._exports_config_list = []
+
+    def init(self):
+        self._exports_config_list.extend(
+            ast.literal_eval(self.get_property("exports", "[]").strip())
+        )
+
+        self.__exports = dict.fromkeys(self._exports_config_list, {})
+
+        if self.__exports:
+            self._get_type_annotations()
+
+    def _get_type_annotations(self):
+        _models = {}
+
+        for attr_name, _ in self.__exports.items():
+            self._exported_attributes[attr_name] = {}
+            self.__exports[attr_name] = []
+            self.__pydantic_models[attr_name] = {}
+            fdict = {}
+
+            for _n, _t in typing.get_type_hints(getattr(self, attr_name)).items():
+                # Skipp return typehint
+                if _n != "return":
+                    self.__exports[attr_name].append(_n)
+                    fdict[_n] = (_t, pydantic.Field(alias=_n))
+
+            _models[attr_name] = (
+                pydantic.create_model(attr_name, **fdict),
+                pydantic.Field(alias=attr_name)
+            )
+
+            self.__pydantic_models[attr_name] = _models[attr_name][0]
+            self._exported_attributes[attr_name]["signature"] = self.__exports[attr_name]
+            self._exported_attributes[attr_name]["schema"] = self.__pydantic_models[attr_name].schema_json()
+
+
+        model = pydantic.create_model(self.__class__.__name__, **_models)
+        self.__pydantic_models["all"] = model
+
+    def execute_exported_command(self, cmd_name, args):
+        if cmd_name in self.__exports.keys():
+            cmd = getattr(self, cmd_name)
+            cmd(**args)
+        else:
+            self.log.info(f"Command {cmd} not exported, check type hints and configuration file")
+
+    @property
+    def pydantic_model(self):
+        return self.__pydantic_models
+
+    @property
+    def exported_attributes(self):
+        return self._exported_attributes
 
     def __getstate__(self):
         return self.name()
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -434,6 +434,24 @@ class HardwareObjectMixin(CommandContainer):
         # Internal object-specific state attribute, used to check for state changes
         self._specific_state = None
 
+        # Dictionary on the form:
+        # key: The exporterd member function name
+        # value: The arguments of the exported member
+        self._exports = {}
+
+        # Dictionary containing list Pydantic models for each of the exported member
+        # functions arguemnts. The key is the member name and the value a list of the
+        # pydantic models.
+        self._pydantic_models = {}
+
+        # Dictionary on the form:
+        # key: The exporterd member function name
+        # value: dictionary on the form {signautre: [<arguments>] schema:<JSONSchema>}
+        self._exported_attributes = {}
+
+        # List of member names (methods) to be exported (Set at configuration stage)
+        self._exports_config_list = []
+
     def __bool__(self):
         return True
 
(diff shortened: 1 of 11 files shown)


mxcubecore/BaseHardwareObjects.py
```diff
@@ -21,10 +21,13 @@
 
 from __future__ import absolute_import
 
+import typing
+import ast
 import enum
 from collections import OrderedDict
 import logging
 from gevent import event, Timeout
+import pydantic
```
marcus-oscarsson, line 30: Pydantic models can be used with typehints handles constraints and allows for conversion to JSONSchema

mxcubecore/BaseHardwareObjects.py
```diff
@@ -680,6 +683,63 @@ def __init__(self, rootName):
...
+        self._exports_config_list = []
+
+    def init(self):
+        self._exports_config_list.extend(
+            ast.literal_eval(self.get_property("exports", "[]").strip())
+        )
+
+        self.__exports = dict.fromkeys(self._exports_config_list, {})
+
+        if self.__exports:
```
marcus-oscarsson, line 698: Augment the `exported` commands with typehint information so that the UI can ask for it

mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
```diff
@@ -386,3 +401,15 @@ def set_phase(self, phase, timeout=None):
 
     def get_point_from_line(self, point_one, point_two, index, images_num):
         return point_one.as_dict()
+
+    def abort(self) -> None:
```
marcus-oscarsson, line 405: Example of type hinted members that can be exported

mxcubecore/configuration/mockup/diffractometer-mockup.xml
```diff
@@ -18,4 +18,5 @@
   <sample_mount_mode>sample_changer</sample_mount_mode>
   <!-- MD3 -->
   <grid_direction>{"fast": (0, 1), "slow": (1, 0), "omega_ref": 0}</grid_direction>
+  <exports>["abort", "status", "my_fancy_function", "my_other_funny_function"]</exports>
```
marcus-oscarsson, line 21: Configuring this hardware object to export: "abort", "status", "my_fancy_function", "my_other_funny_function"

mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -0,0 +1,262 @@
+import sys
```
marcus-oscarsson, line 1: Added the possibility to use type hinted beamline actions by unifying the proposed feature with the already existing code: [ESRF BeamCmds.py](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/ESRF/BeamCmds.py).  The ESRF specific code can than be replaced but this more general BeamlineActions.

mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
```diff
@@ -386,3 +401,15 @@ def set_phase(self, phase, timeout=None):
...
+        return None
+
+    def status(self) -> str:
+        return "READY"
+    
+    def my_fancy_function(self, speed: float, num_images:int, exp_time:float, phase:PhaseEnum) -> bool:
```
marcus-oscarsson, line 411: A better example of a type hinted member

mxcubecore/BaseHardwareObjects.py
```diff
@@ -680,6 +683,63 @@ def __init__(self, rootName):
...
+
+            self.__pydantic_models[attr_name] = _models[attr_name][0]
+            self._exported_attributes[attr_name]["signature"] = self.__exports[attr_name]
+            self._exported_attributes[attr_name]["schema"] = self.__pydantic_models[attr_name].schema_json()
+
+
+        model = pydantic.create_model(self.__class__.__name__, **_models)
+        self.__pydantic_models["all"] = model
+
+    def execute_exported_command(self, cmd_name, args):
```
marcus-oscarsson, line 729: Once validated the data is passed back and the command is exeuted.

rhfogh (changes requested):
This is lovely. So good, in fact, that I think we need to discuss whether and how  Pydantic fits in more widely.

One problem. Could the Pydantic code not be put in HardwareObjectMixin rather than HardwareObject? Or, in other words, is there a technical reason why this can only be supported in xml-configured hardware objects, have we abandoned the possibility of moving to yaml configuration eventually,  or should this be available for both?

As I understand it, this has the specific limited scope of allowing UI-export of functions with type checking. That is very good, and the tool fits.

Pydantic looks so powerful that I think we need to discuss whether we could use it for data classes everywhere, or even for coding more generally. And here I nave a big problem with the Pydantic approach of silently coercing data, so that an int variable will transform "3" or 3.14 to integer 3. I am afraid that this might encourage people not to worry about what type they are passing in, and lead to tricky long term bugs. Quite likely there is no valid alternative to Pydantic. To a quick look there seems to be some tools for validation-of-input, but they seem to be somewhat cumbersome and demanding to use. 

From 'The Zen of Python':
Errors should never pass silently
Explicit is better than implicit

Should we talk about it?

marcus-oscarsson:
> This is lovely. So good, in fact, that I think we need to discuss whether and how Pydantic fits in more widely.
> 

I think it would be really nice if we could use typehints, data classes and possible Pydantic as well more widely. 
 
> One specific problem. Could the Pydantic code not be put in HardwareObjectMixin rather than HardwareObject? Or, in other words, is there a technical reason why this can only be supported in xml-configured hardware objects, have we abandoned the possibility of moving to yaml configuration eventually, or should this be available for both?

Ah, no there is no reason why it's implemented in HardwareObject instead of HardwareObjectMixin. I can definitely move this to HardwareObjectMixin it would make more sense as we would like to go towards YAML configured objects.

> 
> As I understand it, this has the specific limited scope of allowing UI-export of functions with type checking. That is very good, and the tool fits.

Yes this PR was intended for that specifically but yes it would be nice to think about how something like this could be used
more widely.

> 
> Pydantic looks so powerful that I think we need to discuss whether we could use it for data classes everywhere, or even for coding more generally. 

It's something that we have discussed before as well and I would be very happy to continue those discussions. I would love to see this for instance in a future version of queue_model_obects 

>And here I have a problem with the Pydantic approach of silently coercing data, so that an int variable will transform "3" or 3.14 to integer 3. I am afraid that this might encourage people not to worry about what type they are passing in, and lead to tricky long term bugs. Quite likely there is no valid alternative to Pydantic. To a quick look there seems to be some tools for validation-of-input, but they seem to be somewhat cumbersome and demanding to use.

I agree that the silent type coercion is something we would like to avoid for some of our usecases. I think there is an ongoing discussion in the Pydantic project about this and it seems like they might implement some sort for strict mode. In the meantime I think we could use the `Strict*` types to avoid the type coercing, [see](https://pydantic-docs.helpmanual.io/usage/types/#strict-types). I think we could eventually add our own enough `Strict` types if we are missing some. Do you think that would be enough ?

> 
> From 'The Zen of Python': Errors should never pass silently Explicit is better than implicit
> 
> Should we talk about it?

Yes, we can discuss the more wider use of this approach on our next meeting.  I'd be very happy to continue with this specific feature in the meantime though.

rhfogh:
There is indeed an ongoing discussion on Pydantic about silent coercion - and the unanimous and loud opinion of the people who run Pydantic from 2017 to today seems to be that silent coercion is a FEATURE! If you pass 3.14 to a string attribute you get "3.14"; if you pass it to an int attribute you get 3; this is as it should be and it is not going to change. 

You can define your attributes as strict, but 1) that is more work, 2) I notice that StrictFloat does not accept 1 as an input - 1 is an integer and not a float. That is also horrible.

To be honest, this feature of Pydantic fills me with horror. I mean it. On the other hand it looks beautiful as a way to handle serialisation, and a quick Google search suggests that there is nothing else that could fill that slot as well. The nearest alternative for in-code use would seem to be DataClasses, but they seem to lack the serialisation support. 

I wonder what to do:

- Use pydantic, but limit it rigidly to functions that are supposed to be accessed remotely with JSON input?
- Do some kind of custom pydantic implementation, maybe with our own AlmostStrictFloat class? Not very transferable. 
- Use Pydantic BaseModel where you need the JSON support, but Pydantic Dataclasses (which support validators) for more general specification - if I have got it right, a pydantic dataclass does not do the coercion. It just does not do the type checking either, unless you put that in by hand. Clean, but more work.
- - Give up and just use Pydantic and accept its behaviour (shudder)?

You must surely have looked at this in detail, Marcus - what do you think?

marcus-oscarsson:
> There is indeed an ongoing discussion on Pydantic about silent coercion - and the unanimous and loud opinion of the people who run Pydantic from 2017 to today seems to be that silent coercion is a FEATURE! If you pass 3.14 to a string attribute you get "3.14"; if you pass it to an int attribute you get 3; this is as it should be and it is not going to change.
> 

It indeed seems to be the case, I wonder a bit why people prefer that behaviour. I guess its consistent with the way Python already handles types passing; 3.14 to str() makes it the string "3.14" and passing it to int makes it 3. I agree that losing information without a warning is never a good thing, especially if it passes a validation step.

> You can define your attributes as strict, but 1) that is more work, 2) I notice that StrictFloat does not accept 1 as an input - 1 is an integer and not a float. That is also horrible.

That's true but you need to define your object at some point and using Float or StrictFloat would not mean more work in that respect but I agree that its less elegant than simply using Float. I imagine that we could implement our own Float that simply accept the integer 1 as input, on the other hand something strictly being float somehow also means that its exactly that. In general I guess it would be possible to define the types where we think it's important to get a warning when information is lost.

> 
> To be honest, this feature of Pydantic fills me with horror. I mean it. On the other hand it looks beautiful as a way to handle serialisation, and a quick Google search suggests that there is nothing else that could fill that slot as well. The nearest alternative for in-code use would seem to be DataClasses, but they seem to lack the serialisation support.

I'm perhaps not as horrified as you by this and as you I have not found anything that handles this in a better way. I think data classes might suffer from the same problem if not worse because no checking at all is done (its after all only used as annotations), for instance consider the code:

`
from dataclasses import dataclass

@dataclass
class SomeDataClass:
    number: float
    string: str
    integer: int

> SomeDataClass("3.14", 3.14, 3.14)
SomeDataClass(number='3.14', string=3.14, integer=3.14)
`
> 
> I wonder what to do:
> 
> * Use pydantic, but limit it rigidly to functions that are supposed to be accessed remotely with JSON input?
That is, as you understand, how it's currently being used and I think that is already very nice.

> * Do some kind of custom pydantic implementation, maybe with our own AlmostStrictFloat class? Not very transferable.
> * Use Pydantic BaseModel where you need the JSON support, but Pydantic Dataclasses (which support validators) for more general specification - if I have got it right, a pydantic dataclass does not do the coercion. It just does not do the type checking either, unless you put that in by hand. Clean, but more work.

I think we need to take a closer look at exactly where and how we would like to use these features, I'm sure we can find a good alternative after some discussion :). We can add the point to the agenda for one of our coming developer meetings, the next one is quite full already though.

Pydantic dataclasses with validators might indeed be a solution for data models like the ones in queue_model_objects.

> * Give up and just use Pydantic and accept its behaviour (shudder)?
> 
> You must surely have looked at this in detail, Marcus - what do you think?

I've come to more or less the same conclusions as yourself, I've not really found anything better. The models can be used as typehints, there is data class integration with validation which is lacking in the standard data classes and you can easily serialize the data.

marcus-oscarsson:
I would finally suggest that we continue the discussion regarding the general usage of DataClasses/Model objects using for instance Pydantic on one of our upcoming developers meetings. And, that we can continue the work with this PR that concerns the the creation/usage of Pydantic models for the specific use of generating UIs in the meantime. It might mean that the later needs to be updated depending on the outcome of the discussions.

What do you think ?

rhfogh (approved):

rhfogh:
@marcus-oscarsson I agree with your latest comment. This is too good to leave out, and there is no proper alternative. Let us use it for this purpose and discuss before we expand to other things.

marcus-oscarsson:
Wonderfull :)

PR #683, opened 2022-04-25
Bug fix when parsing reference objects

beteva:
Bug has been introduced (when?), so it was only possible to do self.get_roles(). This PR sets back the lost functionality self["xxx"].get_roless(), when reference objects are defined as:

```
<object class=MyObject>
   <xxx>
        <object role=aaa href=/my_reference_object1 />
        <object role=bbb href=/my_reference_object2 />
   </xxx>
</object>
```


marcus-oscarsson:
Very good thanks !

rhfogh (approved):
Fine.

PR #684, opened 2022-04-26
[WIP] Minor fixes

marcus-oscarsson:
Dear all, 

This PR contain some fixes that we had to apply to make things run smoothly with hardware.

 - Attempt to patch py4j in a different way that does not interfere with other parts of the code using socket
 - Fixed some return types and made some code more robust with try/catch and added log messages
 - Added missing get_motros method to GenericDiffractometer
 
This is still WIP as it needs to be confirmed that the socket patch also works for py4j

### DIFF
--- a/mxcubecore/Command/Sardana.py
+++ b/mxcubecore/Command/Sardana.py
@@ -28,7 +28,6 @@ from mxcubecore.dispatcher import saferef
 
 import gevent
 from gevent.event import Event
-from gevent import monkey
 
 try:
     import Queue as queue
@@ -55,8 +54,6 @@ try:
     import taurus
 except Exception:
     logging.getLogger("HWR").warning("Sardana is not available in this computer.")
-
-monkey.patch_all(thread=False, subprocess=False)
 
 
 __copyright__ = """ Copyright © 2010 - 2020 by MXCuBE Collaboration """
--- a/mxcubecore/Command/exporter/StandardClient.py
+++ b/mxcubecore/Command/exporter/StandardClient.py
@@ -20,10 +20,9 @@
 
 """ ProtocolError and StandardClient implementation"""
 import sys
+import socket
 import gevent
 import gevent.lock
-
-from mxcubecore.HardwareRepository import original_socket as socket
 
 __copyright__ = """ Copyright © 2019 by the MXCuBE collaboration """
 __license__ = "LGPLv3+"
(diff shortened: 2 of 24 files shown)


marcus-oscarsson:
@rhfogh , could you let me know how the Py4J patch worked ?

rhfogh (changes requested):
Regrettably the py4j patch does not work t all, and breaks the GPhL workflow. 
AS I understand it:
- The socket module is a single global object
- Patch and unpatch change this object in place. Making sure socket it unpatched when it is attached to the WorkflowConnection has no effect, since  what matters is that the socket module attached to the java gateway must be unpatched *when the ocde is executed*. 
- I tested patching and unpathing when the GPhL workflow staops and starts. This works, but it sounds very risky to turn gevent patching on and off in teh middle of a running program. It kight mess up threads, and/or break in teh case of errors.
- The ideal would be to have two different socket objects to use for code that needs patched rel. unpatched sockets. COuld we not try to make that to work?

rhfogh:
I just tried out an alternative fix with Olof. It works at first look, but we shall see. See mxcubecore, rhfogh fork, branch rhfogh_develop_py4jfix, GphlWorkflowCOnnection.py of you are interested.

marcus-oscarsson:
> Regrettably the py4j patch does not work t all, and breaks the GPhL workflow. AS I understand it:
> 
> * The socket module is a single global object
> * Patch and unpatch change this object in place. Making sure socket it unpatched when it is attached to the WorkflowConnection has no effect, since  what matters is that the socket module attached to the java gateway must be unpatched _when the ocde is executed_.
> * I tested patching and unpathing when the GPhL workflow staops and starts. This works, but it sounds very risky to turn gevent patching on and off in teh middle of a running program. It kight mess up threads, and/or break in teh case of errors.

I dont think this would work, for the reasons that you already mention.

> * The ideal would be to have two different socket objects to use for code that needs patched rel. unpatched sockets. COuld we not try to make that to work?

Yes, I think that is an alternative, but I tired this as the first attempt to do so did not work.  I'm sure we will work out something else

marcus-oscarsson:
> I just tried out an alternative fix with Olof. It works at first look, but we shall see. See mxcubecore, rhfogh fork, branch rhfogh_develop_py4jfix, GphlWorkflowCOnnection.py of you are interested.

What I might suggest is that Ill mark this as WIP for now,  you ad Olof try ouy your new idea and if that works for you I will apply that fix to this PR, test it and then we merge this.

marcus-oscarsson:
superseded by #688

PR #686, opened 2022-05-11
[WIP] Queue improvement

marcus-oscarsson:
This is still ongoing work but I thought that Id share it while it's still ongoing as its could be interesting to
discuss.

The main idea of this is to make the queue entries / tasks easier to manage by making the "data model"
more clearly defined and the task logic easier to express.

With this suggestion the queue_entry is defined in a single file see for [example](https://github.com/mxcube/mxcubecore/blob/mo-queue/mxcubecore/HardwareObjects/queue_entry/ssx_chip_collection.py)

The data model, based on Pydantic, is defined in the same file as the task logic. It's possible that we 
might need to enhance the model to fit our exact needs (especially for validation).

A new queue entry can simply made available by adding a single file, similar to the example above, and it
will be made available automatically and a generic input form presented if there are no custom form
available.

In more detail:

  -  queue_entry.py was split into smaller pieces

  - Each QueueEntry object is in its own file named [queue_entry_name]queue_entry
    snake cased name with a queue_entry suffix

  - The QueueEntry object is named [QueueEntryName]QueueEntry
    camel cased name with a QueueEntry suffix (or simply the camel cased filename)

  - Created queue_entry sub-module that automatically imports any *_queue_entry.py
    file in addition to the constants and base classed defined in base_queue_entry

  - Import of queue_entry submodule is more or less a drop in replacement of the
    previous queue_etnry.py

  - New queue entrie can simply be added by adding a new <name>_queue_entry.py file

  - Added an example of a queue entry using a Pydantic model


rhfogh:
This looks really good. It is a fairly big change, but should definitely be worth it.

While we are at it, could we split up the queue_model_objects file as well? Would it make sense to put the queue_model_object in the same file as the queue_entry, or should they be in separate files? Should the QMOs and the queue_entries be in the same directory? should there be a separate directory for each entry?  Just thinking aloud here.

marcus-oscarsson:
Yes, indeed queue models objects should then also be split into smaller pieces. This PR is as you can see in quite an early stage but I wanted to put it out for discussion (even if it's close to ideas that we have discussed before).

I have a few thoughts that I would like to discuss its;

- Validation of models

- Division of queue model objects

- Do we want to keep models and logic in the same file (My idea is the same file for models specific to the entry and a common models file for common models). 

- When doing this should we keep the old queue entries for backward compatibility or should we aim
  to convert all of them. My initial suggestion would be to keep the old ones, and use them with the
  old AbstractCollect (as Collect is the major consumer of these models) and then write a new AbstractCollect
  and use that with the "new style" queue entries.

What id suggest is that I will try to advance a bit more on this, and if you are interested or have ideas please don't hesitate to express them here

PR #690, opened 2022-06-09
Added AbstractXrayCentring and XrayCentringMockup

rhfogh:
Added two missing files

### DIFF
--- a/mxcubecore/HardwareObjects/QueueManager.py
+++ b/mxcubecore/HardwareObjects/QueueManager.py
@@ -381,7 +381,7 @@ class QueueManager(HardwareObject, QueueEntryContainer):
 
         if use_async:
             task = gevent.spawn(self.__execute_entry, entry)
-            task.link((lambda _t: self._queue_end()))
+            # task.link((lambda _t: self._queue_end()))
         else:
             self.__execute_entry(entry)
 
--- a/mxcubecore/HardwareObjects/mockup/XrayCentringMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/XrayCentringMockup.py
@@ -1,0 +1,67 @@
+#! /usr/bin/env python
+# encoding: utf-8
+# 
+# This file is part of MXCuBE.
+# 
+# MXCuBE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+# 
+# MXCuBE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+# 
+# You should have received a copy of the GNU Lesser General Public License
+# along with MXCuBE.  If not, see <https://www.gnu.org/licenses/>.
+"""
+"""
+
+from __future__ import division, absolute_import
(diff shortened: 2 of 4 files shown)


mxcubecore/HardwareObjects/mockup/XrayCentringMockup.py
```diff
@@ -0,0 +1,67 @@
...
+    def execute(self):
+        """Dummy execution. Adds new entries to queue, to serve as example.
+        """
+
+        queue_manager = self._queue_entry.get_queue_controller()
+
+        logging.getLogger("HWR").debug("Execute mock Xray centring ")
+
+        # Add Dummy task 1
+        dummy = queue_model_objects.DummyTask(delay=5)
```
marcus-oscarsson, line 48: :)

mxcubecore/HardwareObjects/QueueManager.py
```diff
@@ -381,7 +381,7 @@ def execute_entry(self, entry, use_async=False):
 
         if use_async:
             task = gevent.spawn(self.__execute_entry, entry)
-            task.link((lambda _t: self._queue_end()))
+            # task.link((lambda _t: self._queue_end()))
```
marcus-oscarsson, line 384: We have to solve this in another way, because this change will break the current `run_now` feature, which relays on that `_queue_end` is called.

rhfogh: Indeed. I had a mail ready to send, but got distracted. Here it is:

Let me walk through the details:

The problem we found at the test was that QueueManager,execute_entry called self._queue_end at the end of the spawned async task, if use_async=True. 

execute_entry calls from GphlWorkflow and XrayCentringMockyup have use_async=False (by default) so no problem here. There are only two calls that can set use_acync=True:

XMLRPCServer.queue_execute_entry_with_id calls execute_entry with a value for use_async passed into the function. 
components.queue.execute_entry_with_id calls execute_entry with use_async-True if a task index is s passed into the function. 

execute_entry_with_id is only called from queue_start (use_async=False) and from the execute_entry_with_id route (use_async=True)

My proposal would be to call queue_end explicitly from XMLRPCServer.queue_execute_entry_with_id and components.queue.execute_entry_with_id as needed, to put parameters in those functions to control when that happened, and to remove the call to _queue_end from the execute_entry function. But since I am not really clear when those (queue_)execute_entry_with_id happen or how they are used, I could not well propose how that should be organised. Could either of you take care of  that?

marcus-oscarsson: Hm, I think the best would be to change `execute` so that it takes an optional entry argument and if called with this optional entry it means "run_now". Basically by calling something that is equivalent to the snippet below instead of running the entire queue:

```
            task = gevent.spawn(self.__execute_entry, entry)
            task.link((lambda _t: self._queue_end()))
````

Then all code that uses `execute_entry` for "run now" has to call execute(entry).

It's only used in mxcube-web in one or two places, and it's done "manually" in the Qt UI by repeating the code similar to the on above.

rhfogh: Sounds fine with me, but maybe some additional checks would be  needed. 
What should happen if you call execute and the queue is already executing? An Error?
Also, when you call the new execute_entry presumably the queue should be executing already? Should that be checked? Maybe the setting and checking of _is_running etc. needs to be modified with this change?

marcus-oscarsson: Yes, that's a good idea we could throw a RuntimeError in those cases

PR #691, opened 2022-06-14
Moved logic that was in execute_entry to execute

marcus-oscarsson:
The execute method now runs the entire queue or a single entry as one "run" of the queue and manages the various states such as "running", "paused" and "stopped". It further assumes that the queue is not already running

The execute_entry method executes an entry while the queue is already running


rhfogh (approved):
I have not tested it - still struggling with getting my home version to run, but this looks very good.

PR #693, opened 2022-06-21
Release/1.0.0

marcus-oscarsson:
Prepared the release branch for version 1.0.0, added a simple way to "single source" the version number
and made some minor changes to the setup.py. The routine around managing the version number and
packaging will very likely evolve but this is atleast a simple start that allows us to release the first version(s).

One idea would to investigate the [Poetry utility](https://python-poetry.org/docs/), does anyone have
experience with this ?

This is consequently  **not**  intended as a pull request to `develop` but simple a way to visualize the
changes in the release branch in a simple way.

We will tag and merge this with `master` once approved.


requirements.txt
```diff
@@ -0,0 +1,21 @@
...
+pylint==2.14.4
+ruamel.yaml==0.17.21
+scipy==1.8.1
+suds-py3==1.4.5.0
+psutil==5.9.1
+gipc==1.4.0
+
+# GPhL: f90nml, mgen, and py4j are required for GPhL Workflow and emulation
+py4j==0.10.9.5
+mgen==1.2.0
```
rhfogh, line 20: mgen is no longer needed

setup.py
```diff
@@ -1,39 +1,51 @@
...
+    requirements = [x for x in requirements if not x.startswith('#')]
+
+    return requirements
 
+requirements = get_requirements(
+    os.path.join(Path(os.path.abspath(__file__)).parent.as_posix(), "requirements.txt")
+)
+setup_requirements = []
+tests_requirements = []
+extras_requirements = {'gphl': ['f90nml', 'py4j', 'mgen']}
```
rhfogh, line 31: mgen no longer needed.

setup.py
```diff
@@ -1,39 +1,51 @@
...
 
 entry_points = {
-        'console_scripts': console_scripts,
-        'gui_scripts': gui_scripts
-    }
+    'console_scripts': console_scripts,
+    'gui_scripts': gui_scripts
+}
+
+version = get_version(
```
rhfogh, line 40: Presumably there is a reason to do this (with get_version function) instead of just 
from mxcube.__version__ import __version as version
?

rhfogh (approved):

setup.py
marcus-oscarsson: Yes its because the module `mxcubecore` does is probably not installed when you run setup.py so you cant run from `from mxcubecore import`. Even if a relative import could work, one needs to be able to run setup.py from any directory.

rhfogh: Makes sense. Thanks.

PR #709, opened 2022-09-30
Improved detector tests and fixed a minor bug in AbstractDetector

marcus-oscarsson:


test/pytest/test_detector.py
```diff
@@ -45,3 +46,52 @@ def test_detector_atributes(self, test_object):
         ), "Detector hardware object is None (not initialized)"
         exp_time_limits = test_object.get_exposure_time_limits()
         has_shutterless = test_object.has_shutterless()
+
```
rhfogh, line 49: I do not think you can do this. You are assuming that the beam is perfectly centred if extrapolated to distance 0, but how do you know that? In reality it could drift by even quite a few pixels - and, as it happens, the current configuration of DetectorMockup would fail that test. Also, people might even move the detector off centre deliberately. Basically you are testing that get_beam_position has not been overridden, and that the metadata values bx and by that define the beam_position/distance relationship are exactly equal to half the detector width. I think this needs discussion.

rhfogh (changes requested):

test/pytest/test_detector.py
marcus-oscarsson: > I do not think you can do this. You are assuming that the beam is perfectly centred if extrapolated to distance 0, but how do you know that? In reality it could drift by even quite a few pixels - and, as it happens, the current configuration of DetectorMockup would fail that test. 

Not sure I really get what you mean with this, this test is simply testing that beam is in the center of the detector
assuming that the detector distance is at 0 and the beam is not configured to be "off centered", 

>Also, people might even move the detector off centre deliberately. Basically you are testing that get_beam_position has not been overridden, and that the metadata values bx and by that define the beam_position/distance relationship are exactly equal to half the detector width. I think this needs discussion.

Well we are testing that we read the configuration files and convert from pixels to mm, assuming that the beam
is configured to be in the center of the detector at distance 0. 

What you are after is to also read the configured beam center `bx` and `by` and include them in the test ?

How would you test it ?

marcus-oscarsson: I can understand that you get in to difficulties if you intend to override `get_beam_center` and add some other logic instead of interpolation. In that case I guess we need to modify the test make sure that the test for that alternate logic is executed instead of the default implementation ?

rhfogh: We seem to be talking past each other.

Surely the beam position is part of the instrument description? Whether it is in the centre of the detector or not on a physical beamline would depend on the setting of various mirrors and motors. This could be a question of drift, or people could deliberately have  set the beam position to be off centre; I gather that for native data collection the beam *should* ideally be a bit off centre, to avoid the same reflections being lost in the module gaps on different parts of the detector. Also, it seems to be a reasonably common occurrence that the beam position as reflected in the image files is not always a precisely correct reflection of what is actually on the images. 

Anyway, the point of these tests is to see if the code works. Whether the beam hits the detector centre or not is a question of the physical status of the machinery or, in mock mode, of the kind of physical situation you have configured the mock to reflect. 

Or, to put it another way, what you are actually testing is the relationship between two configuration values, to wit that
get_metadata()["bx"] == width / 2.0
and
get_metadata()["by"] == height / 2.0
Why is that part of the code testing? And what is the physical and scientific basis for introducing that constraint?

marcus-oscarsson: > We seem to be talking past each other.
> 
> Surely the beam position is part of the instrument description? Whether it is in the centre of the detector or not on a physical beamline would depend on the setting of various mirrors and motors. This could be a question of drift, or people could deliberately have set the beam position to be off centre; I gather that for native data collection the beam _should_ ideally be a bit off centre, to avoid the same reflections being lost in the module gaps on different parts of the detector. Also, it seems to be a reasonably common occurrence that the beam position as reflected in the image files is not always a precisely correct reflection of what is actually on the images.

That's correct, but I'm afraid you are loosing me with this, not sure entirely what your point is ?

> 
> Anyway, the point of these tests is to see if the code works. Whether the beam hits the detector centre or not is a question of the physical status of the machinery or, in mock mode, of the kind of physical situation you have configured the mock to reflect.
> 
> Or, to put it another way, what you are actually testing is the relationship between two configuration values, to wit that get_metadata()["bx"] == width / 2.0 and get_metadata()["by"] == height / 2.0 Why is that part of the code testing? And what is the physical and scientific basis for introducing that constraint?

The test tests that:
1. The regression logic works (I can add a few more points)
2. Values are read from configuration (with the assumption that the beam is in the center of the detector at distance 0)

Yes, it tests that `get_metadata()["bx"] == width / 2.0 ` and `get_metadata()["by"] == height / 2.0` with the current mockup configuration. But its not constraint in itself you can configure it differently but yest the assumption of the test will fail. We can modify the test to include a beam that is not in the center if that is what you are after ?

marcus-oscarsson: So I updated the test a bit and remove the assumption that the beam is in the center of the detector and reading it from configuration instead. I'm also testing more than two points. Not sure that it was what you were after.

rhfogh: The point of the tests, I thought, is that if they fail something is wrong,
So if the tests fail if get_metadata()["bx"] != width / 2.0 and get_metadata()["by"] != height / 2.0
we are saying that these values must be equal, and it is the duty of programmers to make sure that they remain equal. And I think that this is wrong.

It is true that this tests that the results follow the current mock configuration, but it should not do that. It should test that the results correctly reflect the configuration *whatever that happens to be*, not that they reflect the values we happen to have set on 20220930. So, if you want to check that the calculation is correct, you should compare test_object.get_beam_position(distance=0) with bx and by.
If that fails it means the calculation is wrongly executed. If the current test fails, it just means that the configuration is different from what you had expected (but still perfectly valid). 

If you want to enforce the constraints that bx == width/2.0 and  by == height/2.0, I think you should first argue why that is a scientifically valid constraint, and second put it in a separate test, Or, even better, get rid of bx and by altogether and simply use width/2.0 and height/2.0 in the formula. I do not think that would be the right thing to do, but if you do want to do it, that would be the way.

marcus-oscarsson: > The point of the tests, I thought, is that if they fail something is wrong, So if the tests fail if get_metadata()["bx"] != width / 2.0 and get_metadata()["by"] != height / 2.0 we are saying that these values must be equal, and it is the duty of programmers to make sure that they remain equal. And I think that this is wrong.

I would agree, it was just a simple way of providing some kind of basic assumption to test from.

> 
> It is true that this tests that the results follow the current mock configuration, but it should not do that. It should test that the results correctly reflect the configuration _whatever that happens to be_, not that they reflect the values we happen to have set on 20220930. So, if you want to check that the calculation is correct, you should compare test_object.get_beam_position(distance=0) with bx and by. If that fails it means the calculation is wrongly executed. If the current test fails, it just means that the configuration is different from what you had expected (but still perfectly valid).
> 
> If you want to enforce the constraints that bx == width/2.0 and by == height/2.0, I think you should first argue why that is a scientifically valid constraint, and second put it in a separate test, Or, even better, get rid of bx and by altogether and simply use width/2.0 and height/2.0 in the formula. I do not think that would be the right thing to do, but if you do want to do it, that would be the way.

I'm not sure if you find the tests better with the last changes ?

test/pytest/test_detector.py
```diff
@@ -45,3 +46,42 @@ def test_detector_atributes(self, test_object):
...
+        bx = test_object.get_metadata()["bx"]
+        by = test_object.get_metadata()["by"]
+        ax = test_object.get_metadata()["ax"]
+        ay = test_object.get_metadata()["ay"]
+
+        for _d in range(0, 100):
+            val = test_object.get_beam_position(distance=_d)
+            beam_position = (_d * ax + bx, _d * ay + by)
+
+            assert beam_position == val, "Beam position should be slightly off center"
```
rhfogh, line 60: Should the message be changed?

rhfogh:
@marcus-oscarsson The current version takes care of the objections I was making.

I shall have to come back to this after the weekend, though. 

For one thing, I need the DetectorMockup.get_beam_position for the GPhL collection emulation, as a pair with _set_detector_position. At least I have so far, I shall see whether I can adapt my code to doing without both functions, or whether I shall have to come back to you on this.

For another thing there is a philosophical issue with this. AbstratctDetector is an abstract class, but it also contains a default implementation, for get_beam_position. The question is how far should we test the default implementation (that might be overridden lower down), and how far should we limit tests to aspects that define the behaviour of the abstract class as such, regardless of implementation?

test/pytest/test_detector.py
marcus-oscarsson: Yes indeed :)

marcus-oscarsson:
Okey, good that we solved your main concerns then. Yes lets continue on Monday. 

I think its a good idea to test the default implementation of abstract classes, that is after all (at least in this case) what most will use. If someone adds a new class that overrides the default, I guess that has to be tested in a test specific for that class. However the implementation of `AbstractDetector` don't change, there is nothing that says we can't have several detectors associated with `beamline` that we can test separately. We can simply assume that the `HWR.beamline.detector` inherits `AbstractDetector` directly (or at-least does not override any of the methods with default implementation) and for instance `HWR.beamline.my_other_detector` is of a another kind and can be tested as such. Its one way it could be done.

rhfogh:
@marcus-oscarsson That sounds quite reasonable.
It might be later than Monday - I believe we have another live test then, but I'll see how it pans out.
Rasmus

test/pytest/test_detector.py
```diff
@@ -45,3 +46,42 @@ def test_detector_atributes(self, test_object):
...
+
```
rhfogh, line 61: No objection to these, but I wonder about how they fit in, philosophically. In effect you are just duplicating in the test the code that was already in the functions themselves. Not sure how much that helps, and the test would break if you ever introduced a non-rectangular detector shape, or off-centre detectors. But OK, we can always commit this and change it later if needed.

test/pytest/test_detector.py
```diff
@@ -45,3 +46,42 @@ def test_detector_atributes(self, test_object):
         ), "Detector hardware object is None (not initialized)"
         exp_time_limits = test_object.get_exposure_time_limits()
         has_shutterless = test_object.has_shutterless()
+
+    def test_get_beam_position(self, test_object):
```
rhfogh, line 50: Again, this is just replicating the calculation in the parent function. It will break if anybody introduces a wavelength-dependent beam position, as one might legally do. Let us commit it and think about alternatives later, though.

mxcubecore/HardwareObjects/mockup/DetectorMockup.py
```diff
@@ -46,9 +46,6 @@ def has_shutterless(self):
         """
         return True
 
-    def get_beam_position(self, distance=None, wavelength=None):
```
rhfogh, line 49: The combination of get_beam_position, _set_beam_position, and self._beam_centre I introduced because I needed it for image simulation. It will be several days before I can test if I can manage without it. 1) Could we leave it in as a hack for now, possibly with a switch so that get_beam_position call the superclass function unless self._beam_centre has been set explicitly?  2) If we want to clean this up, self._beam_centre and _set_beam_position would have to go as well.

test/pytest/test_detector.py
marcus-oscarsson: I agree that these tests are maybe not adding much more certainty. It tests that the current beam position follows the linear regression model in AbstractDetector (that most of us are using). A change in that code that breaks the regression or how the values are read from configuration would fail the test.

Which is at least better than not having any test at all.

I think this would work for both non-rectangular detectors and if the beam is calibrated to not be in the middle of the detector (if that whats meant by off center detectors). If its not the case I think you have to explain why it would not be the case

test/pytest/test_detector.py
marcus-oscarsson: Same as, above its at least testing the basic functionality which is better than nothing. We don't have an implementation that accounts for wavelength so I don't really see the use of testing it ?

mxcubecore/HardwareObjects/mockup/DetectorMockup.py
marcus-oscarsson: Sure we can try that

marcus-oscarsson:
Ok, ill make some final adjustments, fixing a assert message that you pointed out and Ill do something for `get_beam_position`. Then I think Ill commit this and we can change it if we find some better ways of testing this or if we find some problems with it,

rhfogh:
@marcus-oscarsson As you say "_we can change it if we find some better ways of testing this or if we find some problems with it,_"
I agree with that. As you say, it is a lot better than nothing, as I, for one, am not sure how we should do it better. We just have to be prepared to change this at need.


For the get_beam_position I could do the fix after check-in, if you prefer. Or you can.

I would propose something like 
    def get_beam_position(self, distance=None, wavelength=None):
        if self. self._beam_centre == (None, None):
            return super().get_beam_centre(distance, wavelength)
        else:
            return  self._beam_centre

and then initialise self._beam_centre to (None, None)

Maybe we do not need to talk about this after all?

marcus-oscarsson:
Okey, thanks for the proposition. Sounds good I think we can continue with this then. Ill merge this and Ill count on you to do the change you suggest above. It would be easier for you to test it with your code as you are using set_beam_position from your simulator.

PR #713, opened 2022-10-13
Very early version of a PyISPyB client

marcus-oscarsson:
This is an attempt to use a client that was generated from the Open API specification provided by the `PyISPyB` server (via `FastAPI`). The advantage would be that API updates are fairly easy to apply by simply re-running the code generating 
command and the validation is provided by the generated client.

The HardwareObject  PyISPyBClient uses a completely generated client. The wrapped client, located in `mxcubecore/utils/pyispyb_client`, was generated with:

```
     openapi-generator-cli generate --skip-validate-spec -i http://py-ispyb-development:8000/ispyb/api/v1/openapi.json -g python
```

The HardwareObject PyISPyBClient is currently only implementing a very limited set of the SSX web services but the hope is to extend this relatively soon and make it possible to replace the older SOAP based client with this one.


rhfogh:
Could you enlarge a bit on how this fits in? Is it web-only or also for the Qt interfaces? is it for testing/mockup or intended for production use, and how close is it o useful? What are the dependencies between this very large and detailed PR and what we have already - in other words how will changes in the Py-ISPyB ASPI condition changes in mxcube - and vice versa? Moving to Py-ISPyB is very much a 'good thing', but some context would help.

marcus-oscarsson:
At the moment we have three ISPyB clients a SOAP client, a REST client and a the new Py-ISPyB client (also REST). This is probably detailed better by a ISPyB developer but a brief summary: 

-   The [ISPyB SOAP](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/ISPyBClient.py) interface covers most of the **standard MX-experiment needs** as well as handling **sessions** and **proposals**
-   The [REST interface](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/ISPyBRestClient.py) extends the SOAP interfaces with a few things like being able to retrieve processing results
-   The Py-ISPyB interface will eventually cover everything and will replace the SOAP and the old REST interface. However **only SSX experiments** are supported by Py-ISPyB for the moment

This Py-ISPyB client was created so that we are be able to store SSX experiments. The intention is that it will eventually replace the other two clients but there needs to be some work done on both client and server side for that to be possible. This is to be seen as more of a proposition **(but we do need it for the SSX experiments)**. The details can be refined and improved, perhaps using the auto generated client is not good in the end. The changes in the Py-ISPyB API will definitely have an impact but we have no other way to deal with SSX data at the moment so have to trust our ISPyB colleagues to deal with such changes with care :)

rhfogh:
@marcus-oscarsson Thanks for a very good explanation. I do not think I should formally approve this, as I am not competent to evaluate it, but I shall be happy to see it merged in.

marcus-oscarsson:
Thanks @rhfogh. I will merge it for now and at the same time say that it's still preliminary and that we are open to improvements and that we are aware that the Py-ISPyB API itself might change during the course of time :). This is at-least a common basis to work from as the community moves to Py-ISPyB

PR #714, opened 2022-10-13
Queue improvement - Replacing #686

marcus-oscarsson:
This PR replaces #686 [WIP] Queue Improvement

The main idea of this is to make the queue entries / tasks easier to manage by making

- **Each queue entry a separate entity that can be loaded dynamically if/when needed (from configuration)**

- **The data model (used by this new queue_entry) an object that is more well defined compared to simple dictionary**. This has the advantage that its easy to get an idea of what data is expected and how to access it (and attributes cant be added when passed around)

- **It possible to use automatically add a method/entry to the UI by simply creating a queue_entry file.** 

The total work will be split into several PRs and we hope to have some discussion around this between each. This is the first PR and just puts the mechanism of being able to load queue_entries "dynamically" from configuration. It also re-structures the files queue_entry.py and queue_model_objects.py into smaller pieces. This introduces a new queue entry and model keeping the old queue entries and queue_model objects intact (thus backward compatible).

With this suggestion the queue_entry is defined in a single file see for [example](https://github.com/mxcube/mxcubecore/blob/5c4489b86b0d187df43710266eb16a4d35994f3a/mxcubecore/HardwareObjects/ESRF/queue_entry/ssx_foil_collection.py)

The data model, based on Pydantic, is defined in the same file as the task logic. It's possible that we
might need to enhance the model to fit our exact needs (especially for validation).

A new queue entry can simply made available by adding a single file, similar to the example above, and it
will be made available automatically and a generic input form presented if there are no custom form
available.

In more detail:

- **The file queue_entry.py was split into smaller pieces and moved one folder up to the root of mxcubecore** ([mxcubecore/queue_entry](https://github.com/mxcube/mxcubecore/tree/queue-entry/mxcubecore/queue_entry)) Its also possible to add site specific queue entries on the HardwareObjects/[SITE]/queue_entry folder (this can ofoucrse be discussed). The site folders to load can be configured in the queue.xml file

- **queue_model objects was also moved one directory up and and placed in** [mxubcore/model.](https://github.com/mxcube/mxcubecore/tree/queue-entry/mxcubecore/model) 

- Each QueueEntry object is in its own file **named [queue_entry_name]queue_entry snake cased name with a queue_entry suffix**

- The QueueEntry object is **named [QueueEntryName]QueueEntry camel cased name with a QueueEntry suffix** (or simply the camel cased filename)

-  Created queue_entry sub-module that automatically imports any *_queue_entry.py file in addition to the constants and base classed defined in base_queue_entry

- Import of queue_entry submodule is more or less a drop in replacement of the previous queue_etnry.py

- New queue entry can simply be added by adding a new _queue_entry.py file


rhfogh:
This PR includes a lot of work and provides many things that we definitely want. Is is very much a step forward. Unfortunately, as things stand I do not think I can approve it for merging into the main develop branch on my own, without input from other people - even after discussing the details with @marcus-oscarsson . Most of this PR is very good indeed, but as I see it, there are things here that break with the basic organisation of the queue handling and pre-empt a future reorganisation of the way we handle the queue. Even these things might actually turn out to be quite good, but I do not think I should approve the PR on my own without first having seen at least a schematic description of how we want the queue to work in the future. 
For more details please see Issue #718 

Since the changes in this PR are already (close to?) production use at the ESRF we clearly need a decision quite soon. We might be able to make a lightning zoom meeting, alternatively it would be really great if people could contribute to the issue, so we can get some kind of consensus on how to proceed.

marcus-oscarsson:
Thank you for your input @rhfogh and the discussions we had the other day. They were indeed really useful and pointed out some important aspects. This PR does add a few new concepts, none (just to be clear) which brakes anything currently existing. What concerns the queue, these can be made completely compatible with today's Qt UI, however one has to add the UI bit for the generic dialogs exported as JSONSchemas.

> This PR includes a lot of work and provides many things that we definitely want. Is is very much a step forward. Unfortunately, as things stand I do not think I can approve it for merging into the main develop branch on my own, without input from other people - even after discussing the details with @marcus-oscarsson . Most of this PR is very good indeed, but as I see it, there are things here that break with the basic organisation of the queue handling and pre-empt a future reorganisation of the way we handle the queue. Even these things might actually turn out to be quite good, but I do not think I should approve the PR on my own without first having seen at least a schematic description of how we want the queue to work in the future. For more details please see Issue #718

I don't regard any of this to pre-empt a future queue reorganization it does indeed add a kind of "generic" `QueueModel`, by adding an attribute `task_data` to `DataCollectionQueueModel`. The `task_data` member is a Pydantic model that replaces the explicit attributes used for data (i.e `acquisition`, or `crystal`). However there is little difference between using a `DataCollectionQueueModel` with the `task_data` from deriving a new Node. It does provide a more dynamic way to implement `QueueModels` and `QueueEntires` . The intent of this PR was to move and split the queue_entry/model modules, fix the imports and add a more dynamic queue_entry. The later could actually be done so that it even can be added transparently from the workflows (which I agree is not the case in this PR). We were actually hoping to promote a discussion around the future queue organization with this PR and this work was seen as a minimal set of changes to get started (that of course can be re-worked if needed).
> 
> Since the changes in this PR are already (close to?) production use at the ESRF we clearly need a decision quite soon. We might be able to make a lightning zoom meeting, alternatively it would be really great if people could contribute to the issue, so we can get some kind of consensus on how to proceed.

We are already making use of this at ESRF and we will need this feature or something similar in the future. I'm personally on vacation this week but back next. I'm very happy to discus this further, a meeting is perhaps more appropriate as its a lengthy subject. All contributions are otherwise welcome. Our next developers meeting will take place the 8th of November and will send
out more information about this in topic in the mail dealing with that meeting.

rhfogh:
> it does indeed add a kind of "generic" QueueModel, by adding an attribute task_data to DataCollectionQueueModel. The task_data member is a Pydantic model that replaces the explicit attributes used for data (i.e acquisition, or crystal). However there is little difference between using a DataCollectionQueueModel with the task_data from deriving a new Node. It does provide a more dynamic way to implement QueueModels and QueueEntires . The intent of this PR was to move and split the queue_entry/model modules, fix the imports and add a more dynamic queue_entry.

I completely agree with everything you say in this quote (even if I think there are a couple of not-quite-minor points outstanding). Moreover, I think it is very good, and I am glad it is being done.

The main reason for pushing for a faster reaction was that I understood this was urgent, the discussion seemed to indicate that it might take a certain amount of thinking to find a consensus solution, and I got the impression that you wanted the PR to be sorted out and merged by November 8th. This suggested to me that we had better bring some more developers on board right away, so we could resolve any disagreements and have time to settle on a decision. But if you are OK with waiting till November 8th for a discussion, I agree that it would be better to avoid rushing things and look at this calmly at our next meeting.

rhfogh (approved):
I obviously have not tested this, but it looks very good.

marcus-oscarsson:
I'll merge this, seeing that the main concern is fixed. The discussion on how to structure the queue needs to continue, see  [Issue 718 ](https://github.com/mxcube/mxcubecore/issues/718) for more details.

PR #715, opened 2022-10-14
Package Management, Linting / Code Quality and GitHub Actions Improvements

oldfielj-ansto:
This PR makes changes in several places, none of which impact the overall functionality of the MXCuBE Core package.
These changes aren't meant to be merged in their current state, rather they are intended to start a discussion around package management tools, linting/code quality tools and package distribution.

Summary of Changes:
- Switch to Poetry "pyproject.toml" over traditional "setup.py" and "requirements.txt" package management.
    - Thanks to [PEP 517](https://peps.python.org/pep-0517/) this change would have no impact on deployment, `pip install mxcubecore` will work the same way as it always has.
    - The only real negative that I can see here is that "bump2version" will stop working and will need to be re-implemented, but the good news here is that [Poetry has a feature](https://python-poetry.org/docs/cli/#version) to support project version bumping automatically, so a new GitHub action could be implemented to instruct Poetry to increment the minor version each time a PR is merged into the "develop" branch.
    - Switching to Poetry provides access to two new useful commands [`poetry build`](https://python-poetry.org/docs/cli/#build) and [`poetry publish`](https://python-poetry.org/docs/cli/#publish), the build command will automatically build the sdist/wheel archives for distribution, the publish command will update a PyPi repository listing and upload the new version attaching the archives generated by the build command. This simplifies distribution overhead and opens up the opportunity to automate the entire process via GitHub actions.
    - A question that might come up as a result of this PR:
      > What effect will this have on MXCuBE3, considering we are using Conda to manage the environment?
        - It will have no impact, as mentioned above `pip install mxcubecore` will work the same and we can even use the Poetry tool directly with Conda as there is a Poetry package maintained on [Conda-Forge](https://anaconda.org/conda-forge/poetry) and running `poetry install` from within the MXCuBE3 Conda environment will install the MXCuBE Core project into the existing Python environment create by Conda.
- Changes to linting/code quality tools.
    - Note that none of the changes here are intended to effect facility specific code under "mxcubecore/HardwareObjects", it should be up to each facility to first decide, if following these standards matters to them, which standards they want to enforce and how they want to implement that enforcement.
    - Implement [Pre-Commit](https://pre-commit.com/) and associated ".pre-commit-config.yaml" configuration file.
        - Pre-Commit is a Python package that can be installed from [PyPi](https://pypi.org/project/pre-commit/) and it's designed to simplify management of GIT Pre-Commit hooks.
        - The use case here is to specify linting/code quality requirements for Black and PEP conformity, minimising the need to manually run multiple commands before raising a new PR, all Black flags are specified here and Pre-Commit can be run from inside the virtual environment by calling `pre-commit run -a`.
        - Since it can be easy to forget to run the commands associated with linting/code quality tools, Pre-Commit offers the ability to generate and install GIT Pre-Commit hook scripts to your local development repository [`pre-commit install`](https://pre-commit.com/#3-install-the-git-hook-scripts) and once installed these scripts will cause new commits to fail if Pre-Commit has not been run and the problems resolved, this generally acts as a final sanity check before trying to raise a PR and realising there is an issue at that point.
    - Take Black out of "safe-mode", current contributing guidelines specify that Black should be run in safe-mode, this prevents Black from making structural changes to the code in areas such as module imports, making this change will mean Black makes large changes to the codebase, in the order of hundreds of files, but it will mean legacy code features such as [star-imports](https://www.geeksforgeeks.org/why-import-star-in-python-is-a-bad-idea/) will be cleaned up and removed.
    - Add the [Flake8](https://pypi.org/project/flake8/) code quality tool, similar to Black, Flake8 is an opinionated code style enforcement tool, it's introduction here with the current settings will mean a large refactor of the entire codebase would be required, the issues it did complain about are too numerous to list here, we should debate if its introduction is acceptable given the workload of changes it will generate and if so, how that workload can be shared so as not to burden any one person with what is likely weeks of work.
- Introduce new/update existing GitHub actions.
    - Pytest (tests.yml)
        - Executes Pytest test cases for Python [3.7, 3.8, 3.9, 3.10].
        - Tests action has been updated to install the environment using Poetry.
        - Tests will run as status checks against PR's and will provide immediate feedback to reviewers if there is an issue that has been found.
        - Tests are currently configured to only run against a PR when a code change is present in a ".py" file.
    - Linting & Code Quality (pre-commit.yml)
        - Executes `pre-commit run -a` to run linting/code quality tools, essentially this is a check to tell the reviewer if the contributor has remembered to run these tools before submitting the PR.
    - Coverage Report (coverage_report.yml)
        - Executes the Pytest test cases again, this time with the "pytest-cov" plugin enabled, if the tests are passing the action will post a comment on the PR with total project coverage and a summary of each changed files coverage, listing any lines of code that have been missed while running the test cases.
        - Each time the PR is updated the action will re-run and update the posted comment.
    - Create/Update Tag (create-tag.yml)
        - Creates/Updates GIT tags anytime a new commit is pushed to the "develop" branch, this will trigger for both PR's being merged and commits pushed directly to the branch.
        - Tags are named based on the version specified in the "pyproject.toml" file, so a project version defined as "1.0.0" would result in a tag being created called "v1.0.0".


oldfielj-ansto:
The "Linting & Code Quality" action fails here due to the aforementioned Pre-Commit settings for Black, etc.

The "Coverage Report" action fails due to the repository GitHub action permissions as they are currently configured, this can be rectified without changing those settings by generating and providing the action with a new token that has sufficient permissions to comment on a PR.

An example of the what the "Coverage Report" action comment looks like can be found here.
https://github.com/AustralianSynchrotron/mxcubecore/pull/3#issuecomment-1266292792

marcus-oscarsson:
Thanks a lot @oldfielj-ansto for this very nice PR, It is very much appreciated :+1:

marcus-oscarsson:
I like the idea of trying out `Poetry`, I did actually consider it when we introduced `bump2version` but did not have the time to investigate it enough. It seems to me to provide a more complete solution over `setup.py`, I like the overview the `pyproject.toml` gives and that it **handles virtual environments** (basically a must when working with Python)

Over all it looks like a good update of our current routines, I suspect that we might need to discuss what to do with the Flake8 warnings/errors. I think we should try to get as much of this as possible in place, perhaps excluding Flake8 to start with. Then come up with a plan on how to perform the necessary changes for Flake8, if we think it will bring us some additional value.

rhfogh:
It is very good that someone is doing this. It will be painful to incorporate, but then taking your medicine always is.

Could I propose that someone runs the full black/flake8 on a separate branch and makes the linted branch and the linter report available to all of us? We could then look at the kind of changes that would await us, digest it, and then discuss what checks we might want to disable (if any, of course).

pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "Intended Audience :: End Users/Desktop",
+    "Natural Language :: English",
+    "Topic :: Scientific/Engineering",
+]
+
+[tool.poetry.dependencies]
+python = ">=3.7,<3.11"
+gevent = "21.12.0"
```
beenje: I don't think pinning all the dependencies like that is a good idea in `pyproject.toml`.
I haven't used `poetry` but I think that's what `poetry.lock` is for: giving a file to reproduce an environment.

If you pin exactly all versions in `pyproject.toml` it makes it very difficult to install in an existing environment.
If you for example have a conda environment with some of the dependencies needed, pip/poetry will overwrite the conda packages with the pypi ones if all the versions don't exactly match, leaving the env in an unknown state.
It also makes it difficult to test with some new versions of those packages (some that could have bug fixes).


My opinion is that you should only put the version constraints that are really needed. Min version is usually fine. Upper bound **only** if you know that a more recent version is breaking things.
This blog post summarises well the issue: https://iscinumpy.dev/post/bound-version-constraints/

- restrict versions as little as possible in the package -> this makes it easy to install in different environments and test if it's working or not
- provide a lock file that can be used for deployment to have something reproducible (poetry.lock)

This is of course only my personal opinion :-)

pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+  | \.venv
+  | _build
+  | buck-out
+  | build
+  | dist
+)/
+'''
+
+[build-system]
+requires = ["poetry-core>=1.0.0", "setuptools>=45", "setuptools_scm[toml]>=6.2"]
```
beenje, line 78: Isn't `poetry` replacing  `setuptools` as build system?
I don't think you can use `setuptools_scm` with poetry: https://python-poetry.org/docs/pyproject#poetry-and-pep-517

beenje (commented):
Nice to see improvements regarding packaging!

I haven't used poetry but made small comments.
Regarding the bump version part, we use [setuptools_scm](https://github.com/pypa/setuptools_scm) at MAX IV and this has been working very well. It takes the version from the git tag and when you install from a branch you get a "dev" version with the commit id. Obviously this only works with `setuptools` and not `poetry`.
A quick search showed this plugin for poetry doing something similar: https://github.com/mtkennerly/poetry-dynamic-versioning
It might be worst investigating?

pyproject.toml
marcus-oscarsson: This was discussed and presented in depth during one of our code camps and following developers meetings. I think most of what you outline above have been covered. What @oldfielj-ansto did in this PR was to put a basic infrastructure into place so that we can get started with this. The pining have been covered as well and we obviously (in most cases) do not need to pin with a lock file. I think it was simply done like this to be compatible with the requirements.txt to not break anything as to somehow stay compatible.

marcus-oscarsson:
Looks great :) thanks !

marcus-oscarsson:
Ill go a head and merge this, there might be some adjustments later on but it puts a lot of good things in place.

pyproject.toml
oldfielj-ansto: As far as I'm aware Poetry will still call setuptools to install package wheels, I believe the reason the specific version of setuptools is specified here was to resolve some issue I saw previously, but I cannot recall which issue in particular.

oldfielj-ansto:
@marcus-oscarsson Cool OK, I'll need to raise another PR to fix an issue I've been troubleshooting with Python3.11.

Seems that the "gipc" package is failing to install as it is not yet compatible with the latest release of Gevent.
Since it looks like this package is only used by sites requiring the "TangoLimaVideo" and "TangoLimaVideoLoopback" hardware objects, I should be able to lock it to Python3.10 and below, allowing sites not using these hardware objects to start experimenting with Python3.11.

I've also been playing around with the code quality / linting tool configs and have gotten the number of automatic refactors down to 112, which I think is a more manageable starting point than the 800+ we started with in this PR.

Just working on adjusting the Flake8 configuration to document what the warning being disabled are, and muting some of the more nuisance spam warnings.

pyproject.toml
oldfielj-ansto: @beenje @marcus-oscarsson So I adjusted the dependency pinning before this was merged and opened things up where I could.

The way it is currently configured will allow most dependencies to install the latest non major update available.
So a version specification of "^1.2.3" will allow any package up to, but excluding "2.0.0".
If the semantic version has not yet reached "1.0.0", a version specification of "^0.2.3" will allow any package up to, but excluding "0.3.0".
[https://python-poetry.org/docs/dependency-specification/#caret-requirements](https://python-poetry.org/docs/dependency-specification/#caret-requirements)

The reason this wasn't opened up initially was due to professional paranoia, generally I don't trust package managers not to break things in a minor update and have run into this issue out in the wild.

For our side, internally we'll be manually updating each package when a new update is released to avoid unexpected issues popping up in the production environment, but I understand this kind of time investment isn't practical at all sites using MXCuBE Core.

As for the lock file, it's there to speed up the installation process, when it's generated it resolves all of the dependency versions and package hashes and saves them to the lock file, the versions in this lock file should not be depended on and should be considered fluid, and dependency pinning should occur inside the "pyproject.toml" file.

marcus-oscarsson:
@oldfielj-ansto. Ok I see. I had the same problem and simply pinned gevent to "21.12.0" which I think should work fine as well.

We will tune these things a bit as we go, I guess its the easiest way. Sounds great for the linting, 112 errors sounds manageable :),

pyproject.toml
marcus-oscarsson: > The reason this wasn't opened up initially was due to professional paranoia, generally I don't trust package managers not to break things in a minor update and have run into this issue out in the wild.

We have been using a similar philosophy/(paranoia :) ) here as well. Interesting to know that the lock file is there for only for performance I thought it was also used to pin the versions.  Lets see/discuss if there are any special considerations regarding that in our next meeting.

pyproject.toml
beenje: Good to know. I didn't know that poetry was still using `setuptools`. I guess dependencies should be managed by poetry itself, but maybe that was an issue in previous versions. Anyway no objection to that.
But from what I read `setuptools_scm` won't work with poetry.

pyproject.toml
beenje: I totally understand the philosophy/paranoia :-)
But I disagree about "dependency pinning should occur inside the "pyproject.toml" file".

It's just that my packaging hat doesn't like version pinning at the package level. As soon as you want to add another package to the same environment you might end up having conflicts (like with `gipc`). The blog post I linked to explains that and this is why I don't like the caret-requirement encouraged by poetry. I think it is still too restrictive.

For me, the pinning should be done for deployment using a different file (requirements.txt or environment.yaml) but not encoded in the package itself.

Here you will have to maintain those versions both in `mxcubecore` and `mxcubeweb` and make sure you update both at the same time. If you are less strict, there is less work at the package level but you can still have one file used for deployment with pinned version. And there is only one place to update.

This is just my two cents. This advice is quite general. I understand that mxcube is only deployed in a specific environment and that pinning the versions at the package level can help deployment.

marcus-oscarsson: I guess its partly matter of opinion.  Its interesting to know how people manage these kind of situations. We will probably continue parts of these discussions in a developers meeting. Perhaps, Mikel or Mehgdad can let you know when that takes place.

I've used both approaches for a few python projects in the past, pinning and leaving the dependence unpinned.  

Leaving the dependencies unpinned have worked fine with general purposes libraries/applications with few dependencies. (like [pyqtconosle](https://github.com/pyqtconsole/pyqtconsole). I had no problems at all with these dependencies but there are only two well used dependencies.

However I had quite a lot of issues with dependencies for applications using a larger set of dependencies. The problem have not been that the requirement constraints are wrong but that there are small incompatibilities and API changes in these dependencies, breaking the application.

I see no other option than strictly pinning at least a subset of the dependencies to avoid that situation. I definitely don't see an issue with doing it in the mxcubecore context as mxcubeweb and mxcubeqt are the only users of this library and that environment should be fairly well controlled. Then perhaps, sure, the requirements will sometimes cause us some issue, but we are completely in control of the projects and environment.

For curiosity, what use does it make to pin dependencies in a release but not in the package. You can install the package but with what the author considers to be the "wrong" dependencies ?

beenje: It's true that pinning for applications (like mxcube) is much easier than for libraries (when you don't know how people will use). So I'm quite sure this can work fine for mxcube.

My view is that you can't predict which version of a dependency will break your app (due to API change). You can suspect that a major release will, but it could just as well be a patch release. And in many cases, a major release won't break your app because even if there are API changes, you might not used the ones affected.

The only way to know if a new version will break your app or not is to test it. If you have a good coverage with automatic tests, you can run those tests nightly. It can give you an easy way to automatically detect if a new version breaks the tests or not (if installing your package installs newest version of the dependencies). If it does break the tests, you can decide to pin that dependency and/or fix it later.
If you pinned before knowing, it makes it more difficult to test with newer version. You can easily end up staying with old dependencies versions without good reasons.

From the user point of view, if a new version of a dependency breaks an app, you usually can quite easily change the version in your env. If you want to add a new package and there is conflict because you need a newer version it's much more difficult to fix. You usually have to wait for the maintainer to make a new release (or force to install the package ignoring the deps).

>  what the author considers to be the "wrong" dependencies

If you pin only things that you know don't work, this is of course fine. If you don't and just think it might break, it could create conflicts for the user.
Again much less likely to be an issue for an app than for a library.

In the end, both approach have advantages/disadvantages. It depends what matter the most to you.

oldfielj-ansto: I agree that minor versions can break the installation of the package, but it is a rarer occurrence than with major releases.
This is the reason why I pin dependencies to a specific version in production, that way there are no surprises.

The way bigger projects would normally handle this kind of issue is to make use of a tool such as [Dependabot](https://github.com/dependabot).
Dependencies get pinned at a specific version, Dependabot periodically checks if a new version is available and raises a PR on the project, if the tests pass, the PR is often merged automatically.

The problem for MXCuBE Core is that the tests haven't been fully fleshed out yet, so this approach wouldn't really work, as some issues present in a new version of a dependency would not be found.

I don't really see a valid reason to completely unpin all dependencies, even if things worked without issue, it's generally good practice to separate projects such as MXCuBE into virtual environments to avoid dependency clashes with other Python applications. I would not recommend installing MXCuBE Core into the OS default or user default Python environments, it should either be installed into the MXCuBE-Web Conda environment or a new Python virtual environment.

For cases where a dependency problem is identified in a deployed environment and you would like to change the version installed, you can always modify the "pyproject.toml" to match the version you are looking and "pip" install the MXCuBE Core package again to apply the changes.

PR #716, opened 2022-10-14
Some updates after move to mxcubecore

marcus-oscarsson:
Some fixes we made after the move to mxcubecore. No major improvements or changes but makes things work nicely has a whole :). This PR relies on #714, so the imports to queue_entry and models are changed meaning that the tests will fail until its been merged/rebased with #714.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -18,7 +18,6 @@
 #  You should have received a copy of the GNU General Lesser Public License
 #  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
 
-
 from __future__ import absolute_import
 
 import typing
@@ -32,7 +31,6 @@ import pydantic
 from mxcubecore.dispatcher import dispatcher
 from mxcubecore.CommandContainer import CommandContainer
 from mxcubecore.utils.conversion import string_types
-
 
 __copyright__ = """ Copyright © 2010-2020 by the MXCuBE collaboration """
 __license__ = "LGPLv3+"
@@ -477,7 +475,7 @@ class HardwareObjectMixin(CommandContainer):
         self.update_state(self.STATES.UNKNOWN)
 
     def init(self):
-        """ "'public' post-initialization method. Override as needed
+        """'public' post-initialization method. Override as needed
 
         For ConfiguredObjects called after loading contained objects"""
         self._exports = dict.fromkeys(self._exports_config_list, {})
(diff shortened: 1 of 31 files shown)


mxcubecore/BaseHardwareObjects.py
```diff
@@ -465,11 +477,18 @@ def _init(self):
         self.update_state(self.STATES.UNKNOWN)
 
     def init(self):
-        """"'public' post-initialization method. Override as needed
+        """ "'public' post-initialization method. Override as needed
```
beteva: Maybe one " less

mxcubecore/HardwareObjects/LimaJungfrauDetector.py
```diff
@@ -0,0 +1,211 @@
...
+from mxcubecore import HardwareRepository as HWR
+from mxcubecore.CommandContainer import ConnectionError
+
+from mxcubecore.HardwareObjects.abstract.AbstractDetector import (
+    AbstractDetector
+)
+
+from mxcubecore.BaseHardwareObjects import HardwareObjectState
+
+MaskDir = '/data/id29/inhouse/opid291/Jungfrau/tests/ahoms/processing/20220928'
```
beteva: Hardcoded. Either it is configurable or move the Hardware object to the  ESRF directory.

beteva (approved):
Thank you @marcus-oscarsson for this PR - not only the new functionalities, but also effort to clean the rest.
The PR still uses Microdiff and GenericDiffractometer, which need to be replaced by AbstractDiffractometer and MicroDiffarctoemeter shortly.

mxcubecore/HardwareObjects/LimaJungfrauDetector.py
marcus-oscarsson: Oh, yes. This seems to have snuck in without me noticing ....

mxcubecore/HardwareObjects/DESY/NanoDiff.py
```diff
@@ -496,18 +496,20 @@ def omega_reference_add_constraint(self):
             return
         if self.omega_reference_par["camera_axis"].lower() == "x":
             on_beam = (
-                (self.beam_position[0] - self.zoom_centre["x"])
-                * self.omega_reference_par["direction"]
-                / self.pixels_per_mm_x
-                + self.omega_reference_par["position"]
-            )
+                self.beam_position[0] - self.zoom_centre["x"]
```
rhfogh, line 499: I assume this is black. It looks horrible. Might there be a way of avoiding this change?

marcus-oscarsson: It is indeed black, not sure its easy to change the rules for how code is split between lines. We should probably not use Black if we don't like it I guess ?

marcus-oscarsson: I do agree that it does not look very nice but I think we should continue to use Black. I guess the best thing in these cases is to rewrite the code a bit so that it becomes easier to read when split over several lines.

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -99,7 +119,7 @@ def set_ui_snapshot_cb(self, fun):
         self._ui_snapshot_cb = fun
 
     def get_snapshot(self, overlay=True, bw=False, return_as_array=False):
-        """ Get snappshot(s)
+        """Get snappshot(s)
```
rhfogh: -> snapshot

rhfogh (changes requested):
This seems to be almost 'all-black', and it looks fine, with one exception:
Cpuld you undo the black changes in the files whose names start with 'Gphl'?
I am in the middle of a maojr refactoring, and it will be a lot of work for me to apply such omnipresent, detailed changes on code that is already heavily changed on my side. I can do my own black runs once my code stabilises.

marcus-oscarsson:
No problem, I removed the black run from the Gphl files

rhfogh (approved):
Excellent. Thanks!

PR #717, opened 2022-10-19
Abstract Hardware Objects cleanup

beteva:
Python 3 syntax. black, flake8 and pylint applied.
Mark AbstractMCA, AbstractAperture and AbstractSlits as deprecated.


mxcubecore/HardwareObjects/abstract/AbstractCharacterisation.py
```diff
@@ -1,38 +1,63 @@
...
         Returns:
-            (str) The Characterisation result
+            (str) The Characterisation result.
         """
-        pass
 
     @abc.abstractmethod
     def get_html_report(self, output):
-        """
+        """Get the path to the chrasterisation html report, generated by the
```
rhfogh: Typo

mxcubecore/HardwareObjects/abstract/AbstractCharacterisation.py
```diff
@@ -103,44 +128,33 @@ def input_from_params(self, ref_parameters, char_params, path_str):
...
         """
-        pass
 
     @abc.abstractmethod
     def get_default_characterisation_parameters(self, defualt_input_file):
         """
         Args:
-            defualt_input_file (str): Path to file containing default input
-
+            defualt_input_file (str): Path to file containing default input.
```
rhfogh: Typo ' defualt...'

mxcubecore/HardwareObjects/abstract/AbstractFlux.py
```diff
@@ -17,38 +17,32 @@
...
+        # flux by default is read only
+        self.read_only = True
 
     def init(self):
         """Initialise some parameters."""
-        super(AbstractFlux, self).init()
+        super().init()
         self.read_only = self.get_property("read_only") or True
 
-    def _set_value(self, value):
```
rhfogh, line 48: Is this unnecessary now (is _set_value not an abstract method in the superclass?)

mxcubecore/HardwareObjects/abstract/AbstractSampleChanger.py
```diff
@@ -531,36 +527,32 @@ def change_mode(self, mode, wait=True):
             SampleChangerState.ChangingMode, wait, self._do_change_mode, mode
         )
 
-    @task
+    @dtask
     def scan(self, component=None, recursive=False):
-        """
-        Scan component or list of components for prescence.
-
+        """Scan component or list of components for prescence.
```
rhfogh: Do you8 mean 'presence'?

mxcubecore/HardwareObjects/abstract/AbstractSampleView.py
```diff
@@ -17,6 +17,9 @@
 #  You should have received a copy of the GNU General Lesser Public License
 #  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
 
+"""Anstract class for a Sample View.
```
rhfogh: Typo

mxcubecore/HardwareObjects/abstract/AbstractSampleView.py
```diff
@@ -255,53 +240,44 @@ def get_shape(self, sid):
...
         return
 
     @abc.abstractmethod
     def get_points(self):
-        """
-        Get all Points currently handled.
-
+        """ Get all currently handled centred points.
         Returns:
-            (list[Point]): All points currently handled
```
rhfogh, line 273: Is there no way to specify 'list of points' as a type?

rhfogh (approved):
Very Nice.

For discussion (should not hold up this PR): 
1) can/should we consider putting the type annotations in the def statement?
2) Coudl we not use typea notations like Sequence[Optional[Point]] to be more precise?

mxcubecore/HardwareObjects/abstract/AbstractFlux.py
beteva: Yes indeed. but as it is again an abstract class and we have nothing to implement in it, I've removed it. This is the case for other abstract methods as well, but as we expect them to be implemented in the concrete Hardware Object, we do not care in the abstract child class.
Hope nobody has an objection, but am ready to change, if needed.

beteva:
Thanks @rhfogh for the review. This is only a partial attempt to be Python3 compliant and go towards CI which can include a check of the style.

beteva:
Another point for discussion would be which of the hardware objects can be marked as deprecated and potentially removed from the package.

mxcubecore/HardwareObjects/abstract/AbstractFlux.py
rhfogh: Good point. I had missed that. I approve.

PR #721, opened 2022-11-08
take_snapshot instead of get_snapshot

meghdadyazdi:
I think to be consistent with [sampleView hardwareObject](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/SampleView.py#L140), we should define take_snapshot instead of get_snapshot.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractVideoDevice.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractVideoDevice.py
@@ -292,6 +292,13 @@ class AbstractVideoDevice(HardwareObject):
             open(filename, "w").write(jpgstr)
 
     def take_snapshot(self, bw=None, return_as_array=True):
+        """Take the snapshot.
+        Args:
+            bs(bool): Return grayscale image (True)
+            return_as_array(bool): Return the image as array. Default is True.
+        Returns:
+            (): Snapshot image.
+        """
         self.get_snapshot(bw, return_as_array)
     
     def get_snapshot(self, bw=None, return_as_array=True):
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
I have nothing against the name change per say, it is indeed more consistent. We need, however, to be a bit careful as this would be an API breaking change, so in theory we should then handle it as such.

If we would like this name change we should at least also need to make sure that all the "client" objects using this is updated as this is an AbstractClass we are modifying the API. Did you investigate in how many places in the code `get_snapshot` is used ?

rhfogh (changes requested):
This is clearly a good idea, but like @marcus-oscarsson I think a bit more is required.
How about:
- changing also all the places where this function is called (as MO said)
- putting in a get_snapshot function that calls take_snapshot and have it raise a DeprecationWarning. Then we can get rid of it later, and meanwhile we are not breaking the API.

marcus-oscarsson:
Yes, adding the deprecation warning is a good idea. I agree that it would be necessary

mxcubecore/HardwareObjects/abstract/AbstractVideoDevice.py
```diff
@@ -291,13 +292,19 @@ def save_snapshot(self, filename, image_type="PNG"):
...
+        self.get_snapshot(bw, return_as_array)
+    
+    def get_snapshot(self, bw=None, return_as_array=True):
         """Get the snapshot.
         Args:
             bs(bool): Return grayscale image (True)
             return_as_array(bool): Return the image as array. Default is True.
         Returns:
             (): Snapshot image.
         """
```
marcus-oscarsson, line 311: Perhaps move the doc string to `take_snapshot` then, if we eventually will remove `get_snapshot`

PR #722, opened 2022-11-09
Replaced the global_role with "dotted/attribute path"

marcus-oscarsson:
The "dotted/attribute path" of a HardwareObject within the Beamline hierarchy can be retrieved through `beamline.get_id`.

### DIFF
--- a/mxcubecore/HardwareObjects/Beamline.py
+++ b/mxcubecore/HardwareObjects/Beamline.py
@@ -33,7 +33,7 @@ __author__ = "Rasmus H Fogh"
 
 import logging
 
-from mxcubecore.BaseHardwareObjects import ConfiguredObject
+from mxcubecore.BaseHardwareObjects import ConfiguredObject, HardwareObject
 
 # NBNB The acq parameter names match the attributes of AcquisitionParameters
 # Whereas the limit parmeter values use more udnerstandable names
@@ -122,9 +122,13 @@ class Beamline(ConfiguredObject):
         # List of undulators
         self.undulators = []
 
+        # Dictionary with the python id of hardwareobject as key 
+        # and the "dotted/attribute path" to hardwareobject from the 
+        # Beamline object
+        self._hardware_object_id_dict = {}
+
     def init(self):
         """Object initialisation - executed *after* loading contents"""
-
         # Validate acquisition parameters
         for acquisition_type, params in self.default_acquisition_parameters.items():
             unrecognised = [x for x in params if x not in self.SUPPORTED_ACQ_PARAMETERS]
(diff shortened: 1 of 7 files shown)


mxcubecore/HardwareRepository.py
```diff
@@ -299,6 +299,7 @@ def init_hardware_repository(configuration_path):
     _instance = __HardwareRepositoryClient(configuration_path)
     _instance.connect()
     beamline = load_from_yaml(BEAMLINE_CONFIG_FILE, role="beamline")
+    beamline._hwr_init_done()
```
marcus-oscarsson, line 302: The dictionary with the "dotted paths" have to be built after all HardwareObjects have been initialized. They are otherwise None when iterating through.

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -144,6 +148,64 @@ def init(self):
...
+
+        Args:
+            ho (HardwareObject): The hardware object for which to get the id
+
+        Returns:
+            (str): Returns "dotted path/attribute"
+        """
+        return self._hardware_object_id_dict.get(id(ho))
+
+    def _get_id_dict(self) -> dict:
```
marcus-oscarsson, line 190: Build a structure used for looking up the "dotted path" for a specific HardwareObject.

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -41,11 +41,7 @@ def init(self):
         super(SampleView, self).init()
 
         self._camera = self.get_object_by_role("camera")
-        self._focus = self.get_object_by_role("focus")
-        self._zoom = self.get_object_by_role("zoom")
-        self._frontlight = self.get_object_by_role("frontlight")
-        self._backlight = self.get_object_by_role("backlight")
-        self._diffractometer = self.get_object_by_role("diffractometer")
+        self._diffractometer =  HWR.beamline.diffractometer
```
rhfogh: Why would you want this attribute, instead of just using HWR.beamline.diffractometer every time you need to get hold of the diffractometer?

marcus-oscarsson: No real reason, I simply made the easiest fix.

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -161,12 +163,29 @@ def get_id(self, ho: HardwareObject) -> str:
...
         Returns:
-            (str): Returns "dotted path/attribute"
+            "dotted path/attribute"
+        """
+        return self._hardware_object_id_dict.get(ho)
+
+    def get_ho(self, _id: str) -> Union[HardwareObject, None]:
         """
-        return self._hardware_object_id_dict.get(id(ho))
+        Returns the HardwareObject with the given id
```
rhfogh, line 175: Look, I'd approve it. But why do you not just use the following one-line code?:

return self._hardware_object_id_dict.get(_id)

I do not know if any American programmers would rebel against the prostitution-related connotations of getting a ho ;-)  - but then we do not have any Americans on board

marcus-oscarsson: Because the key of the dictionary is a `reference` to a HardwareObject if one wants to do what you suggest we also need a map where the "dotted_paths" are they keys.

I can change the name to `get_hardware_object` instead if you find it less offending

rhfogh: Ah, me bad. I did not read it properly (or got led astray by how I expected it to work, maybe). Comment withdrawn.

rhfogh (approved):

marcus-oscarsson:
Ok, thanks

PR #725, opened 2022-11-23
Fixed and simplified image_number precision configuration

rhfogh:
This removes duplicates setting of image number ndigits, and refactors to specify the precision as a digit rather than as previously
<precision>'05'</precision>

It seems to work in qt (for a very quick test), but could not be tested in the web version, where it requires the pending PR  mxcube3 #942


marcus-oscarsson:
Looks good as far as I can see. I guess you left the commented code to make it more clear what changed ? Otherwise Id simply remove it ?

rhfogh:
Indeed. I shall remove it.

PR #726, opened 2022-11-24
Fixed inconsistent valueChanged and getValue "return types"

marcus-oscarsson:
The **FluxMockup** "emits a value changed with a complex structure which is inconsistent with its `get_value`. I would expect the two to be consistent. Which one to use is however maybe more of a debate. I thought simply the "value" as returned by `get_value`  would be sufficient.

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/FluxMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/FluxMockup.py
@@ -51,8 +51,13 @@ class FluxMockup(AbstractFlux):
         self.measure_flux()
         return self.current_flux_dict["flux"]
 
-    def measure_flux(self):
-        """Measures intesity"""
+    def measure_flux(self) -> None:
+        """
+        Measures intesity
+
+        Emits:
+           valueChanged (float): The new flux value
+        """
         beam_size = HWR.beamline.beam.get_beam_size()
         transmission = HWR.beamline.transmission.get_value()
         flux = self.default_value * (1 + 0.001 * random())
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/mockup/FluxMockup.py
```diff
@@ -51,8 +51,13 @@ def get_value(self):
         self.measure_flux()
         return self.current_flux_dict["flux"]
 
-    def measure_flux(self):
-        """Measures intesity"""
+    def measure_flux(self) -> None:
+        """
+        Measures intesity
+
+        Emits:
```
marcus-oscarsson, line 58: Not sure how to document this with google style doc strings ?, I found no reference for it ?

oldfielj-ansto: @marcus-oscarsson There are a couple of places I've found this documented, I'll link them below.
https://google.github.io/styleguide/pyguide.html
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Normally we wouldn't bother documenting a "None" return value, so we wouldn't include "Returns:" in this docstring, we also wouldn't include "Args:", as there are no input parameters for this method definition.

At least as far as the Google Python style guide is concerned, this would be all that is required in the docstring.

```python
def measure_flux(self) -> None:
    """Measures intensity"""
```

There isn't really any reason that you can't go above what the Google style requires, so including "Emits:" is probably fine.

marcus-oscarsson: Hm, yes, my comment was maybe a bit unclear :). I tried to find out how/if events are supposed to be documented with google style doc strings.  I think it would be useful to document the events as well.

rhfogh (approved):

PR #727, opened 2022-11-30
Logs - now with color

marcus-oscarsson:
- Added some colors in the logs, colors used can of course be discussed.
- Updated dependencies

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ lucid3 = { git = "https://github.com/mxcube/lucid3.git", branch = "master"}
 PyTango = { version = "^9.3.6", optional = true }
 python-ldap = "^3.4.3"
 requests = "^2.28.1"
+colorama = "^0.4.6"
 
 [tool.poetry.dev-dependencies]
 pylint = [ { version = "==2.13.9", python = "<=3.7.1" }, { version = "^2.15.6", python = ">=3.8" } ]
--- a/mxcubecore/__init__.py
+++ b/mxcubecore/__init__.py
@@ -4,6 +4,7 @@ import logging
 from logging.handlers import RotatingFileHandler
 import os
 import sys
+from colorama import Fore, Back, Style
 
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore import __version__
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -41,10 +41,13 @@ suds-py3 = "1.4.5.0"
 psutil = "^5.9.4"
 gipc = "^1.4.0"
 importlib_metadata = { version = ">=4.12.0", python = "3.7" }
-py4j = { version = "0.10.9.7", optional = true }
-f90nml = { version = "1.4.3", optional = true }
+py4j = "0.10.9.7"
+f90nml = "1.4.3"
```
oldfielj-ansto: Shouldn't the "py4j" and "f90nml" packages remain optional?

Seems like the only place they are used is in the GPHL workflow code.

You can choose to install optional extras via `pip install package[extra]`,
So if you needed to install MXCuBE Core with the GPHL dependencies in a development environment it would be `pip install -e ./mxcubecore[gphl]`.

pyproject.toml
```diff
@@ -41,10 +41,13 @@ suds-py3 = "1.4.5.0"
...
+lucid3 = { git = "https://github.com/mxcube/lucid3.git", branch = "master"}
```
oldfielj-ansto: Same with the "lucid3" package, wouldn't this be an optional extra?

marcus-oscarsson: Yes, you are right that they are not strictly necessary I can make them optional again. I  thought that the majority of us will be using those libraries and installing them for those who don't will probably not be much of a problem.

rhfogh: Hopefully a majority will ;-) But the code is set up so that those libraries are only imported if gphl_workflow is present in the beamline config, so there is no problem with leaving them optional, unless you think it makes for simpler maintenance to always include them.

marcus-oscarsson: I thought that it made the installation simpler but I don't mind making them optional, it was just to simplify things.

marcus-oscarsson:
Tests will pass when #729 is resolved/merged :)

pyproject.toml
oldfielj-ansto: It does simplify installation, I might be being a bit overzealous with trying to minimise dependency bloat.

marcus-oscarsson: :), thats alright I think the intention is appreciated and its good to be careful.

Ive added them back as extra dependencies, lets keep it like this and we will see how it goes. We can remove the optional bit if it turns out to be too much of a  hassle

PR #728, opened 2022-11-30
Simple utility for checking XML files

marcus-oscarsson:
Started to implement a very simple tool for checking XML files. For now its simply checks which class is used for configuration but this will of course be expanded. The idea is to create a configuration schema for each HardwareObject and then add a functionality to convert to YAML


rhfogh (approved):
I cannot really evaluate this, but for what it is worth, it looks fine.

marcus-oscarsson:
Well, there is not much to evaluate at the moment really :)

PR #729, opened 2022-11-30
Making sure the basic packages are installed, most importantly openldap

marcus-oscarsson:
The poetry install fails on the python-ldap installation if openldap is not installed

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,7 +15,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        conda env update --file conda-environment.yml --name base
+        #conda env update --file conda-environment.yml --name base
         pip install --upgrade pip
         pip install poetry
         poetry config virtualenvs.create false
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -23,4 +23,5 @@ jobs:
 
     - name: Run Pre-Commit
       run: |
-        poetry run pre-commit run --all-files
+        # Disable linting for now
+        # poetry run pre-commit run --all-files
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,9 +13,11 @@ jobs:
       with:
         python-version: "3.7"
 
+    - name: Install openldap
+      run: sudo apt-get install -y libldap2-dev libsasl2-dev slapd ldap-utils tox lcov valgrind
+
     - name: Install dependencies
       run: |
-        conda env update --file conda-environment.yml --name base
         pip install --upgrade pip
         pip install poetry
         poetry config virtualenvs.create false
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,7 +15,7 @@ jobs:
 
     - name: Install dependencies
       run: |
-        #conda env update --file conda-environment.yml --name base
+        conda env update --file conda-environment.yml --name base
         pip install --upgrade pip
         pip install poetry
         poetry config virtualenvs.create false
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -15,6 +15,7 @@ jobs:
 
     - name: Install dependencies
       run: |
+        conda env update --file conda-environment.yml --name base
         pip install --upgrade pip
         pip install poetry
         poetry config virtualenvs.create false
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,6 +20,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          conda env update --file conda-environment.yml --name base
           python -m pip install --upgrade pip
           python -m pip install poetry --user
           python -m poetry install
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,9 +18,11 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install openldap
+        run: sudo apt-get install -y libldap2-dev libsasl2-dev slapd ldap-utils tox lcov valgrind
+
       - name: Install dependencies
         run: |
-          conda env update --file conda-environment.yml --name base
           python -m pip install --upgrade pip
           python -m pip install poetry --user
           python -m poetry install
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -1,7 +1,0 @@
-name: mxcubcore
-channels:
-  - conda-forge
-dependencies:
-  - openldap
-  - pip
-  - poetry
--- a/conda-environment.yml
+++ b/conda-environment.yml
@@ -1,0 +1,7 @@
+name: mxcubcore
+channels:
+  - conda-forge
+dependencies:
+  - openldap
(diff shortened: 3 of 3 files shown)


oldfielj-ansto:
@marcus-oscarsson  Seems a bit redundant to have a secondary Conda environment for MXCuBE Core, when one is already setup for MXCuBE-Web, is this more for QT interface beamlines?

.github/workflows/pre-commit.yml
```diff
@@ -15,11 +15,13 @@ jobs:
 
     - name: Install dependencies
       run: |
+        conda env update --file conda-environment.yml --name base
```
oldfielj-ansto: A Conda environment is not required for pre-commit, it just perform static code analysis, no code is executed.

.github/workflows/tests.yml
```diff
@@ -20,6 +20,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          conda env update --file conda-environment.yml --name base
```
oldfielj-ansto: A Conda environment may be required if and when the tests require an OS package to be installed as a dependency, but as far as I can tell there is no such dependency and this is likely not required.

marcus-oscarsson: hm, I think python-ldap needs openldap (at least it did for me) which is a OS package. That's why I added it :)

.github/workflows/pre-commit.yml
marcus-oscarsson: `poetry install` below will failed when I ran it because there were no `openldap` installed.

marcus-oscarsson:
> @marcus-oscarsson Seems a bit redundant to have a secondary Conda environment for MXCuBE Core, when one is already setup for MXCuBE-Web, is this more for QT interface beamlines?

Hm, yes I agree. Is there maybe a better way to manage the OS libraries like openldap ?

marcus-oscarsson:
@oldfielj-ansto, I just gave it a try to make sure I was not completely out in the "outback" ;) So you can see that this run [run](https://github.com/mxcube/mxcubecore/actions/runs/3599800151/jobs/6063768954) fails without `openldap`. But there is perhaps another better way to install it (without using a conda environment)

marcus-oscarsson:
Okey, so I had to fiddle a bit with this ;), I finally removed the conda-environment file and are using `apt-get` to install the dependencies instead. Not certain if that is actually any better, although it removes the need for the `conda install` we need to repeat the `apt-get `line instead :)

.github/workflows/tests.yml
oldfielj-ansto: Yeah, so this looks to be down to the "python-ldap" package requiring additional OS packages.
The reason I didn't include this package initially, was due to it not being present in the "requirements.txt" file before moving to defining the dependencies in the "pyproject.toml" file.

If "python-ldap" is a hard requirement for any MXCuBE Core environment, maybe it should exist within a Conda environment.

oldfielj-ansto: The Conda version included in the container seems to be quite out of date, so the first step would be to update that.
```
 ==> WARNING: A newer version of conda exists. <==
  current version: 4.12.0
  latest version: 22.11.0

Please update conda by running

    $ conda update -n base -c defaults conda
```

Using Conda also means we can't use the default environment in the test container, we'll need to run Poetry install from within the created Conda environment.

The "pre-commit.yml" action YAML would look something like this.
```yaml
name: Linting & Code Quality
on:
  pull_request:
    types: [opened, reopened, synchronize]
jobs:
  pre-commit:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v3

    - name: Set up Python 3.7
      uses: actions/setup-python@v4
      with:
        python-version: "3.7"

    - name: Setup Conda environment
      run: |
        wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ./miniconda.sh
        /bin/bash ./miniconda.sh -y -b -p ./.venv && rm ./miniconda.sh
        ./.venv/bin/conda update -y -n base -c defaults conda
        ./.venv/bin/conda env update -f conda-environment.yml -n base --prune

    - name: Install Python dependencies
      run: |
        ./.venv/bin/python -m pip install --upgrade pip
        ./.venv/bin/poetry install

    - name: Run Pre-Commit
      run: |
        ./.venv/bin/pre-commit run --all-files
```

And the "tests.yml" action would look like this.
```yaml
name: Pytest
on:
  pull_request:
    types: [opened, reopened, synchronize]
    paths:
      - '**.py'
jobs:
  Python:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        python-version: ["3.7", "3.8", "3.9", "3.10"]
    steps:
      - uses: actions/checkout@v3

      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v4
        with:
          python-version: ${{ matrix.python-version }}

      - name: Setup Conda environment
        run: |
          wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ./miniconda.sh
          /bin/bash ./miniconda.sh -y -b -p ./.venv && rm ./miniconda.sh
          ./.venv/bin/conda update -y -n base -c defaults conda
          ./.venv/bin/conda env update -f conda-environment.yml -n base --prune

      - name: Install Python dependencies
      run: |
        ./.venv/bin/python -m pip install --upgrade pip
        ./.venv/bin/poetry install

      - name: Test with pytest
        run: |
          ./.venv/bin/pytest --no-cov
```

The extra steps to setup the Conda environment may slow the tests down, if necessary we could look at options such as the [conda-lock](https://github.com/conda-incubator/conda-lock) tool to speed things back up.

Another thing to consider here is how this new Conda environment will effect the existing MXCuBE-Web Conda environment, would the intended use for this environment be strictly for testing, development and QT deployments or would it be intended to install the Conda dependencies from this environment into the existing MXCuBE-Web Conda environment?

oldfielj-ansto:
> Okey, so I had to fiddle a bit with this ;), I finally removed the conda-environment file and are using `apt-get` to install the dependencies instead. Not certain if that is actually any better, although it removes the need for the `conda install` we need to repeat the `apt-get `line instead :)

Since it would seem that the "python-ldap" package is a hard requirement, it might be better to go with a Conda environment here, at least for development and testing purposes.

.github/workflows/tests.yml
marcus-oscarsson: Yes, the `python-ldap `requirement was for some reason missing in the original requirements, thats why you did not notice it earlier.

The conda environment for mxcubecore would as I see it only be used for testing and development of mxcubecore independently of either of the UI's. In most situations one would get these system dependencies from the Web or Qt environments. Its perhaps easier to simply keep the apt-get lines instead. I don't mind adding the conda environment back but I'm not sure it has any real advantage ? 

We could perhaps try out `mamba` its usually a bit faster than `conda`, if you still think its worth using a conda environment ?

oldfielj-ansto: OK, yeah so if it's mostly just an issue on the testing and development side of things, it will probably be easier to stick with "apt-get" installing the OS libraries.

Should probably mention these OS package requirements somewhere, maybe as a note on the "README.md" file.

marcus-oscarsson: Okey, lets do like this for now and then we can add a conda environment later if we have more of these or discover that its useful for some other reason.

PR #711, opened 2022-10-07
[WIP] Arinax md bzoom

danielhomsa:
The files being added are:
- One modification done to the Microdiff HardwareObject class inside the get_motors() method, the modification is regarding the returned dictionary so when there is no MiniKappa nozzle on the MD this is not returned as motor, avoiding a NoneType exception when using a different headtype.
- Arinax directory: Comtains the HardwareObjects for Arinax Microdiffractometers, Bzoom Camera device and zoom, as well as developments made for the Eiger detector, Vortex detector and other HardwareObjects developed during the work for the SSRF.
- SSRF directory: Contains HardwareObjects modified and developments for the SSRF and the X17UM beamline.

PR #730, opened 2022-12-05
Linting & Code Quality Refactor

oldfielj-ansto:
As was previously discussed in the developer meetings, I've adjusted the pre-commit configuration to bring down the number of files being refactored from 800+ to just over 100.

The changes look to almost entirely structural and shouldn't effect the functionality of the code having been refactored.
Tests are passing, but they don't cover the full scope of the files being touched here and I don't have an up to date MXCuBE-Web environment to fully test these changes.

I've disabled the GitHub action for Python3.11 tests for the moment, as one of the dependencies [(gipc)](https://pypi.org/project/gipc/) is causing the Python3.11 environment to break due to requiring an older version of [Gevent](https://pypi.org/project/gevent/), which is not compatible with Python3.11.

I'm also excluding the "utils/pyispyb_client" directory for now as it seems to be a work in progress, the configuration can be updated at a later date to include this directory.

Once this PR is merged and things have had some time to settle, we can look at turning some of the linting & code quality tools back on in a future PR.


marcus-oscarsson:
Looks good as far as I can see. I browsed through the files quite quickly but the linting fixes seemed to be manageable, mostly line breaks, quotes and spaces.

I did not know this problem with Gevent and Python 3.11 Ill try to investigate what we can do with that situation.

We said in one of our meetings that would wait for linting updates until the developers with pending updates have had the opportunity to submit PR's. I consider these linting fixes fairly easy to manage in merge so I would be fine with merging this, what do you think @beteva @rhfogh ?

oldfielj-ansto:
@marcus-oscarsson This isn't something that necessarily needs to be merged immediately if it's going to cause issues, we can wait for any large open PRs to be closed, then I can rerun the pre-commit scripts and merge it in.

marcus-oscarsson:
Sounds good, Id be happy to get it done as fast as possible, but we should of course wait if it causes someone an unnecessarily complicated situation :)

rhfogh:
I have no outstanding PRs. so while this may make merging harder, we cannot hold it up forever.

Just one thing: Would it be possible to exclude files starting with GPhL and the GPhL subdirectory? Then (with a minimum of help on the procedure) I could run the linting myself and save the trouble of merging?

oldfielj-ansto:
@marcus-oscarsson @rhfogh I've rebased and updated the config to exclude both code bases for Gphl and Arinax in [PR #711](https://github.com/mxcube/mxcubecore/pull/711/files#diff-16888d62f61bfc283027199e84a3fd6749e3925b5190171f5c4f323d897f4820).

@marcus-oscarsson The coverage report action was failing on the "python-ldap", due to missing OS packages, I've fixed that issue, but the action is now failing on what looks to be a permissions issue. This could be due to the actions not having pull-request write permissions assigned, as the coverage report action needs write access to the PR to post a comment.

marcus-oscarsson:
Ok, looks good to me :). Is there a way to give the "coverage report action" rights to add a comment ?

oldfielj-ansto:
@marcus-oscarsson I think there's an option for default access restrictions to actions in the repository settings.

marcus-oscarsson:
Hm that seems alright, all actions have read and write access, I think we need to use a token to get write access though, something like:

```
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
```

Ill see what I can do :)

oldfielj-ansto:
@marcus-oscarsson OK, the "GITHUB_TOKEN" secret is the default, but maybe that doesn't have enough access, so let's try a new token and see if that fixes the issue, if it doesn't, maybe GitHub has made a change and broken some functionality the action was using and I'll go do some more testing to try and debug it.

PR #734, opened 2022-12-19
complex equipment handling

meguiraun:
extra bits for handling the complex equipment ;)

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
@@ -92,6 +92,7 @@ class AbstractDiffractometer(HardwareObject):
         super().__init__(name)
         self.motors_hwobj_dict = {}
         self.nstate_equipment_hwobj_dict = {}
+        self.complex_hwobj_dict = {}
         self.username = name
         self.current_phase = None
         self.head_type = None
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -121,6 +122,14 @@ def init(self):
...
                     "No nstate (discrete positions) equipment configured"
                 )
+        for role in self["complex_equipment"].get_roles():
+            try:
+                self.complex_hwobj_dict[role] = self[
+                    "complex_equipment"
+                ].get_object_by_role(role)
+
+            except Exception as ex:
+                logging.getLogger("HWR").warning("Diffractometer: could not load complex equipment {}".format(ex))
```
beteva, line 132: Hi @meguiraun, could you, please, explain the need for this. Is there any equipment you need to control, but is not possible with the current design?  We've actually removed it some time ago to make the code less "complex". The idea is that everything handled by the diffractometer is either a motor type, or nstate type.

meguiraun: Then I definitely have an outdated reference to the camera in the diffractometer xml. I will remove that and close this PR. Thanks!

But, where should I define the camera hwobj? in the sample_view.xml?

Issue #440, opened 2019-12-09
Finish motor PR


marcus-oscarsson:
Finish the remaining issues with #436

marcus-oscarsson:
Also add read_only attribute

marcus-oscarsson:
Well done @beteva and @rhfogh we can close this one too

PR #740, opened 2023-02-09
Rhfogh develop

rhfogh:
Bug fixes, to match MXCuBEQt PR #440.
Works with Python 3.10

Includes changing collections.Callable to collections.abc.callable
and changing the type of some Qt call parameters from float to int.


marcus-oscarsson:
Hi @rhfogh. Looks like lots of good work ! Are you sure the PR is in the shape you would like it. Seems to be some WIP commits and some that have already been committed ?

rhfogh:
Hi @marcus-oscarsson . I'd take your advice. The WIP commits are all in GPhL-specific files. The problem is that there is no functioning GPhL workflow in develop. The versions that are functioning are all in various side branches (SOLIEIL, EMBL-HH, ALBA, ...). I am in a medium-term project of getting GPhL workflow actually running under mxcubeqt, as a preliminary step to getting it running in mxcubeweb. And until it is done there will be nothing completed to push. I suppose one could argue that I should have made a rigid separation between bug fixes on develop and my development project and cherrypicked bug fixes between them, but I have not done so, and it would be a lot of work, certainly at this point. Quite apart from anything else I would have to set up a separate testing environment for the official develop, which I have not got at the moment.

Considering that there is no working GPhL-functionality to break and that the WIP parts are all GPhL-specific I am happy with the commit, as a stepping stone to the future, and pretty sure it will not break anything for anybody. But I will listen if you tell me I ought to do things differently.

marcus-oscarsson:
Alright, well done for all the work :). The effective changes looks good to me. 

I usually simply remove the WIP messages and squash them together before making a PR but I know that its fiddly to do ;) its clear what it is like this so I think its fine.

How about the commits with two authors ?, are those really needed I think they are already merged once (I'm a bit surprised that they are present). I think your changes really starts with hash [0b36dcc](https://github.com/mxcube/mxcubecore/pull/740/commits/0b36dcc3c9d6205b37275d7138f3afc26d9ab61a) ?. I simply want to avoid  that we don't get the same commit twice.

I'm off for long weekend so Ill take a closer look on Tuesday, maybe @beteva has some hints in the meantime ?

rhfogh:
@marcus-oscarsson 
There is definitely room for improvement, here. I can see that a long list of my commits appear twice, and there is along list of commits from other people than me, from between 27/9 and 20/10 2022 that I do not know either why they are there. They clearly come from some kind of merge, but ought to be present already.

So, this is not how we want things to look - no quarrel there - but I do not know how  to do better. My practice is to keep working locally, rebase on develop at more or less regular intervals, and when rebasing gets too complex in one way or the other - merge. I do get more merges than I would expect, sometimes. Since I am working on mxcubecore in 2-3 different local copies (mxcubeqt, mxcubeweb, and masssif1) some confusion may creep in that way too. 

My one main goal is to get a final code state that actually contains what it is supposed to. Can anyone advise me ?

rhfogh:
I shall get rid of this one and replace it with another with the same content but a sane commit history

PR #742, opened 2023-02-15
Rhfogh develop

rhfogh:
Bug fixes, to match MXCuBEQt PR https://github.com/mxcube/mxcubecore/issues/440.
Works with Python 3.10

Includes changing collections.Callable to collections.abc.callable
and changing the type of some Qt call parameters from float to int.

Also includes a large step in the WIP of refactoring the GPhL workflow integration (which should have no effect on anything else).

Replaces and earlier PR with a bad and messy commit structure.


marcus-oscarsson:
Looks good !

PR #743, opened 2023-02-17
update python-ldap

meghdadyazdi:
Updating conda-environment.yml based [#954]( https://github.com/mxcube/mxcubeweb/issues/954) discussion.

I ran `poetry lock --no-update` to regenerate the `poetry.lock` but as it just changed the `content-hash` and nothing else so I don't commit it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ py4j = "0.10.9.7"
 f90nml = "1.4.3"
 lucid3 = "^3.0.0"
 PyTango = { version = "^9.3.6", optional = true }
-python-ldap = "^3.4.3"
+python-ldap = "^3.4.0"
 requests = "^2.28.1"
 colorama = "^0.4.6"
 


pyproject.toml
```diff
@@ -45,7 +45,7 @@ py4j = "0.10.9.7"
 f90nml = "1.4.3"
 lucid3 = "^3.0.0"
 PyTango = { version = "^9.3.6", optional = true }
-python-ldap = "^3.4.3"
+python-ldap = "^3.4.0"
```
marcus-oscarsson, line 48: Hm, should this not be in the _conda requirements file_, as you proposed earlier for `mxcube-web`

meghdadyazdi: To keep the mxcubecore independent. If mxcubecore needs to be used stand-alone, one have all the packages to run that (?)

marcus-oscarsson: I see, I'm wondering if we should not simply create a [conda-environment.yml](https://github.com/mxcube/mxcubeweb/blob/develop/conda-environment.yml) similar to the one in mxcube-web. It would be more consistent with the discussion we had in #954.  As I understand it this does not really solve your problem right ?

meghdadyazdi: Installing from `pyproject.toml` requires installing `gcc` for us therefore we install `python-ldap=3.4.0` from [conda-environment.yml](https://github.com/mxcube/mxcubeweb/blob/develop/conda-environment.yml) so in this way `python-ldap=3.4.0` will be installed with no problem in our site.

marcus-oscarsson: Oh, thats an interesting dependency relation so you are expecting that `python-ldap=3.4.0` is already installed from `mxcube-web` by conda and that that requirement is already met when you install mxcubecore via `pyproject.toml ` ?.

rhfogh:
I do not know if this helps, but to get something that worked for my docker PyQt install I had 
    apt-get install -y apt-utils curl git sudo wget \
    libldap2-dev libsasl2-dev libmemcached-dev ldap-utils tox lcov valgrind
in the DOckerfile (corresponds to system install),

openldap in conda-environment.yml

and 
python-ldap >= 3.4.3, < 4.0
in the requirements_mxcubecore file.

pyproject.toml
beenje: Yes, exactly. We use the same conda env to install mxcube-web and mxcubecore. `python-ldap` is a python dependency to mxcubecore so it should definitively be in the `pyproject.toml` here.

We just had to lower the version here from `^3.4.3` to `^3.4.0` because on conda-forge, `3.4.3` doesn't exist yet.
If a compatible version (`3.4.0`) is already installed in the conda env, `pip install mxcubecore` won't try to install another version.

marcus-oscarsson: I see, quite a subtle version relationship I would say. But I think the rest of us wont suffer from this

PR #745, opened 2023-02-28
Fix bug in HC_over_E

rhfogh:
Redone PR in order to clean up commit structure

PR also contains WIP work on GPhL-specific parameter querying


rhfogh:
ERROR. I cannot read. 
HC_OVER_E = h * c / e * 10e6
is actualy CORRECT, but I read it as 
HC_OVER_E = h * c / e * 1.0e6
which is wrong

WITHDRAWN. 
I shall change the line to 
HC_OVER_E = h * c / e * 1.0e7
just to avoid sim8ilar misunderstandings in the future

PR #746, opened 2023-03-02
Update base classes with type hinting and add test cases

oldfielj-ansto:
This PR adds type hinting and low level hardware object test cases.

Currently the goal is to add type hinting and test cases for each class defined in the modules [BaseHardwareObjects.py](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/BaseHardwareObjects.py) and [BaseHardwareObjects.py](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/CommandContainer.py), which has not yet been deprecated.

Once these test cases are completed it will be a lot simpler to create test cases for the abstract hardware objects, as all the low level logic can be patched over using MagicMocks.


oldfielj-ansto:
Still a work in progress, not ready to merge.

mxcubecore/BaseHardwareObjects.py
```diff
@@ -54,38 +58,43 @@ class DefaultSpecificState(enum.Enum):
...
         """Object initialisation - executed *before* loading contents"""
         pass
 
-    def init(self):
+    def init(self) -> None:
         """Object initialisation - executed *after* loading contents"""
         pass
 
-    def replace_object(self, role, new_object):
+    def replace_object(self, role: str, new_object: Self) -> None:
```
rhfogh, line 108: Could this give trouble if the new object is  different subclass? Not  that this function will ever be used much.

mxcubecore/BaseHardwareObjects.py
```diff
@@ -134,79 +141,132 @@ def procedures(self):
...
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.__properties_changed: Dict[str, Any] = {}
+        self.__properties_path: Dict[str, Any] = {}
 
-        self.__properties_changed = {}
-        self.__properties_path = {}
+    def set_property_path(self, name: Union[str, Any], path: Union[str, Any]) -> None:
```
rhfogh: This I do not understand. Why Union[str, Any] when str is part of Any? And when is a name not a string, anyway? The same question applies for other functions.

mxcubecore/BaseHardwareObjects.py
```diff
@@ -134,79 +141,132 @@ def procedures(self):
...
-        self.__references = []
-        self._xml_path = None
+        self._property_set: PropertySet
+        self.__objects_names: List[Union[str, None]] = []
+        self.__objects: List[List[Union["HardwareObject", None]]] = []
+        self._objects_by_role: Dict[str, "HardwareObject"] = {}
+        self._path: str = ""
+        self.__name: str = node_name
+        self.__references: List[Tuple[str, str, str, int, int, int]] = []
+        self._xml_path: Union[str, None] = None
```
rhfogh, line 227: Better: 
Optional[str]
Same in many other places

mxcubecore/BaseHardwareObjects.py
```diff
@@ -134,79 +141,132 @@ def procedures(self):
...
+    def set_user_file_directory(user_file_directory: str) -> None:
+        """Set user file directory.
+
+        Args:
+            user_file_directory (str): User file directory path.
+        """
         HardwareObjectNode.user_file_directory = user_file_directory
 
-    def name(self):
+    def name(self) -> str:
```
rhfogh, line 238: This should ABSOLUTELY be changed to a property or an attribute - but not necessarily in this PR

mxcubecore/BaseHardwareObjects.py
```diff
@@ -134,79 +141,132 @@ def procedures(self):
...
+    def get_xml_path(self) -> Union[str, None]:
+        """Get XML file path.
+
+        Returns:
+            Union[str, None]: XML file path.
+        """
         return self._xml_path
 
-    def __iter__(self):
+    def __iter__(self) -> Generator[Union["HardwareObject", None], None, None]:
```
rhfogh, line 279: I wiosh we could get rid of the '__' functions - but not necessarily in this PR.

mxcubecore/BaseHardwareObjects.py
```diff
@@ -365,10 +480,21 @@ def get_object_by_role(self, role):
                 else:
                     break
 
-    def objects_names(self):
+    def objects_names(self) -> List[Union[str, None]]:
```
rhfogh, line 501: Can we ever have None in that list (apart from the usual 'Optional...)?

rhfogh (approved):
This is very, very good. As yuou see I have some suggestions/questions, but I would not insist on holding this up - we can always come back to them.

In a way it is too bad to have done so much work on (e.g.) HardwareObjectNode when,. (IMHO) it ought to be deprecated. But it is great to have a start on introducing type annotations.

mxcubecore/BaseHardwareObjects.py
marcus-oscarsson: Looks really nice all of this. I have the same question as Rasmus for the **Any** bit

marcus-oscarsson:
Looks really nice !. I'm not entirely up to speed with typhints so maybe there is something I did not understand with how `Union[str, Any]` is used. Would not simply `Any` be enough ?. Or is it more to signal that it should be str but its in reality in many cases not a str and so `Union[str, Any]`  is partly used for documentation ?

marcus-oscarsson:
Hey @oldfielj-ansto, do you think you can rebase this when you get the time to solve the conflicts, then Ill merge. Thanks !

mxcubecore/BaseHardwareObjects.py
oldfielj-ansto: @rhfogh When methods are overridden or extended in sub-classes, the type hints need to be redefined.
If the method is inherited by the sub-class without being overridden, the "Self" type adapts so that it always references an instance of the current level class.

mxcubecore/BaseHardwareObjects.py
oldfielj-ansto: A lot of these type hints I came up with from looking at the current code base and seeing how people were using the methods, like what types they were passing through to the methods.

Basically if I wasn't sure I could rule out other types, I just picked the most likely type while also allowing other types via "Any". The "str" type being first in the union order implying that it is preferred.

The "Any" type is pretty much a catch all, that would include the "str" type, one reason to use:
```python
name: Union[str, Any]
```
Instead Of:
```python
name: str
```
Is that when used in a Pydantic model or via the [validate_arguments](https://docs.pydantic.dev/latest/usage/validation_decorator/) decorator, Pydantic will attempt to coerce the type of the parameter to each type referenced in the union block and adding "Any" to the end of that union block will prevent a validation error if type coercion fails, but that doesn't really apply here.

The "name" parameter should probably should be typed to "str", I'll update it.

mxcubecore/BaseHardwareObjects.py
oldfielj-ansto:
```python
Optional[str]
```
Is the shorthand equivalent for:
```python
Union[str, None]
```

Generally we'd use "Optional[str]" when typing method arguments, then "Union[str, None]" when typing method returns and elsewhere.

mxcubecore/BaseHardwareObjects.py
oldfielj-ansto: I did run into some logic paths while writing tests where items in the list could be "None".
That behaviour might be the result of a bug though.

oldfielj-ansto:
> This is very, very good. As yuou see I have some suggestions/questions, but I would not insist on holding this up - we can always come back to them.
> 
> In a way it is too bad to have done so much work on (e.g.) HardwareObjectNode when,. (IMHO) it ought to be deprecated. But it is great to have a start on introducing type annotations.

Thanks, I still think it's worth doing it while these objects aren't fully deprecated, even then, they may be around for a while yet for legacy compatibility reasons while facilities update their codebases.

oldfielj-ansto:
> Looks really nice !. I'm not entirely up to speed with typhints so maybe there is something I did not understand with how `Union[str, Any]` is used. Would not simply `Any` be enough ?. Or is it more to signal that it should be str but its in reality in many cases not a str and so `Union[str, Any]` is partly used for documentation ?

Thanks, the section @rhfogh commented on has been updated to simply be typed as "str".
As mentioned above, I mostly just did this to indicate that I was fairly sure a string was the expected type, but wasn't entirely sure, so left the typing open.

oldfielj-ansto:
> Hey @oldfielj-ansto, do you think you can rebase this when you get the time to solve the conflicts, then Ill merge. Thanks !

Sorry @marcus-oscarsson I've been pretty busy with other parts of the MX3 implementation, haven't had a chance till now to follow up on this.

I've tidied things up and rebased my changes onto the latest commit.

mxcubecore/BaseHardwareObjects.py
rhfogh: I think that could be a problem. For a function called replace_object, I do not think you can count on the new object being the same subclass as the current one.  There could be motor or camera or detector objects of different types that were exchanged.

mxcubecore/BaseHardwareObjects.py
rhfogh: Indeed. In general I would be in favour of deciding what types *should* be allowed and fixing the code, rather than allowing type 'Any', which sort of nullifies much of the advantage of typing.

mxcubecore/BaseHardwareObjects.py
rhfogh: OK (though 'Optional' does sound neater). Just by the by: If you write
self._xml_path: str = None
is that taken as an error, or understood to mean Optional[str]?

mxcubecore/BaseHardwareObjects.py
rhfogh: Might be worth checking

mxcubecore/BaseHardwareObjects.py
oldfielj-ansto: Hmm OK, fair point. I'll update the type hint from "Self" to "object" since these objects won't always share a common base.

mxcubecore/BaseHardwareObjects.py
oldfielj-ansto: Currently, violating any of these type hints will not raise an error, all we're doing here is hinting to IDE static type checkers what the intended usage is. At runtime, these type hints will be ignored and no exception will be raised.

We can change this behaviour to raise a validation exception instead, but I would be wary about doing this without some wider discussion, as enforcing these types could have the potential to break facility code.

Technically, both
```python
some_var: Optional[str]
# AND
some_var: Union[str, None]
```
are now considered deprecated ways of describing
```python
some_var: str | None
```

Reason I haven't used the newer [binary operator way of defining union types](https://peps.python.org/pep-0604/) is that it's only available from Python3.10 onward.

I'm happy to update the type hints to `Optional[str]` rather than `Union[str, None]`, but I guess the bigger issue is maintaining consistency. @marcus-oscarsson Maybe we should add a section to the contributing guidelines for type hinting and some general guidelines on their use.

mxcubecore/BaseHardwareObjects.py
marcus-oscarsson: Sounds like a well thought approach to me and clarifies whats preferred and that there might be some doubt of whats actually being passed around. I agree that it would be very nice to fix the code but I think this is a great intermediate step and as mentioned indicates where we need to look into what types are used.

mxcubecore/BaseHardwareObjects.py
marcus-oscarsson: Thanks for the update on type hints, I think we also had a lesson on the latest development here (at least me :) ) Sounds like a great idea to update the contributing guidelines.  It would be nice if we could increase the usage of type hints and then maybe turn on some sort of warnings (I guess thats possible), to indicate where we need to update the code.

marcus-oscarsson:
Great !, thanks for all this nice work !

PR #748, opened 2023-03-03
Minor fix and video streaming update

marcus-oscarsson:
This PR contains

 - Deletion of comments that was forgotten in last PR
 - Replaced Javascript based streaming with a more versatile python based solution

Removed the old node.js based stream relay  and replaced it with a new python based streamer now available
at: [video-streamer](https://github.com/mxcube/video-streamer). It can be customized to our needs and formats
can of course be added if needed. It currently supports MPEG1 (via FFMPEG) and MJPEG. It can definitely be improved so don't hesitate to come with ideas and suggestions. (Documentation is on its way)


rhfogh (approved):
Well, I cannot really check the video streaming stuff, but it looks fine.

PR #751, opened 2023-04-21
ESRF HO cleanup and optimisation

beteva:
1. BlissMotor uses try/except when checking for the state
2. ExporterMotor to use the application State for equipment, which has no proper motor state, but is used as a motor.
3. Sent parameters to bliss to optimise the load/unload sequence.
4. Cleanup.


rhfogh (approved):
Fine to a quick look

PR #755, opened 2023-05-14
Rhfogh develop wfgui

rhfogh:
Improved GPhL UI, and further tested/debugged


marcus-oscarsson:
Thanks, looks good.  Well done !. Looks like we have to dig into why the coverage report is not generated but its not for this PR.

PR #756, opened 2023-05-17
Rhfogh a1 develop

rhfogh:
This includes changes to GPhL-relevant files made in esrf-a1-develop, and new additions on the GPhL side.

It does *not* include [WIP] tmp (JBF 20230509) that allows for passing string parameters to functions requiring floats; this probllem should be considered separately


mxcubecore/HardwareObjects/XMLRPCServer.py
```diff
@@ -707,6 +707,7 @@ def addXrayCentring(self, parent_node_id, **centring_parameters):
     def addGphlWorkflow(self, parent_node_id, task_dict, workflow_id):
         """Add GPhL owrkflow to queue."""
         self.workflow_id = workflow_id
+        from mxcubecore.model import queue_model_objects as qmo
         gphl_model = qmo.GphlWorkflow()
         from mxcubecore.model import queue_model_objects as qmo
```
marcus-oscarsson, line 712: This looks a bit strange, seems like there is one import to many, the one below is wrong :)

marcus-oscarsson: Ill manage that as it was there before

PR #759, opened 2023-05-25
Added additional test cases to cover command container classes

oldfielj-ansto:
Even more test cases to cover command container classes, I've also added type hinting for these classes where possible.
Getting fairly close to having all the base classes covered by test cases.

There are some areas where full test coverage couldn't be achieved, some of the methods may need to be refactored before we can get to 100% coverage. I also ran into some areas of code where there seem to be missing imports, these might be present when MXCuBE-Web is also installed.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -93,7 +93,7 @@ class ConfiguredObject:
 
         self.name = name
 
-        self._objects: TOrderedDict[str, Union[Self, None]] = OrderedDict(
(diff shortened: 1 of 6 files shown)


marcus-oscarsson:
@oldfielj-ansto thanks !. Ill have a closer look during the day. Looks really nice from a first look :+1:

marcus-oscarsson:
Thanks ! Looks like there are some issues with the GitHub Virutal machine/env again, seems to be come and go somehow. Does anyone know what this is ?

PR #760, opened 2023-05-26
Fixes and additions done on ID29 and A1

marcus-oscarsson:
- Data collections and fixes done on ID29 for SSX experiments
 - Adapted to latest workflow code GPHL and ESRF for A1
 - General bug fixes


mxcubecore/HardwareObjects/ESRF/queue_entry/ssx_injector_collection.py
```diff
@@ -62,48 +60,47 @@ class SsxInjectorCollectionQueueEntry(SsxBaseQueueEntry):
...
-    NAME = "SSX Injector Collection (Lima1)"
+    NAME = "SSX Injector Collection"
     REQUIRES = ["point", "line", "no_shape", "chip", "mesh"]
 
     def __init__(self, view, data_model: SsxInjectorCollectionQueueModel):
         super().__init__(view=view, data_model=data_model)
 
     def execute(self):
         super().execute()
+        MAX_FREQ = 925.0
```
rhfogh: On general principles, a configuration constant like this should not be defined inside a function, but in a configuration file - or failing that at module or at least class level.

mxcubecore/HardwareObjects/PyISPyBClient.py
```diff
@@ -106,15 +109,44 @@ def get_current_beamline_values(self):
                 "wavelength": HWR.beamline.energy.get_wavelength(),
                 "resolution": HWR.beamline.resolution.get_value(),
                 "transmission": HWR.beamline.transmission.get_value(),
-                "detector_distance": HWR.beamline.detector.distance.get_value(),
+                "detector_distance": 177.70,
```
rhfogh: How do you justify putting hardwired constants inside a function that is named get_current_beamline_values - and that is not site-specific?

Also, if this is only called from ssx_base_queue_entry, should it have such a generic name?

mxcubecore/HardwareObjects/PyISPyBClient.py
```diff
@@ -106,15 +109,44 @@ def get_current_beamline_values(self):
...
+                "beam_size_x": 4,
+                "beam_size_y": 2,
+                "beam_shape": "gaussian",
+                "energy_bandwidth": 1,
             }
         )
 
-    def create_ssx_data_collection_group(self, session_id=None):
-        self._update_token()
+    def get_additional_lims_values(self):
```
rhfogh, line 124: For this function too - why hardwired values, and why such a generic name if only needed in a special case?

mxcubecore/HardwareObjects/mockup/MotorMockup.py
```diff
@@ -62,6 +62,9 @@ def init(self):
         """Initialisation method"""
         # get username, actuator_name and tolerance
         super().init()
+        global DEFAULT_VALUE
```
rhfogh: Surely this is unnecessary, given that the default value is already handled elseqhere?

mxcubecore/utils/conversion.py
```diff
@@ -44,8 +44,7 @@
     binary_type = bytes
 
 # Conversion from kEv to A, wavelength = HC_OVER_E/energy
-# NB This is * 10e7 (and NOT * 10e6), because the unit is A, not nm.
-HC_OVER_E = h * c / e * 1.0e7
```
rhfogh, line 48: Do you mind keeping this at 1.0e7? the values is the same as 10e6, but I think the latter is confusing - It tripped me up once already.

rhfogh (changes requested):
The sooner this is merged the better - even if it is still WIP it would make it a lot easier for me to make subsequent PRs if this was in. Just a few  points raised that I( think would merit a look or at least a comment first.

mxcubecore/HardwareObjects/PyISPyBClient.py
marcus-oscarsson: None other really except from that this is still WIP and I did not yet go over that bit. I've moved the values to either configuration or reading of hardware. At the moment the PyISPyBClient have only been used on ID29 were we for instance did not have a motorized detector so it was simply easier to hard coded this value, anyways ...

The intent with this method is otherwise to create something from where we can get all those values that we somehow use together, quite often,  but that comes from various places. Thus the name `get_current_beamline_values`, I though it could be moved to some more appropriate place like the `Beamline` object  in the future but as its still a bit of work in progress.

mxcubecore/HardwareObjects/mockup/MotorMockup.py
marcus-oscarsson: Not really, as `DEFAULT_VALUE` is defined on module level above. We need to do something like this if we still want to use "predefined" mock value and also be compatible with the  `default_value` property of AbstractActuator. I could simply removed the `DEFAULT_VALUE` above, if that is what you mean ?

mxcubecore/utils/conversion.py
marcus-oscarsson: Yes, but then we need to correct the comment "because the unit is A, not nm."  because the comment and the code is not consistent. `HC_OVER_E` is still in Angstrom

rhfogh: Indeed, the comment is obsolete - and wrong besides. I put it in as part of a fix back when I had not understood what was happening. By all means let's remove it.

mxcubecore/HardwareObjects/PyISPyBClient.py
rhfogh: All makes sense

mxcubecore/HardwareObjects/mockup/MotorMockup.py
rhfogh: Id say that we do not need either the 'global' declaration or the resetting. As I understand it, self.default_value is set in a superclass init. If that is not done, self.default_value will be reset to DEFAULT_VALUE by the code as it was:
        if self.default_value is None:
            self.default_value = DEFAULT_VALUE
            self.update_value(DEFAULT_VALUE)
Again as I understand it, you only need 'global' if you want to *modiify* the module value locally.
And resetting DEFAULT_VALUE would have the effect of  resetting the module level default to the configured default_value of the first mock motor to have one. Which seems to me to be not only unnecessary but wrong. We do not want to change teh moduile level default value at all, AFAICS.

mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -502,7 +587,7 @@ def oscilScanMesh(
             "startRasterScanEx",
         )
 
-        self._wait_ready(900)  # timeout of 15 min
+        self._wait_ready(900)  # timeout of 30 min
```
rhfogh, line 590: No real problem here - but is 900s not equal to 30min?

mxcubecore/HardwareObjects/PyISPyBClient.py
```diff
@@ -203,20 +239,20 @@ def create_ssx_data_collection(
...
                 "resolutionAtCorner": 0.0,
-                "flux_end": 0.0,
-                "detector_id": 4,
-                "startTime": datetime.datetime.now(),
-                "endTime": datetime.datetime.now(),
-                "repetitionTate": 0.0,
-                "energyBandwidth": 0.0,
-                "monoStripe": "mono_stripe_example",
+                "flux_end": extra_lims_values.flux_end,
+                "detector_id": 83,
```
rhfogh: Ideally that should be in configuration and not hardwired in.

rhfogh (approved):
Nice and welcome work. Please deal with remaining comment in what you think is the best way, and then you are welcome to merge

rhfogh:
Thanks!

PR #766, opened 2023-06-08
Fixed casing issue, camel to snake

marcus-oscarsson:
Fixed casing of `rejectCentring`


rhfogh (changes requested):
This also needs changing in core/components/samplechanger.py (line 211), and possibly also in docs/source/SampleCentring.rst

PR #767, opened 2023-06-21
PR is For Using Difractometer in Plate Mode - Making PlateManipulator available in Mxcube-web

jbflo:
This PR is for When Using Difractometer in Plate Mode 
This includes changes In PlateManipulator class and Mockup

Should not have breaking Change ,
 PR  #1005 in The MXCuBE-web is related to this one .

### DIFF
--- a/mxcubecore/HardwareObjects/PlateManipulator.py
+++ b/mxcubecore/HardwareObjects/PlateManipulator.py
@@ -159,11 +159,6 @@ class Drop(Container):
         sample = self.get_components()
         return sample[0]
 
-    # def get_index(self):
-    #    """
-    #    Descript. Drop index is relative to the row
-    #    """
-    #    return self._well_no
 
 
 class Cell(Container):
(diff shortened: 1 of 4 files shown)


beteva:
Dear @jbflo, thanks for this PR. Unfortunately the new  PlateManipulatorMaintenance  is incompatible with the latest mxcubecore standarts, namely inheriting from Equipment is deprecated. Could you, please, change the file so it corresponds to the latest mxcubecore standards (you may have a look at the contributing guidelines).

jbflo:
Dear @beteva 
Thanks for your commented.
I had a look at the contributing guidelines & mxcubecore/deprecated/,  but could not really see the information that will give a me a clue.

 I noticed the old maintenance classs are still implemented as Equipment.

will this change  be enough  ?

image


Many Thanks
JB

rhfogh:
This is too far from what I know about for me to review. Just a question: I notice that the PlateManipulator class seems to be strongly coupled to the Crims class. Are plate manipulators in general always attached to Crims? Or are there people who use plate manipulators who might not have Crism installed? Should the Crims dependency be moved to a site-specifric class?

jbflo:
You Probably right @rhfogh, CRIMS is not mandatory for the Class to work. You can still use the class without CRIMS .
EMBL Hambourg who started this had CRIMS and EMBL Grenoble/"ESRF" who is currently using it has CRIMS.
MAX LAB @JieNanMAXIV who are maybe Interested at the  PlateManipulator probably not.  We can maybe think of moving the Crims dependency  to a site-specifric class. later

PR #768, opened 2023-06-23
Linting and minor fixes

marcus-oscarsson:
While testing I discovered a few things that needed to be fixed.

- Delegating the retrieval of the image filename to the detector `HardwareObject` previously in the `PathTemplate`. However the later knows nothing about HDF5 file structure such as number of images per file or how the images are stored, so its more appropriate to have this method  on the detector.

- Some linting and camel to snake case issues fixed 

- Added function `adxv_notify` on AbstractMultiCollect that can be used for following images

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -1567,15 +1567,7 @@ class PathTemplate(object):
         return prefix
 
     def get_image_file_name(self, suffix=None):
-        template = "%s_%s_%%0" + str(self.precision) + "d.%s"
-
-        if suffix:
-            file_name = template % (self.get_prefix(), self.run_number, suffix)
-        else:
-            file_name = template % (self.get_prefix(), self.run_number, self.suffix)
-        if self.compression:
-            file_name = "%s.gz" % file_name
-
+        file_name = HWR.beamline.detector.get_image_file_name(self)
         return file_name
 
     def get_image_path(self):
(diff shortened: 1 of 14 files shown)


mxcubecore/model/queue_model_objects.py
```diff
@@ -1567,15 +1567,7 @@ def get_prefix(self):
...
-        template = "%s_%s_%%0" + str(self.precision) + "d.%s"
-
-        if suffix:
-            file_name = template % (self.get_prefix(), self.run_number, suffix)
-        else:
-            file_name = template % (self.get_prefix(), self.run_number, self.suffix)
-        if self.compression:
-            file_name = "%s.gz" % file_name
-
+        file_name = HWR.beamline.detector.get_image_file_name(self)
```
meguiraun, line 1570: why this change? I always assumed the PathTemplate is the one deciding the name

marcus-oscarsson: Yes, it used to be :). As i described above I decided to delegate it to the detector because its the object that knows the details of how the files are written, size of data set, format and other internals that are needed to get the image filename. The `PathTemplate` is more thought of as a template to be filled out, so it follows this logic somehow 

So the `PathTemplate` is decoupled from the detector used, perhaps one uses different detectors

meguiraun: ahaaa... sorry! I went directly to the code 😋

marcus-oscarsson: :), too much into the code :+1:

PR #769, opened 2023-06-26
Allow open/close methods to have timeout

beteva:
Allow the AbstractShutter to use the  AbstractNState class timeout mechanism.
Cleanup the test and the Mockup class


marcus-oscarsson:
Thanks :)

meguiraun (approved):
thanks!

PR #771, opened 2023-07-06
Fixed bug in new handling of image file names

rhfogh:
This fixes a bug in the new handling of image file names.


marcus-oscarsson:
Oh yes, thanks @rhfogh. I noticed this yesterdays as well and was preparing a PR. Thanks again for the fast reaction !

PR #773, opened 2023-07-11
Rhfogh develop

rhfogh:
Integration between MiniKappaCorrection and GPhL transcal; Updated to GPhL June 16 workflow branch.

Apart from some minor fixes and reformatting should not affect non-GPhL operation


marcus-oscarsson:
Thanks !

PR #774, opened 2023-07-18
Add test for other than open/close values.

beteva:
Implemented more values in the ShutetrMockup to allow different than OPEN/CLOSED values to be set and tested.
Reversed a line in AbstractShutter, removed by inadvertence.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractShutter.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractShutter.py
@@ -43,6 +43,7 @@ class AbstractShutter(AbstractNState):
     """Abstract base class for shutter type objects."""
 
     __metaclass__ = abc.ABCMeta
+    VALUES = BaseValueEnum
 
     @property
     def is_open(self):
--- a/mxcubecore/HardwareObjects/mockup/ShutterMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/ShutterMockup.py
@@ -40,6 +40,7 @@ class ShutterStates(Enum):
 class ShutterMockup(AbstractShutter, ActuatorMockup):
     """
     ShutterMockup for simulating a simple open/close shutter.
+    Fake some of the states of the shutter to correspong to values.
     """
 
     SPECIFIC_STATES = ShutterStates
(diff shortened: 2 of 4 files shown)


rhfogh (changes requested):
This may be just me misunderstanding, but near as I can see, AbstractShutter.BaseValueEnum is not imported or used anywhere, Which would mean that the BaseValueEnum used in ShutterMockup comes from AbstractNState, and contains only the value UNKNOWN. Which does not look right.

beteva:
> This may be just me misunderstanding, but near as I can see, AbstractShutter.BaseValueEnum is not imported or used anywhere, Which would mean that the BaseValueEnum used in ShutterMockup comes from AbstractNState, and contains only the value UNKNOWN. Which does not look right.

In AbstractShutter, there is a new BasaeValueEnum defined, which has OPEN and CLOSE members on top of the UNKNOWN.  This is what makes it shutter, rather than simply  NState - you guarantee OPEN and CLOSE. The ShutterMockup can be used as a model to implement other values, if needed, e.g. at ESRF we need MOVING and EXTERNAL.

rhfogh:
Yeah, I did notice that, but AFAICS the new BaseValueEnum is not used anywhere. It is defined in AbstractShutter.py, but it is not used inside that file, and it is not imported anywhere else. All imports are of AbstractNState.BaseValueEnum - and since BaseValueEnum is not inside the class but only in the file, I cannot see how it could  be inherited anywhere either.
When ShutterMockup references self.VALUES, near as I can see it inherits the VALUES class attribute that is set in AbstractNState, where BaseValueEnum is defined as containing only "UNKNOWN". 

Would you not need a line saying
    VALUES = BaseValueEnum
in AbstractShutter?

beteva:
> Would you not need a line saying VALUES = BaseValueEnum in AbstractShutter?

yes, there is such a line, right after the declaration of the class

rhfogh:
Me bad. I was looking in the code (locally) of the develop branch, not of the shutter_test branch.

My apologies.

beteva:
> Me bad. I was looking in the code (locally) of the develop branch, not of the shutter_test branch.
> 
> My apologies.

No problem. With the previous PR this line has been removed by inadvertence. I've put it back.

PR #776, opened 2023-07-24
Fixed bug in recent commit

rhfogh:
The 'Open' button did not work in the Qt mock version; removing ShutterMockup._set_value restored previous function

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/ShutterMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/ShutterMockup.py
@@ -19,7 +19,6 @@
 
 """ Mockup shutter implementation"""
 
-from warnings import warn
 from enum import Enum, unique
 from mxcubecore.HardwareObjects.abstract.AbstractShutter import AbstractShutter
 from mxcubecore.BaseHardwareObjects import HardwareObjectState
--- a/mxcubecore/HardwareObjects/mockup/ShutterMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/ShutterMockup.py
@@ -63,7 +63,3 @@ class ShutterMockup(AbstractShutter, ActuatorMockup):
         )
         self.VALUES = Enum("ValueEnum", values_dict)
 
-    def _set_value(self, value):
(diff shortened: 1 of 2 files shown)


mxcubecore/HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -63,7 +63,3 @@ def _initialise_values(self):
         )
         self.VALUES = Enum("ValueEnum", values_dict)
 
-    def _set_value(self, value):
```
rhfogh, line 66: With this function, valueChanged was not emitted and so Qt UI was not updated. Now the function is removed, you instead trigger ActuatorMockup._set_value, with the function body 
        self.update_value(value)
which seems to work

beteva: @rfogh, did you run the tests.

rhfogh: I certainly did. With the develop version I could not run a session, because clicking 'Open' had no effect. When I commented out that function the mock installation again worked as normal. I would not start removing bits of an already accepted PR without just cause. Did you get a different result? It would be interesting to see what was different between our two set-ups.

Actually, I could imagine that there might be reasons that you do need a _set_value function there (possibly having to do with there being an abstract function in the superclass, testing and linting, etc.), but you would be the person to know about that. In that case, maybe one could put a copy of the _set_value function from the superclass, where is uses update_value?

beteva: You are right, adding _set_value was more of a concern to overload an abstract method and also to serve as an example. When implementing a real hardware, this is the method to fill in with the real action. I also see why the update_value did not work with the mock-up - assigning the _nominal_value makes update_value believe that the value has not changed.
Thank you for sorting this out.

rhfogh: Would you want me to retire the PR and instead modify the _set_value function so it is a copy of the one in the superclass?

PR #778, opened 2023-08-14
make sure poller callback is called

elmjag:
Trigger poller's async watcher, so that new values que is processed. The watcher will invoke poller callback with new value(s).


marcus-oscarsson:
Thanks !

PR #781, opened 2023-08-25
Updates to XRF and EnergyScan abstract classes, Characterisation and minor fixes

marcus-oscarsson:
Updates to XRF and EnergyScan abstract classes, Characterisation and minor fixes:

 - General fixes related to linting and name changes  
 - Updates made to XRF and EnergyScan abstract classes.
 - Speed up and options added to characterization
 - Handling of run number at directory level


mxcubecore/HardwareObjects/LNLS/LNLSCollect.py
```diff
@@ -138,30 +172,38 @@ def loop(self, owner, data_collect_parameters_list):
...
+                print("[SCAN-UTILS] output : " + stdout.decode("utf-8"))
                 # stderr
-                logging.getLogger("HWR").error('[SCAN-UTILS] errors : ' + stderr.decode('utf-8'))
-                print('[SCAN-UTILS] errors : ' + stderr.decode('utf-8'))
+                logging.getLogger("HWR").error(
+                    "[SCAN-UTILS] errors : " + stderr.decode("utf-8")
+                )
+                print("[SCAN-UTILS] errors : " + stderr.decode("utf-8"))
 
             except BaseException:
```
rhfogh, line 208: That ought to be just plain 'except:'. But even 'except Exception:' would be better than using BaseException

mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -371,69 +371,37 @@ def set_light_in(self):
         logging.getLogger("HWR").info("Moving backlight in")
         light_hwobj = self.get_object_by_role("BackLightSwitch")
         light_hwobj.set_value(light_hwobj.VALUES.IN)
-        self.wait_ready(20)
+        self._wait_ready(20)
```
rhfogh, line 374: For curiosity (no objection and nothing wrong with it) why use _wait_ready instead of HardwareObjectMixin.wait-ready? If there is something wrong or insufficient with the latter maybe we ought to change it, so the same system could be used throughout

mxcubecore/HardwareObjects/Session.py
```diff
@@ -189,18 +210,14 @@ def get_base_process_directory(self):
             self.get_base_data_directory(), self.processed_data_folder_name
         )
 
-    def get_image_directory(self, sub_dir=None):
+    def get_image_directory(self, sub_dir: str = "") -> str:
```
rhfogh, line 213: Are we officially moving to that instead of (or as an alternative to) Google annotations for setting types? I'd be happy with that, but what is the status?

mxcubecore/HardwareObjects/XSDataMXCuBEv1_4.py
```diff
@@ -182,7 +181,7 @@ def setImageFile(self, imageFile):
                 "ERROR! XSDataMXCuBEDataSet.setImageFile argument is not list but %s"
                 % imageFile.__class__.__name__
             )
-            raise Exception(strMessage)
+            raise BaseException(strMessage)
```
rhfogh, line 184: Why BaseException?

rhfogh (approved):
Looks good, and nice with the linting/cleanup.

My only major beef is why the use of BaseException - I thought that was supposed to be a sort of underlying base class, not for actual use - but since it surely works there is no reason to hold up the PR for thata.

mxcubecore/HardwareObjects/XSDataMXCuBEv1_4.py
marcus-oscarsson: I don't really know this is automatically generated code that comes with EDNA, Ill check with @olofsvensson there might be a reason for it but I otherwise agree that it should be simply `Exception`

mxcubecore/HardwareObjects/Session.py
marcus-oscarsson: I think one usually omits the type in the doc string if the method is type hinted, its otherwise a bit of duplication and the tools should be able to pick it up. For instance https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

mxcubecore/HardwareObjects/Microdiff.py
marcus-oscarsson: Its because this `_wait_ready` has some very specific MD related logic in it. I think we could/can make it usre the ready_event and use wait_ready, but I think we thought this was simpler.

mxcubecore/HardwareObjects/LNLS/LNLSCollect.py
marcus-oscarsson: Oh, yes. Its a line that its not really touched by this PR so I guess I did not notice but I can change it. Its also a LNLS site specific file, I would in anyways not mind updating it in this case :)

mxcubecore/HardwareObjects/Session.py
rhfogh: Which would mean that you have the choice where to put it. That would be fine with me, but I would also be wiling to standardise on either variant, if we agreed to do so.

mxcubecore/HardwareObjects/Microdiff.py
rhfogh: Makes perfect sense - and does not suggest we need to refactor anything on either side.

mxcubecore/HardwareObjects/LNLS/LNLSCollect.py
rhfogh: Cool. Thanks for answering all those queries.

mxcubecore/HardwareObjects/Session.py
marcus-oscarsson: Oh yes, good point, we should revisit our contributing guidelines in anyways (long due) and maybe add something about it

marcus-oscarsson:
:), thanks for the review. Enjoy the WE !

PR #782, opened 2023-08-29
P11 update 2 - Separate set of changes to some general files

agruzinov:
Here is another part of P11 files and additional cherry-picking on the P11-related HO as a part of cumulative update from work on P11 with Bixente and Jan. 

There were also couple of commits on queue_entry.py, like disabling the wait for a result from EDNA because it is not fully implemented yet. Where it could be placed? Should there be separate entry for the queue entry implementation on P11?

For example:

  collection_plan = strategy_result.getCollectionPlan()
            else:
                collection_plan = None
            log.debug("collection plan is %s. IGNORD FOR NOW" % str(collection_plan))

            # if collection_plan:
            if False:
                if char.auto_add_diff_plan:
                    # default action
                    self.handle_diffraction_plan(self.edna_result, None)

Thanks!

Andrey


mxcubecore/Command/Tango.py
```diff
@@ -324,9 +325,11 @@ def get_info(self):
 
     def update(self, value=Poller.NotInitializedValue):
 
-        if value == Poller.NotInitializedValue:
+        if isinstance(value, numpy.ndarray):
```
rhfogh: 'value=value' is perfectly OK, but a no-op. ONe could jsut check for tuple and NOtInitialised and otherwise do nothing.

mxcubecore/Command/Tango.py
```diff
@@ -340,7 +343,11 @@ def get_value(self):
         else:
             value = self.device.read_attribute(self.attribute_name).value
 
-        if value != self.value:
+        
+        if isinstance(value, numpy.ndarray):
+            if any(value != self.value):
```
rhfogh: I suspect that this will not work. I tested, and any(ndarray1 != ndarray2)  gives an error.
(ndarray1 != ndarray).any()  gives a boolean, at least, but then (4 == ndarray1).any() also gives a boolean.

Numpy arrays have effective broken Python comparison, list membership etc., so if you have to have values that are numpy arrays you need to be exceedingly careful how you handle them. Nested lists might be easier.

mxcubecore/HardwareObjects/DESY/P11Session.py
```diff
@@ -104,42 +111,45 @@ def get_current_proposal_number(self):
...
 
     def read_info(self, filename):
         buf = open(filename).read()
+    def load_info(self,filename):
         try:
             # json_str = buf[buf.index("{"):buf.index("}")+1]
             # return json.JSONDecoder().decode(json_str)
             json_str = buf
             return json.loads(json_str)
+            return json.loads(open(filename).read())
```
rhfogh: The second return statement never gets executed.

mxcubecore/HardwareObjects/TangoMotor.py
```diff
@@ -142,40 +177,59 @@ def motstate_to_state(self, motstate):
         return state
 
     def motor_state_changed(self, state=None):
-        if state is None:
-            state = self.chan_state.getValue()
+        if state == None:
```
rhfogh: On principle this should be 'is'. There is only one None, and '==' could trigger a __eq__ function on a state object, which coudl even return True, in theory.

mxcubecore/model/queue_model_objects.py
```diff
@@ -331,6 +331,8 @@ def get_display_name(self):
         display_name = HWR.beamline.session.get_default_prefix(self)
         if self.lims_code:
             display_name += " (%s)" % self.lims_code
+        if self.get_name() != "":
```
rhfogh: Maybe consider'if self.get_name():' Which result would you want if self.get_name() returned None?

mxcubecore/HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -51,14 +51,9 @@ def init(self):
...
-        """Add additional, known in advance states to VALUES"""
-        values_dict = {item.name: item.value for item in self.VALUES}
-        values_dict.update(
-            {
-                "MOVING": "In Motion",
-                "UNUSABLE": "Temporarily not controlled",
-            }
-        )
-        self.VALUES = Enum("ValueEnum", values_dict)
+    def open(self):
```
rhfogh: open and close should have the same signature as the function they override. Ideally the mockup should handle timeouts as well (why can't it?), but failing that the function should at least take a timeout value, even if it does not do anything with it. In fact, why are these function s necessary?

mxcubecore/HardwareObjects/mockup/MotorMockup.py
```diff
@@ -119,3 +119,7 @@ def _move(self, value):
             self.update_specific_state(None)
 
         return value
+
```
rhfogh, line 122: @beteva : Maybe we should consider adding this function to AbstractMotor?

mxcubecore/HardwareObjects/mockup/BeamMockup.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, name):
 
         self._beam_size_dict["slits"] = [9999, 9999]
         self._beam_size_dict["aperture"] = [9999, 9999]
-        self._beam_position_on_screen = [318, 238]
+        self._beam_position_on_screen = [680, 512]
```
rhfogh, line 41: Should this be congfigured (eventually, not now).

mxcubecore/HardwareObjects/ISPyBClient.py
```diff
@@ -1598,6 +1604,9 @@ def disable(self):
     def enable(self):
         self._disabled = False
 
+    def is_connected(self):
```
rhfogh, line 1607: This should be put in the ISPyBClient mockups as well.

mxcubecore/Command/Tango.py
agruzinov: I've added the changes, this might also fix the sometimes happenning infinite loop error. Some testing will be needed.

mxcubecore/HardwareObjects/DESY/P11EDNACharacterisation.py
```diff
@@ -0,0 +1,612 @@
...
+
+import triggerUtils
+
+# from edna_test_data import EDNA_DEFAULT_INPUT
+# from edna_test_data import EDNA_TEST_DATA
+
+
+class P11EDNACharacterisation(AbstractCharacterisation):
+    def __init__(self, name):
+        super(EDNACharacterisation, self).__init__(name)
```
rhfogh: EDNACharactersation is not defined and is not one of the superclasses.
Also you just need super() with the newer syntax.
Actually it might be easier if you did make this a subclass of EDNACharactersation, since there is a lot of duplication between those classes. But As this is WIP, I gather, this need not hold up the merge.

mxcubecore/Command/Tango.py
rhfogh: I had in mind something more like

       lif value == Poller.NotInitializedValue:
            value = self.get_value()
        elif isinstance(value, tuple):
            value = list(value)

There is no point in setting value = value - it does nothing.
A different point is that having a value that is a numpy array is a high bug risk because those behave in such strange ways.  If you need it you must do the necessary checking, of course, but it woudl be easier to have teh value as a list or tuple (lisat, being mutable, have their own problems when they are exported)

rhfogh (changes requested):
So far so good.

mxcubecore/Command/Tango.py
agruzinov: Thanks for noticing that!

mxcubecore/Command/Tango.py
```diff
@@ -346,11 +346,10 @@ def get_value(self):
 
         
         if isinstance(value, numpy.ndarray):
-            if any(value != self.value):
+            if (value != self.value).any():
```
rhfogh: This will work.

But, because of the peculiar behaviour of numpy the following all return True:

(numpy.zeros((3,2)) == numpy.array((0,0,0))).any()

(0 == numpy.zeros((3,2)) ).any()

If you use numpy arrays as values, you need to manually ensure that you will not get into problems with this.

agruzinov: I've added check using numpy.array_equal(value, self.value). This seems to work better.

mxcubecore/model/queue_model_objects.py
agruzinov: I've added the change, but maybe checking explicitly against empty string was intentional. If None, it should return default prefix.

mxcubecore/Command/Tango.py
rhfogh: That is neat. I did not realise one could do that.

mxcubecore/HardwareObjects/DESY/P11EDNACharacterisation.py
agruzinov: I wanted to do it anyway and now it is a good excuse :) Thanks!

mxcubecore/Command/Tango.py
```diff
@@ -340,8 +345,12 @@ def get_value(self):
         else:
             value = self.device.read_attribute(self.attribute_name).value
 
-        if value != self.value:
-            self.update(value)
+        
+        if isinstance(value, numpy.ndarray):
+            if not numpy.array_equal(value, self.value):
+                self.update(value)
+            elif value != self.value:
```
rhfogh: I think these last two lines should be de-indented. It si if value is not a numpy array that you want to do the value != self.value check, right?

mxcubecore/Command/Tango.py
```diff
@@ -326,9 +327,13 @@ def update(self, value=Poller.NotInitializedValue):
 
         if value == Poller.NotInitializedValue:
             value = self.get_value()
-        if isinstance(value, tuple):
+        elif isinstance(value, tuple):
             value = list(value)
 
+        #Explicit conversion to list to avoid numpy array.
+        if isinstance(value, numpy.ndarray):
```
rhfogh: You could do this as an elif to the previous if statement, but this of course works equally well. Anyway, the whole is a nice solution to the numpy problem.

mxcubecore/HardwareObjects/DESY/P11Session.py
```diff
@@ -104,42 +111,46 @@ def get_current_proposal_number(self):
...
             return None
+            if ety.is_file() and ety.name.startswith('commissioning-metadata'):
+                self.beamtime_info.update( self.load_info(ety.path) )
+                self.beamtime_info['rootPath'] = PATH_COMMISSIONING
 
         return self.read_info(fname)
 
     def read_info(self, filename):
         buf = open(filename).read()
+
```
rhfogh, line 137: I still think this one is wrong
read_info reads into the variable 'buf' and does nothing with it.
Meanwhile load_info starts with a variable called 'buf' that has not been initialised
It looks like read_info and load_info should be combined into a single function, that should be used instead of both (with whichever name you prefer)

mxcubecore/HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -51,14 +51,9 @@ def init(self):
...
-        """Add additional, known in advance states to VALUES"""
-        values_dict = {item.name: item.value for item in self.VALUES}
-        values_dict.update(
-            {
-                "MOVING": "In Motion",
-                "UNUSABLE": "Temporarily not controlled",
-            }
-        )
-        self.VALUES = Enum("ValueEnum", values_dict)
+    def open(self, timeout=0):
```
rhfogh: These functions should be unnecessary - they are identical to the functions in the superclass. You could just remove them

Also there seems to be a typo: it should be 
    def close(self, timeout=0):
not
    def close(self, open(self, timeout=0)):

mxcubecore/Command/Tango.py
agruzinov: Yes! I was too fast. Thanks!

mxcubecore/Command/Tango.py
agruzinov: Good point. Modified.

mxcubecore/HardwareObjects/DESY/P11Session.py
agruzinov: Agree, a new combined function is added.

mxcubecore/HardwareObjects/mockup/ShutterMockup.py
agruzinov: I've lost a bit. The commit here 119fee4b6fc7268d974934443b13016f1feceda6 looks different from the current version in the dev branch. Maybe I should just revert the changes and live the last version. I guess those changes has been made originally to simulate the shutter at the beginning.

agruzinov: sorry, in this commit: d8189637212ba14ce6dfb01705436a71ca513793

rhfogh (approved):
Guess we are there

agruzinov:
Thanks for all suggestions!

mxcubecore/HardwareObjects/abstract/AbstractShutter.py
```diff
@@ -45,6 +46,14 @@ class AbstractShutter(AbstractNState):
...
     VALUES = BaseValueEnum
 
+    def __init__(self, name):
+        AbstractNState.__init__(self, name)
+
+    def init(self):
+        """Initilise the predefined values"""
+        self.set_value(self.VALUES.UNKNOWN)
+        AbstractNState.init(self)
+
```
beteva: Looks like you have some old version of the shutter, which wants to put back things which have been removed or added for the AbstractShutter.py and ShutterMockup.py

mxcubecore/HardwareObjects/mockup/ShutterMockup.py
```diff
@@ -51,14 +51,9 @@ def init(self):
         self.update_value(self.VALUES.CLOSED)
         self.update_state(self.STATES.READY)
 
-    def _initialise_values(self):
-        """Add additional, known in advance states to VALUES"""
```
beteva: Looks like you have some old version of the shutter, which wants to put back things which have been removed or added for the AbstractShutter.py and ShutterMockup.py

beteva (changes requested):

mxcubecore/HardwareObjects/mockup/ShutterMockup.py
beteva: @agruzinov This has been changed in PR#774, merged on 19/07. Could you, please, make sure that AbstractShutter and ShutterMockup in your PR are up to date.

agruzinov: Thanks for noticing that. I was synching with the develop branch, something went wrong on the rebase maybe.

beteva (approved):

beteva:
@agruzinov thanks for updating the Tango part.

agruzinov:
Main contributions came from Bixente and Jan. I've jumped in later :) Hopefully, it can be merged soon.

rhfogh:
I'd say you could merge it now, if you want - both Antonia and I have OK'ed it. Unless you think that Marcus has to express himself as well?

marcus-oscarsson:
Looks good :)

PR #787, opened 2023-09-07
Default editor configurations and environment for development

marcus-oscarsson:
Hi

 - A conda file with the  basic dependencies for development
 - Editor configuration files with sensible defaults 
 
If you prefer that we don't commit the .vscode folder Ill simply rename it to be a "template/example" folder


rhfogh:
Looks fine.

Two questions:

Why are we not trimming tailing whitespace?
Which ediitor/editors is this used for (I use Pycharm which has its own config system)?

dist/mxcubecore-1.0.0.tar.gz
fabcor-maxiv, line 1: Was this file committed on purpose? Could it have been a mistake?

marcus-oscarsson:
@rhfogh Ah, no we should trim white-spaces

I think there are few editors that can use it, I just added it for those who a are not using either pycharm or vscode

rhfogh (approved):

dist/mxcubecore-1.0.0.tar.gz
meguiraun: a commit removing it is not enough... file is gonna be in the git history

fabcor-maxiv: I believe `/dist/` should be added to `.gitignore` to prevent this in the future. Same for `/build/` and other names like this. Also `.vscode/` could be removed from `.gitignore` if you really want to go forward with adding files from this directory, otherwise it seems a bit weird, doesn't it?

marcus-oscarsson: Yes you are probably right, I actually kept .vscode in gitignore so not "local" changes get committed by accident. You are right that this is probably not a good idea, but not sure how to do it differently ?

fabcor-maxiv: @marcus-oscarsson I do not use VS Code, so I am not sure what a typical `.vscode/` directory contains.

Typically for the case where we want to allow only specific files to be committed, we can use the allow-list pattern instead of the default block-list pattern of `.gitignore`. Something like this:

`/.vscode/.gitignore`

```gitignore
# Block everything...
*
# ... but allow the following:
!launch.json
!setting.json
```

Not sure if it makes sense for our use case here.

marcus-oscarsson: Hm, I see. In general I don't think we would like any changes to the .vscode to get committed unless one explicitly wants to. The problem I see is that any personal preferences would be overridden at pull but I so far don't actually see it as a problem my self (one for all so to speak). In anyways I'm happy to make it just a template and people do what they want in their .vscode but lets try this for now :)

marcus-oscarsson: @fabcor-maxiv, which editor are you using ?
@rhfogh, are there a similar file for PyCharm that we can upload maybe ?

fabcor-maxiv: @marcus-oscarsson From what I can see, there is not much that needs to be in VS Code specific settings. But again, I am not a VS Code user, so I might be completely wrong.

For example, as far as I know pytest settings belong in `pytest.ini`.

On the other hand `.editorconfig` seems like it is not tied to a specific editor, so it could maybe make sense to share this file in the repo. I guess most mainstream editors might have a setting or plugin to take this file into account.

In the end, testing/formatting/linting should be enforced by some git pre-commit hook and/or by the the CI workflows (GitHub actions). So I recommend to tighten things up at this level in priority (in the settings of the test runner, linters, and formatter that should be all taken into account by git pre-commit, which should be run in the GitHub actions) -- rather than in the editors.

For the launchers we could maybe agree to use one of those (but only if really there is a need that is shared by more contributors):
* https://pypi.org/project/invoke/
* https://pypi.org/project/taskipy/

rhfogh: Pycharm has an entire settings directory with a zillion files and subdirectory, but nothing that looks like a simple .config file. Id guess we are better off having people individually make the right settings; I'd find it hard to separate out the agred settings from teh colour schemes and other individual stuff.

marcus-oscarsson: Sure, I agree that tighten up the pre-commit hook is a good thing. We have one that we have disabled for the moment because we are also tightening up the linting which caused us some issues. So we are not entirely there yet but eventually :). We are happy with some additional help if you are up for it ;)

However, providing some editor config with sensible defaults is also good id say. We can't fix everything in the pre-commit hooks right ?  and good tooling in the editor can help out in the process.

fabcor-maxiv: Agreed
I had in mind to look at all this at some point :)

marcus-oscarsson: Cool thanks :)

marcus-oscarsson: @rhfogh, ah I see. Well if thats how it is.

PR #791, opened 2023-09-17
Rhfogh develop

rhfogh:
GPhL bug fixes

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
@@ -53,8 +53,6 @@ spectrum_test_data = [
 
 
 class XRFSpectrumMockup(AbstractXRFSpectrum):
-    # def __init__(self, name):
-    #     super.__init__(name)
 
     def init(self):
         pass
--- a/mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
@@ -52,10 +52,9 @@ spectrum_test_data = [
 ]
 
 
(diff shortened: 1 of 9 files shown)


mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
```diff
@@ -52,10 +52,9 @@
 ]
 
 
-class XRFSpectrumMockup(AbstractXRFSpectrum, HardwareObject):
-    def __init__(self, name):
-        AbstractXRFSpectrum.__init__(self)
-        HardwareObject.__init__(self, name)
+class XRFSpectrumMockup(AbstractXRFSpectrum):
+    # def __init__(self, name):
+    #     super.__init__(name)
```
fabcor-maxiv: Why push code that is commented out?
Also that code seems incorrect to me. I think it should be `super().__init__(name)`.

mxcubecore/model/queue_model_objects.py
```diff
@@ -2470,8 +2470,9 @@ def calculate_dose(self, transmission=None):
         :return:
         """
 
-        if transmission is None:
-            transmission = 100.0 * self.transmission
+        # if transmission is None:
+        #     transmission = 100.0 * self.transmission
```
fabcor-maxiv: Same here. Why push code that is commented out?

mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
rhfogh: Because this is a clean-up operation - of someone else's code. I think it is wrong to have a subclass function that does nothing but call the superclass function, but I left it in in case the original author had complaints. I shall remove it now.

marcus-oscarsson:
Thanks @rhfogh

PR #792, opened 2023-09-19
tango attribute polling: fix issue with numpy array attributes

elmjag:
In TangoChannel.update(), we'll get an exception if attribute value is of numpy.ndarray. 

Value of numpy.ndarray type can't be compared to Poller.NotInitializedValue, Comparison raised following exception:
    
      ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
    
Check _first_ if value is ndarray, and only if it's not, compare it with Poller.NotInitializedValue.
https://github.com/elmjag/mxcubecore/blob/tango_numpy/mxcubecore/Command/Tango.py#L325

Also, some minor clean-up.

### DIFF
--- a/mxcubecore/CommandContainer.py
+++ b/mxcubecore/CommandContainer.py
@@ -276,7 +276,7 @@ class ChannelObject:
                 if cmdobj is not None:
                     cmdobj(value)
 
-    def get_value(self, force: bool = False) -> None:
+    def get_value(self, force: bool = False):
         """Get channel value.
 
         Args:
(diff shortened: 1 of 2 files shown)


mxcubecore/CommandContainer.py
```diff
@@ -276,7 +276,7 @@ def update(self, value: Any) -> None:
                 if cmdobj is not None:
                     cmdobj(value)
 
-    def get_value(self, force: bool = False) -> None:
+    def get_value(self, force: bool = False):
```
marcus-oscarsson, line 279: No real reason to remove the type hint, but as its returns None we can leave it as such :)

marcus-oscarsson:
Wonderful, thanks !

PR #793, opened 2023-09-21
Fix login issues in ISPyB clients (and mockups)

fabcor-maxiv:
For test purposes `ISPyBRestClientMockup` should fail authentication when the password is `wrong`.
Also there were multiple ways to access the "login type" value, that were not always in sync.
The one true way should be to go through `get_property("loginType")`.

GitHub: relates to https://github.com/mxcube/mxcubeweb/issues/1045
GitHub: relates to https://github.com/mxcube/mxcubeweb/pull/1049

### DIFF
--- a/mxcubecore/HardwareObjects/ISPyBClient.py
+++ b/mxcubecore/HardwareObjects/ISPyBClient.py
@@ -40,6 +40,9 @@ if sys.version_info > (3, 0):
 logging.getLogger("suds").setLevel(logging.INFO)
 
 
+LOGIN_TYPE_FALLBACK = "proposal"
+
+
 # Production web-services:    http://160.103.210.1:8080/ispyb-ejb3/ispybWS/
 # Test web-services:          http://160.103.210.4:8080/ispyb-ejb3/ispybWS/
 
@@ -167,7 +170,6 @@ class ISPyBClient(HardwareObject):
         self._disabled = False
 
         self.authServerType = None
-        self.loginType = None
         self.loginTranslate = None
 
         self.ws_root = None
@@ -194,7 +196,6 @@ class ISPyBClient(HardwareObject):
             if self.ldapConnection is None:
                 logging.getLogger("HWR").debug("LDAP Server is not available")
 
-        self.loginType = self.get_property("loginType") or "proposal"
         self.loginTranslate = self.get_property("loginTranslate") or True
         self.beamline_name = HWR.beamline.session.beamline_name
 
@@ -326,7 +327,15 @@ class ISPyBClient(HardwareObject):
                 except AttributeError:
                     pass
 
+    @property
+    def loginType(self):
+        return self.get_property("loginType", LOGIN_TYPE_FALLBACK)
+
     def get_login_type(self):
+        warnings.warn(
+            "Deprecated method `get_login_type`. Use `loginType` property instead.",
+            DeprecationWarning,
+        )
         return self.loginType
 
     def translate(self, code, what):
--- a/mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
@@ -20,6 +20,9 @@ except Exception:
 # to simulate no session scheduled, use "nosession" for password
 
 
+LOGIN_TYPE_FALLBACK = "proposal"
+
+
 class ISPyBClientMockup(HardwareObject):
     """
     Web-service client for ISPyB.
@@ -30,7 +33,6 @@ class ISPyBClientMockup(HardwareObject):
         self.__translations = {}
         self.__disabled = False
         self.__test_proposal = None
-        self.loginType = None
         self.base_result_url = None
         self.lims_rest = None
         self.login_ok = True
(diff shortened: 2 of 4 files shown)


rhfogh (approved):
Nothing to say, really, but no objections

marcus-oscarsson:
Thanks :)

mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
```diff
@@ -87,8 +88,12 @@ def init(self):
             "Laboratory": {"laboratoryId": 1, "name": "TEST eh1"},
         }
 
+    @property
+    def loginType(self):
+        return self.get_property("loginType") or LOGIN_TYPE_FALLBACK
+
     def get_login_type(self):
-        self.loginType = self.get_property("loginType") or "proposal"
+        # This method is unnecessary. The `loginType` property can be used instead.
```
marcus-oscarsson: Hm, we could add a deprecation warning or simply refactor the code to not use get_login_type ?

mxcubecore/HardwareObjects/ISPyBClient.py
```diff
@@ -326,7 +327,12 @@ def init(self):
                 except AttributeError:
                     pass
 
+    @property
+    def loginType(self):
+        return self.get_property("loginType") or LOGIN_TYPE_FALLBACK
+
     def get_login_type(self):
+        # This method is unnecessary. The `loginType` property can be used instead.
```
marcus-oscarsson: Same as below :)

mxcubecore/HardwareObjects/ISPyBClient.py
```diff
@@ -326,7 +327,12 @@ def init(self):
                 except AttributeError:
                     pass
 
+    @property
+    def loginType(self):
+        return self.get_property("loginType") or LOGIN_TYPE_FALLBACK
```
marcus-oscarsson: `get_property` takes a second argument (default parameter) but using `or` also works :)

mxcubecore/HardwareObjects/ISPyBClient.py
beteva: Maybe it is also a good time to get rid of the camel case variables.

mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
fabcor-maxiv: @marcus-oscarsson I want to avoid doing any bigger refactoring in this pull request.

In a first version I had removed `get_login_type` because I could not see it used anywhere. But maybe someone somewhere still does really rely on it, I can not know for sure. I am fine with removing `get_login_type`, but I do not want to break anyone else.

I would be fine with deprecation as well, of course. How are deprecations usually handled in mxcubecore?

Someone else than me should make the call. But no big refactoring please :)

mxcubecore/HardwareObjects/ISPyBClient.py
fabcor-maxiv: Ah... Good to know. I will update the pull request to take advantage of this.

mxcubecore/HardwareObjects/ISPyBClient.py
fabcor-maxiv: @beteva I rather not do the renaming in this pull request. The `loginType` attribute/property is used in many places that I can see, and who knows how many more places that I can not see.

mxcubecore/HardwareObjects/ISPyBClient.py
marcus-oscarsson: :), it was more for information, as you like really

mxcubecore/HardwareObjects/ISPyBClient.py
marcus-oscarsson: Ok

mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
marcus-oscarsson: Ok, lets do this in two rounds then. We will remove `get_login_type` and change `loginType` to `login_type` in a nother PR. we have used `warning.warn` before, like below 
```
warnings.warn(
            "Deprecated method, <name> ", DeprecationWarning
)
````

marcus-oscarsson:
Looks good to me, we will try to follow up with another PR to take care of `loginType` and etc (feel free to give it go if you feel inspired :) ).

mxcubecore/HardwareObjects/ISPyBClient.py
fabcor-maxiv: I prefer it with the default parameter. Updated.

mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
fabcor-maxiv: I added the deprecation warning.

fabcor-maxiv:
Thanks all for the reviews.

I think I have addressed what can/should be addressed within this pull request. I think it is ready for merge :)

marcus-oscarsson:
Cool, thanks

PR #794, opened 2023-09-25
Bug fix and GPhL release version update

rhfogh:
Simple bug fix (missing import statement in recent commit).

An editor like Pycharm that highlights errors in red is nice to have here.

### DIFF
--- a/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
+++ b/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
@@ -349,7 +349,7 @@ class ChooseLattice(Payload):
 
     @property
     def priorCrystalClasses(self):
-        """set of crystal class names"""
+        """tuple of crystal class names"""
         return self._priorCrystalClasses
 
     @property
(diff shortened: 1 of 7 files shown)


rhfogh:
Added some real changes - upgraded handling of crystal symmetry for GPhL workflow automation

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
```diff
@@ -349,7 +349,7 @@ def __init__(
 
     @property
     def priorCrystalClasses(self):
```
marcus-oscarsson, line 351: How about a nice type hint ;)

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -2083,17 +2093,41 @@ def select_lattice(self, payload, correlation_id):
...
+                )
+            params["space_group"] = space_group
+            params["crystal_classes"] = crystal_classes
+
+        else:
             params = self.query_pre_strategy_params(choose_lattice)
             if params is StopIteration:
                 return StopIteration
             indexing_solution = params["indexing_solution"]
+        #
```
marcus-oscarsson: Extra hash :)

marcus-oscarsson:
Thanks, looks good to me. Just a typo to fix perhaps.

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
rhfogh: The entire file is minus type hints and doc strings, as are a couple of other GPhL files. Fixing it is on my TODO list (as indeed it should be), but I would rather do them all at once and *then* keep them up to date than add them piecemeal.

PR #807, opened 2023-10-16
fix sardana attribute limits

meguiraun:


fabcor-maxiv:
Linting issues. You might want to install pre-commit hooks.

fabcor-maxiv:
~~Not sure why the `pages-deploy` is considered as "cancelled" (and why it is marked as red), since this job is not supposed to run in a case like this one.~~ Fixed, now. I restarted the job. Checks are all green. I am still confused about what happened. BUt whatever, we will see if it happens again on other pull requests.

marcus-oscarsson:
:), Ok, thanks alot guys ! (Also dont worry to much about the piplines ;))

PR #808, opened 2023-10-17
Moved logic from mxcube-web to mxcubcore

marcus-oscarsson:
Moved logic that was previously in `mxcubeweb` down to mxcubecore and "unified" the interface for calling `execute` and `abort` between "annotated" and "standard" `BeamlineActions`.

Improved the AnnotatedBamlineAction mockups

### DIFF
--- a/mxcubecore/HardwareObjects/BeamlineActions.py
+++ b/mxcubecore/HardwareObjects/BeamlineActions.py
@@ -1,5 +1,4 @@
 import sys
-import typing
 import ast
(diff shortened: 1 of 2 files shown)


fabcor-maxiv:
Seems like this PR branched out from a quite old commit of `develop`. Might be the reason why the *Read the Docs* build failed.

rhfogh (commented):
Fine to a quick look, but I do not know enough to judge.

PR #813, opened 2023-10-30
Rhfogh develop

rhfogh:
Added improved test mode for GPhL UI testing without image simulation.

Also a few bug fixes.


mxcubecore/model/queue_model_objects.py
```diff
@@ -26,6 +26,7 @@
 import copy
 import os
 import logging
+import ruamel.yaml as yaml
```
marcus-oscarsson, line 29: We have to be carefull with imports in this file, if you read the comment below (in the file)

```
# This module is used as a self contained entity by the BES
# workflows, so we need to make sure that this module can be
# imported eventhough HardwareRepository is not avilable.
```

The non standard python modules needs to be moved to a try 
except similar to the one below, for this to work with the BES wokflows.

marcus-oscarsson: That would include `rumel` and `crystal_symmetry`

marcus-oscarsson:
Looks good, only had one comment about the import, let me know if its unclear :) ?

mxcubecore/model/queue_model_objects.py
rhfogh: I moved them to the function where they are needed instead. Cleaner, and if those imports do not work those functions can never be called anyway.

BTW, If we want the model to become a pure data holder, various GPhL functionality can (and would have to) be moved elsewhere. Do you have any suggestions where - those functions seem to have their natural home in the QMO?

PR #814, opened 2023-11-01
add documentation URL to pyproject specifications

elmjag:
It seems that poetry started to insist that 'documentation' field must contain a valid URL. Poetry will refuse to install package with an empty 'documentation' field.

Let's point to our generated docs at readthedocs.io site.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ maintainers = [
 readme = "README.md"
 homepage = "http://github.com/mxcube/mxcubecore"
 repository = "http://github.com/mxcube/mxcubecore"
-documentation = ""
+documentation = "https://mxcubecore.readthedocs.io/"
 keywords = ["mxcube", "mxcube3", "mxcubecore"]
 classifiers = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -12,7 +12,7 @@ maintainers = [
 readme = "README.md"
 homepage = "http://github.com/mxcube/mxcubecore"
 repository = "http://github.com/mxcube/mxcubecore"
-documentation = ""
+documentation = "https://mxcubecore.readthedocs.io/en/latest/"
```
fabcor-maxiv: I would remove the `en/latest/` suffix.

fabcor-maxiv (changes requested):

pyproject.toml
elmjag: yes, updated to link to https://mxcubecore.readthedocs.io/

fabcor-maxiv (approved):

marcus-oscarsson:
Thanks again ;)

PR #816, opened 2023-11-14
ISARA maint hardware object

elmjag:
The sample changer maintenance hardware object for ISARA2 robots. This hardware object support ISARA2 sample changes exposed via tango device server.

### DIFF
--- a/mxcubecore/HardwareObjects/ISARAMaint.py
+++ b/mxcubecore/HardwareObjects/ISARAMaint.py
@@ -1,0 +1,265 @@
+from typing import Optional
+from tango import DeviceProxy, DevFailed
+from mxcubecore.BaseHardwareObjects import Equipment
+
+"""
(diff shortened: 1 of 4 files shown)


mxcubecore/HardwareObjects/ISARAMaint.py
fabcor-maxiv, line 1: Can we name things `IsaraMaint` instead of `ISARAMaint` (file and class)? It is much easier to read at least for me.

fabcor-maxiv (commented):
Great!

mxcubecore/HardwareObjects/ISARAMaint.py
elmjag: Unfortunately it have already become an established naming convention for ISARA related files and classes. It's used in other places as well.

I don't think it worth the effort now to change to Isara in all the places.

PR #818, opened 2023-11-15
Cleanup and documentation.

beteva:


rhfogh (approved):
Looking good. Thanks.

mxcubecore/HardwareObjects/MachCurrent.py
```diff
@@ -10,67 +30,70 @@
...
         if opmsg and opmsg != self.opmsg:
             self.opmsg = opmsg
             logging.getLogger("user_level_log").info(self.opmsg)
-        self.emit("valueChanged", (mach, opmsg, fillmode, refill))
+        self.emit("valueChanged", (value, opmsg, fillmode, refill))
 
     def get_current(self) -> float:
+        """Read the ring current.
+        Returns:
+            (float): Ring current [mA]
```
marcus-oscarsson, line 79: You dont have to add the return type here if you also have type hint

beteva: This is for more complete documentation  (like we have the units of the return value)

PR #820, opened 2023-11-21
Remove `cancel-in-progress` in `pages` workflow

fabcor-maxiv:
This should prevent cancelled workflows to be considered as failed from the point of view of the pull requests checks.


fabcor-maxiv:
Same as https://github.com/mxcube/mxcubeweb/pull/1133

marcus-oscarsson:
Nice, thanks !

PR #825, opened 2023-12-04
sardana motors: handle limit 0.0 correctly

elmjag:
The taurus Quantity(0.0) value evaluates to false. Thus when motor's limit is set to exactly 0.0 the code:

    if minval:
       <save limit>

will not run the if-body. Check explicitly for None values of minval and maxval variables to avoid this problem.


beteva:
Fine with me. Indeed this is better handling of the limits.

PR #826, opened 2023-12-06
MASSIF-1 Crystal Direct Harvester implementation

jbflo:
This code related to the implementation of the Crystal Direct Harvester in Massif-1 ESRF

This should not have breaking change. 
Beamline who does not have an Harvester should not have any new things/modification to do 
A PR in Mxcube-web related to this one need to be reviewed as well .


mxcubecore/Command/exporter/ExporterClient.py
```diff
@@ -127,7 +127,7 @@ def __process_return(self, ret):
             ProtocolError
         """
         if ret[:4] == RET_ERR:
-            msg = "Diffractometer: {}".format(str(ret[4:]))
+            msg = f"{self.get_server_object_name()} : {str(ret[4:])}"
             logging.getLogger("HWR").error(msg)
```
fabcor-maxiv, line 141: As far as I know, the recommendation for Python logging is to always use printf style formatting (the one with the `%` interpolation) to take advantage of lazy evaluation:

* https://realpython.com/python-f-strings/#lazy-evaluation-in-logging
* https://docs.python.org/3/library/stdtypes.html#old-string-formatting

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -414,18 +414,42 @@ def sample_changer_maintenance(self):
...
 
     __content_roles.append("sample_changer_maintenance")
+    
+    @property
+    def harvester(self):
+        """Harvester Hardware object
+        can be a sample or plate holder
+
+        Returns:
+            Optional[AbstractHarvester]:
```
fabcor-maxiv, line 465: Why not use type hints?

fabcor-maxiv (commented):
A couple of random comments. I did not review the whole thing, we do not seem to have such harvester.

mxcubecore/HardwareObjects/Beamline.py
jbflo: As the other method are like this, i'll leave it  for now. maybe someone can make the change for all the methods in this class in a separate PR.

mxcubecore/HardwareObjects/abstract/AbstractSampleChanger.py
```diff
@@ -715,6 +715,10 @@ def _do_unload(self, sample_slot=None):
     @abc.abstractmethod
     def _do_reset(self):
         return
+    
+    @abc.abstractmethod
```
beteva: I would remove the abstract decorator, as it is only specific for the harvester and should not require that all the sample changers implement this method.

mxcubecore/Command/exporter/ExporterClient.py
```diff
@@ -127,7 +127,7 @@ def __process_return(self, ret):
             ProtocolError
         """
         if ret[:4] == RET_ERR:
-            msg = "Diffractometer: {}".format(str(ret[4:]))
+            msg = "{} : {}".format(self.get_server_object_name(), str(ret[4:]))
```
beteva: You might as well consider use the f-string (see with pylint).

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -458,10 +456,36 @@ def sample_changer_maintenance(self):
 
     __content_roles.append("sample_changer_maintenance")
 
+    @property
+    def harvester(self):
+        """Harvester Hardware object
+        can be a sample or plate holder
+
+        Returns:
+            Optional[AbstractHarvester]:
```
beteva, line 465: Is this needed? It can be just seen as a sample changer in beamline.

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -458,10 +456,36 @@ def sample_changer_maintenance(self):
...
+    @property
+    def harvester_maintenance(self):
+        """harvester maintenance Hardware object
+
+        Returns:
+            Optional[Harvester]:
+        """
+        return self._objects.get("harvester_maintenance")
+
+    __content_roles.append("harvester_maintenance")
```
beteva, line 480: Is this needed? It can be just seen as a sample changer in beamline.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
```diff
@@ -183,31 +189,58 @@ def init(self):
         self._update_selection()
         self.state = self._read_state()
 
+    def _mount_from_harvester(self):
+        return self._harvester and self._harvester_hwo
+
```
beteva: @jbflo, please see my comment for the AbstractSampleChanger.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
```diff
@@ -183,31 +189,58 @@ def init(self):
...
+                    if sample.get_address() == (
+                        str(present_sample[0])
+                        + ":"
+                        + str(present_sample[1])
+                        + ":"
+                        + "%02d" % int(present_sample[4])
+                    ):
+                        present_sample_list.append(sample)
 
         return present_sample_list
```
beteva: For better readability, it would be good to split this in three methods - the get_sample_list, which is the standard method, which calls  _get sample_list_flex and _get_sample_list_harvester.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
```diff
@@ -531,25 +608,38 @@ def _do_load(self, sample=None):
...
             logging.getLogger("HWR").info(f"Waited SC activity {time.time() - _tt}")
-        except:
+        except Exception:
+            logging.getLogger("user_level_log").error(
+                "ERROR While Waited SC activity to start"
+            )
             for msg in self.get_robot_exceptions():
                 logging.getLogger("user_level_log").error(msg)
-            raise
+            return False
```
beteva: Attention, this is changing the behaviour of the method.

mxcubecore/HardwareObjects/Harvester.py
```diff
@@ -0,0 +1,501 @@
...
+[Commands]
+
+ - getSampleList : Get list of available sample from Harvester
+ - Harvest : Harvest sample make it ready to load
+    <equipment class="Harvester">
+        <username>harvester</username>
+        <exporter_address>wid30harvest:9001</exporter_address>
+    </equipment>
+-----------------------------------------------------------------
+"""
```
beteva: It is great that there is some explanation, Could you, please, add the standard MXCuBE license message (see EMBLFlexHCD.py) Also the xml example should use object instead of equipment.

mxcubecore/HardwareObjects/HarvesterMaintenance.py
```diff
@@ -0,0 +1,140 @@
+"""
+Harvester Maintenance maintenance .
+"""
```
beteva: Same as for Harvester.

mxcubecore/HardwareObjects/abstract/AbstractSampleChanger.py
```diff
@@ -679,6 +678,13 @@ def _resolve_component(self, component):
             return comp
         return component
 
+    def _mount_from_harvester(self):
+        """
+        Mount from Harvester or from default dewar
+        Returns: (Bool) true Mount from Harvester
+        """
+        return
+
```
beteva: Is this used to check if harvester or sample changer is used? If yes, than the name is misleading (maybe something like is_harvester is more appropriated) and moreover it should be a property.

beteva (changes requested):

mxcubecore/Command/exporter/ExporterClient.py
jbflo: I did, but then followed Fabcor suggestion 
https://github.com/mxcube/mxcubecore/pull/826#discussion_r1417609839

mxcubecore/HardwareObjects/Beamline.py
jbflo: Yes, yes but it is a bit more complex.
- This Hardware can not be treated as a sample changer as it is more like a sample Holder. 
-Accessing it is possible through the Sample Changer but In Massif-1 they still wanted to use the Harvester even the Sample Changer is not available.

mxcubecore/HardwareObjects/Beamline.py
jbflo: https://github.com/mxcube/mxcubecore/pull/826#discussion_r1449178378

mxcubecore/HardwareObjects/Beamline.py
beteva: My question was more can't you specify the harvester hardware object instead of the sample changer in the beamline configuration. Thus you can also benefit from the checks which are done for mounting/unmounting.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
jbflo: Yes, thanks  We have to see on Other Beamlines, but in Massif-1 this behavoir is OK in since
Just raise will return None which consider a loading success.
But to avoid  to collect on the previous mounted sample.  a return False fit better.

rhfogh: I would not agree here, J-B. The behaviour of load (which is determined by _do_load) must be the same as the behaviour of SampleChanger.load(). Replacing a raised error with a return of False does not look like the same behaviour.

Also - if MASSIF-1 really interprets a call that raises an error as a successful load, that sounds like a serious error that should eb fixed on the MASSIF-1 side, no?

rhfogh (changes requested):
Surely a file called EMBLFlexHCD.py should be in HardwareObjects/EMBL, and not in HardwareObjects/ ?

beteva:
Not quite @rhfogh, in this case it refers that the flex is manifactured by EMBL, just as Cats, Marvin ...:-)

mxcubecore/HardwareObjects/Beamline.py
rhfogh: Ah well, in that case it should stay where it is. I do not seem to have a very high hit-rate today )-:

mxcubecore/HardwareObjects/EMBLFlexHCD.py
jbflo: not only   MASSIF-1,  I mentioned Massif-1 because it's there I found the issue. 
this code bellow is not quite OK  ? why if load return None , you make it True ?
https://github.com/mxcube/mxcubeweb/blob/a958e7d1d3e4ffcc354974d07685522088f7cd23/mxcubeweb/core/components/samplechanger.py#L198C1-L205C31

rhfogh: I'd say the problem is that happens if load raises an error. The way this would normally work in Python, the raised error would stop further execution of the code and jump right down to line 224 where it says:

        except Exception as ex:
            logging.getLogger("MX3.HWR").exception("[SC] sample could not be mounted")
            raise RuntimeError(str(ex)) from ex

Since this again raises an error, you would never get to line 244 
        return res
but would have the error propagate upwards to the next error handler. 
With your change you would get a different behaviour: you would have the function return False instead of raising an error. If you are making that change I would say you need  a very careful analysis of the calling functions and their error handling to be sure that you want the new behaviour rather than the old one.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
jbflo: method removed from AbstractSampleChanger

jbflo:
> Surely a file called EMBLFlexHCD.py should be in HardwareObjects/EMBL, and not in HardwareObjects/ ?

@rhfogh  do you still request this change  ?

mxcubecore/HardwareObjects/Beamline.py
jbflo: > Is this needed? It can be just seen as a sample changer in beamline.

@beteva  As we discuss a while ago, could you have look again ?

rhfogh (approved):

marcus-oscarsson (approved):

marcus-oscarsson:
Looks good as far as I can see :), (its a big one ;) )

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -78,15 +75,23 @@ def __init__(self, *args):
...
             crims_url
             + barcode
             + "/plans/xml"
-            + "?harvester-key=%s" % harvester_key
         )
-        f = urlopen(url)
-        xml = f.read()
+
+        headers = { 
+            'User-Agent' : "Mozilla/5.0 (Windows NT 6.1; Win64; x64",
```
beteva: Should this really be hard-coded?

beteva:
Hi @jbflo, Most of the changes are Ok with me. It still bothers me, though, that you have to check for harvester in the EMBLFlexHCD.py.

mxcubecore/HardwareObjects/Harvester.py
```diff
@@ -0,0 +1,643 @@
...
+        sample_states = self.get_samples_state()
+        crystal_uuids = self.get_crystal_uuids()
+
+        for index, x_tal in enumerate(crystal_uuids):
+            if crystal_uuid == x_tal:
+                return sample_states[index]
+
+        return None
+
+    def check_crystal_state(self, crystal_uuid: str) -> str | None:
```
jbflo: @beteva @marcus-oscarsson I Just Notice  this Type hints
require Python 1.10 and above.
Is this a problem ?

marcus-oscarsson: You mean ` str | None` ?

You should be able to use 

```
from typing import Optional

 ... -> Optional[str]
```

jbflo: > You mean ` str | None` ?
Yes
> You should be able to use
> 
> ```
> from typing import Optional
>  ... -> Optional[str]
> ```
Yeah,

jbflo:
> Hi @jbflo, Most of the changes are Ok with me. It still bothers me, though, that you have to check for harvester in the EMBLFlexHCD.py. 

as we We check because depending of where the SC take the Sample , the Sample List and the Next Actions are different 

maybe it's because the Dewar is (consider) a part of the FLEX.

Now we should probably think of 2 things.
1- The Sample changer ( the Robot ) FLEX
2 The Sample Holder (Can be a Dewar(HCD) or Harvester)

Or maybe we should have a Flex and a HCD Class ?  Or an EMBLFlexHavester CLass ?
having them both implementing as one is not much help understanding the Harvester.

Yeah maybe an EMBLFlexHavester CLass but this will have must of the EMBLFlexHCD methode In.

I am thinking about it @beteva , but for now i don't have a better clue.
for now I Think this class (EMBLFlexHCD) is the best place. 

please let me know if you are thinking  of a better logic.

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -122,5 +129,38 @@ def get_processing_plan(barcode, crims_url, harvester_key):
...
+            + str(crystaluuid)
+            + "/dcgroupid/"
+            + str(datacollectiongroupid)
+            + "/dcid/"
+            + str(dcid)
+            + "/mx/"
+            + str(proposal)
+            + "/token/"
+            + str(rest_token)
+            + "?janitor_key=kbonvRqc8"
```
fabcor-maxiv: Maybe better as an f-string?

Also is the key supposed to be here? Is that a secret?

marcus-oscarsson:
I agree with @beteva that it would be nice if the sample changer and the Harvester code could be de-coupled. I think that would be a nice followup PR.

beteva:
> I am thinking about it @beteva , but for now i don't have a better clue. for now I Think this class (EMBLFlexHCD) is the best place.
> 
> please let me know if you are thinking of a better logic.

@jbflo, it might be tempting to split Robot and Holder, but I am afraid that this will add an extra complication to already complicated device. I would rather go for the other idea - having a specific  EMBLFlexHavester CLass, which inherits from the EMBLFlexHCD and only overloads the methods which differ for the harvester. Than in the beamlise-setup.yml it will be easy to choose which "sample changer" you have at given moment.What do you think?

jbflo:
> > I am thinking about it @beteva , but for now i don't have a better clue. for now I Think this class (EMBLFlexHCD) is the best place.
> > please let me know if you are thinking of a better logic.
>
> @jbflo, it might be tempting to split Robot and Holder, but I am afraid that this will add an extra complication to already complicated device. I would rather go for the other idea - having a specific  EMBLFlexHavester CLass, which inherits from the EMBLFlexHCD and only overloads the methods which differ for the harvester. Than in the beamlise-setup.yml it will be easy to choose which "sample changer" you have at given moment.What do you think?

Yes I also think it's a better Idea , I'll go on that side. Thanks !

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
```diff
@@ -0,0 +1,207 @@
...
+
+from mxcubecore.HardwareObjects import EMBLFlexHCD
+
+
+
+class EMBLFlexHarvester(EMBLFlexHCD):
+    __TYPE__ = "Flex Sample Changer"
+
+    def __init__(self, *args, **kwargs):
+        super(EMBLFlexHarvester, self).__init__(self.__TYPE__, True, *args, **kwargs)
```
beteva: You can do
`super().__init__(*args, **kwargs)` 
Also self._loaded_sample = (-1, -1, -1) is better to be here, than in init(). you can also add self._harvester = None (not compulsory, just style).

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
```diff
@@ -0,0 +1,207 @@
...
+    def get_loaded_sample(self):
+        sample = None
+
+        loaded_sample_addr = self._hw_get_mounted_sample()
+
+        for s in self.get_sample_list():
+            if s.get_address() == loaded_sample_addr:
+                sample = s
+
+        return sample
```
beteva: get_loaded_sample() is identical to the parent class. no need to overload it

beteva (approved):
Well done @jbflo. All looks nice and neat. I only have few style remarks. Have not checked the EMBLFlexHCD code, but I trust you it is the original one.

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
beteva: Please, check the style with black as well.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
```diff
@@ -521,6 +521,16 @@ def defreeze(self):
         self.enable_power()
         self._execute_cmd_exporter("defreezeGripper", command=True)
 
+    def get_room_temperature_mode(self):
```
marcus-oscarsson: I guess these two methods should not be here then, if we now have the `EMBLFlexHarvester` object

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -774,9 +798,9 @@ def get_default_acquisition_parameters(self, acquisition_type="default"):
 
         acq_parameters = queue_model_objects.AcquisitionParameters()
 
-        #logging.getLogger("HWR").debug(f"""
-            #Beamline object. Getting acquisition parameters for acquisition type {acquisition_type}
-        #""")
+        # logging.getLogger("HWR").debug(f"""
+        # Beamline object. Getting acquisition parameters for acquisition type {acquisition_type}
+        # """)
```
marcus-oscarsson: Remove or un-comment ?

mxcubecore/HardwareObjects/EMBLFlexHCD.py
```diff
@@ -378,7 +378,7 @@ def chained_load(self, old_sample, sample):
     def _set_loaded_sample_and_prepare(self, sample, previous_sample):
         res = False
 
-        if not -1 in sample and sample != previous_sample:
```
marcus-oscarsson, line 381: Looks good :) :+1:

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
```diff
@@ -0,0 +1,207 @@
+# encoding: utf-8
```
marcus-oscarsson, line 1: Looks good as far as I can see

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
```diff
@@ -0,0 +1,207 @@
...
+        return self._execute_cmd_exporter("getRoomTemperatureMode", attribute=True)
+
+    def set_room_temperature_mode(self, value):
+        self._execute_cmd_exporter("setRoomTemperatureMode", value, command=True)
+        logging.getLogger("user_level_log").info(
+            f"setting Robot Room temperature to {value}"
+        )
+        return self.get_room_temperature_mode()
+
+    def mount_from_harvester(self):
```
marcus-oscarsson, line 64: Does this mean that there should be an equivalent method for the AbstractSampleChanger that returns false by default, or how is this used ?

mxcubecore/model/queue_model_objects.py
```diff
@@ -329,6 +329,14 @@ def has_lims_data(self):
             return False
         return False
 
+    def has_protein_acronym(self):
```
marcus-oscarsson, line 341: Can't we somehow merge this with has_lims_data so that we dont need to check for both here :https://github.com/mxcube/mxcubecore/pull/826/files#diff-bf4145a2e91e5b6388dbc69591ea8c2fc19b54fcdf10087cd77b0087ccb3b423R283

mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -260,6 +269,8 @@ def init(self):
         self.centringVertical = self.get_object_by_role("centringVertical")
         self.centringFocus = self.get_object_by_role("centringFocus")
 
+        self.saved_motor_position = {}
```
marcus-oscarsson, line 272: How is `saved_motor_position` used ?

mxcubecore/HardwareObjects/EMBLFlexHCD.py
```diff
@@ -546,7 +546,7 @@ def _do_load(self, sample=None):
             _tt = time.time()
             self._wait_busy(300)
             logging.getLogger("HWR").info(f"Waited SC activity {time.time() - _tt}")
-        except:
+        except Exception:
```
rhfogh, line 549: I think this should actually be 'except:' - whatever the linter says. 
The difference is that 'except:' also catches 'built-in system-exiting exceptions', and in this case you are re-raising the exception anyway and you  would still want to log the robot_exceptions before exiting.

mxcubecore/model/queue_model_objects.py
jbflo: Hmm I think not . 
unfortunately  has_lims_data logic  is  not suitable for all cases. 
in this case protein acronym is known even there is no lims data .

And The scientist would ask for it again if we remove the second check.
Having the acronym as parts in the prefix was a mandatory for them.

we could maybe merge  has_protein_acronym with has_lims_data ?
 
 
 PS : thanks i should not use TypeError here i'll fix it,

mxcubecore/HardwareObjects/Microdiff.py
jbflo: - For calibration purpose (to find find offset position )
 1 -  move those motors to a specific position and saved 
 2 - move again .. and again example(3 click centering)
 3-  then we need the saved position for offsets calculation
 
it being used here : https://github.com/mxcube/mxcubeweb/pull/1145/files#diff-5ea1dcf48d30f15039fd0b4b1bd66d2b46bc68ee3aa35f4f0db10259723ffa79R280

mxcubecore/HardwareObjects/EMBLFlexHCD.py
jbflo: hmm should we not care about linter anymore ?
PR should be merge if linter are passed right ?

Not an expert, i'll leave this for the other guys :smiley:

mxcubecore/HardwareObjects/Beamline.py
jbflo: I could but i Prefer not making the remove part of this PR. 

it Was already commented.
by @ Andrey Gruzinov Maybe ?
I just applied black on it so it could pass the Linter.

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
jbflo: HAHA :smiley: , Having an AbstractSampleChanger  equivalent method that returns false by default for 
mount_from_harvester is  indeed what I wanted.

but could not agree on this with @beteva 
Or maybe my explanation was not clear enough 
SO I find a work around and ending up removing it from AbstractSampleChanger.

mxcubecore/HardwareObjects/EMBLFlexHCD.py
marcus-oscarsson: You can use `BaseException` if you want to catch 'built-in system-exiting exceptions' and still keep the linter happy. However, I don't think we should be catching those, unless there is a good reason for it, even if as in this case we re-raise them.

mxcubecore/HardwareObjects/Beamline.py
marcus-oscarsson: I see, black is good but Id say commented code like this can simply be removed ;)

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
marcus-oscarsson: But how then are you using this method of the base class does not have this method any other sample changer would throw an `AttributeError`

mxcubecore/HardwareObjects/Microdiff.py
marcus-oscarsson: Hm, are you sure, I could not find any reference to `saved_motor_position`

mxcubecore/model/queue_model_objects.py
marcus-oscarsson: > we could maybe merge has_protein_acronym with has_lims_data ?

yes, it that would be nice if we could try that

mxcubecore/HardwareObjects/EMBLFlexHCD.py
rhfogh: In this case I think there is a good reason. We want to log those accumulated messages before we exit.
Also, frankly, following the linter in this case is a matter of the tail wagging the dog. That particular linter point was put in to 'encourage' people to put in specific exceptions rather than catching everything willy-nilly. If we want to allow a bare 'except:' in some cases (and I think we do) surely we should either remover that check from the linter or use the mechanisms for making exceptions for a given line of code. Making convoluted python code to be allowed to do what we want while by passing the linter does not sound like the right way to go.

marcus-oscarsson: Well, in this case the sample changer wont raise a system exiting exception and if it does we probably can't ask it to give us the reason for it. If it does we would know what happened by looking at the stacktrace or the sample changer log file.

The second point would be that, its really not a good idea to catch system exiting exceptions because we might trap them and our program might not exit or behave in strange ways. Its problematic when we use CTRL-C or for us more commonly decide to stop the queue by raising `GreenletExit`

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
beteva: I still do not like to have anything about harvester in the AbstractSampleChanger class. For me, unless I do not get it right, the _do_load method should be overloaded with this one.

jbflo: > But how then are you using this method of the base class does not have this method any other sample changer would throw an `AttributeError`

`    
       sc = HWR.beamline.sample_changer

        try:
            mount_from_harvester = sc.mount_from_harvester()
        except AttributeError:
            mount_from_harvester = False
`
            
I did this way every time.  not elegant but  SC who does not this method should pass the AttributeError

marcus-oscarsson: But, why do you need to call `mount_from_harvester` specifically would it not be enough to call `mount` and the `mount` of the `Harvester` object knows what to do ?

mxcubecore/HardwareObjects/Microdiff.py
jbflo: > Hm, are you sure, I could not find any reference to `saved_motor_position`

Yes, Just tagged you , https://github.com/mxcube/mxcubeweb/pull/1145#discussion_r1565904742

mxcubecore/HardwareObjects/EMBLFlexHCD.py
rhfogh: OK, I'll take that as a convincing argument. Point withdrawn.

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
jbflo: There are a lot of mounting logic there : mxcubeweb/core/components/samplechanger.py  

And mount_from_harvester help us prevent or add  behavior that not or are specific when mounting from an Harvester.  
 
have a look at this PR: [https://github.com/mxcube/mxcubeweb/pull/1145/files](url)
FIle :  mxcubeweb/core/components/samplechanger.py  


PS : those two PR are a big big (Sorry about that), but reviewing them both together maybe better

marcus-oscarsson: I see, how about simply setting a flag on the harvester hardware object and deal with the mounting specifics in the harvester hardware object. I assume that you know precisely when you need to use harvester. You could  add something to the sample address so that you know that its a harvester sample. In this case the harvester hardware object would know if its a standard sample or a harvester sample.

jbflo: That will  unfortunately not enough :  mxcubeweb.mxcubeweb.core.components.samplechanger.py
has specific logic related to queuing, loading and  centering that are by default.

Things we don't want : default autoloop centring
Things we want do to : add logic while execution of queue 

it's not just mounting, have another look at  this file : [https://github.com/mxcube/mxcubeweb/blob/JB-develop-harvester/mxcubeweb/core/components/samplechanger.py](url)
check on method like queue_mount_sample or mount_sample_clean_up.

mxcubecore/model/queue_model_objects.py
jbflo: @marcus-oscarsson I had another look and i think a merge is not a good fit for this logic.

has_lims_data method name is too generic would have been better as has_lims_id ..
has_protein_acronym is a better fit on get_default_prefix as you only wanted to care about the protein_acronym


`  i would reccommend something like that : 

      if sample_data_node:
            name = sample_data_node.name
            if (sample_data_node.has_protein_acronym()):
                protein_acronym = sample_data_node.crystals[0].protein_acronym
                if name:
                    prefix = "%s-%s" % (protein_acronym, name)
                else:
                    prefix = protein_acronym
            else:
                prefix = name or ""
      elif generic_name:
            prefix = "<acronym>-<name>"
`

and  not check for has_lims_data()

mxcubecore/HardwareObjects/Beamline.py
jbflo: comment removed

PR #827, opened 2023-12-07
link to developer docs from README file

elmjag:
Link to the generated developer docs from the readme file. This way these docs will get some visibility. Right now I suspect most are not aware of the dev docs existance.


rhfogh (approved):
Good idea

PR #829, opened 2023-12-12
Introduced AbstractAuthenticator

marcus-oscarsson:
**Greetings from the MXCuBE meeting :)**

Introduced `AbstractAuthenticator` and made necessary updates to `LdapLogin`

This also means that we should move the `authenticate` call from `ISPyBclient` to the the authenticate/login
logic. 

All `*LdapLogin` classes have been updated so that they care now call `*Authenticator`

### DIFF
--- a/mxcubecore/HardwareObjects/ISPyBClient.py
+++ b/mxcubecore/HardwareObjects/ISPyBClient.py
@@ -670,7 +670,8 @@ class ISPyBClient(HardwareObject):
         # Authentication
         if self.authServerType == "ldap":
             logging.getLogger("HWR").debug("LDAP login")
-            ok, msg = self.ldap_login(login_name, psd, ldap_connection)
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
So, now with the syntax fix for the ISPyBClient

rhfogh (approved):

fabcor-maxiv:
I do not understand what is going on with the pull requests now :D

Why is this new pull request not on top of `develop` (i.e. on top of the revert of the previous one)?

fabcor-maxiv:
By the way, indeed it was a syntax error, thanks for the fix :)

marcus-oscarsson:
@fabcor-maxiv, yes as you noticed that there was a syntax error, I thought it was better to revert the PR and then add the fix. There's been sometime since I created the PR and I forgot that I also needed to also test with the `ISPyBClient` and not only `ISPyBClientMockup`.

PR #830, opened 2023-12-14
AbstractAuthenticator fix after revert

marcus-oscarsson:
It turns out that the revert and re-apply I made yesterday removed most of the important changes to the PR. I'm very sorry for that. Here is the PR again, its the same PR but with all the files included.


rhfogh (approved):
I did not have time to read through this again (and anyway I was never the best judge for an authentication PR). But given that this was already approved once, I am happy to transfer approval to the resubmission.

Issue #837, opened 2024-01-16
error using sqlalchemy 2.0.25


elmjag:
If sqlalchemy 2.0.25 package is installed, mxcube will fail to start with this error message:

    Traceback (most recent call last):
      File "/app/web/mxcubeweb-server", line 3, in <module>
        from mxcubeweb import main
      File "/app/web/mxcubeweb/__init__.py", line 14, in <module>
      from mxcubeweb.server import Server as server  # noqa: E402
      File "/app/web/mxcubeweb/server.py", line 19, in <module>
        from mxcubeweb.core.components.user.database import (
      File "/app/web/mxcubeweb/core/components/user/database.py", line 4, in <module>
        from sqlalchemy import create_engine
      File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/__init__.py", line 12, in <module>
        from . import util as _util
      File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/util/__init__.py", line 15, in <module>
        from ._collections import coerce_generator_arg as coerce_generator_arg
      File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/util/_collections.py", line 38, in <module>
        from .typing import is_non_string_iterable
      File "/opt/conda/lib/python3.9/site-packages/sqlalchemy/util/typing.py", line 56, in <module>
        from typing_extensions import TypeAliasType as TypeAliasType  # 3.12
    ImportError: cannot import name 'TypeAliasType' from 'typing_extensions' (/opt/conda/lib/python3.9/site-packages/typing_extensions.py)

The problem is that `sqlalchemy` 2.0.25  requires `typing_extensions` >= 4.9.0.

However, it's not possible to install that version of `typing_extensions`. This is due to another transient dependency to `typing_extensions` package via `ispyb-client` package.

`Ispyb-client` have a pinned dependency on `typing_extensions` 4.3.0. Mxcubecore is already using latest version of `ispyb-client`.

elmjag:
Is it possible to drop ispyb-client from mxcubecore depencies? Is it used a lot?

Or, can we update ispyb-client to work with typing_extensions >= 4.9.0?

Who maintains this package now? Its pypi page does not say a lot: https://pypi.org/project/ispyb-client/

fabcor-maxiv:
[...]

PR #838, opened 2024-01-17
remove ispyb-client package from dependencies

elmjag:
Remove package ispyb-client from mxcubecore's dependencies.

It is not used anymore, and also it is causing issues due to it's 'pinned' dependency to typing_extensions package at version 4.3.0.

Solves: https://github.com/mxcube/mxcubecore/issues/837


fabcor-maxiv (approved):

fabcor-maxiv:
My assumption is that it is due to a change introduced in version [ReadTheDocs version 10.15.1](https://docs.readthedocs.io/en/stable/changelog.html#version-10-15-1), and in particular this [pull request](https://github.com/readthedocs/readthedocs.org/pull/10979).

fabcor-maxiv:
The ReadTheDocs build is broken. I am investigating in this ticket: https://github.com/mxcube/mxcubecore/issues/839. But I do not see any reason why it should prevent this change from being merged.

PR #841, opened 2024-01-18
GPhL update and bug fixes

rhfogh:
Updated to newest GPhL workflow
Fixed some GPhL bugs
Cleaned up some queue exceptions


marcus-oscarsson:
Looks good to me :)

PR #842, opened 2024-01-19
Simplify dependencies

fabcor-maxiv:
Remove formatting and linting dependencies from the conda environment since they are in pre-commit already.

Remove test dependencies from the conda environment since they are in the Poetry dev dependencies already.

Remove code related to `importlib_metadata` on Python "<3.8".

Remove dependency version constraints for Python "<3.8" since we only support Python ">=3.8,<3.11".

Remove setuptools and setuptools-scm from build dependencies since they are superfluous.

### DIFF
--- a/conda-environment-dev.yml
+++ b/conda-environment-dev.yml
@@ -9,11 +9,3 @@ dependencies:
   # Dev
   - pip
   - poetry
-  # Format, lint
-  - black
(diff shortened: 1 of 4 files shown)


rhfogh (approved):
Very nice. I have not tested this, but the idea is exactly right

PR #844, opened 2024-01-25
Update Python version for the `pre-commit` hooks

### LINKED ISSUE #836: Old Python version (3.8) in pre-commit hooks
As highlighted in https://github.com/mxcube/mxcubecore/pull/834#issuecomment-1891719681 the pre-commit hooks seem to rely on an old version of Python, which is not the same version that conda installs, so it makes things a bit more annoying than necessary.

We should probably address this.

I am not sure we need to have a version range for Python in the `conda-environment-dev.yml` file. Maybe it is better to choose a version (major minor, for example `3.10`), and then use the same for pre-commit hooks and everywhere else it makes sense.

fabcor-maxiv:
Since the Python range is `>=3.8,<3.11` in `conda-environment.yml` (and in the Poetry config in `pyproject.toml` as well), the version installed by default will be Python 3.10.

It makes sense to use the same version in the pre-commit hooks.

GitHub: fix https://github.com/mxcube/mxcubecore/issues/836

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,11 @@
 default_language_version:
-  python: python3.8
+  python: python3.10
 repos:
   - repo: https://github.com/python-poetry/poetry
     rev: 1.5.0
     hooks:
       - id: poetry-check
-        language_version: python3.8
 #      - id: poetry-lock
-#        language_version: python3.8
   - repo: https://github.com/myint/autoflake
     rev: v1.6.0
     hooks:
@@ -24,7 +22,6 @@ repos:
     hooks:
     - id: black
       name: Black
-      language_version: python3.8
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
     hooks:


rhfogh (changes requested):
Not sure I understand this. The point of the pre-commit hooks is to run tests, right? And the agreed minimum version of Python is 3.8, right? Then, surely, you should run the tests under Python 3.8, so that you catch it  if someone is using code that does not work in Python 3.8? Personally I would love it if we could upgrade the minimum version to something higher, but until we can do that, can we upgrade it in the tests?

Or have I misunderstood something?

marcus-oscarsson:
I agree with @rhfogh and as I said previously if I'm not mistaken the `pre-commit` scripts runs in a separate environment specified by the pre-commit config file so it should not be effected by what's in the conda environment. That is also why I previously asked what the exact problem was so that we can deal with it correctly.

*Edit*: I might of-course be mistaken and there is a Python version issue to deal with but it then becomes a bit more complicated for the reason @rhfogh highlights above, but we need to know what the issue was.

PR #846, opened 2024-01-29
Use default Python version in pre-commit hooks

fabcor-maxiv:


fabcor-maxiv:
There is the following at the top of the file, so there is no need to set the Python version again for individual hooks:

```yaml
default_language_version:
  python: python3.8
```

marcus-oscarsson:
:+1:

PR #847, opened 2024-01-30
Code quality improvements to the TangoShutter Class

HilbertCorsair:
A more formal way to retreve unconventional shutter states from the xml configuration file and add them to the class standard values dictionary with conventional names.


mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -31,19 +31,27 @@
   <command type="tango" name="Open">Open</command>
   <command type="tango" name="Close">Close</command>
   <channel type="tango" name="State" polling="1000">State</channel>
-  <values>{"open": "OPEN", "cloded": "CLOSED", "DISABLE" : "DISABLE"}</values>
+  <values>{"OPEN": "MYOPEN", "NEWSTATE": [HardwareObjectState.BUSY, "MYSTATE"]}</values>
```
fabcor-maxiv: If this is meant to be JSON, it is not valid JSON.

Maybe it is supposed to read:

```suggestion
  <values>{"OPEN": "MYOPEN", "NEWSTATE": ["HardwareObjectState.BUSY", "MYSTATE"]}</values>
```

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -109,33 +110,38 @@ def _initialise_values(self):
...
+        try:
+            config_values = json.loads(self.get_property("values"))
+            for key, val in config_values.items():
+                if isinstance(val, (tuple, list)):
+                    values_dict.update({key: val[1]})
+                    states_dict.update({key: tuple(val)})
+                else:
+                    values_dict.update({key: val})
+        except (ValueError, TypeError) as err:
+            print(f"Exception in _initialise_values(): {err}")
```
fabcor-maxiv: Maybe use logging?

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -109,33 +110,38 @@ def _initialise_values(self):
...
-            if self.config_values:
-                _state = self.config_values[str(self.state_channel.get_value())]
-            else:
-                _state = str(self.state_channel.get_value())
-
-        except (AttributeError, KeyError):
+            _state = self.get_value().name
+            return self.SPECIFIC_STATES[_state].value[0]
+        except (AttributeError, KeyError) as err:
+            print(f"Exception in get_state(): {err}")
```
fabcor-maxiv: Logging instead of print?

mxcubecore/HardwareObjects/TangoShutter.py
HilbertCorsair: Thanks for pointing that out. That part of the commentry slipped through. I didn't need to use a list for the configuration but I will look into it and do some tests with a configuration that takes a list.

beteva: This could eventually be changed to
`<values>{"OPEN": "MYOPEN", "NEWSTATE": ["MYSTATE", "BUSY"]}</values>`

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -31,19 +31,27 @@
...
-convantional states defined in the TangoShutter Class. This tag is not necessay in cases where the tango shutter states
-are all covered by the TangoShuter class conventional states.
+In the example the <values> property contains a dictionary that redefines or
+adds specific tango shutter states.
+When redefining a known state, only the VALUES Enum will be updated.
+When defining a new state (new key), the dictionary value should be a
+list. The new state is added to both the VALUES and the SPECIFIC_STATES Enum.
+Attention:
+ - do not use tuples or the python json parser will fail!
+ - the first element of the list should be a standard HardwareObjectState
```
beteva, line 44: Complient with the above change, the text would be:
` - the second element of the list should be a standard HardwareObjectState
   name (UNKNOWN, WARNING, BUSY, READY, FAULT, OFF - see in BaseHardwareObjects.py)!`

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -109,33 +110,38 @@ def _initialise_values(self):
                 "FAULT": "FAULT",
             }
         )
+        try:
+            config_values = json.loads(self.get_property("values"))
+            for key, val in config_values.items():
+                if isinstance(val, (tuple, list)):
+                    values_dict.update({key: val[1]})
+                    states_dict.update({key: tuple(val)})
```
beteva: I guess the easiest would be to convert the string to a HardwareObjectState member, so instead of
 `states_dict.update({key: tuple(val)})`
one could do
`states_dict.update({key: (HardwareObjectState[val[1]], val[0])})`

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -31,19 +31,27 @@
...
-are all covered by the TangoShuter class conventional states.
+In the example the <values> property contains a dictionary that redefines or
+adds specific tango shutter states.
+When redefining a known state, only the VALUES Enum will be updated.
+When defining a new state (new key), the dictionary value should be a
+list. The new state is added to both the VALUES and the SPECIFIC_STATES Enum.
+Attention:
+ - do not use tuples or the python json parser will fail!
+ - the first element of the list should be a standard HardwareObjectState
+   member (see in BaseHardwareObjects.py)!
```
beteva: Compliant with the above example, the text would be:
` - the second element of the list should be a standard HardwareObjectState name (UNKNOWN, WARNING, BUSY, READY, FAULT, OFF - see in BaseHardwareObjects.py)!`

HilbertCorsair:
Just pushed the latest updates. I added logging as well.

HilbertCorsair:
I don't know how relevant this is but, if somene needs to use a list in the xml config file , make sure to use double quotes ("). I copy-pasted the configuration from the comment to the xml file for a test and it had replaced the double quotes (") with simple quotes (') resulting in a json parsing error. No (') are allowed in the json inside the xml ! Only ("). :)

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -31,19 +31,28 @@
   <command type="tango" name="Open">Open</command>
   <command type="tango" name="Close">Close</command>
   <channel type="tango" name="State" polling="1000">State</channel>
-  <values>{"open": "OPEN", "cloded": "CLOSED", "DISABLE" : "DISABLE"}</values>
+  <values>{"OPEN": "MYOPEN", "NEWSTATE": ["HardwareObjectState.BUSY", "MYSTATE"]}</values>
```
beteva: Hi @HilbertCorsair, I think it will be good if there is the hint about the double quotes you've mentioned. This is quite useful to know.
Small correction - with the current code this line should look like
`<values>{"OPEN": "MYOPEN", "NEWSTATE": ["MYSTATE", "BUSY"]}</values>`

HilbertCorsair:
Just updated the comment section with the example provided by @beteva as well as a warning about using only double quotes.

rhfogh (commented):
I am sorry, but I really, really hate this. I can see that this particular PR is not making the difference;  the main point is already merged in, so it is too late. But really:

Putting JSON strings inside XML is bad in itself. But worse, we have here a major change to the configuration system, that has not been discussed as such, and that is only valid for TangoShutters and only implemented inside TangoShutter.py. How are we to keep track of what is happening, if every class uses its own custom configuration system?  What happens when someone has to make a new class and looks for examples on how to configure it? What happens if someone else reads inside those XML properties (you are allowed to) and tries to untangle this on his own, maybe to figure out (from the configuration files) what the applicable states for a TangoShutter are?   

Supposedly, you can store any data structure you need with XML. It is not very practical, which is why moving to another format seems such a good idea, but while we are agreed on using XML surely the correct procedure is to take the trouble and use XML, not to put in your own hacks.

What I would like to see at this point is either 

1) a proposal for how we are to use JSON-inside-XML as the overall configuration system for MXCuBE in the future, complete with rules for how it is applied and a rationale for why it is a good idea. 

or 

2) An explanation for why this is a temporary aberration, and how we can make sure that it will go away again without serving as a pattern for future implementations.

fabcor-maxiv:
@HilbertCorsair, what was the issue with XML here again? If this can be expressed with XML, this should be expressed with XML. I looked again at the previous pull request, but could not find the rationale why this does not use XML all the way. Did I miss it? Is there a reasoning written down somewhere?

_[It might be my fault that we focused on the "`eval` vs. JSON" topic, instead of focusing on the "why not XML?", and I am sorry for that.]_

elmjag:
As a side effect for testing this code, we wrote some unit tests for this implementation of TangoShutter:

https://github.com/elmjag/mxcubecore/commit/60769564260604be18d51a0dede9b61ba5ecbacf

Fell free to include this into this PR. If that complicates thing to much, I'll make a separate PR in the future.

HilbertCorsair:
I don't like the current configuartion solution either. I'm allready looking into configuration by yaml but this will take some time. I can try a different configuration using only XML. I will test this Monday. Bottom line is that we need a way to mapp any unconventional local tango shutter value : state pare to the conventional states definded in the class. I's not always necesarry and the code doesn't change anything in the configuration of  Tango shutters that have conventional values. This is why the values tag in the xml file is optional, as a solution for any shuters with unconventional local tango values (CLOSE instead of CLOSED for example). I don't think there is an issue using just XML for this however if we want to have a general Tango shutter class I don't see an easy way arround have to  use 2 configuration sysems (one for shutters with conventional values and one for shutters with unconventional values). The current version does this by importing the json dictionary whenever a shutter has nonconvetional values. I saw this in an example somwhere but I guess we don't neceserly need json.

fabcor-maxiv:
Oh! But I see now, that this is already done with `ast.literal_eval` in one of the parent classes:

https://github.com/mxcube/mxcubecore/blob/e4b2434d1f2aa98dfba3ff10463ed89bebd5795e/mxcubecore/HardwareObjects/abstract/AbstractNState.py#L80-L88

Then we should definitely have stuck to this!

Can we see concrete examples of what we want to write in the XML file that can not be handled by `AbstractNState.py` already?

marcus-oscarsson:
@fabcor-maxiv, @HilbertCorsair, In the end I merged the PR so Im perhaps the one that is responsible for not being more vigilant.  The discussion around `eval` was good an welcome, so nothing to worry about there. @rhfogh Is also completely right that we should definitely avoid sidestepping the current configuration system more than we are already doing with `ast.literal_eval`.  Which in itself is not a XML construct and something we have done because we thought it was convenient.

@fabcor-maxiv  also points out something important, what does `json.loads` solve that can't be done with whats already implemented in the base class `AbstractNState` ?. If the answer is very little I'd say that is worth using whats implemented there and maybe make the necessary changes in `TangoShutter`.

As we said during our last developers meeting, we should now come up with a plan how to migrate from XML to YAML. Ill create an issue for this within shortly where we can discuss how to best proceed.

HilbertCorsair:
I need some advice here before submiting further modifications to this TangoShutter class. As pointed out by @fabcor-maxiv the current configuration for the shuter class uses a json dictionary inside the values tag of the XML configuration file. This can be evidenced in mxcubeweb/tast/HardwareObkectsMockup.xml/safety_shutter.xml and it's not something I came up with @rhfogh . I can change the TangoShutter class to take a different configuration based on pure XML, something like this : 

```
<object class="ShutterMockup">
    <actuator_name>safety_shutter</actuator_name>
    <username>SafetyShutter</username>
    <values>
        <value state="OPEN">open</value>
        <value state="CLOSED">closed</value>
    </values>
</object>
```
This removes the need for a json dictionary but will probably break the initialize_values() method of the AbstractNclass and I think this would be a more significant change to the configuration. 
If we opt for this solution I will have to refactor both the AbstractNState and the TangoShutter class.
Another aspect I'm not sure I understand is : why do we need to account for lists in the values of the configuration file? If my understanding is correct we shouldn't need to use lists in the configuration unless some tango shutter somewhere has a list as specific state. Is this the case? 
This could be configuerd in XML in the following way : 

```
<object class="ShutterMockup">
    <actuator_name>safety_shutter</actuator_name>
    <username>SafetyShutter</username>
    <values>
        <value state="OPEN">open</value>
        <value state="CLOSED">
            <item>"MYSTATE"</item>
            <item>"BUSY"</item>
        </value>
    </values>
</object>
```
Another simpler temporary option would be to live the configuration and AbstractNstate as tehy are and just change the Tango shutter class to use `values = ast.literal_eval(self.get_property("values")) ` as in the AbstractNstate. This would avoid using json but I wouldn't say it's a bretter solution. 
How should I move foreward with this? Do I refactor the AbstractNstate and TangoShutter to remove the json from the configutation? (shouldn't be too hard or impactfull as a change) do I only change the TangoShutter class to use literal_eval instead of json? Living things as they are for now is also an option. 
Any recomandations? @marcus-oscarsson @rhfogh

fabcor-maxiv:
If one of the parent classes already uses `ast.literal_eval()` (`AbstractNState`), then my recommendation is to also use this in the `TangoShutter` class (whether the Tango shutter can use the parent method as-is or if it needs to be overridden with some tweaks, I have not understood).

HilbertCorsair:
@fabcor-maxiv Done!

fabcor-maxiv:
@HilbertCorsair Thanks for your work :)

We plan on testing this latest version tomorrow.

Have you considered adding the [basic unit test](https://github.com/mxcube/mxcubecore/pull/847#issuecomment-1921274838) to this PR?

fabcor-maxiv (approved):

HilbertCorsair:
I just added the basic unit test proposed by @elmjag : elmjag@6076956 . Thanks @fabcor-maxiv

test/pytest/test_hwo_tango_shutter.py
```diff
@@ -0,0 +1,119 @@
...
+from tango.server import Device, command
+from tango.test_context import DeviceTestContext
+from mxcubecore.HardwareObjects import TangoShutter
+
+"""
+Test the HardwareObjects.TangoShutter shutter hardware object.
+"""
+
+
+VALUES_JSON = """
```
elmjag, line 15: Please rename this variable to for example `VALUES_LITERAL`.  The `VALUES_JSON` is now confusing, as technically speaking we are not supporting JSON strings but pythong literal expressions. Event if it so happens in this case that text is identical.

elmjag:
I have tested this PR, it works fine! 

Just one addition nitpick comment. Can we please squash all commits in this PR? 20+ commits feels a bit, ahum, noisy in the git history.

elmjag (approved):

beteva:
Thank you everybody for the efforts to make the perfect TangoShutter Hardware Object. I think this has been a very useful not only for the TangoShutter, bur also to serve as a model not only for similar hardware objects, which would inherit from AbstractNState, but also for hardware objects, using Tango commands and properties.
Being one of the authors of the AbstractNState, I need to explain what is the idea behind _initialise_values and ast,literal_eval.
The VALUES enum is the one, used by any GUI to set/get the value of the HardwareObject.
In the case of NStates, we expect the particular values, not defined in the AbstractNState class to be defined in the configuration file as a dictionary, with keys being the new name for the enum member and values,  which can be of type str, bool, int, float, bur also tuple. We than convert this dictionary to an eval and add it to the already defined members. The dictionary can contain not only new members of the eval, but also redefine the value of already existing ones. For this we use ast.literal_eval and defined the _initialise_values method To make this more clear, here are some real case examples:
`<values>{"IN": "BEAM", "OUT": "OFF"}</values>` - beamstop
`<values>{"IN": False, "OUT": True}</values>` - front and back light in/out
`<values>{"A30": (1, 30, 0.94), "A15": (2, 15, 0.78)}</values>` - aperture  (index,size,factor)
We've chosen a tuple because of the conversion to enum, as VALUES is an enum.
Maybe this will make more clear the <values> implementation for the TangoShutter and its parent class.

test/pytest/test_hwo_tango_shutter.py
beteva: Hi @elmjag, I am a bit puzzled that we need a special test for the TangoShutter and not using the generic shutter test. Do we miss something in the generic shutter test or the ShutterMockup. If such is the case, maybe it is worth adding your work to the generic shutter test, so it would serve other shutter types, not just the Tango ones.

elmjag: These are tests to test code in TangoShutter.py. The generic shutter test code does not cover it, as it is using ShutterMockup hardware object.

More specifically these tests set-up an actual tango device and instantiate a TangoShutter hardware object that talks to that device. This way you check that opening/closing shutter work, with TangoShutter hwo, using real tango stack.

They way I read the code in `test/pytest/test_shutter.py`, it is testing the code in the abstract classes starting with AbstractShutter.

The tests in `test_shutter.py` and `test_hwo_tango_shutter.py` are testing related, but separate aspects of the code base.

marcus-oscarsson:
I'm happy to merge this but there is a conflict, could you have a look @HilbertCorsair ?

PR #850, opened 2024-02-05
Fix ReadTheDocs configuration

fabcor-maxiv:
Use the `--force` flag of `mamba env create` in case environment exists.

Skip `poetry run` when running `sphinx-build` since Poetry installs in the conda environment directly.

Remove the workaround for the unwarranted error about "missing `conda.environment`" configuration setting, because it has been fixed already. See:
* https://github.com/readthedocs/readthedocs.org/pull/10979#issuecomment-1896036953
* https://github.com/readthedocs/readthedocs.org/pull/11040

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -20,15 +20,9 @@ build:
     python: "mambaforge-22.9"
 
   commands:
-    - "mamba env create --file conda-environment-dev.yml"
+    - "mamba env create --file conda-environment-dev.yml --force"
     - "mamba run --name mxcubecore poetry install --only=docs,main"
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
For some reason that I can not pinpoint yet, the ReadTheDocs builds complain that the conda environment already exists, like in this one for example: https://readthedocs.org/projects/mxcubecore/builds/23347508/

I hope using the `--force` flag will smooth things out.

fabcor-maxiv:
I will probably need to do the same for *mxcubeweb*.

fabcor-maxiv:
In this case I will merge myself before external review, because:
* it seems straightforward to me
* it is kind of blocking other pull requests I want to make
* I am more or less the only one who knows this stuff

PR #851, opened 2024-02-06
Clean `conda-environment-dev.yml`

fabcor-maxiv:
Fix the pinned version for `python-ldap`.

Add some explanatory comments.

Remove `pip` since it is already a dependency of `python`.

Remove `openldap` since it is already a dependency of `python-ldap`.


rhfogh (commented):
Good work, much appreciated.

marcus-oscarsson:
:+1:

PR #855, opened 2024-02-22
Minor refactoring of get_object_by_role() in HardwareObjectNode class

HilbertCorsair:
There are probably many aspects that can be improved in the HardwareObjectNode class, particularly in error handling and logging. For now, I just made some minor changes to the get_object_by_role() method.

These changes eliminate unnecessary looping, they also eliminate the infinite loop and hopefully makes the core prettier.

I really hope I didn't miss anything. Please let me know what you think?


mxcubecore/BaseHardwareObjects.py
```diff
@@ -472,31 +472,20 @@ def get_object_by_role(self, role: str) -> Union["HardwareObject", None]:
...
-            if role in obj._objects_by_role:
-                return obj._objects_by_role[role]
+        while objects:
+            if role in self._objects_by_role:
+                return self._objects_by_role[role]
 
-            for object in obj:
-                objects.append(object)
+            try :
+                self = objects.pop()
```
rhfogh: Reassigning self to another value is really confusing (and arguably unpythonic). 

Also the combination of making and popping lists confuses at least me (though it was already there in the previous version), so I might be misunderstanding what this does, but assuming I got it right, I would propose the following code as a better simplification. This does breadth-first traversal of the tree of objects in a single loop without recursive calls

        role = str(role).lower()

        objects = [self]

        for curr in objects:

			result = curr._objects_by_role.get(role)

			if result is None:

				objects.extend(obj for obj in curr if obj)

			else:

				return result

Of course the thing we most need to do with HardwareObjectNode is to replace it with yaml-configured objects, so I would not do any more work than necessary on cleaning it up.

rhfogh (changes requested):
See comment to get_object_by_role

mxcubecore/BaseHardwareObjects.py
HilbertCorsair: Excellent! Thank you, that is indeed more elegant. I tested it and didn't notice any issues. I commited your recomanded changes.

beteva:
Hi @HilbertCorsair,
cleaning the code is a  a good initiative,  I guess, though, having the TangoShutter HO and test here is by mistake (maybe a rebase is missing). Could you, please, remove it. Thank you

rhfogh (approved):

HilbertCorsair:
You're right @beteva sory about that. I did the rebase.

rhfogh (changes requested):
Something is really weird here. The PR now registers as having 75 files changed instead of 3. If you look at the actual changes (most have to do with  GPhL files) it looks like the red code (supposedly the previous state) is *not* the code that is in there now.  The green code looks to a quick look like  it is what is there already, which should be correct. Before we merge this, can someone explain what is going on?

HilbertCorsair:
When I pushed the changes after the rebase I used the --force-with-lease option so all the devellopments that were made on the upstream are kept. The 2 commits related to my refatoring are included with all the other commits made in the meantime by other developers.  I think that's why all those other files show up.
However I would fill better is someone checks if this is ok. I made these last changes on the same local develop branch as I did last time for the TangoShutter. Now that I did a rebase and removed the previous commits I'm not sure the previous changes to  the TangoShutter class are still kept anywhere on the upstream. I made a backup of the branch before the rebase just in case.

HilbertCorsair:
I ran into this method while trying to find a silent error that was causing the app to fail. It turns out it was a JavaScript error on the front end having to do with detector configuration issues. Nevertheless, I thought it would be a good idea to propose a quick refactoring and eliminate unnecessary looping .

rhfogh:
As Antonia said - yes, this is a good initiative. Don't let me discourage you ;-) 

On the rebasing - I tested a bit locally, and I think you should be able to get the kind of tree you want if you make a new branch as a copy of upstream develop, and cherrypick the commits you want to it. What you have now looks like you are recommitting a lot of commits that are already in the develop branch, and even if it could give the right final result, it would be a  horribly complex commit history. Could you give it a try?

HilbertCorsair:
Yes, that sounds like a better idea. Just to make sure I understand, I wil have to close this current PR and create a new pull request from the new branch right ?

HilbertCorsair:
This refactoring will be submitted from a new branch.

PR #856, opened 2024-02-24
Updated versions of DESY P11 HO (PR 800, 854)

agruzinov:
As discussed in PR 800/854 here is the DESY P11 HO objects cumulative PR.


beteva:
Hi @agruzinov, all looks fine for me. Thanks for the effort.
I will wait for another day if somebody has anything to say and will merge the PR.

beteva (approved):

mxcubecore/HardwareObjects/DESY/P11AlbulaView.py
```diff
@@ -26,10 +27,8 @@
 import time
 from PIL import Image
 
-
 sys.path.append("/opt/dectris/albula/4.0/python/")
 
-from mxcubecore.BaseHardwareObjects import HardwareObject
```
marcus-oscarsson, line 33: I like the idea of this the AlbulaView, as I understand it it starts Albula on the same machine as MXCuBE is running and gets the images from the detector zero-mq stream ?

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -374,35 +353,57 @@ def collect_characterisation(
...
-            self.omega_mv(init_pos, self.default_speed)
-            self.collect_std_collection(start_angle, stop_angle)
 
-            diffr.set_omega_velocity(self.default_speed)
-            self.acq_window_off_cmd()
-            self.acq_off_cmd()
+            # self.omega_mv(init_pos, self.default_speed)
+            self.collect_std_collection(start_at, stop_angle)
+
+            # This part goes to standard collection. Otherwise it produces phantom openings.
```
marcus-oscarsson: A general good practice is to remove commented code

mxcubecore/HardwareObjects/DESY/P11EDNACharacterisation.py
```diff
@@ -236,15 +267,24 @@ def input_from_params(self, data_collection, char_params):
         acquisition_parameters = data_collection.acquisitions[0].acquisition_parameters
         path_template = data_collection.acquisitions[0].path_template
 
+        # Make sure there is a proper path conversion between different mount points
+        print(
```
marcus-oscarsson: Consider using log instead of print ?

marcus-oscarsson:
Looks good to me, just a few comments.

mxcubecore/HardwareObjects/DESY/P11AlbulaView.py
agruzinov: Yes, but it is not completely implemented. Also in the python 3.11 this line leads to the segfault anyway. This will require a bit more work yet.

marcus-oscarsson: Ok, i see. Did you also foresee to display a certain image on demand ?

agruzinov: Ideally it is working right now with in-house software as a standard viewer (made by Jan). It shows the monitor interfaces during data collection and than changes to the recently collected file and shows it afterwards, so one can scroll back if needed. I was trying to understand whether the Adxv can show anything during data collection, but before hdf5 is saved, there is no way I guess.

mxcubecore/HardwareObjects/DESY/P11EDNACharacterisation.py
agruzinov: Done, thanks!

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Done, thanks!

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -340,11 +340,11 @@ def collect_characterisation(
         self.omega_mv(start_angle, self.default_speed)
 
         for img_no in range(nimages):
-            print("collecting image %s" % img_no)
+            logging.info("collecting image %s" % img_no)
```
fabcor-maxiv, line 342: I guess here in this case it would be better to use `self.log` instead of `logging` . And maybe also `debug` instead of `info`. So maybe `self.log.debug`.

_[Side note: Logging is all over the place in mxcube, we should seriously look at it at some point. Maybe set up a clear policy, and if possible enforce it with some linting. Feels like everyone writes whatever they feel like on the moment, no one looks at the existing code base or at the docs.]_

mxcubecore/HardwareObjects/DESY/P11AlbulaView.py
marcus-oscarsson: Hm, not sure. I don't think so

mxcubecore/HardwareObjects/DESY/P11Collect.py
marcus-oscarsson: It would indeed be very nice to cleanup the logging, then I think we should be bit more liberal for site specific files like this one.

marcus-oscarsson:
The linting step is failing, Ill merge when its fixed :). Thanks !

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Thanks for the comment. I still need all these logs/debugs because the system is still in active change. Once it is stabilised, the excessive logs should indeed go. But for now, it is hard to debug things without them.

agruzinov:
black and autopep8 are ran on the code but it is not yet happy still... What could it be?

fabcor-maxiv:
> black and autopep8 are ran on the code but it is not yet happy still... What could it be?

Autoflake: https://github.com/mxcube/mxcubecore/actions/runs/8063131751/job/22024333705?pr=856#step:6:29

You should be able to run `pre-commit run` locally.

marcus-oscarsson:
@agruzinov do the follwing in your shell `pre-commit run --all-files` then check the changes with `git gui` and commit the updates.

marcus-oscarsson:
Great ! :+1:

PR #857, opened 2024-02-26
refactoring get_objects_by_role() in HardwareObjectNode class

HilbertCorsair:
This is the refactoring that was discused in my previous PR. This new PR should eliminate the coplications of the previous atempt.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -464,39 +464,24 @@ class HardwareObjectNode:
                 yield obj
 
     def get_object_by_role(self, role: str) -> Union["HardwareObject", None]:
-        """Get hardware object by role.
+            """Get hardware object by role.
 
-        Args:
-            role (str): Role.
+            Args:
+                role (str): Role.
 
-        Returns:
-            Union[HardwareObject, None]: Hardware object.
-        """
-        object = None
-        obj: Self = self
-        objects = []
-        role = str(role).lower()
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Thanks, its a clear improvement very nice :+1: 

There seems to be some linting errors. Apart from that it would be really nice with a test case for this, especially since its a very frequently used piece of code and also because it might get changed when we start working on the xml to yml work.

fabcor-maxiv:
Some of the linting issues are unrelated to this pull request, so I fixed them here: https://github.com/mxcube/mxcubecore/pull/858

rhfogh (approved):
Excellent! Thanks for taking the trouble

marcus-oscarsson:
@rhfogh I think there were still some linting not passing, no problem of fixing them in a follow up PR though.

rhfogh:
Sorry, I guess I jumped the gun there

marcus-oscarsson:
Oh wow, working late :). No problem, I think it happened a few times already and it should be easy enough to fix :) What I think Ill do is to enable the option that blocks merging if the pipeline is failing. However it means that we need to disable the test coverage step for now.

I'm on vacation until Monday, so It has to wait for next week.

rhfogh:
If we start insisting on the pipelines succeeding, then I guess someone would have to help me with mine - it seems to fail every time for my pushes and has never worked as far as I know. But lots of time to look at that later.

PR #858, opened 2024-02-26
Fix end of lines in `doc/json_schema_gui/*.json`

fabcor-maxiv:


marcus-oscarsson:
Excellent !, thanks :)

PR #859, opened 2024-02-28
P11 develop debug

agruzinov:
Debug after upgrading to the latest development.

### DIFF
--- a/mxcubecore/HardwareObjects/DESY/P11Collect.py
+++ b/mxcubecore/HardwareObjects/DESY/P11Collect.py
@@ -66,6 +66,9 @@ class P11Collect(AbstractCollect):
         self.acq_off_cmd = self.get_command_object("acq_off")
         self.acq_window_off_cmd = self.get_command_object("acq_window_off")
 
+        self.latest_frames = None
+        self.acq_speed = None
+
         if None in [
             self.lower_bound_ch,
             self.upper_bound_ch,
@@ -89,15 +92,14 @@ class P11Collect(AbstractCollect):
         HWR.beamline.diffractometer.move_motors(motor_position_dict)
 
     def _take_crystal_snapshot(self, filename):
-        diffr = HWR.beamline.diffractometer
         self.log.debug("#COLLECT# taking crystal snapshot.")
 
-        if not diffr.is_centring_phase():
+        if not HWR.beamline.diffractometer.is_centring_phase():
             self.log.debug("#COLLECT# take_snapshot. moving to centring phase")
-            diffr.goto_centring_phase(wait=True)
+            HWR.beamline.diffractometer.goto_centring_phase(wait=True)
 
         time.sleep(0.3)
-        if not diffr.is_centring_phase():
+        if not HWR.beamline.diffractometer.is_centring_phase():
             raise RuntimeError(
                 "P11Collect. cannot reach centring phase for acquiring snapshots"
             )
@@ -155,7 +157,7 @@ class P11Collect(AbstractCollect):
             )
 
             logging.getLogger("HWR").info(
-                "Collection parameters: %s" % str(self.current_dc_parameters)
+                "Collection parameters: %s", str(self.current_dc_parameters)
             )
 
             log.info("Collection: Storing data collection in LIMS")
@@ -245,9 +247,8 @@ class P11Collect(AbstractCollect):
             log.info("Collection: Updating data collection in LIMS")
             self.update_data_collection_in_lims()
 
-        except:
-            exc_type, exc_value, exc_tb = sys.exc_info()
-            failed_msg = "Data collection failed!\n%s" % exc_value
+        except RuntimeError as e:
+            failed_msg = "Data collection failed!\n%s" % str(e)
             self.collection_failed(failed_msg)
         else:
             self.collection_finished()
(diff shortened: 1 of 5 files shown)


mxcubecore/HardwareObjects/DESY/P11SampleChanger.py
```diff
@@ -486,3 +486,19 @@ def _update_selection(self):
                 s._set_loaded(False)
 
         self._set_selected_sample(None)
+    
+    def _set_loaded_sample(self, sample):
```
beteva, line 490: @agruzinov, is defining this method solves the recurrence problem you've mentioned?

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -144,6 +136,8 @@ def data_collection_hook(self):
 
         osc_pars = self.current_dc_parameters["oscillation_sequence"][0]
         file_info = self.current_dc_parameters["fileinfo"]
+        osc_pars["kappa"] = 0
+        osc_pars["kappa_phi"] = 0
```
beteva, line 276: Do you have minikappa head on your diffarctometer?

agruzinov: No, this is just not to have none values, just in case. Does it make sense?

beteva: It is OK if you would never use minikappa.

mxcubecore/HardwareObjects/DESY/P11SampleChanger.py
agruzinov: Yes, it solves the "loading failed" problem.

agruzinov:
Funny enough after synchronising with the latest develop on github webpage, linting test is failing now, although I did not change a contenent.

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -330,33 +444,29 @@ def collect_characterisation(
...
             stop_angle = start_at + img_range * 1.0
 
-            logging.info("collecting image %s, angle %f" % (img_no, start_at))
+            self.log.debug("collecting image %s, angle %f" % (img_no, start_at))
 
+            # Keep it here for now. It is not clear if it is needed.
             if start_at >= stop_angle:
-                init_pos = start_at
-
+                init_pos = start_at  # - self.acq_speed * self.turnback_time
```
rhfogh: Currently both branches of the 'if' statement are identical - and anyway init_pos is not used so both do nothing. Harmless, but maybe a clean-up or comment-out?

rhfogh (approved):
Fine with me

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Did it, thanks! It is still not clear if this is an issue or not.

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
```diff
@@ -333,7 +333,7 @@ def manual_centring(self, phi_range=120, n_points=3):
             self.user_clicked_event = AsyncResult()
             x, y = self.user_clicked_event.get()
             if click < 2:
-                phi_mot.set_value_relative(phi_range)
+                phi_mot.set_value_relative(phi_range, timeout=5)
```
beteva: If your phi_mot  inherits from AbstractMotor, than only means that you will wait max 5s for the motor to move an will raise TimeoutError if the movement takes longer. It does not give you an extra time for the clicking.
Maybe your problem was that by default the timeout=0, meaning, do not wait for the movement to finish.

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -470,18 +577,19 @@ def add_h5_info(self, h5file):
...
+                #     1.0,
+                #     0.0,
+                #     0.0,
+                # ]
+                # h5fd["entry/instrument/detector/module/fast_pixel_direction"].attrs[
+                #     "vector"
+                # ] = [1.0, 0.0, 0.0]
+                # h5fd["entry/instrument/detector/module/slow_pixel_direction"].attrs[
+                #     "vector"
+                # ] = [0.0, 1.0, 0.0]
```
beteva, line 592: This represents the direction of the diffractometer head in your metadata image. If your detector is Eiger and you use the dectris filewriter, these values are set by default  as horizontal orientation, so you need to change to something else if your diffractometer is not the same.

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
agruzinov: Thanks for the comment! What would be the best way to mitigate it? I do not see any "wait" option there.

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Thanks!

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
beteva: timeout=None waits forever. But you can as well put some big value, like 2 min (timeout=60). This is only security - the command will return as soon as the movement has finished. 

You may want to have a look at the AbstractMotor.py. It says:
` timeout (float): optional - timeout [s],
                             If timeout == 0: return at once and do not wait;
                             if timeout is None: wait forever
`

agruzinov: OK, thanks for the tip. Cheers!

PR #860, opened 2024-02-28
New XRF implementation

beteva:
- Change the AbstractXRFSpectrum class:
  - snake case method names.
  - remove unnecessary methods.
  -  use the standard HardwareObjectState states and emit stateChanged only.
- Change the xrf queue model to call the new methods and to use the strateChanged signal only
- Add tests


beteva:
WIP as testing not finished yet.

mxcubecore/HardwareObjects/mockup/XRFSpectrumMockup.py
```diff
@@ -17,53 +17,28 @@
...
-    250,
-    300,
-    200,
-    100,
-    0,
-    0,
-    0,
-    90,
-]
+__copyright__ = """ Copyright © 2010-2023 by the MXCuBE collaboration """
```
fabcor-maxiv: Skip the years maybe?

```suggestion
__copyright__ = """ Copyright © by the MXCuBE collaboration """
```

beteva: yes, you are right. I've missed this one

meguiraun:
will try to test this in the beamline next week, on a quick look it looks good!

beteva:
Thanks @meguiraun. For information it works on the ESRF beamlines.

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
```diff
@@ -69,33 +67,34 @@ def start_xrf_spectrum(
...
             prefix (str): File prefix
             session_id (int): Session ID number (from ISpyB)
             blsample_id (int): Sample ID number (from ISpyB)
         """
         self.spectrum_info_dict = {"sessionId": session_id, "blSampleId": blsample_id}
-
         integration_time = integration_time or self.default_integration_time
         self.spectrum_info_dict["exposureTime"] = integration_time
-
+        self.spectrum_info_dict["filename"] = ""
```
marcus-oscarsson, line 92: It would probably nice to document which keys that are used in the spectrum_info_dict, even nicer would be to use a Pyndatic model.

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
```diff
@@ -105,22 +104,41 @@ def start_xrf_spectrum(
...
+        gevent.spawn(
+            self.execute_xrf_spectrum,
+            integration_time,
+            self.spectrum_info_dict["filename"],
+        )
         return True
 
-    def execute_xrf_spectrum(self):
-        """Do the acquisition"""
+    def execute_xrf_spectrum(self, integration_time=None, filename=None):
```
marcus-oscarsson: I guess we don't need to add `xrf_spectrum` in the method names as its on a `XRFSpectrum` object maybe `execute`, `start` and so on is enough ?

mxcubecore/HardwareObjects/ESRF/ESRFXRFSpectrum.py
```diff
@@ -18,94 +18,122 @@
...
         self.default_erange = literal_eval(
             self.get_property("default_energy_range", "[2.0, 15.0]")
         )
         self.cfg_energies = literal_eval(
             self.get_property("cfg_energies", "[7, 9, 12, 15]")
         )
 
-    def _doSpectrum(self, ctime, filename, wait=None):
-        return self.choose_attenuation(ctime, filename)
+    def _doSpectrum(self, ctime, filename):
```
marcus-oscarsson, line 85: watch out camel case :)

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
```diff
@@ -27,10 +27,8 @@
  - _execite_xrf_scan, spectrum_analyse, 
 
 Emit signals:
- - xrfSpectrumStarted
- - xrfSpectrumFinished
- - xrfSpectrumFailed
- - xrfSpectrumStatusChanged
+ - stateChanged
+ - xrfSpectrumStatusChanged (with error message)
```
marcus-oscarsson: You can perhaps use the new way of documenting signals that @meguiraun added ?

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
beteva: @marcus-oscarsson,  these are ISPYB keys, which I do not know how they will evolve. For time being I would not document this part.

mxcubecore/HardwareObjects/ESRF/ESRFXRFSpectrum.py
beteva: This is here for compatibility reasons. I've put a big deprecation warning and should remove it when everybody is OK.

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
beteva: Once we agree on the model. This was still under discussion., I believe.

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
beteva: Maybe. Have not thought about the implication.

mxcubecore/HardwareObjects/ESRF/ESRFXRFSpectrum.py
marcus-oscarsson: Ok

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
marcus-oscarsson: I think we eventually decided to put this on the Class level right @meguiraun, @fabcor-maxiv

marcus-oscarsson:
ok, looks fine to me then. We can follow up with a PR for the signal documentation and perhaps eventually something with a model instead of the dictionary for the ISPyB values

mxcubecore/HardwareObjects/abstract/AbstractXRFSpectrum.py
fabcor-maxiv: > This was still under discussion., I believe.

What exactly?


> we eventually decided to put this on the Class level

If that is about that, then yes, we agreed it is better at the class level: https://github.com/mxcube/mxcubecore/pull/866#issuecomment-2014761065

If help is needed regarding the exact formatting of the docstring for emitted signals, ping me.

beteva: OK, I'll change it.

PR #862, opened 2024-03-04
Linted BaseHardwareObjects.py left out in #857

marcus-oscarsson:
Linted BaseHardwareObjects.py left out in #857


rhfogh (approved):
Ooops! Good it is fixed

PR #864, opened 2024-03-07
Update GPhL wokflow centring handling

rhfogh:
Added capability to pass actual (and not just predicted) centrings to GPhL workflow. All changes are in GPhL-specific code


rhfogh:
Thanks!

PR #866, opened 2024-03-14
[Documentation Code Camp] docstrings for emitted signals

meguiraun:
Initial work for documenting emitted signals. I initially focused  on the abstract classes (not all done yet 🤪)

Example output:
image


fabcor-maxiv:
Wouldn't it make more sense to document the emitted signals at the class level rather than at the method level?

If I use an *actuator* I want to know what signals it emits and when (under what circumstances, i.e. what is the meaning of the signal), but I do not need to know which of the actuator's methods actually emits the signal, do I?

And same thing, if I want to implement a new (concrete) actuator, I want to know what signals my actuator should emit, but I will choose which methods will actually emit those signals.

cc @elmjag

beteva:
@meguiraun thanks for the work. Good start. There is one slight problem - I think AbstractAperture is an obsolete class, There will be very soon a new PR to handle aperture test and mock up, but they inherit from AbstractNState. In this context, would you mind if we see together which are the obsolete abstract classes.
@fabcor-maxiv I guess it is all right if we document the signals on the class level only. Actually this is the case right now, but certainly needs a bit of explanation which signal means what.

Maybe we should make a general signals documentation file, which lists all the "standard" signals, like valueChanged, limitsChanges, stateChanged, specificStateChanged, as well as the methods defined in  BaseHardwareObjects.py - the emit method (namely its signature in order to explain how to emit a signal) and the force_emit_signals method,  available for overloading . If we go for this, than we could explain the usage of these signals and in the classes only mention which are the emitted  signals,  and only document if there are other signals, specific for the class. What do you think?

rhfogh:
I agree with @beteva - The biggest need is what are the general kinds of signals that are used to tie the application together, where they come from, and what they do. In a way the biggest question is not so much 'which signals does this class emit' but 'which signals should I listen for' and 'which signals do I need to emit in my own class?'.

meguiraun:
okey, I see your point. I will leave the methods out.

yes, totally agree with a general documentation page (why, where in code are coming, some basic example, etc.), I had it in mind but I thought to start with documenting the code

meguiraun:
1. Added a new md with signal documentation, very early, please @beteva @rhfogh  have a look, and if you feel so commit any improvement
2. Updated, let me know if the "new" AbstracActuator class documentation is ok:

image

mxcubecore/HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -37,7 +44,25 @@
...
+        value specified by xml property, otherwise None
+    _nominal_limits : tuple
+        values specified by xml property, otherwise None
+    actuator_name : str
+        actuator name specified by xml property, otherwise None
+    read_only : bool
+        read only flag, specified by xml property, otherwise False
+
+    username : int
+        the number of legs the animal has (default 4)
```
fabcor-maxiv: What!? :D

```suggestion
```

mxcubecore/HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -37,7 +44,25 @@
...
+        Current actuator value
+    default_value : float
+        value specified by xml property, otherwise None
+    _nominal_limits : tuple
+        values specified by xml property, otherwise None
+    actuator_name : str
+        actuator name specified by xml property, otherwise None
+    read_only : bool
+        read only flag, specified by xml property, otherwise False
+
```
fabcor-maxiv:
```suggestion
    Attributes:
        _nominal_value (float):
            Current actuator value
        default_value (float):
            value specified by XML property, otherwise `None`
        _nominal_limits (tuple):
            values specified by XML property, otherwise `None`
        actuator_name (str):
            actuator name specified by XML property, otherwise `None`
        read_only (bool):
            read only flag, specified by XML property, otherwise `False`
```

mxcubecore/HardwareObjects/abstract/AbstractBeam.py
```diff
@@ -213,6 +219,13 @@ def evaluate_beam_info(self):
     def re_emit_values(self):
         """
         Reemits beamSizeChanged and beamInfoChanged signals
+
+        Emits:
+            beamSizeChanged:  ('beamSizeChanged', (width, height)))
```
fabcor-maxiv:
```suggestion
            beamSizeChanged:  ('beamSizeChanged', (width, height))
```

fabcor-maxiv (changes requested):

docs/source/dev/hwobj_signals.md
fabcor-maxiv, line 1: Personal preference, we do not have a policy for this as far as I know: I tend to prefer "_[Semantic Line Breaks](https://sembr.org/)_".

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
+# MXCuBECore HardwareObject signals
+
+:Created: 20240318
+
+MXCuBE relies heavily on signals beeing emmited and listened to by many elements. For example, a hardware object may listen to other lower level hardware objects in order to update values after some calculation. But it is also critical in the UI (both web and QT), in both cases they expect periodic signal updates for displaying the most recent information to the user (e.g. motor positions, data collection state, etc.)
```
fabcor-maxiv:
```suggestion
MXCuBE relies heavily on signals being emitted and listened to by many elements. For example, a hardware object may listen to other lower level hardware objects in order to update values after some calculation. But it is also critical in the UI (both web and Qt), in both cases they expect periodic signal updates for displaying the most recent information to the user (e.g. motor positions, data collection state, etc.)
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Implementation
+
+Depending on the installed modules signals are emitted using [Louie](https://pypi.org/project/Louie/) or [PyDispatcher](https://pypi.org/project/PyDispatcher/). The former being based on the later. The developer does not need to deal with the differences between those two modules as it is already being handled in the file [dispatcher](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/dispatcher.py).
```
fabcor-maxiv:
```suggestion
Depending on the installed modules, signals are emitted using [Louie](https://pypi.org/project/Louie/) or [PyDispatcher](https://pypi.org/project/PyDispatcher/). The former being based on the later. The developer does not need to deal with the differences between those two modules as it is already being handled in the file [dispatcher](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/dispatcher.py).
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+ **_NOTE:_**  can we remove of of those dependencies?
```
fabcor-maxiv:
```suggestion
 **_NOTE:_**  can we remove any of those dependencies?
```

docs/source/dev/hwobj_signals.md
rhfogh: Fabien only just introduced me to semantic line breaks, but I am heartily in favour.

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+
+
+> PyDispatcher provides the Python programmer with a multiple-producer-multiple-consumer signal registration and routing infrastructure for use in multiple contexts
+
+
+When certain events or conditions occur within a hardware object, corresponding signals are emitted to inform connected components or modules about these changes.
+
+The [BaseHardwareObject](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/BaseHardwareObjects.py) class serves as the base class for all hardware objects in MXCuBE. It includes methods for defining and emitting signals, allowing derived classes to customize signal emission based on their specific requirements.
```
fabcor-maxiv:
```suggestion
The {py:class}`mxcubecore.BaseHardwareObjects.HardwareObject` class serves as the base class for all hardware objects in MXCuBE. It includes methods for defining and emitting signals, allowing derived classes to customize signal emission based on their specific requirements.
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Emit
+
+Signals are typically emitted when the state of a hardware object changes, such as when it becomes ready for operation, encounters an error, or completes a task. Additionally, signals may be emitted to indicate changes in parameters or settings of the hardware, such as new setpoints, values, or configuration options.
+
+To emit a signal, derived classes can use the `emit` method provided by the BaseHardwareObject class. This method takes the name of the signal as an argument and optionally includes additional data or parameters to pass along with the signal. This method calls the `dispatcher.send` method.
```
fabcor-maxiv:
```suggestion
To emit a signal, derived classes can use the {py:meth}`mxcubecore.BaseHardwareObjects.HardwareObjectMixin.emit` method provided by the {py:class}`HardwareObject` class. This method takes the name of the signal as an argument and optionally includes additional data or parameters to pass along with the signal. This method calls the `dispatcher.send` method.
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+
+
+From the _BaseHardwareObject_ class (removing extra lines for brevity):
```
fabcor-maxiv:
```suggestion
From the {py:class}`HardwareObject` class (removing extra lines for brevity):
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+    def emit(self, signal: Union[str, object, Any], *args) -> None:
+        signal = str(signal)
+
+        if len(args) == 1:
+            if isinstance(args[0], tuple):
+                args = args[0]
+        dispatcher.send(signal, self, *args)
+```
+
+So, in a custom hardware object, since inhnerits from _BaseHardwareObject_, one only needs to call:
```
fabcor-maxiv:
```suggestion
So, in a custom hardware object, since it inherits from {py:class}`HardwareObject`, one only needs to call: 
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+So, in a custom hardware object, since inhnerits from _BaseHardwareObject_, one only needs to call: 
+```
+self.emit('my_signal', new_value)
+```
+
+### Receive
+
+_BaseHardwareObject_ implements the following ```connect```method, built around the homonymous method of _PyDispatcher_. Making it more convenient to use. The functions provides syntactic sugar ; Instead of ```self.connect(self, "signal", slot)``` it is possible to do ```self.connect("signal", slot)```
```
fabcor-maxiv:
```suggestion
{py:class}`HardwareObject` implements the following {py:meth}`HardwareObjectMixin.connect` method, built around the homonymous method of _PyDispatcher_. Making it more convenient to use. The functions provides syntactic sugar: instead of `self.connect(self, "signal", slot)` it is possible to do `self.connect("signal", slot)`.
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+
+From the [BaseHardwareObject](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/BaseHardwareObjects.py#L878) class (removing extra lines for brevity):
```
fabcor-maxiv:
```suggestion
From the {py:meth}`HardwareObjectMixin.connect` method (removing extra lines for brevity):
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+            sender.connect_notify(signal)
+```
+
+And an example usage on a custom hardware object would be:
+
+```
+    self.connect(some_other_hwobj, "a_signal", callback_method)
+```
+
+This assumes that ```some_other_hwobj``` is linked in the custom hwobj initialization, and _callback_method_ must exist, otherwise an exception will happen once the signal is received.
```
fabcor-maxiv:
```suggestion
This assumes that `some_other_hwobj` is linked in the custom hardware object initialization, and `callback_method` must exist, otherwise an exception will happen once the signal is received.
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+
+If the sender hwobj has a method named ```connect_notify```, it will be called on connect. Since this connect happens at application init, this typically triggers the emission of all signals during initialization, and thus all receivers start with the most recent value.
```
fabcor-maxiv:
```suggestion
If the sender hardware object has a method named `connect_notify`, it will be called on connect. Since this connect happens at application initialization, this typically triggers the emission of all signals during initialization, and thus all receivers start with the most recent value.
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,208 @@
...
+2024-03-18 12:21:30.426785 | valueChanged emitted, new value: 0.3473955083798488
+2024-03-18 12:21:30.427018 | valueChanged callback, arguments: (0.3473955083798488,)
+2024-03-18 12:21:33.427715 | valueChanged emitted, new value: 0.9503048610962694
+2024-03-18 12:21:33.427902 | valueChanged callback, arguments: (0.9503048610962694,)
+In [3]: ho1.stop()
+```
+
+As you can see, the second hardware object receives and processes first one's signal.
+
+> At least one entry must appear in the beamline config yml file, in this case I left the procedure mockup, all the other mockups are commented. Thats why only a few items appear in the loading table.
```
fabcor-maxiv:
```suggestion
> At least one entry must appear in the beamline's YAML configuration file. In this case I left the procedure mockup only, all the other mockups are commented. That is why only a few items appear in the loading table.
```

fabcor-maxiv (changes requested):
I do not think there is a `BaseHardwareObject` class. So I replaced with `HardwareObject`. But in most cases what we talk about is actually in the `HardwareObjectMixin`.

I tried the Sphinx/docutils cross-reference roles, we'll see if we get better links.

In principle we should be able to configure Sphinx to keep a copy of the source code, so that we can have links to source code in the documentation (not links to GitHub).

And in general, not only for (Sphinx) documentation, we should avoid links to a branch, but use permalinks that link to a specific commit.

fabcor-maxiv:
In many of the modified docstrings, it feels like adding the correct type hints would be nearly as efficient (especially for the return values).

rhfogh:
@fabcor-maxiv  HardwareObjectMixin woudl be the right choice throughout. HardwareObjects inherit from *either* HardwareObject *or* HardwareObjectYaml, but both inherit from HardwareObjectMixin,.

docs/source/dev/hwobj_signals.md
elmjag: wait, semantic break is the same things as putting each sentence on it's own line?

fabcor-maxiv: > semantic break is the same things as putting each sentence on it's own line?

No.

I never read the full document, only the introduction, I do not see much point in being strict about this. But to me, the important thing is to break long paragraphs where it makes sense logically.

So in Markdown I would write like this:

```none
I never read the full document, only the introduction,
I do not see much point in being strict about this.
But to me, the important thing is to break long paragraphs
where it makes sense logically.
```

meguiraun:
preview in https://mxcubecore--866.org.readthedocs.build/en/866/dev/hwobj_signals.html

beteva:
Very good job @meguiraun 👍
The device and the equipment signals are definitely deprecated. Quite some time ago there was a decision not to use Device and Equipment class any more.
You are right, we should remove the update signal and only use valueChanged.
You can also remove the xrf related signals - in my WIP for xrf I've changed them all to stateChanged

marcus-oscarsson:
Well done ! very nice !. As Antonia already pointed out, we should really start to remove the Device and Equipment classes.

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,244 @@
...
+Depending on the installed modules, signals are emitted using [Louie](https://pypi.org/project/Louie/) or [PyDispatcher](https://pypi.org/project/PyDispatcher/). The former being based on the later. The developer does not need to deal with the differences between those two modules as it is already being handled in the file [dispatcher](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/dispatcher.py).
+ **_NOTE:_**  can we remove any of those dependencies?
+
+
+> PyDispatcher provides the Python programmer with a multiple-producer-multiple-consumer signal registration and routing infrastructure for use in multiple contexts
+
+
+When certain events or conditions occur within a hardware object, corresponding signals are emitted to inform connected components or modules about these changes.
+
+The {py:class}`mxcubecore.BaseHardwareObjects.HardwareObject` class serves as the base class for all hardware objects in MXCuBE. It includes methods for defining and emitting signals, allowing derived classes to customize signal emission based on their specific requirements.
```
rhfogh, line 18: Strictly speaking it is the HardwareObject  OR HardwareObjectYaml class (both inherit from HardwareObjectMixin). Once we unify the YAML and XML configuration this distinction should hopefully disappear.

meguiraun: added this clarification in the text

meguiraun:
while the signals are still there, I simply marked as deprecated. 
Also added queue and diffractometer

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,339 @@
+# MXCuBECore HardwareObject signals
+
+:Created: 20240318
+
+MXCuBE relies heavily on signals being emitted and listened to by many elements. For example, a hardware object may listen to other lower level hardware objects in order to update values after some calculation. But it is also critical in the UI (both web and Qt), in both cases they expect periodic signal updates for displaying the most recent information to the user (e.g. motor positions, data collection state, etc.)
+
+## Implementation
+
+Depending on the installed modules, signals are emitted using [Louie](https://pypi.org/project/Louie/) or [PyDispatcher](https://pypi.org/project/PyDispatcher/). The former being based on the later. The developer does not need to deal with the differences between those two modules as it is already being handled in the file [dispatcher](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/dispatcher.py).
```
axelboc:
```suggestion
Depending on the installed modules, signals are emitted using [Louie](https://pypi.org/project/Louie/) or [PyDispatcher](https://pypi.org/project/PyDispatcher/). The former being based on the latter. The developer does not need to deal with the differences between those two modules as it is already being handled in the file [`dispatcher.py`](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/dispatcher.py).
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,339 @@
...
+
+
+> PyDispatcher provides the Python programmer with a multiple-producer-multiple-consumer signal registration and routing infrastructure for use in multiple contexts
+
+
+When certain events or conditions occur within a hardware object, corresponding signals are emitted to inform connected components or modules about these changes.
+
+The {py:class}`mxcubecore.BaseHardwareObjects.HardwareObject` class serves as the base class for all hardware objects in MXCuBE. It includes methods for defining and emitting signals, allowing derived classes to customize signal emission based on their specific requirements.
+
+>Strictly speaking it is the HardwareObject OR HardwareObjectYaml class (both inherit from HardwareObjectMixin). Once we unify the YAML and XML configuration this distinction should hopefully disappear.
```
axelboc:
```suggestion
> Strictly speaking it is the `HardwareObject` OR `HardwareObjectYaml` class (both inherit from `HardwareObjectMixin`). Once we unify the YAML and XML configuration, this distinction should hopefully disappear.
```

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,339 @@
...
+
+```
+    def emit(self, signal: Union[str, object, Any], *args) -> None:
+        signal = str(signal)
+
+        if len(args) == 1:
+            if isinstance(args[0], tuple):
+                args = args[0]
+        dispatcher.send(signal, self, *args)
+```
```
axelboc, line 39: Not convinced this code sample is needed. Same with `connect` below. One can always go look in the classes mentioned. There is also the risk that the docs might fall out of sync in the future if the code ever changes.

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,339 @@
...
+2024-03-18 12:21:30.426785 | valueChanged emitted, new value: 0.3473955083798488
+2024-03-18 12:21:30.427018 | valueChanged callback, arguments: (0.3473955083798488,)
+2024-03-18 12:21:33.427715 | valueChanged emitted, new value: 0.9503048610962694
+2024-03-18 12:21:33.427902 | valueChanged callback, arguments: (0.9503048610962694,)
+In [3]: ho1.stop()
+```
+
+As you can see, the second hardware object receives and processes first one's signal.
+
+> At least one entry must appear in the beamline's YAML configuration file. In this case I left the procedure mockup only, all the other mockups are commented. That is why only a few items appear in the loading table.
```
axelboc: Maybe reword to avoid using the first person "I"?

docs/source/dev/hwobj_signals.md
```diff
@@ -0,0 +1,339 @@
...
+
+## General signals List
+
+The following tables list the generic signals as well the signals from the some of the most important hardware objects.
```
axelboc:
```suggestion
The following tables list the generic signals as well as the signals from some of the most important hardware objects. 
```

mxcubecore/HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -22,7 +22,14 @@
 Defines the set/update value, get/set/update limits and validate_value
 methods and the get_value and _set_value abstract methods.
 Initialises the actuator_name, username, read_only and default_value properties.
-Emits signals valueChanged and limitsChanged.
+
+Emits:
+    valueChanged: during hwobj initialization and when setting a new value. Format: (   valueChanged, new_value)
```
axelboc:
```suggestion
    valueChanged: during hwobj initialization and when setting a new value. Format: (valueChanged, new_value)
```

mxcubecore/HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -37,7 +44,21 @@
...
+        _nominal_value (float):
+            Current actuator value
+        default_value (float):
+            value specified by XML property, otherwise `None`
+        _nominal_limits (tuple):
+            values specified by XML property, otherwise `None`
+        actuator_name (str):
+            actuator name specified by XML property, otherwise `None`
+        read_only (bool):
+            read only flag, specified by XML property, otherwise `False`
```
axelboc:
```suggestion
            read-only flag, specified by XML property, otherwise `False`
```

mxcubecore/HardwareObjects/abstract/AbstractActuator.py
```diff
@@ -175,10 +204,10 @@ def re_emit_values(self):
         super(AbstractActuator, self).re_emit_values()
 
     def force_emit_signals(self):
-        """Forces to emit all signals.
+        """
+        Forces to emit all signals.
```
axelboc: "Force" seems implied.

```suggestion
        Emit all signals.
```

Alternatively, in the next file, you've used the word "Reemit".

mxcubecore/HardwareObjects/abstract/AbstractAperture.py
```diff
@@ -53,21 +53,24 @@ def init(self):
             logging.getLogger("HWR").error("Aperture: no position list defined")
 
     def get_diameter_size_list(self):
-        """
+        """Get diameters size list
```
axelboc: Might as well be a bit more verbose:

```suggestion
        """Get list of diameter sizes.
```

mxcubecore/HardwareObjects/abstract/AbstractAperture.py
```diff
@@ -53,21 +53,24 @@ def init(self):
...
+
         Returns:
             list: list of diameter sizes in microns
         """
         return self._diameter_size_list
 
     def get_position_list(self):
-        """
+        """Get positions list
```
axelboc:
```suggestion
        """Get list of positions.
```

and so on ...

It's not always needed; for instance, the next one reads fine: "Get current diameter index"

mxcubecore/HardwareObjects/abstract/AbstractAperture.py
```diff
@@ -171,15 +181,20 @@ def set_out(self):
...
+
         Returns:
             bool: True if aperture is in the beam, otherwise returns false
         """
         pass
 
     def force_emit_signals(self):
-        """
-        Reemits all signals
+        """Reemits all signals
```
axelboc: I would stick with the indicative mood everywhere:

```suggestion
        """Reemit all signals.
```

mxcubecore/HardwareObjects/abstract/AbstractCharacterisation.py
```diff
@@ -151,7 +155,8 @@ def get_default_characterisation_parameters(self, default_input_file):
     def prepare_input(self, _input):
         """Called by characterise before characterisation starts, can be used
         to manipulate edna_input object before characterisation starts.
-        Example: to set a site specific output directory
+        Example: to set a site specific output directory.
```
axelboc:
```suggestion
        Example: to set a site-specific output directory.
```

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -143,7 +148,15 @@ def collect(self, owner, dc_parameters_list):
 
     def do_collect(self, owner):
         """
-        Actual collect sequence
+        Actual collect sequence.
```
axelboc:
```suggestion
        Actually collect the sequence.
```

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -317,7 +340,12 @@ def collection_stopped(self):
             self.data_collect_task.kill(block=False)
 
     def collection_finished(self):
-        """Collection finished beahviour"""
+        """Collection finished beahviour
```
axelboc:
```suggestion
        """Collection finished behaviour
```

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -747,14 +834,16 @@ def move_to_centered_position(self):
...
+
+        Args:
+            motor_position_dict (dict): key value with motor names and desired position
         """
         return
 
     def take_crystal_snapshots(self):
         """
-        Descript. :
+        Take crytal snapshots of the currently loaded sample
```
axelboc:
```suggestion
        Take crystal snapshots of the currently loaded sample
```

docs/source/dev/hwobj_signals.md
meguiraun: yeah I know, but I still think it makes the description easier to understand. That code has not changed much anyway in the recent mxcube history

fabcor-maxiv (approved):

beteva (commented):
Very nice.
Apparently we need changes in many python files for better documentation. I suggest we see together with the other developers who does what, but I like the class description you've made for AbstractActuator. We should follow the same way.

meguiraun:
AbstractBeam with docstrings in class [here](https://mxcubecore--866.org.readthedocs.build/en/866/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractBeam.html)
AbstracActuator with docstrings outside class [here](https://mxcubecore--866.org.readthedocs.build/en/866/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractActuator.html)

I can remove all changes to the abstract classes if it makes the other PR easier to handle

ping @beteva

fabcor-maxiv:
> AbstractBeam with docstrings in class [here](https://mxcubecore--866.org.readthedocs.build/en/866/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractBeam.html)
> AbstracActuator with docstrings outside class [here](https://mxcubecore--866.org.readthedocs.build/en/866/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractActuator.html)

For me, this belongs in the class docstring : )

meguiraun:
> > AbstractBeam with docstrings in class [here](https://mxcubecore--866.org.readthedocs.build/en/866/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractBeam.html)
> > AbstracActuator with docstrings outside class [here](https://mxcubecore--866.org.readthedocs.build/en/866/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractActuator.html)
> 
> For me, this belongs in the class docstring : )

I think same way, it seems its place

beteva:
@meguiraun, agree the best place is the class docstrings.

marcus-oscarsson:
Hey @meguiraun, do you have an update on this one, it seemed almost done ?

meguiraun:
completely lost track of this one... so sorry...

If people are happy I am also happy. Not all the hwobjs are covered tough

fabcor-maxiv:
Not all checks are successful. After this is fixed, I think it is fine to merge. And then I might do follow-ups as I had already started in <https://github.com/mxcube/mxcubecore/pull/947>.

meguiraun:
all checks are green now!

marcus-oscarsson:
Sorry my mistake. I must have missed the last comments, was this ready to be merged ?

marcus-oscarsson:
@fabcor-maxiv, @elmjag. Sorry to take such a long time with this one. Is this one ready to be merged now. I think there's been a conflict for some time now ?

fabcor-maxiv:
It is still on my mind. I plan to take over this one and also integrate some of the work I started in <https://github.com/mxcube/mxcubecore/pull/947>. I assigned it to me.

fabcor-maxiv:
Done in <https://github.com/mxcube/mxcubecore/pull/1116>.

PR #867, opened 2024-03-14
Set `sphinx-material` theme for documentation

fabcor-maxiv:


fabcor-maxiv:
Preview: https://mxcubecore--867.org.readthedocs.build/en/867/

Prompted by: https://github.com/mxcube/mxcubeweb/issues/1116

marcus-oscarsson:
It looks quite nice, but there seems to be items missing in the contents right ?

marcus-oscarsson:
thanks, It looks nice but it has the same issue as `furo` with the API contents I believe (not all the text is shown)

fabcor-maxiv:
There has to be a way to improve this. I have some ideas, but have not had the time to try them yet. Well, one or two I tried and it did not improve things.

Maybe there is another theme out there that fits our needs better.

marcus-oscarsson:
I'm sure there is some way as well, lets go with whats easiest for now unless you really want to dig into this.

marcus-oscarsson:
Input on this one, do we keep the current layout or do we try this one, seeing that both are suffering from similar limitations when it comes to the left side menu ?

fabcor-maxiv:
The one we have at least does not truncate the names in left and right columns, and in general I think I prefer the one we currently have.

So if it were only me, for now, I would leave as is (furo).

But I still intend to spend some time fiddling with configuration settings and themes and maybe I find a better compromise.

`furo` (current):

image

`sphinx-material`:

image

marcus-oscarsson:
Ok, We will keep it open for now then, you can close it if you consider it as done

marcus-oscarsson:
Closing this for now, please feel free to re-open if you like

PR #868, opened 2024-03-14
Documentation for Yaml/Xml configured files

rhfogh:


docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+HardwareObjects can be configured to contain both other hardware objects and properties. The former are identified relative to the container by a role.
+
+## Finding the files
+Configuration files are searched for by name in a series of directories given as a lookup path. This is specified either with the --coreConfigPath command line parameter to MXCuBE, or through the MXCUBE_CORE_CONFIG_PATH environment variable. The files live under the mxcubecore/mxcubecore/configuration/ directory, typically with a directory for each beamline (which may or may not be up to date relative to the actual bamline environment).
```
fabcor-maxiv:
```suggestion
Configuration files are searched for by name in a series of directories given as a lookup path. This is specified either with the `--coreConfigPath` command line parameter to MXCuBE, or through the `MXCUBE_CORE_CONFIG_PATH` environment variable. The files live under the `mxcubecore/mxcubecore/configuration` directory, typically with a directory for each beamline (which may or may not be up to date relative to the actual beamline environment).
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
+# Configuration files
+
+The MXCuBE core is organised as a tree of *HardwareObjects*, and configuration is organised with a file per HardwareObject. The (mandatory) topmost object in the tree is the *Beamline* object, which is configured in the beamline_config.yml file. This in turn contains the names of other configuration files, recursively, which may be a mixture of yaml and xml configuration files. The Beamline object can be accessed as an attribute of the HardwareRepository module ('*HardwareRepository.beamline*'). As of 20240314 mxcube is in the middle of a change-over to a new system for configuration that uses yaml files instead of xml files, which has different ways of accessing the configuration data from inside the code.
```
fabcor-maxiv:
```suggestion
The MXCuBE core is organised as a tree of *HardwareObjects*, and configuration is organised with a file per HardwareObject. The (mandatory) topmost object in the tree is the *Beamline* object, which is configured in the `beamline_config.yml` file. This in turn contains the names of other configuration files, recursively, which may be a mixture of YAML and XML configuration files. The Beamline object can be accessed as an attribute of the HardwareRepository module (`HardwareRepository.beamline`). As of 2024-03-14 MXCuBE is in the middle of a change-over to a new system for configuration that uses YAML files instead of XML files, which has different ways of accessing the configuration data from inside the code.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+HardwareObjects can be configured to contain both other hardware objects and properties. The former are identified relative to the container by a role.
+
+## Finding the files
+Configuration files are searched for by name in a series of directories given as a lookup path. This is specified either with the --coreConfigPath command line parameter to MXCuBE, or through the MXCUBE_CORE_CONFIG_PATH environment variable. The files live under the mxcubecore/mxcubecore/configuration/ directory, typically with a directory for each beamline (which may or may not be up to date relative to the actual bamline environment).
+
+## Yaml-configured objects
+### Code and file structure
+Each yaml-configured object has a 'name' attribute, which is equal to the role that identifies the object within the containing object (the name of the Beamline object is 'beamline').
```
fabcor-maxiv:
```suggestion
Each YAML-configured object has a `name` attribute, which is equal to the role that identifies the object within the containing object (the name of the Beamline object is `beamline`).
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
...added in the configuration must match a pre-defined attribute coded in the class. This means that you can look in the class code to see which attributes are  available.  The only exceptinon is the *_initialise_class* attribute at the start of the file. This dictionary contains the import name of the class that is to be created, and optionally parameters to be passed to the init() function of that class. The *_objects* attribute gives the HardwareObjects that are contained in (i.e. children of) the object. The dictionary key is the role name, and the value is the name of the configuration file. Each role_name must match a read-only property coded in the body of the class, and must be added to the *__content_roles* list of the class by the class code. Note that classes are loaded and initialised in the order given by this list, so that there is a reproducible loading order. Contained objects can be defined as procedures, so that they are added to the list of procedures. Each Yaml-configured class has an _init() function that is executed immediately after the object is created, and and init() function that is executed after configured parameters and contained objects have been loaded.
```
fabcor-maxiv:
```suggestion
YAML-configured objects must be subclasses of the `BaseHardwareObjects.ConfiguredObject` class. The most complete example is the Beamline object and the comments in `beamline_config.yml` are the best guide to the syntax of YAML configuration files. It is a key principle of YAML-configured classes that **all** attributes added in the configuration must match a pre-defined attribute coded in the class. This means that you can look in the class code to see which attributes are available. The only exception is the `_initialise_class` attribute at the start of the file. This dictionary contains the import name of the class that is to be created, and optionally parameters to be passed to the `init()` method of that class. The `_objects` attribute gives the HardwareObjects that are contained in (i.e. children of) the object. The dictionary key is the role name, and the value is the name of the configuration file. Each `role_name` must match a read-only property coded in the body of the class, and must be added to the `__content_roles` list of the class by the class code. Note that classes are loaded and initialised in the order given by this list, so that there is a reproducible loading order. Contained objects can be defined as procedures, so that they are added to the list of procedures. Each YAML-configured class has an `_init()` method that is executed immediately after the object is created, and and `init()` function that is executed after configured parameters and contained objects have been loaded.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Accessing configuration data
+The Beamline object ('*HardwareRepository.beamline*') is a yaml-configured object, and is the starting point for finding other hardware objects. These may in turn contain other objects, so you can do e.g. *HardwareRepository.beamline.detector.distance* to get the detactor distance motor object. Configured properties are similarly accessed as simple atributes, e,g,  *beamline.default_acquisition_parameters*. Each object has three special properties and one function to deal with the objects contained within it. These are:
```
fabcor-maxiv:
```suggestion
The Beamline object (`HardwareRepository.beamline`) is a YAML-configured object, and is the starting point for finding other hardware objects. These may in turn contain other objects, so you can do e.g. `HardwareRepository.beamline.detector.distance` to get the detector distance motor object. Configured properties are similarly accessed as simple attributes, e,g, `beamline.default_acquisition_parameters`. Each object has three special properties and one function to deal with the objects contained within it. These are:
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+- *all_roles*: a list of the roles (attribute names) of contained HardwareObjects, in loading order.
+- *all_objects_by_role*: an ordered dictionary of role:HardwareObject
+- *procedures* an ordered dictionary of role:HardwareObject for procedures
+- *replace_object()* A function to replace an existing configured object at runtime with a new object.
```
fabcor-maxiv:
```suggestion
- `all_roles`: a list of the roles (attribute names) of contained HardwareObjects, in loading order;
- `all_objects_by_role`: an ordered dictionary of HardwareObjects;
- `procedures` an ordered dictionary of HardwareObjects for procedures;
- `replace_object()`: a function to replace an existing configured object at runtime with a new object.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+## XML-configured objects
+### Code and file structure
+Xml-configured objects have a name() function, that returns the name of the configuration file used to specify it (without the '.xml' suffix). It is this name that is used in internal data structures and a number of access functions.
```
fabcor-maxiv:
```suggestion
XML-configured objects have a `name()` method, that returns the name of the configuration file used to specify it (without the `.xml` suffix). It is this name that is used in internal data structures and a number of access functions.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+Xml-configured objects must be subclasses of the BaseHardwareObjects.HardwareObject class. A good example is mxcubecore/configuration/mockup/detector-mockup.xml (note that 'hwrid' is an alias for what is normally wirtten as 'href'). In xml configuration contained objects are given using the "object" element, with the 'href' attribute giving the configuration file name to pick up (you can use a similar syntax to redirect the topmost element to another file) and the 'role' attribute giving the role name. Properties (both simple, and nested structures) are given as contained xml elements. The configuration data are kept in complex internal data structures, with links back to the original xml
```
fabcor-maxiv:
```suggestion
XML-configured objects must be subclasses of the `BaseHardwareObjects.HardwareObject` class. A good example is `mxcubecore/configuration/mockup/detector-mockup.xml` (note that `hwrid` is an alias for what is normally written as `href`). In XML configuration contained objects are given using the "object" element, with the `href` attribute giving the configuration file name to pick up (you can use a similar syntax to redirect the topmost element to another file) and the `role` attribute giving the role name. Properties (both simple, and nested structures) are given as contained XML elements. The configuration data are kept in complex internal data structures, with links back to the original XML.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+The important functions can be found in the BaseHardwareObjects,HardwareObjectNode class. XML-configured files have no limits on the attributes or objects thay can contain. This leads to greater flexibility, since you can add a new attribute when needed without modifying the class code; it also means that there is no way to check which attributes are supported without looking into the configuration files, and gives more scope for local and potentially conflicting implementations. The functions have quite complex behaviour that amounts to overloading.
```
fabcor-maxiv:
```suggestion
The important functions can be found in the `BaseHardwareObjects,HardwareObjectNode` class. XML-configured files have no limits on the attributes or objects they can contain. This leads to greater flexibility, since you can add a new attribute when needed without modifying the class code; it also means that there is no way to check which attributes are supported without looking into the configuration files, and gives more scope for local and potentially conflicting implementations. The functions have quite complex behaviour that amounts to overloading.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Accessing configuration data
+
+The recommended way to access contained objects is through the *get_object_by_role* function, since it works on role names rather than the less predictable file names. As implemented the function will look recursively in contained objects for a given role name if the topmost object does not contain it. The *get_roles* function returns a list of roles that are defined on the object itself.
```
fabcor-maxiv:
```suggestion
The recommended way to access contained objects is through the `get_object_by_role` function, since it works on role names rather than the less predictable file names. As implemented the function will look recursively in contained objects for a given role name if the topmost object does not contain it. The `get_roles` method returns a list of roles that are defined on the object itself.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+You can get and set the values of properties by normal obj.attr syntax, which will also get you non-property attributes. The *get_properties* functions returns a dictionary of all properties and their values, and the *get_property* function behaves as get_properties().get. Direct setting of properties internally calls the set_property function, and this function automatically converts strings to int, float or bool if possible.
```
fabcor-maxiv:
```suggestion
You can get and set the values of properties by normal `obj.attr` syntax, which will also get you non-property attributes. The `get_properties` method returns a dictionary of all properties and their values, and the `get_property` method behaves as `get_properties().get`. Direct setting of properties internally calls the `set_property` function, and this function automatically converts strings to `int`, `float` or `bool` if possible.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,37 @@
...
+
+There are additional ways of accessing contained objects. *get_objects* and *has_object* take as input the object name, i.e. the name of the configuration file that specifies the object (or objects) (which is by definition globally unique). An Xml-configured object is also coded to mimic a Python list and dictionary of contained objects, so that anObject[ii] (ii: integer) returns the ii'th containd object, whereas anObject[key] (key: string) returns the contained object defined by the name (i.e. config file name)
```
fabcor-maxiv:
```suggestion
There are additional ways of accessing contained objects. `get_objects` and `has_object` take as input the object name, i.e. the name of the configuration file that specifies the object (or objects) (which is by definition globally unique). An XML-configured object is also coded to mimic a Python list and dictionary of contained objects, so that `anObject[ii]` (`ii` being an integer) returns the `ii`'th contained object, whereas `anObject[key]` (key being a string) returns the contained object defined by the name (i.e. config file name).
```

fabcor-maxiv (commented):
Thanks!

I made some suggestions (that you should be able to apply directly via the GitHub UI), hopefully they do not introduce any error.

(Not a policy as far as I can tell, maybe it is just a preference of mine) I think it would be nice to cut long sentences over multiple lines. Usually I tend to follow something like [*Semantic Line Breaks*](https://sembr.org/). From my point of view the gain is when reviewing *diffs*.

rhfogh:
Thanks, good improvement. I have added an additional commit introducing semantic line breaks.

rhfogh:
The first version was incorrect on some points having to do with the XML-configured files. This should hopefully be better.

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+Direct setting of properties internally calls the `set_property` function,
+and this function automatically converts strings to `int`, `float` or `bool` if possible.
+
+There are additional ways of accessing contained objects.
+`get_objects` and `has_object` take as input the object name
+As currently coded (was it always thus?) the name is equal to the role name used to add the object.
+An XML-configured object is also coded to mimic a Python list and dictionary of contained objects,
+so that `anObject[ii]`
+(`ii` being an integer) returns the `ii`'th contained object,
+whereas `anObject[key]` (key being a string) returns the contained object defined by the name (i.e. the role name).
```
fabcor-maxiv, line 124:
```suggestion
whereas `anObject[key]` (`key` being a string) returns the contained object defined by the name (i.e. the role name).
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+
+For XML-configured HardwareObjects (but not for YAML:-configured ones)
```
fabcor-maxiv, line 126:
```suggestion
For XML-configured HardwareObjects (but not for YAML-configured ones)
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+there are two additional ways of getting hold of HardwareObjects.
+
+Beamline.get_hardware_object lets you get a HO from a dotted list of rolenames (e.g. 'detector.distance')
```
fabcor-maxiv, line 129:
```suggestion
`Beamline.get_hardware_object` lets you get a hardware objects from a dotted list of role names (e.g. `detector.distance`)
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+This is essentially a convenience function to avoid repeated get_object_by_role calls.
```
fabcor-maxiv, line 130:
```suggestion
This is essentially a convenience function to avoid repeated `get_object_by_role` calls.
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+For YAML-configured objects the same could be done by direct attribute access.
+
+HardwareReposotory.get_hardware_object, on the other hand,
```
fabcor-maxiv, line 133:
```suggestion
`HardwareReposotory.get_hardware_object`, on the other hand,
```

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+(you can use a similar syntax to redirect the topmost element to another file)
+and the `role` attribute giving the role name.
+Simple properties are given as contained XML elements,
+and complex properties (dictionaries) are given as elements of type 'object' without a 'href' attribute.
+
+
+The configuration data are kept in complex internal data structures,
+with links back to the original XML.
+
+The important methods can be found in the `BaseHardwareObjects,HardwareObjectNode` class.
```
fabcor-maxiv, line 96:
```suggestion
The important methods can be found in the `BaseHardwareObjects.HardwareObjectNode` class.
```

fabcor-maxiv (commented):
Some more suggestions. Note that the changes can all be "batched" together in one single commit via the GitHub web interface (or at least it should be possible, I have seen/done it before).

fabcor-maxiv (approved):

docs/source/dev/configuration_files.md
```diff
@@ -0,0 +1,127 @@
...
+that uses YAML files instead of XML files,
+which has different ways of accessing the configuration data from inside the code.
+HardwareObjects can be configured to contain both other hardware objects and properties.
+The former are identified relative to the container by a role.
+
+## Finding the files
+Configuration files are searched for by name in a series of directories given as a lookup path.
+This is specified either with the `--coreConfigPath` command line parameter to MXCuBE,
+or through the `MXCUBE_CORE_CONFIG_PATH` environment variable.
+The files live under the `mxcubecore/mxcubecore/configuration` directory,
```
elmjag: Is --coreConfigPath parameter specific for MxCube-QT? I'm getting `unrecognized arguments` error when trying to use this with mxcubeweb:

```
$ ./mxcubeweb-server --coreConfigPath foo
usage: mxcubeweb-server [-h] [-r HWR_DIRECTORY] [-s STATIC_FOLDER] [-l LOG_FILE] [-L LOG_LEVEL] [-el ENABLED_LOGGER_LIST] [-w] [-t]
mxcubeweb-server: error: unrecognized arguments: --coreConfigPath foo
```

elmjag: _The files live under the `mxcubecore/mxcubecore/configuration` directory,_

Is that specific for MxCube-QT?

For the record, we keep our config files under /opt/mxcube/conf when deploing our MxCube-web.

rhfogh: Yes, needs clarification.

rhfogh: Good point - I'll fix this.

rhfogh: Thanks for noticing.

docs/source/dev/json-schema-generated-user-interface.md
```diff
@@ -10,7 +10,7 @@ We have a system for code-generated interfaces that can be plugged in to both a
...
 
-The form of the JSON schemas used has been agreed between Rasmus Foghy, Jean-Baptiste FLorial, an MArcus Oscarsson. Hopefully it fits with standard practices.
+The form of the JSON schemas used has been agreed between Rasmus Fogh, Jean-Baptiste Florial, an Marcus Oscarsson. Hopefully it fits with standard practices.
```
fabcor-maxiv:
```suggestion
The form of the JSON schemas used has been agreed between Rasmus Fogh, Jean-Baptiste Florial, and Marcus Oscarsson. Hopefully it fits with standard practices.
```

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -306,11 +306,14 @@ def get_available_workflows(self):
...
         Used for both characterisation, diffractcal, and acquisition
 
-        :param data_model (GphlWorkflow): GphlWorkflow QueueModelObjecy
-        :param choose_lattice (ChooseLattice): GphlMessage.ChooseLattice
-        :return: -> dict
+        Args:
+            choose_lattice (:obj: `ChooseLattice`, optional): ChooseLattice message
+
+        Returns:
+            dict: Parameter value dictionary
```
fabcor-maxiv, line 316: General comment

I wonder if maybe it is more efficient to fix the type hints in the function signature, rather than adding the types to the docstring (which I find is a bit awkward to do). If I am not mistaken (needs to be checked), the generated documentation should be able to pick the type hints and format it all nicely. I do not know what happens when the type in the annotation is not the same as the type in the docstring.

fabcor-maxiv (approved):

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
rhfogh: I think that will indeed be standard procedure, but I am postponing it till the time where I fix all my docstrings. For now I only do enough to make the test pass.

fabcor-maxiv:
Seems like we should merge this, because the latest updates seem to be code changes rather than documentation.

rhfogh:
You are right. I was waiting for @oldfielj-ansto, but it is time now. Merging.

PR #870, opened 2024-03-15
Fix generic workflow queue entry execute method

### LINKED ISSUE #869: GenericWorkflowQueueEntry execute method loops infinitely
I found an issue when using the GenericWorkflowQueueEntry class.

The `execute` method of the `GenericWorkflowQueueEntry` class  should stop  when the worfklow hardware object `state` is changed to `ON`, but in the current implementation the `execute` method enters an infinite loop and is never marked as completed

fhernandezvivanco:
Resolves #869


rhfogh (commented):
Looking at the code, the  original intent seems to be that the test is done on workflow_running, and that the change with changes of state are dealt with by the workflow_state_handler. If that is not working properly (as apparently it is not) it would be preferable to either fix the workflow-state_handler (instead of just bypassing it), or removing it entirely and working off the states directly. This change is rather a short-term fix, leaving the code complicated to the next person. Could we check with @marcus-oscarsson , who apparently wrote this to start with?

marcus-oscarsson:
This was done before all hardware objects had a standardized `state`. The proposed change should in theory work for handling the state RUNNING/BUSY however the `workflow_state_handler` also handles a state called OPEN which means that the workflow requests a dialog to be displayed. I don't think the later is used today (it was used in the Qt interface with EDNA based Workflows).

I think we can try this change but we need to test it properly, and in that case remove the `workflow_state_handler` method. However if your workflow emits `stateChanged` then the `self.running` attribute should be correct, does `workflow_state_handler` get called at all ?

marcus-oscarsson:
What Id suggest is, if you feel comfortable with it (it would make the clean up work for me easier) remove the `workflow_state_handler` and the connect in `pre_execute`. Like this we have removed this obsolete piece of code and Ill  make a note to test this to make sure that the `running` state is correct in the end.

rhfogh:
@marcus-oscarsson Your last proposal sounds fine. Anyway, the GPhL workflow does not use this class, so whatever works, ...

marcus-oscarsson:
@fhernandezvivanco: What did you decide to do, are you going for the cleanup ?

fhernandezvivanco:
@marcus-oscarsson , sorry for the delay. Happy to clean the class. For some reason I misunderstood and thought you were going for the code cleanup.

marcus-oscarsson:
@fhernandezvivanco sorry to bother you with this again, do you think you can rebase this on develop ? and then Ill merge

fhernandezvivanco:
> @fhernandezvivanco sorry to bother you with this again, do you think you can rebase this on develop ? and then Ill merge

Hi Marcus, sorry for the late reply. I've rebased the branch, thanks! :)

marcus-oscarsson:
@fhernandezvivanco, Oh its me who is sorry we should have merged this ages ago !

PR #873, opened 2024-03-16
[WIP] P11 develop 2

agruzinov:
Small changes to the P11 HO after testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mxcubecore"
-version = "1.77.0"
+version = "1.78.0"
 license = "LGPL-3.0-or-later"
 description = "Core libraries for the MXCuBE application"
 authors = ["The MXCuBE collaboration <mxcube@esrf.fr>"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mxcubecore"
-version = "1.78.0"
+version = "1.79.0"
 license = "LGPL-3.0-or-later"
 description = "Core libraries for the MXCuBE application"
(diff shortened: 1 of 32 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "mxcubecore"
-version = "1.76.0"
+version = "1.77.0"
```
fabcor-maxiv: As far as I understood the process, that should not be done from within a pull request.

```suggestion
version = "1.76.0"
```

fabcor-maxiv (changes requested):

pyproject.toml
marcus-oscarsson: Its indeed strange that the PR includes a previous commit, from the CI (#4f6c5d9)

agruzinov:
Is there a cleaner way to incorporate all the changes and having a fork without having all the commits that happened in between after rebase? 

For tnow it is like this:

There is a branch of core/qt under mu user.
Than:

-Synchronise all the changes that happened in between with the develop branch through the web interface. 
Than on the develop maschine:
git checkout develop
git pull (with rebase)

than:

git checkout p11_develop_2
git rebase develop
git push

Rhetorical question: what I'm doing wrong? :)

Or it is easier just to to work on the main repo than?

Any comments will be appreciated? Sorry for the basic question, but I think I've lost the understanding of the git behaviour.

Cheers!



git branch p11_develop_2

pyproject.toml
agruzinov: That happens if I get the latest updated from the upstream develop to agruzinov/develop through github web interface.

agruzinov:
Resolved through another PR#905. Closing.

PR #874, opened 2024-03-18
DOCS: starting with the Command Package and its modules

AnnieHeroux:
Some failure during import prevented some of the modules to be rendered in the docs. WIP since I have not looks through every module. To address the issue just added a try/except around the import.

Not sure if all the methods in the classes need to have docstring. It is misleading to only have some documentation for the get_value() and is_connected() (inherited?)

### DIFF
--- a/mxcubecore/Command/Epics.py
+++ b/mxcubecore/Command/Epics.py
@@ -37,6 +37,8 @@ __license__ = "LGPLv3+"
 
 
 class EpicsCommand(CommandObject):
+    """ Epics Command """
+    
     def __init__(self, name, pv_name, username=None, args=None, **kwargs):
         CommandObject.__init__(self, name, username, **kwargs)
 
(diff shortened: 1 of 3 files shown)


fabcor-maxiv:
@AnnieHeroux, indeed... `PyTango` is only on optional dependency:

https://github.com/mxcube/mxcubecore/blob/c316987b8e480258d14a818a6642c0624662ca23/pyproject.toml#L48

We use PyTango at MAX IV, so I guess I always install it and thus probably was never bothered by the issue you encountered.

Running a command like the following should help:

```
poetry install --all-extras
```

Can you provide more details about the issue you noticed? What kind of warning and/or error message you saw? Can you provide an example of a name of a document that is not built correctly or at all?

If that is the same error/warning message that I see now, then it looks like it also impacts other things that can not be imported: `redis`, `qt`, `SpecClient`, and so on...

There might be a better way to fix this than wrapping imports within a `try` block. Let's investigate further... : )

AnnieHeroux:
I am not quite sure how to proceed with the specific site dependencies. I will leave it for now.

fabcor-maxiv:
> I am not quite sure how to proceed with the specific site dependencies. I will leave it for now.

Site-specific dependencies can be handled separately. When it is about documentation my attitude is to never get ourselves blocked on details, better have some documentation, even if incomplete or imperfect ,than no documentation at all.

fabcor-maxiv (approved):

marcus-oscarsson:
Thanks @AnnieHeroux !

Issue #784, opened 2023-08-30
AbstractCharacterisation is broken


rhfogh:
Lookiing at both EDNACharacterisation, and P11EDNACharacterisation, it seems that neither follows the rules of their superclass, AbstractCharacterisation. There are several function signatures in the subclasses that do not follow those in the superclass. In theory the superclass definitions should always take precedence, but in this case it looks like maybe the abstract superclass should be modified to conform to the working implementations. 

Anyway, could someone who uses EDNACharacterisation take at look at harmonising this?

PR #876, opened 2024-03-18
EDNA Characterisation - Setting characterization result via XMLPRC

marcus-oscarsson:
Setting characterization result via XMLPRC instead of reading from file system


rhfogh (commented):
Looks fine to a quick look.

While you are at it, could you take a look at AbstractCharacterisation and match the method signatures in the abstract class to those in the subclasses? Not much work, but it does require someone who unde4rstrands how this class works.

marcus-oscarsson:
Ill try to that in a follow up PR

marcus-oscarsson:
I think you already created an issue for fixing the `AbstractCharacterisation`  issue #784

PR #877, opened 2024-03-18
AbsractMultiCollect - Removed number of passes and passing number of images

marcus-oscarsson:
Passing number of images and removed number of passes

 - The MD (and not only the detector) need number of images to be passed for some scans. 
 - Removed number of passes

### DIFF
--- a/deprecated/HardwareObjects/mockup/MultiCollectMockup.py
+++ b/deprecated/HardwareObjects/mockup/MultiCollectMockup.py
@@ -157,10 +157,15 @@ class MultiCollectMockup(AbstractMultiCollect, HardwareObject):
     ):
         return
 
-    def prepare_oscillation(self, start, osc_range, exptime, npass):
+    def prepare_oscillation(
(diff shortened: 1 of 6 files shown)


rhfogh (commented):
Nice clean-up. Looks fine by me, to a quick look, but my IDE flags up some problems:

MultiCollectMockup:
- Does not implement all abstract methods.
- start_acquisition does not match superclass function parameters.

AbstractMultiCollect:
- calls to last_image_saved do not match parameters in function definition

ESRFMultiCollect:
- Does not implement all abstract methods.
- Definition of last_image_saved do not match superclass function def

LNLSCollect:
- Does not implement all abstract methods.
- signature of start_acquisition does not match superclass

marcus-oscarsson:
Thanks for the input we will try do something with these, but I don't think it matches what was done in the PR so if you don't mind we will try to address those in a followup PR

rhfogh:
Sure. :-)

marcus-oscarsson:
Created issue #890

PR #878, opened 2024-03-18
ESRF HardwareObjects -  ESRFBeamlineActions and ESRFXRFSpectrum

marcus-oscarsson:
Added ESRFBeamlineActions and ESRFXRFSpectrum courtesy @beteva, the later will probably be updated fairly soon due to the refinements ongoing in #860. Apart from that there are some general fixes.


rhfogh (commented):
I am not the right person to approve, but looks fine to a quick look.

PR #883, opened 2024-03-18
base_queue_entry - Added missing import

marcus-oscarsson:
Added missing import


rhfogh (commented):
Good, thanks. But while you are at it, there is one more error in line 475 ('copy' is not defined):

                            inverse_beam_item = copy(interleave_item)

PR #884, opened 2024-03-18
FlexHCD - Don't change to transfer phase in unload

marcus-oscarsson:
- Don't change to transfer phase in unload (to be done earlier by prepare unload in control system)
 - Emitting gripper changed


rhfogh (approved):
Looks fine to an uninformed look.

PR #887, opened 2024-03-18
BeamlineActions - Only emit commandReplyArrived if action is running

marcus-oscarsson:
Only emit commandReplyArrived if action is running


rhfogh (approved):
Fine to a quick look

PR #889, opened 2024-03-18
[Documentation Code Camp] - Added introductory section for queue documentation - Part 1

marcus-oscarsson:


rhfogh (approved):

marcus-oscarsson:
This is still very much in progress, Ill try to add a bit more introduction and some figures that might make things easier to understand.  Then Ill go on with some more details that will point out where things are implemented and how new tasks are created.

marcus-oscarsson:
If you think about something that you'd like to know Id try to add it, I'm not promising that it will be any clearer but it least it will be documented ;)

rhfogh:
I do have one. As I understood the old system, there was a single source of truth, specifically the tree of QueueModelObjects. But (again as I understood it) MxcubeWeb introduced a new data structure that served as a competing source of truth that could be queried or modified independently. I'd like to understand what both those structures are and how they interact.

marcus-oscarsson:
Ill try to clarify that, the queue is the same in both Qt and Web, still the same old tree of `QueueModelObjects` but ill try to clarify that. The creation of the nodes differ slightly a bit but it should otherwise be the same.

fabcor-maxiv (commented):
Same suggestion I made for the other pull requests: when dealing with markup I tend to prefer [Semantic Line Breaks](https://sembr.org/). It is a personal preference, as far as I know we do not have a policy for this, so feel free to follow it, to whatever degree you feel comfortable with, or not at all.

docs/source/dev/queue.md
```diff
@@ -0,0 +1,12 @@
+## Introduction
+
+Today's MXCuBE queue system was first implemented in version 2.x of MXCuBE. The aim was to create a queuing system that provided means for automation and that integrated well with the sample changer robots, the ISPyB LIMS system, and workflow engines. The queue was designed to have manual and automatic modes of operation.
```
fabcor-maxiv: We are missing a title

```suggestion
# Queue system

## Introduction

Today's MXCuBE queue system was first implemented in version 2.x of MXCuBE. The aim was to create a queuing system that provided means for automation and that integrated well with the sample changer robots, the ISPyB LIMS system, and workflow engines. The queue was designed to have manual and automatic modes of operation.
```

fabcor-maxiv (commented):
Looks good overall

docs/source/dev/queue.md
```diff
@@ -0,0 +1,53 @@
...
+
+The model consists of an object called `QueueModel` that has a tree structure of `TaskNode` objects. The `TaskNode` object is often referred to as a *queue model object* and defines the data associated with a collection protocol. There is a one-to-one mapping between a specific queue model object and a specific queue entry, for example, `DataCollectionTaskNode` and `DataCollectionQueueEntry`. The mapping is defined in a structure called MODEL_QUEUE_ENTRY_MAPPINGS in mxcubecore/queue_entry/base_queue_entry.py.
+
+The combination of a `QueueEntry` and its model (`QueueModel`) is often referred to as an ask *task*, and makes up the entity that will be executed when the queue is executed.
```
axelboc, line 14:
```suggestion
The combination of a `QueueEntry` and its model (`QueueModel`) is often referred to as a *task*, and makes up the entity that will be executed when the queue is executed.
```

docs/source/dev/queue.md
```diff
@@ -0,0 +1,53 @@
...
+```
+# View is either a Qt view or a Mock object
+view = Mock()
+
+# The sample_model and sample_entry belong to the sample to which
+# we would like to add the task.
+# sample_model, sample_entry
+
+# In the Qt verison this item is displayed in the queue while its
+# not displayed at all in the Web version
```
axelboc, line 39:
```suggestion
# In the Qt version, this item is displayed in the queue.
# In the Web version, it's not displayed at all.
```

docs/source/dev/queue.md
```diff
@@ -0,0 +1,53 @@
+# Queue system
+Today's MXCuBE queue system was first implemented in version 2.x of MXCuBE. The aim was to create a queuing system that provided means for automation and that integrated well with the sample changer robots, the ISPyB LIMS system, and workflow engines. The queue was designed to have manual and automatic modes of operation.
+
+To enable automatic data collection over several samples, one needs to associate tasks to be performed on a set of samples. The decision was to represent the queue as a tree of nodes, where each node defines the logic for a certain task or collection protocol. Each node can have a set of child nodes that are executed after the main logic of its parent. A sample node, for example, defines how a sample is mounted, and its child nodes define what is done on that sample once it is mounted.
```
axelboc: I feel it's a bit confusing to describe the entire queue as a tree of nodes that contains sample nodes and task nodes. From what I understand, there's basically a flat queue of sample "entries" at the top, and each sample entry is associated with a tree of task "nodes", right?

docs/source/dev/queue.md
```diff
@@ -0,0 +1,53 @@
...
+
+The combination of a `QueueEntry` and its model (`QueueModel`) is often referred to as an ask *task*, and makes up the entity that will be executed when the queue is executed.
+
+The one-to-one mapping makes it possible to represent a queue and construct the queue entries from the queue model objects. In this way, the tree of queue model objects defines the queue. A task or collection protocol can be added to the queue by adding the *queue model object* to the `QueueModel`. For instance, a rotational data collection is added to the queue by adding a `DataCollectionTaskNode` to the `QueueModel`. This method of adding tasks to the queue is used by the Workflow engines, while the interfaces directly attach a `QueueModel` to the corresponding `QueueEntry` which is then
+*enqueued*
```
axelboc: Yeah ... I'm lost :sweat_smile: — a small diagram would probably be worth a thousand words. :joy:

docs/source/dev/queue.md
marcus-oscarsson: No, its a tree !

marcus-oscarsson:
I'm sorry @axelboc  but I'm not done yet its still very much work in progress this but ill make a diagram as well

docs/source/dev/queue.md
marcus-oscarsson: Its made to look flatter than it really is on the web UI side

docs/source/dev/queue.md
fabcor-maxiv: I also like diagrams! : D

In case it helps... We can enable [graphviz](https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html) if not enabled yet.

There are probably other extensions out there:
* <https://github.com/sphinx-contrib/>
* <https://github.com/yoloseem/awesome-sphinxdoc>
* <https://sphinx-extensions.readthedocs.io>

marcus-oscarsson: Hm, I think Id prefer to create the diagram myself in this case but perhaps for other things

docs/source/dev/queue.md
axelboc: :exploding_head: but then is it a tree of "sample entries" and each entry has its own tree of "task nodes", or are sample nodes and task nodes truly mingled with one another inside the same tree?

docs/source/dev/queue.md
axelboc: I know of [mermaid](https://mermaid.js.org/intro/) (looks like it's Node-based, though)

docs/source/dev/queue.md
marcus-oscarsson: Ill make a figure and maybe it makes things easier to understand, I never tried to put a Sample node on a Group node, but that could actually work (not that it makes alot of sense)

Basically:

- Root
   - Sample 1
     - Group 1
       - Task 1
       - Task 2
       - ....
     - ...
     - Group N
       - Task 1
       - Task 2
  - Sample 2
    -  ...
  - Sample N

marcus-oscarsson: In the Qt version the queue is both displayed and executed like this

marcus-oscarsson: In the web version the Group nodes are not displayed and the queue is displayed one sample at the time, but that is an illusion. When we run the queue in automatic mode, the above is what gets executed

marcus-oscarsson: When we run it sample by sample we explicitly disable all entries except for the entry we want to execute the queue. This is very explicit in the Qt version because each item can be enabled and disabled by a checkbox in the web version this was made implicit. The later was a wish from the scientists  because the prior was considered to be to complicated and confusing when there were lots of samples.

marcus-oscarsson: Maybe it helps, but Ill try to clarify and make a diagram

docs/source/dev/queue.md
fabcor-maxiv: Yes, that is mermaid and plantuml that I had in mind (rather than graphviz, even though it works fine for simple things). The good thing is that we directly have the diagram as code inside the document (inside the Markdown or ReStructuredText). I opened a ticket <https://github.com/mxcube/mxcubecore/issues/908> to gather the info, in case someone ever wants/needs to do a diagram this way.

marcus-oscarsson: Aha, okey you mean like that. I thought that you meant that the diagram would generated from the source. In that case its much more useful ... to have the code inside the document :)

docs/source/dev/queue.md
axelboc: Nice, thanks, it does clear things up a lot!

> I never tried to put a Sample node on a Group node, but that could actually work (not that it makes alot of sense)

I guess this is my point: the fact that it's a very flexible tree structure under the hood does not necessarily need to be documented. If the recommended/actual usage of the queue is that samples are always directly under the root of the tree, then the explanations and terminology can be dumbed down a lot (and arguably the code as well, but that's another debate).

marcus-oscarsson: It was designed flexible on purpose  to make it easy to extend/adapt. There were lots of ideas regarding automation at the time (and there still is) so we needed to be quite flexible and did not want to risk coming up with a system that did not fit what the scientists possibly wanted to do in the future.

The implementation as it is today can be simplified without loosing functionality, and that is partially also what we want to do with the refactoring of the queue that we speak about sometimes. The queue could be implemented in many different ways, it can be a completely flat structure but then there has to be something else that defines the relation between the "entries". 

I understand what you mean, somehow I prefer to document as it really is as its for developers and then perhaps add a additional explanations on what simplifications that are or can be done.

docs/source/dev/queue.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Differences between the Web and the Qt user interfaces
+When the Web interface was designed, the scientists wanted to change some aspects of how the queue operates and how it is presented to the user. The decision was to work on one sample at a time, rather than for the entire queue, as is done in the Qt version. This also means  that the Web interface is centered around displaying the contents of the queue for one sample. It was also decided to not display `DataCollectionGroupQueueEntry` in the queue. The `DataCollectionGroupQueueEntry` still exists in the Web interface, but there is no view created for it.
+
+On a technical level, there is a tight coupling between the `QueueEntry` base class to a Qt view object that makes the creation of the object slightly different between the two user interfaces (Qt and Web). To handle the tight coupling to Qt a `Mock` object used in the Web version instead of a Qt view and the handling of the updating of the view in the web version handled through signals passed over websockets to the client.
```
AnnieHeroux: To handle the tight coupling to Qt a `Mock` object used in the Web version instead of a Qt view and the handling of the updating of the view in the web version handled through signals passed over websockets to the client.

This sentence might be better if split in two????? So the Web has a Mock object to mimic the Qt?

marcus-oscarsson: Yes, you are right. That sentence is a a bit convoluted

marcus-oscarsson: > "This sentence might be better if split in two????? So the Web has a Mock object to mimic the Qt?"

Yes

rhfogh:
AH, one question here:

In the diagram it is shown with Samples always at the top ()makes sense) and with only one level of DataCollectionGroup. Is that a rule, just common practice, or a coincidence? When running multisweep workflow experiments, it arguably makes more sense to have multiple levels or groupings, a group for the entire workflow, containing another group for teh characterisation, and yet another group for the actual acquisition sweeps. And indeed this is how I have coded the GPhL workflow. I think we need to decide whether we accept multiple levels of groupings or not, and make it explicit in the documentation, otherwise people will start making incorrect assumptions and different bits of code can clash.

rhfogh (changes requested):
We need to sort out the question of whether nested groups are allowed or not.

Great with the diagrams, BTW, that makes everything a lot clearer.

docs/source/dev/queue.md
axelboc: Thanks for the diagram!

docs/source/dev/queue.md
```diff
@@ -0,0 +1,67 @@
...
+:alt: Queue overview node concept
+
+The queue is a tree of nodes, where each node defines the logic for a certain task or collection protocol.
+```
+
+The execution order is depth first, so that all the children of a node are executed before the node's siblings. Each node has a `execute` method that defines its main logic and `pre_execute`and `post_execute` methods that are executed before and after the main `execute` method.
+
+This design was thought to map well to the semantics of how data is collected with a sample changer and the upload of metadata to LIMS. The sample changer has a number of samples, each having a set of data collections (child nodes). The results are uploaded to the lims system after each data collection, (`post_execute`). Workflow engines can further group native collection protocols into grouping nodes, which in turn can be used as building blocks for even more complex collection strategies.
+
+```{figure} /assets/queue_execute_methods.svg
```
fabcor-maxiv, line 29: This file is missing

fabcor-maxiv (changes requested):

docs/source/dev/queue.md
marcus-oscarsson: Thanks, I'm aware, I'm still working on it. Also trying to get the scaling to work

marcus-oscarsson:
@rhfogh, I'm still working on this section and ill try to make some notes on whats common practice between what is possible. There is no decision on how deep one can nest data collection groups or samples for that matter. Its left open by purpose. The conventional one crystal type per pin with sample changer collection is whats shown in the figure, Ill make a note on that.

One can imagine that for instance a plate sample changer where one have two levels of samples plate and crystal, I'm not saying that thats how it should be done but its a possibility. For workflows nested groups could make sense and even for multi sweep collections.

rhfogh:
OK, If you will put in something explicit about allowing nested groups, I'll remove my 'Request Changes' immediately and leave the rest to you. I made the point because (until I modified it) there was code assuming that the Sample was  always two levels up from the DataCollection.

I still need to check with Olof exactly what the rules and assumptions about grouping are for MASSIF-1 operation.

rhfogh (approved):
Approved, given that there will be some wording on the fact that e.g. nested DataCollectionGroups are allowed.

marcus-oscarsson:
I'm not sure how to approach this because I make some changes and push mostly to see how it renders, but then I get comments on certain things that are still in progress. As its draft Ill try to answer your general questions regarding the queue like the one @rhfogh asked, and the rest is still very much subject to change.

rhfogh:
You could mark it as WIP, if you prefer that we wait with the comments.

axelboc:
We just got mixed up on what "Draft" means. Marcus opened this to show progress and to preview his work, not to request a review. As discussed during the meeting this morning, "Draft" means that it's not yet ready for review (unless the opener specifically asks for feedback). The removal of "Draft" status by the opener means that reviews are welcome.

marcus-oscarsson:
I happy for the broader questions like the one @rhfogh asked or general questions about the queue that you'd like to know about, it helps me to narrow down on some details. Just to avoid that we waste our time on things that will most likely change. I also mark it as WIP ;)

marcus-oscarsson:
I believe this is starting to get usable, its not complete yet I will add a few sections in a near future. I plan on adding more on how the execution works and then Ill add something about the new queue design and the drawbacks of the current design. However, thats for Part 2 :)

beteva:
Nice with diagrams.

docs/source/dev/queue.md
```diff
@@ -0,0 +1,79 @@
...
+
+```{figure} /assets/queue_overview_node_concept.svg
+:width: 400
+:alt: Queue overview node concept
+
+*Typical structure of a queue for conventional colletions with sample changer containing pins with samples.*
+```
+
+**_NOTE:_** There are no constraints on the structure of the tree. However, today: the top level always contains sample nodes,
+and the nesting occurs at the group level. The depth of nesting is currently maximum 3 - Sample -
```
rhfogh: That is not actually the case for GPhL workflow - we nest more deeply. Maybe we should change that. This probably needs discussion, to make sure that everybody either is or is not ready for deeper nesting.

docs/source/dev/queue.md
```diff
@@ -0,0 +1,79 @@
...
+dc_entry = qe.DataCollectionQueueEntry(view, dc_model)
+
+HWR.beamline.queue_model.add_child(sample_model, group_model)
+HWR.beamline.queue_model.add_child(group_model, dc_model)
+
+sample_entry.enqueue(group_entry)
+group_entry.enqueue(dc_entry)
+```
+
+Each type of data collection protocol has a piece of code similar to the code above associated with it. In the Web version, this is done in mxcubeweb.components.queue.py in the methods called add_**task_type**, for instance `add_data_collection`. In the Qt version, the creation is done in two different places: first in the widget where the queue model object is created, and then through `QueueModel.view_created`.
```
rhfogh: I think we should say explicitly that in the Qt version enqueueing the queue entry is done automatically when the qmo is added to the queue, via the QueueModel.add_child function, which sends out a child_added signal. It seems - to a *very* quick look - that the same is the case in the web version, via the function components.queue.Queue.queue_model_child_added. Maybe this should have a more prominent place in thedocumentation. Actually, is it the rule or an exception when enqueue is called explicitly rather than via signal?

rhfogh (changes requested):
Very nice, but I think a couple of things for a quick discussion

docs/source/dev/queue.md
marcus-oscarsson: That's not an issue, If you know how many levels you nest Ill update with that figure or phrase it like "the most common nesting depth is 3" or similar

docs/source/dev/queue.md
marcus-oscarsson: Its not entirely correct for the web version because its only the case for collections added through workflows that use the signal, the rest is done through `enqueue`. Ill see what I can do

marcus-oscarsson: Thanks, I tried to make clarify a bit, let me know what you think ?

docs/source/dev/queue.md
marcus-oscarsson: So, I added a "disclaimer" for workflows, but let me know if you rather that we add a figure.

PR #895, opened 2024-03-19
ISPyB mock: drop ldap server code

elmjag:
Remove the code that 'deals' with authServerType and ldapServer ISPyBClientMockup properties.

The LDAP mode is not really used in the code. The object does not change behaviour when authServerType is changed. Let's drop LDAP bits, as it creates unnecessary complication when using MxCube with mock hwobjs.


elmjag:
also, see related PR in mxcubeweb: https://github.com/mxcube/mxcubeweb/pull/1185

fabcor-maxiv (approved):
Fine, I guess

marcus-oscarsson:
Looks good to me :)

PR #896, opened 2024-03-20
Added CI_TOKEN to create_tag action

marcus-oscarsson:
I think that whats preventing the tagging to work is the missing CI_TOKEN


fabcor-maxiv:
@marcus-oscarsson, I am not sure this change does actually change anything.

As far as I understood:
* on one hand, the action `checkout` seems to use `${{ github.token }}` as default value for `token`
  * See its README documentation: <https://github.com/actions/checkout>
* on the other hand the `CI_TOKEN` environment variable and `${{ github.token }}` should have the same value
  * See this documentation: <https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret>

marcus-oscarsson:
Its the only difference I could find between the mxcubeweb, where the tagging works, and mxcubecore.  I remember that we had some issue with this in the past and we tried adding this CI_TOKEN and it started to work.

We could perhaps try ?

Not sure about the last point, CI_TOKEN is a token we created our self's, not sure where it says that it should be the same as github.token which is something else.

fabcor-maxiv:
Oh! I got my wires crossed... I read `CI_TOKEN` and assumed it was `GITHUB_TOKEN`. My previous message can be ignored.

PR #903, opened 2024-03-21
Move `.gitignore` out of docs `autosummary` dir

fabcor-maxiv:
Sometimes it is necessary to delete all Sphinx `autosummary` files in order to generate them from scratch.
For example, if a Python file is (re)moved from source code tree its corresponding autosummary `.rst` file remains.

The easy way is to delete the `autosummary` directory. But then the `.gitignore` file is deleted as well, which is inconvenient.


marcus-oscarsson:
:+1:

PR #905, opened 2024-03-21
P11 collection debug

agruzinov:
Repeat of #873. Hopefully this resolves the issues with re-basing. Now this branch is not from the fork under my user.


beteva (approved):
I believe you've tested all this and you have all the fields of the osc_pars assigned and compulsory. Otherwise, beware of var = dict["key"]. it might rise KeyError.

agruzinov:
Thanks for the comment! Cheers!

PR #909, opened 2024-03-21
Add document about the documentation system

fabcor-maxiv:
GitHub: related to https://github.com/mxcube/mxcubeweb/pull/1177

### DIFF
--- a/docs/source/dev/docs.md
+++ b/docs/source/dev/docs.md
@@ -1,0 +1,81 @@
+# Working with documentation
+
+This documentation is automatically built and published from the contents of the [mxcubecore](https://github.com/mxcube/mxcubecore) repository.
+Each time the repository's `develop` branch is updated, documentation is regenerated and published to [https://mxcubecore.readthedocs.io/](https://mxcubecore.readthedocs.io/)
+
+If you want to modify this documentation, make a pull request to the repository with your suggested changes.
+
+## Modifying the documentation
+
+This documentation is built using the [Sphinx](https://www.sphinx-doc.org/) documentation generator.
+The documentation's source and configuration files are located in the `docs` folder of the [mxcubecore](https://github.com/mxcube/mxcubecore) repository.
+Sphinx will also read [Python docstrings](https://peps.python.org/pep-0257/) from the repository's source code.
+
+Refer to the {doc}`contributing guidelines </dev/contributing>`.
+
+
+### Building documentation
+
+Follow the instructions for [Installing a development environment](https://mxcubeweb.readthedocs.io/en/latest/dev/environment.html).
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
The first part is taken from <https://github.com/mxcube/mxcubeweb/pull/1177>.

We should try to avoid duplicates like this. Maybe if it is not specific to Web or Qt then it should be placed in core. Web and Qt could simply link back to core.

elmjag (approved):

axelboc:
> We should try to avoid duplicates like this. Maybe if it is not specific to Web or Qt then it should be placed in core. Web and Qt could simply link back to core.

One doc site? One monorepo? :grin:

fabcor-maxiv:
> > We should try to avoid duplicates like this. Maybe if it is not specific to Web or Qt then it should be placed in core. Web and Qt could simply link back to core.
> 
> One doc site? One monorepo? 😁

Not sure if it is a serious suggestion (at least partly). ***1.*** Unified documentation: maybe that would be nice, I do not know, I have not considered it, seems like it would be some work. ***2.*** Monorepo: having never worked with one (that I can remember) it is hard to judge, but I think I would be against it.

axelboc:
> Not sure if it is a serious suggestion (at least partly). _**1.**_ Unified documentation: maybe that would be nice, I do not know, I have not considered it, seems like it would be some work. _**2.**_ Monorepo: having never worked with one (that I can remember) it is hard to judge, but I think I would be against it.

Mostly joking since I have no clue what a monorepo looks like for a Python project :smile_cat:. In all seriousness, though, I do think it would hugely improve developer experience (one PR in the monorepo, instead of two PRs in two separate repos that have to be reviewed and merged together), and simplify dev ops (dev installation, testing, deployment, etc.). Monorepos are now very common for front-end projects, but that's because the tooling has improved sufficiently; no idea what the state is in the Python world. Either way, a monorepo would be a prerequisite to having a single doc site. Sorry for the huge tangent :sweat_smile:

marcus-oscarsson:
I agree that we should definitely avoid duplication.  

We moved a way from  a monorepo solution using Git submodules a few years back because people struggled with the submodules. Moving back to some sort of monorepo would make some things easier but others harder so I think we should try to keep things as they are unless there some very good arguments for changing.

I think the way it is today works well enough, we could simply add links from the mxcubeweb documentation to the mxcubecore one. I think other solutions like having a separate repository for documentation risk to create unnecessary overhead and complicates things. We could perhaps use https://github.com/mxcube/mxcube for some common documentation and then link to that

axelboc:
Yeah submodules aren't great. Front-end monorepos typically rely on a smart package manager (like pnpm) or some other dedicated tooling.

PR #915, opened 2024-04-09
New XRF Implementation

beteva:
- Change the AbstractXRFSpectrum class:
  - snake case method names.
  - remove unnecessary methods.
  -  use the standard HardwareObjectState states and emit stateChanged only.
- Change the xrf queue model to call the new methods and to use the strateChanged signal only
- Add tests


beteva:
Sorry for the mess. Had to re-create the PR.

marcus-oscarsson:
Discussed with @meguiraun over email to merge

PR #916, opened 2024-04-09
bring back 'Ring Current' widget in mocked beamline

elmjag:
Emit 'valueChanged' signal when mocked Machine Info state changes. Seems that 'machInfoChanged' signal have been replaced depricated in favour of 'valueChanged'.
    
This brings back 'Ring Current' widget when running MXCuBE agains the mocked beamline.

Also, fixes some typos in the docs string.

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/MachineInfoMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/MachineInfoMockup.py
@@ -10,10 +10,8 @@ information by :
 
     - produces a current value that varies with time
     - simulates a control room message that changes with some condition
-      ()
-    - simulates
 
-[Emited signals]
+[Emitted signals]
 machInfoChanged
    pars:  values (dict)
 
--- a/mxcubecore/HardwareObjects/mockup/MachineInfoMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/MachineInfoMockup.py
@@ -88,7 +88,7 @@ class MachineInfoMockup(AbstractMachineInfo):
             values["attention"] = self.attention
             self._mach_info_dict = values
 
-            self.emit("machInfoChanged", values)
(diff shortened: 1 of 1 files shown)


beteva:
Thanks for cleaning up
If we apply the rule we have (get/set_value and valueChanged for the attribute which is in the name of the file, indeed MachineInfo should emit valueChanged. I guess we should also implement get_value that gives back a dictionary with the relevant keys

beteva:
Thanks for cleaning up
If we apply the rule we have (get/set_value and valueChanged for the attribute which is in the name of the file, indeed MachineInfo should emit valueChanged. I guess we should also implement get_value that gives back a dictionary with the relevant keys

elmjag:
> I guess we should also implement get_value that gives back a dictionary with the relevant keys

Looking at AbstractMachineInfo class, get_value() does not seems to be the part of the API for MachineInfo hardware objects. Is there any code that wants to call get_value() method on MachineInfo hardware objects?

beteva:
> Is there any code that wants to call get_value() method on MachineInfo hardware objects?

Actually not yet, but I believe we have to got through the whole chain HO-adapter to make it more consistent. I was looking at the AbstractMachibeinfo and think that we need to implement get_value() method and emit the right signal (valueCjanged) if we want to be sure that it does not depend on the local HO only.. I  suggest to merge this PR as it is and, if you do not mind, I'll take over to make more "standard" MachineInfo chain (I've actually started the implementation for testing mostly)

elmjag:
> I suggest to merge this PR as it is and, if you do not mind, I'll take over to make more "standard" MachineInfo chain (I've actually started the implementation for testing mostly)

Yes, feel free to merge this and it would be much appreciated can refactor MachineInfo suff!

PR #917, opened 2024-04-10
Refactor AbstractMachineInfo.

beteva:
- Method get_value to read all the available parameters.
- Method update_value to specify "standard" signal - valueChanged.
- Possibility to define which accelerator parameters to be watched for update in the configuration file.

### DIFF
--- a/mxcubecore/HardwareObjects/TangoMachineInfo.py
+++ b/mxcubecore/HardwareObjects/TangoMachineInfo.py
@@ -47,7 +47,7 @@ class TangoMachineInfo(AbstractMachineInfo):
         # we only consider the attributes defined in the parameters property
         for name in self._mach_info_keys:
             if hasattr(self.get_channel_object(name), "get_value"):
-              setattr(self, f"{name}", self.get_channel_object(name))
+                setattr(self, f"{name}", self.get_channel_object(name))
             else:
                 # remove the attributes we cannot read
                 self._mach_info_keys.pop(name)
@@ -64,7 +64,7 @@ class TangoMachineInfo(AbstractMachineInfo):
 
         for attr_key in attr_list:
             try:
-               hasattr(self.get_channel_object(attr_key), "get_value")
+                hasattr(self.get_channel_object(attr_key), "get_value")
             except AttributeError:
                 attr_list.remove(attr_key)
         self._mach_info_keys = attr_list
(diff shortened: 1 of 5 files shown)


rhfogh:
Not sure I like 'get_value' for this class. get_value is defined for an AbstractActuator, which actually has a single value that follows from the (name of the) class. Also, those 'values' are generally settable. For what sounds more like a general information dictionary I'd say that a different name would be less confusing.

beteva:
> Not sure I like 'get_value' for this class. get_value is defined for an AbstractActuator, which actually has a single value that follows from the (name of the) class. Also, those 'values' are generally settable. For what sounds more like a general information dictionary I'd say that a different name would be less confusing.

Well, if we consider MachineInfo as an entity, than it has only one value, which happens to be a dictionary. I do not mind renaming the method (something like get_attributes), but than we do not follow the rule get_value corresponds to the class name.
I am fine either way.

mxcubecore/HardwareObjects/TangoMachineInfo.py
```diff
@@ -0,0 +1,96 @@
...
+        # we only consider the attributes defined in the parameters property
+        for name in self._mach_info_keys:
+            if hasattr(self.get_channel_object(name), "get_value"):
+                setattr(self, f"{name}", self.get_channel_object(name))
+            else:
+                # remove the attributes we cannot read
+                self._mach_info_keys.pop(name)
+
+        # we only want to update when the current changes.
+        self.current.connect_signal("update", self._update_value)
```
marcus-oscarsson, line 56: Should this be `valueChanged` ?

test/pytest/test_mach_info.py
```diff
@@ -55,5 +55,20 @@ def test_mach_info_atributes(self, test_object):
             test_object.get_topup_remaining(), (int, float)
         ), "topup_remaining value has to be int or float"
         assert isinstance(
-            test_object.get_mach_info_dict(), dict
+            test_object.get_value(), dict
         ), "Machine info dictionary has to be dict"
+
+    def test_get_value(self, test_object):
```
marcus-oscarsson, line 61: :+1:

marcus-oscarsson: Nice with tests

marcus-oscarsson:
Thanks, looks good to me :),

PR #920, opened 2024-04-18
Fix usage of deprecated `ruamel.yaml` API

### LINKED ISSUE #921: `ruamel.yaml` vs. `PyYAML`?
@rhfogh wrote:

> One question: There is also an import yaml in HardwareObjects.StateMachine. Should that also be modified?

-- https://github.com/mxcube/mxcubecore/pull/920#pullrequestreview-2009186086

In my environment, `import yaml` imports the ***`yaml` import package*** installed by the ***[`PyYaml`](https://pypi.org/project/PyYAML/) distribution package***. But `PyYAML` is not declared as direct dependency in `pyproject.toml`, which is a bad thing. Obviously, all direct dependencies (the things we import directly) should all be declared in `pyproject.toml`.

In my environment, it happens that `PyYAML` is installed as a dependency of `pre-commit` (and also Sphinx-related things, but that is relatively new, while `pre-commit` has been here longer). So things work fine, but...

1. If we depend on `PyYAML`, we should declare it so in `pyproject.toml`.
2. Maybe we do not need to depend on both Yaml libraries: `ruamel.yaml` and `PyYAML`.

fabcor-maxiv:
Also formatted with `black`.

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -31,9 +31,11 @@ from mxcubecore.model import queue_model_enumerables
 
 try:
     from mxcubecore.model import crystal_symmetry
-    import ruamel.yaml as yaml
+    from ruamel.yaml import YAML
 except Exception:
-    logging.getLogger("HWR").warning("Cannot import dependenices needed for GPHL workflows - GPhL workflows might not work")
+    logging.getLogger("HWR").warning(
+        "Cannot import dependenices needed for GPHL workflows - GPhL workflows might not work"
+    )
 
 # This module is used as a self contained entity by the BES
 # workflows, so we need to make sure that this module can be
@@ -1609,7 +1611,6 @@ class PathTemplate(object):
             logging.getLogger("HWR").debug(
                 "PathTemplate (DESY) - (to be defined) directory is %s" % self.directory
             )
-            #archive_directory = self.directory
             archive_directory = HWR.beamline.session.get_archive_directory()
         elif PathTemplate.synchrotron_name == "ALBA":
             logging.getLogger("HWR").debug(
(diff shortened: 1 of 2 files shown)


fabcor-maxiv:
I am navigating a bit blindly here. I have no idea if this part of the code still runs, not sure how I would trigger it. Since it did not fail before, for me there is no difference. @rhfogh, the impacted part of the code seems to be related to Gphl workflows so you might have better visibility here.

There is another usage of `ruamel.yaml` in `mxcubecore/HardwareRepository.py` but it seems to already conform to the new API.

mxcubecore/model/queue_model_objects.py
```diff
@@ -1609,7 +1611,7 @@ def get_archive_directory(self):
             logging.getLogger("HWR").debug(
                 "PathTemplate (DESY) - (to be defined) directory is %s" % self.directory
             )
-            #archive_directory = self.directory
+            # archive_directory = self.directory
```
elmjag: How about just removing this line :)

rhfogh: Please do.

mxcubecore/model/queue_model_objects.py
```diff
@@ -31,9 +31,11 @@
 
 try:
     from mxcubecore.model import crystal_symmetry
-    import ruamel.yaml as yaml
+    from ruamel.yaml import YAML
```
rhfogh, line 34: This line should actually be deleted. For another time, how is it you cam maker changes and apply them to the PR?

rhfogh (approved):
Excellent - and thanks for black-checking this for me.

One question: There is also an 
import yaml
in HardwareObjects.StateMachine. Should that also be modified?

fabcor-maxiv:
> One question: There is also an import yaml in HardwareObjects.StateMachine. Should that also be modified?

Oh! Wow! ... That is another weird one... New ticket, to discuss this: <https://github.com/mxcube/mxcubecore/issues/921>

mxcubecore/model/queue_model_objects.py
fabcor-maxiv: Done

PR #926, opened 2024-04-29
[FIX]  Remove has_protein_acronym  method

jbflo:
- has_protein_acronym was introduce in PR: https://github.com/mxcube/mxcubecore/pull/826
 but it turn Crystals *always* have an attribute called 'protein_acronym', can be None or ""
Thanks to @rhfogh 

- And sample can have a protein_acronym even LIMS data had not been passed in
-  so that would have been the final fix

- and also apply black

Apologize for this  :worried:


rhfogh (approved):
Yes, this works fine. Thanks.

PR #927, opened 2024-05-02
PyTango no longer optional dependency 

marcus-oscarsson:
There are test cases that depend on PyTango and a large majority is using Tango on someway, so the optional flag is a bit unnecessary in my opinoin.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1788,7 +1788,7 @@ files = [
 name = "pytango"
 version = "9.4.1"
 description = "A python binding for the Tango control system"
-optional = true
+optional = false
 python-versions = ">=3.6"
 files = [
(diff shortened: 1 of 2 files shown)


fabcor-maxiv (approved):

fabcor-maxiv:
> There are test cases that depend on PyTango

Maybe we could fix the pipelines, documentations, and whatnot, to make sure that all optional dependencies are installed before running tests. I guess what I mean is that it probably is not strictly necessary to make it a mandatory dependency. But of course, I am also fine if we make it a mandatory dependency (it is probably less work).

marcus-oscarsson:
@fabcor-maxiv Yes, I agree that it would make sense if we had alot of dependencies that are more or less exclusive and only used by a few sites but I think it makes less sense for PyTango at least as I see it.

PR #929, opened 2024-05-07
Tango poller: retry reading on tango errors

elmjag:
When polling tango attributes, catch tango communication errors. On error, log a warning and try again reading the attribute again. In our case, we need this for a tango device that intermittently replies slowly, thus causing timeout errors.

As a small bonus, I tacked on changes to make mxcubecore/Command/Tango.py black compliant.


marcus-oscarsson:
Thanks

PR #940, opened 2024-06-14
Attempt to solve issue 902

### LINKED ISSUE #902: Remove device and equipment classes
A while ago we decided to deprecate and stop using [Device](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/BaseHardwareObjects.py#L1126) and [Equipment](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/BaseHardwareObjects.py#L1216).

Lets's remove them!

marcus-oscarsson:
Attempt to resolve #902. If accepted everybody have to update their local configuration files and use `<object>` instead of `<equipment> `or `<device>`

#issue 902


fabcor-maxiv:
> If accepted everybody have to update their local configuration file and use instead of or

Missing some words here? :D

marcus-oscarsson:
I'm not sure this is the best way to do this; I would suggest that each site goes through their specific files and the common files to check if the proposed changes are alright. Or we can open a PR per site and for common files if its more digestible ?

marcus-oscarsson:
> > If accepted everybody have to update their local configuration file and use instead of or
> 
> Missing some words here? :D

Its the `<` tags that made the "words" go missing :)

beteva:
There are two aspects of this issue - one is to change everywhere in the python code Device and Equipment inheritance and remove the two classes.The other is to change all the device and equipment to object keyword in the xml files.
Agree with @marcus-oscarsson that the local configuration has to be changed by each site. Just to confirm that this can be done even without changing the inheritance in the python files. We've already doing it at ESRF. once this done, we could do in one go the python inheritance,

marcus-oscarsson:
So what do you think, are we ready to make the jump to `Object` in this way or is to drastic ?

marcus-oscarsson:
No regrets :)

PR #942, opened 2024-06-17
utility module for unit conversions (mxcubecore.utils.units)

elmjag:
Adds mxcubecore.utils.units module, so there is a common place for unit conversion utility functions. Adds some initial functions to convert between time, energy, length and current units.

The idea is that we'll add more functions in this module, as the need arises.

### DIFF
--- a/mxcubecore/utils/units.py
+++ b/mxcubecore/utils/units.py
@@ -1,0 +1,71 @@
+"""
+Utility functions for converting between different units.
+"""
+
+#
+# time units
+#
+
+
+def us_to_sec(us: float) -> float:
+    """
+    convert microseconds (μs) to seconds
+    """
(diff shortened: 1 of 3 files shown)


mxcubecore/utils/units.py
```diff
@@ -0,0 +1,71 @@
+"""
+Utility functions for converting between different units.
+"""
+
+#
+# time units
+#
+
+
+def us_to_sec(us) -> float:
```
fabcor-maxiv:
```suggestion
def us_to_sec(us: float) -> float:
```

For all the others as well. Note that the type hint `float` will also accept `int`.

elmjag: Good catch, added parameter type annotations.

PR #945, opened 2024-06-24
Gphl maxiv changes

meguiraun:
Fixes and changes we did while @rhfogh  was with us. @rhfogh , happy to drop this if you have a better approach, I just did not want to lose these changes.

Maybe the interleave stuff should be left out for now?


rhfogh:
@meguiraun This is good, but I have incorporated these changes in my latest PR, as there were a couple of close-by changes for cleaning up, so I am closing this one. Could you look at the upcoming PR to see if it is OK?

PR #936, opened 2024-05-30
[WIP] First attempt to introduce PyDantic configuration model

marcus-oscarsson:

PR #949, opened 2024-07-01
yaml configs: a couple of fixes

elmjag:
I had to make these two changes to be able to run mxcube-web, wiith mixed XML and YAML config files for HWOs.

It possible it overlaps slightly with https://github.com/mxcube/mxcubecore/pull/936. However, as we decided to post-prone the Pydantic changes for later, I'm putting this here.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -24,7 +24,6 @@ import typing
 import ast
 import enum
 import copy
-from collections import OrderedDict
 import logging
 from gevent import event, Timeout
 import pydantic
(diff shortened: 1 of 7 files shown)


marcus-oscarsson:
Thanks !

marcus-oscarsson:
The pre-commit is failing for some reason, it does not seem to be files that you changed. 

I let you decide exactly how to do, either you merge this as it is and we follow up with a lint fix later or you include in this PR

elmjag:
The the pre-commit was failing on the original xml_yaml_conversion branch. I was not sure to do about that. I have added commits to fix reported errors.

Now, I see that the tests are failing, I'll take a look at what is going on.

marcus-oscarsson:
Its ok for the tests I think, its because the syntax in the yml files changed from _class to class and so on

marcus-oscarsson:
Ill merge this and we will take care of the tests later

PR #953, opened 2024-07-02
Added some more queue documentation

marcus-oscarsson:
As promised Ive started to extend the queue documentation. I'm not very imaginative and a bit out of inspiration so I simply called the new queue entry "dynamic" queue entry with lack for a better term. Let me know if you have some nicer ideas :)

I'm working on the section on how to create/add these entries.


docs/source/dev/queue.md
```diff
@@ -77,3 +78,44 @@ When the Web interface was designed, the scientists wanted to change some aspect
 On a technical level, there is a tight coupling between the `QueueEntry` base class and a Qt view object that makes the creation of the `QueueEntry` object slightly different between the two user interfaces (Qt and Web). To handle the tight coupling to Qt, a `Mock` object is used in the web version instead of a Qt View. The Qt interface can access the view directly via a reference, something that is impossible in the Web interface. Because the view and the `QueueEntry` do not execute in the same process, updates to the view in the web version are instead done through signals passed over websockets to the client.
 
 ## Execution
+As mentioned above, the execution order is depth first, so that all the children of a node are executed before the node's siblings. Each node has a `execute` method that defines its main logic and `pre_execute` and `post_execute`.
+
+A queue entry has a state internally called `status` that indicates the state of execution; the state can be one of ["SUCCESS", "WARNING", "FAILED", "SKIPPED", "RUNNING", "NOT_EXECUTED"].
```
fabcor-maxiv:
```suggestion
A queue entry has a state internally called `status` that indicates the state of execution; the state can be one of:
```

docs/source/dev/queue.md
```diff
@@ -77,3 +78,44 @@ When the Web interface was designed, the scientists wanted to change some aspect
...
+ - `FAILED`: The execution failed, indicated with red in the UI
+ - `SKIPPED`: Item was skipped, indicated with red in the UI
+ - `RUNNING`: Item is currently being executed, indicated with blue in the UI
+ - `NOT_EXECUTED`: Item has not yet been executed; the default item color is gray.
+
+While running the queue, it emits a set of signals/events via `self.emit`. The signals are:
+
+- queue_entry_execute_finished: When a queue entry finishes execution for any given reason, the queue entry in question and one of the strings (Failed, Successful, Skipped or Aborted) are passed with the signal.
+- queue_stopped: When the queue was stopped
+- queue_paused: When the queue was/is paused
```
fabcor-maxiv:
```suggestion
- `queue_entry_execute_finished`: When a queue entry finishes execution for any given reason, the queue entry in question and one of the strings (Failed, Successful, Skipped or Aborted) are passed with the signal.
- `queue_stopped`: When the queue was stopped
- `queue_paused`: When the queue was/is paused
```

docs/source/dev/queue.md
```diff
@@ -77,3 +78,44 @@ When the Web interface was designed, the scientists wanted to change some aspect
...
+
+The exception `QueueSkippEntryException` can be raised at any time to skip to the next entry in the queue. Raising QueueSkippEntryException` will skip to the next entry at the same level as the current node, meaning that all child nodes will be skipped as well. The
```
fabcor-maxiv:
```suggestion
The exception `QueueSkippEntryException` can be raised at any time to skip to the next entry in the queue. Raising `QueueSkippEntryException` will skip to the next entry at the same level as the current node, meaning that all child nodes will be skipped as well. The
```

docs/source/dev/queue.md
```diff
@@ -77,3 +78,44 @@ When the Web interface was designed, the scientists wanted to change some aspect
...
+
+## Dynamically loaded queue entries
+After some years of use, the developer community has found that some aspects of the queue can be improved. The queue was originally designed with a certain degree of flexibility in mind; it was thought that the queue was going to need to be extended with new collection protocols occasionally, but that  that the number of protocols would remain quite limited. Furthermore, the model layer `QueueModel` objects were designed as pure data-carrying objects with the intent that these objects could be instantiated from serialized data passed over an RPC protocol. The `QueueModel` objects were also originally designed to run on Python 2.4 (that was used at the time) and had no real support for typing these data-carrying objects. The Python dictionary data structure is often used to pass the data to various parts of the application. An approach that was adapted partly because the dictionary structure is simple to serialize to a string, but also because the already existing collection routines were already using dictionaries to pass data internally.
```
fabcor-maxiv:
```suggestion
After some years of use, the developer community has found that some aspects of the queue can be improved. The queue was originally designed with a certain degree of flexibility in mind; it was thought that the queue was going to need to be extended with new collection protocols occasionally, but that the number of protocols would remain quite limited. Furthermore, the model layer `QueueModel` objects were designed as pure data-carrying objects with the intent that these objects could be instantiated from serialized data passed over an RPC protocol. The `QueueModel` objects were also originally designed to run on Python 2.4 (that was used at the time) and had no real support for typing these data-carrying objects. The Python dictionary data structure is often used to pass the data to various parts of the application. An approach that was adapted partly because the dictionary structure is simple to serialize to a string, but also because the already existing collection routines were already using dictionaries to pass data internally.
```

fabcor-maxiv (commented):
Some suggestions

rhfogh:
Good start. I look forward to the description of how you add an object to the queue - and how you query the queue, in either Qt or Web.

On the nesting of the queue, I do not think it is enough to say that 'there are no limits on the queue, but in practice it is always 'Sample-Group-DataCollection'. If I put a complex nested structure on the queue, will the databases, viewers, user interface etc. reflect it? Or will they compress the structure to have a single group containing all the leaf nodes? Or will some objects be ignored? Unless there is a known contract about how complex structures should be dealt with, it is hard to use them with any confidence.

docs/source/dev/queue.md
fabcor-maxiv: Another suggestion: https://sembr.org/

In this case it would have made it much easier to see what has changed in this diff : )

marcus-oscarsson: True, would that mean that we would need to break on each period "." then ?

docs/source/dev/queue.md
marcus-oscarsson: Thanks

marcus-oscarsson:
Thanks @fabcor-maxiv

marcus-oscarsson:
@rhfogh Ill add something about adding items and querying the queue within shortly :)

I'm not sure that I can answer the question about the nesting entirely as we have not defined any contract on the nesting/composition of the queue entries. I can possibly describe how its handled today but the scope would be somewhat limited to mxcubeweb (and possibly qt). I would probably not be the person to answer for the other part about databases and data-viewers. There is not necessarily a strong relation on the  queue structure and how the data is represented in a database.

docs/source/dev/queue.md
fabcor-maxiv: Each period `.` is an opportunity for a line break, yes. That is the easiest and least controversial. Of course if two sentences fit within 80 or 88 characters, then they can easily stay on the same line.

But also a single sentence can be broken up over multiple lines, like in the example:

```
They are endowed with reason and conscience
and should act towards one another in a spirit of brotherhood.
```

but here the difficulty is in finding good places to put the line breaks : )

rhfogh (approved):
Very good!

I still have some issues, but they are with how the code is, not with how it is documented. At some point we ought to harmonise the queue so that all entries follow the same model (and, dammit,  rename QueueSkippEntryException to QueueSkipEntryException). But at least now it is clear where we are.

fabcor-maxiv (approved):

marcus-oscarsson:
@rhfogh, see #963 :)

marcus-oscarsson:
The QueueSkippEntryException is easy enough to rename :) It would be wonderful to continue with improving/refactoring the queue and harmonizing the queue code. We should in my opinion keep the backward compatibility. In that perspective it would probably be a good idea to keep the "old" entries as they are and translate them into the new system and then phase them out. There are probably other good ways of handling the transition that we need to consider and discuss. We also need to continue to discuss the technical bits.

rhfogh:
@marcus-oscarsson I concur

PR #956, opened 2024-07-04
yaml conf: make 'export yaml config' a user invokable featureExport yaml arg

elmjag:
Let's make it possible for the users run 'export yaml config' without editing the code.

This PR adds the nessesary hooks for mxcubeweb to implement the '--export-yaml-config' command line argument.

For mxcube-web part, see this PR: https://github.com/mxcube/mxcubeweb/pull/1285

### DIFF
--- a/mxcubecore/HardwareRepository.py
+++ b/mxcubecore/HardwareRepository.py
@@ -35,8 +35,9 @@ import os
 import time
 import importlib
 import traceback
-from typing import Union, TYPE_CHECKING
+from typing import Union, Optional, TYPE_CHECKING
 from datetime import datetime
+from pathlib import Path
 
 from ruamel.yaml import YAML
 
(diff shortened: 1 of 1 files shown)


rhfogh (changes requested):
Good idea, but about the implementation: 
I think it is problematic to add a new mandatory argument to a number of functions. First because a directory for once-in-a-rare-while export of yaml-converted files does not deserve to be a mandatory parameter. Second because this will break the Qt implementation, which will also have to introduce this mandatory argument. Could the whole thing not be rewritten so that all necessary arguments defaulted to None?

elmjag:
I have changed the PR so that ` yaml_export_directory` parameter is now optional for `init_hardware_repository()` and `load_from_yaml()` functions. I assume that's the functions that Qt-implementation is using.

marcus-oscarsson:
Thanks Guys have a nice WE !

PR #957, opened 2024-07-04
Rhfogh xml yaml

rhfogh:
Proposed improvement of yaml linting. See comments in .yamllint.yaml

Note that by setting `# % YAML 1.2` at the head of the files, we ensure that only `true` and `false` (with various capitalizations) are treated as Booleans - this is what ruamel-yaml (which follows Yaml 1.2) does anyway. By keeping the `truthy` rule we further limit Booleans to all-lower-case. 

Unfortunately there seems to be no way to ensure a single spelling for either `null`, `.inf`, or `.nan`

I have added further rules to block using empty elements as nulls (so nulls have to be added explicitly), and to ban numbers like `010`.

Anyway, I would propose that there should never be any need to put in yamllint-disable statements (except for line length).


elmjag:
Nice!

How about we merge this to develop branch instead?

This is not needed for adding YAML config files support. Let's keep the scope xml_yaml_conversation branch focused on that.

fabcor-maxiv:
> How about we merge this to develop branch instead?

@rhfogh I agree that this should be done on `develop` branch first.

PR #966, opened 2024-07-08
P11 char

agruzinov:
Code changes in the P11 files to make the characterization properly executed using EDNA.


agruzinov:
I did black several times, and it looks happy on several machines...

agruzinov:
It is ready for a review now. Rebased to the recent develop.

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
```diff
@@ -72,7 +72,8 @@ def init(self):
                 "DigitalZoomMotor: digital zoom is not supported " "by camera object"
             )
 
-        self.set_is_ready(self.get_state() == MotorStates.READY)
+        if MotorStates.READY:
+            self.is_ready()
```
beteva: This is always True. Is this what you want?

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
```diff
@@ -30,7 +30,7 @@
 from mxcubecore.HardwareObjects.abstract.AbstractMotor import MotorStates
 
 
-class DigitalZoomMotor(AbstractMotor, Device):
+class DigitalZoomMotor(AbstractMotor, HardwareObject):
```
beteva, line 33: AbstractMotor inherits from HardwareObject. Did not look at the rest of the code, but is it necessary to inherit from HardwareObject?

mxcubecore/HardwareObjects/DESY/MjpgStreamVideo.py
```diff
@@ -333,7 +333,7 @@ def init(self):
                 sensor_height = int(sensor_info["value"])
             self.sensor_dimensions = (sensor_width, sensor_height)
 
-        self.set_is_ready(True)
+        self.is_ready()
```
beteva: is_ready only returns True/False, does not set the state. This line looks strange.

mxcubecore/HardwareObjects/DESY/MjpgStreamVideo.py
```diff
@@ -46,7 +46,7 @@
 from mxcubecore.BaseHardwareObjects import HardwareObject
 
 
-class MjpgStreamVideo(AbstractVideoDevice, Device):
+class MjpgStreamVideo(AbstractVideoDevice, HardwareObject):
```
beteva: AbstractVideoDevice inherits from HardwareObject. Is it necessary to inherit from HardwareObject?

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -0,0 +1,112 @@
...
+import gevent
+
+from mxcubecore import HardwareRepository as HWR
+from mxcubecore.HardwareObjects.abstract.AbstractMachineInfo import AbstractMachineInfo
+from PyTango import DeviceProxy
+
+
+class P11MachineInfo(AbstractMachineInfo):
+    """Simulates the behaviour of an accelerator information"""
+
```
beteva: Maybe it is worth checking the TangoMachineInfo, which does the same job (except for the simulated values).

mxcubecore/HardwareObjects/DESY/P11Session.py
```diff
@@ -38,14 +38,12 @@
 
 
 class P11Session(Session):
-
     default_archive_folder = "raw"
 
     def __init__(self, *args):
         super().__init__(*args)
```
beteva: You can remove this.

mxcubecore/HardwareObjects/DESY/P11Zoom.py
```diff
@@ -43,7 +43,6 @@ def __init__(self, name):
         AbstractNState.__init__(self, name)
```
beteva: I would use` super().__init__(name)` instead.

mxcubecore/HardwareObjects/DESY/MjpgStreamVideo.py
fabcor-maxiv: That seems strange... Does this method do something or is it a getter?

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -257,10 +236,10 @@ def do_collect(self, owner):
...
 
     def data_collection_hook(self):
-        if not self.init_ok:
-            raise RuntimeError(
-                "P11Collect. - object initialization failed. COLLECTION not possible"
-            )
+        # if not self.init_ok:
+        #     raise RuntimeError(
+        #         "P11Collect. - object initialization failed. COLLECTION not possible"
+        #     )
```
fabcor-maxiv:
```suggestion
```

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -313,13 +303,13 @@ def data_collection_hook(self):
                     + "%s_%d" % (prefix, runno),
                 )
 
-                # Filepath to the EDNA processing
-                # filepath = os.path.join(basepath,"%s_%d" % (prefix, runno))
+                # # Filepath to the EDNA processing
+                # filepath = os.path.join(basepath, "%s_%d" % (prefix, runno))
```
fabcor-maxiv:
```suggestion
```

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -313,13 +303,13 @@ def data_collection_hook(self):
...
 
-                # setting up xds_dir for characterisation (used there internally to create dirs)
-                self.current_dc_parameters["xds_dir"] = os.path.join(
-                    basepath, "%s_%d" % (prefix, runno)
-                )
+                # # setting up xds_dir for characterisation (used there internally to create dirs)
+                # self.current_dc_parameters["xds_dir"] = os.path.join(
+                #     basepath, "%s_%d" % (prefix, runno)
+                # )
```
fabcor-maxiv:
```suggestion
```

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -466,28 +792,71 @@ def collect_characterisation(
...
+        The function prepares a standard collection by setting the start angle and angle increment in
+        the header of the Eiger detector.
+
+        :param start_angle: The start_angle parameter represents the starting angle of the standard collection
+        sequence. It is used to also set the start angle in the header of the detector.
+        :param img_range: The `img_range` parameter represents the range of angles over which the
+        detector will collect single image. It is used to set the angle increment in the header of the Eiger
+        detector
+        :return: a boolean value of True.
+        """
```
fabcor-maxiv: The contribution guideline says to use Google style for docstrings: https://mxcubecore.readthedocs.io/en/stable/dev/contributing.html#style-guidlines

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -466,28 +792,71 @@ def collect_characterisation(
...
+
+    def collect_std_collection(self, start_angle, stop_angle):
+        """
+        The function collects data from a standard acquisition by moving the omega motor from a start
+        angle to a stop angle.
+
+        :param start_angle: The starting angle for the collection
+        :param stop_angle: The stop_angle parameter is the final angle at which the collection should
+        stop
+        """
```
fabcor-maxiv: Use google style for docstrings

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -0,0 +1,112 @@
...
+                                            True (attention raised to the user)
+
+   optional fields:
+      any number of optional fields can be sent over with this signal by
+      adding them in the values dictionary
+
+      for example:
+         values['lifetime']
+         values['topup_remaining']
+"""
```
fabcor-maxiv: Is this docstring correct? It does not seem so, because of the name for example. Anyway it looks like a class docstring so it should be part of the class, and there are some formatting issues in this docstring.

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -0,0 +1,112 @@
...
+pass
```
fabcor-maxiv: This line probably does not belong here.

```suggestion
```

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
```diff
@@ -168,6 +168,33 @@ def init(self):
...
+        # ]:
+        #     self.init_ok = False
+        #     self.log.debug("lower_bound_ch: %s" % self.lower_bound_ch)
+        #     self.log.debug("upper_bound_ch: %s" % self.upper_bound_ch)
+        #     self.log.debug("acq_arm_cmd: %s" % self.acq_arm_cmd)
+        #     self.log.debug("acq_on_cmd: %s" % self.acq_on_cmd)
+        #     self.log.debug("acq_off_cmd: %s" % self.acq_off_cmd)
+        #     self.log.debug("acq_window_off_cmd: %s" % self.acq_window_off_cmd)
+        # else:
+        #     self.init_ok = True
```
fabcor-maxiv:
```suggestion
```

fabcor-maxiv (commented):
Many of the files in this pull request contain no meaningful change, but only formatting (empty lines, and imports). I guess it is fine because they needed to be reformatted one day or another, but that is unexpected and a bit confusing.

The contribution guideline says to use Google style for docstrings: https://mxcubecore.readthedocs.io/en/stable/dev/contributing.html#style-guidlines

agruzinov:
Thanks a lot for everyone for the comments. I've tried to incorporate most of them except the doc strings.

If the rest is fine I would suggests to merge it like now and that I'll submit another PR for the docstrings.

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
agruzinov: There are still some issues with data collection. That is why I still need some of the commented code. Once it will reach the equilibrium, my plan is to do the polishing. Thanks for noting that.

mxcubecore/HardwareObjects/DESY/P11Zoom.py
agruzinov: Thanks for the comments. I've removed the respecting code parts.

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -404,7 +404,7 @@ def close_detector_cover(self):
         pass
 
     def set_transmission(self, value):
-        """
+        """def get_beam_
```
rhfogh: This does not look right. Why 'get_beam' if the function is 'set_transmission'? Better nothing than something wrong.

mxcubecore/HardwareObjects/MotorsNPosition.py
```diff
@@ -182,6 +183,10 @@ def motor_value_changed(self, position=None):
         self.update_multi_value()
 
     def update_multi_value(self):
+        if self._updating_multi_value:
+            return
```
fabcor-maxiv: Why are we not returning anything here, when the last line of this function is `return current_idx`? This does not seem okay.

mxcubecore/HardwareObjects/DESY/P11Beam.py
```diff
@@ -20,19 +20,28 @@
...
+    ELLIPTICAL = "ellipse"
+
+
 class P11Beam(AbstractBeam):
     def __init__(self, *args):
         super().__init__(*args)
 
-        self._beam_size_dict = {"aperture": [9999, 9999], "slits": [9999, 9999]}
-
+        self.pinhole_hwobj = self.get_object_by_role("pinhole")
```
beteva, line 35: Maybe it is more consistent if you move this to init,.

mxcubecore/HardwareObjects/DESY/P11Beam.py
```diff
@@ -45,6 +54,20 @@ def __init__(self, *args):
...
+        }
+
+        self._beam_shape = BeamShape.ELLIPTICAL
+        self._beam_info_dict = {
+            "size_x": self._beam_width,
+            "size_y": self._beam_height,
+            "shape": self._beam_shape,
+            "label": self._beam_label,
+        }
+
```
beteva: Is there a reason that you duplicate part of the `AbstractBeam.__init__` and you change the aperture and slits size values?
Is there a reason you change the beam sshape from UNKNOWN to ELLIPTICAL in `__init__`.

fabcor-maxiv (dismissed):
I reviewed only the common parts, I ignored what has changed in `DESY` directory.

mxcubecore/HardwareObjects/DESY/P11Beam.py
```diff
@@ -105,3 +154,60 @@ def mirror_idx_changed(self, value=None):
...
+        self._beam_width = _size[0]
+        self._beam_height = _size[1]
+        self._beam_shape = _shape
+        self._beam_info_dict["size_x"] = _size[0]
+        self._beam_info_dict["size_y"] = _size[1]
+        self._beam_info_dict["shape"] = _shape
+        self._beam_info_dict["label"] = _label
+
+        return self._beam_info_dict
```
beteva: Is there a reason why you duplicate the  AbstractBeam evaluate_beam_info?

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -118,7 +94,14 @@ def set_energy(self, value):
...
         """
-        HWR.beamline.energy.set_value(value)
+        current_value = HWR.beamline.energy.get_value()
+        if abs(current_value - value) < 0.01:
+            self.log.debug(
+                "The difference between the current and desired energy values is less than 0.01. No change made."
+            )
+        else:
+            HWR.beamline.energy.set_value(value)
+            self.log.debug(f"Energy value set to {value}.")
```
beteva: May I suggest that you put your logic rather in your P11Energy implementation, not in the Collect. You may need the same for moving the energy as a motor or when you change the wavelength. Thus whenever you do set_value, you will have the test for whether to move or not and  `HWR.beamline.energy.set_value(value)` here would be enough.

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -0,0 +1,77 @@
...
+from PyTango import DeviceProxy
+
+
+class P11MachineInfo(AbstractMachineInfo):
+    """Simulates the behaviour of an accelerator information"""
+
+    default_current = 100.0  # [mA]
+    default_message = "Machine message is not updated"
+    default_lifetime = 1.85  # hours Lifetime
+    default_maschine_energy = 6.08  # GeV
```
beteva: Is this a mockup?

beteva (changes requested):

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
agruzinov: No idea how it is landed there, Thanks for checking. Removed.

mxcubecore/HardwareObjects/MotorsNPosition.py
agruzinov: This is due to the previous error when the at some point updating recursion was happening with the error. So this is to prevent it updating if it is updating already.

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
agruzinov: In a sense, it is not completely finished for P11.

mxcubecore/HardwareObjects/MotorsNPosition.py
fabcor-maxiv: But should this function return something or not? The last line of the function returns a value, doesn't it? That is why I am confused...

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: Done

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: No particular reason. Cleaned -up.

mxcubecore/HardwareObjects/MotorsNPosition.py
beteva: I am wondering is it not simpler to configure those motors as NState devices, so this class may not be even needed. Or should it not inherit from AnstractNState, so all the updating mechanism is already in place?
Or I am too naive?

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: I was too fast. The previous realization had this. After update it is not needed. Removed. Thanks for noticing.

mxcubecore/HardwareObjects/DESY/P11Beam.py
```diff
@@ -20,19 +20,28 @@
...
 
 
+class BeamShape(Enum):
+    """Beam shape definitions"""
+
+    UNKNOWN = "unknown"
+    RECTANGULAR = "rectangular"
+    ELLIPTICAL = "ellipse"
+
+
```
beteva: This class is defined in AbstractBeam, simple import should do it.

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Thanks for comment. Moved to P11Energy.

mxcubecore/HardwareObjects/MotorsNPosition.py
agruzinov: IN this case it should not return anything. I've substituted it with pass. 
I need to look into reimplementation. I can not tell right know. I would avoid touching it just yet.

fabcor-maxiv: > IN this case it should not return anything.

But this method `update_multi_value` must always return a value, or not? For example here:

https://github.com/mxcube/mxcubecore/blob/8cd71300604b0646be497e90c8d8aab2799b0de0/mxcubecore/HardwareObjects/MotorsNPosition.py#L135

> IN this case it should not return anything. I've substituted it with pass.

I have no idea what this is supposed to achieve. :D

>  I need to look into reimplementation. I can not tell right know. I would avoid touching it just yet.

Yes, this needs to be rethought, I do not think it should be merged as it is now.

mxcubecore/HardwareObjects/MotorsNPosition.py
```diff
@@ -182,6 +183,10 @@ def motor_value_changed(self, position=None):
         self.update_multi_value()
 
     def update_multi_value(self):
+        if self._updating_multi_value:
+            pass
+        self._updating_multi_value = True
```
fabcor-maxiv: This does not make sense to me. As far as I can tell this `if` clause does not do anything. Maybe the `if` clause needs to be removed entirely.

```suggestion
        self._updating_multi_value = True
```

rhfogh: The change looks problematic too. The basic ideas seems to be to avoid having this function called indirectly while it is already being executed. To achieve this
1) decide what shoudl be returned from these recursive calls. Since the 'no current position' value seems to be -1, maybe that would be a better choice. 

2) I would propose to start with

  current_idx = -1
    if not self._updating_multi_value:
      self._updating_multi_value = True
      try:
        Put the entire body of the function here
        Then end with:

      finally:
    self._updating_multi_value = False
  return current_idx

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
rhfogh: Agree with Antonia

mxcubecore/HardwareObjects/DESY/P11Beam.py
rhfogh: Now nothing is happening in __init__ you chould remove the function altogether. There is no need for a function when the only thing it does is to call teh same superclass function

mxcubecore/HardwareObjects/DESY/P11Beam.py
```diff
@@ -105,3 +132,60 @@ def mirror_idx_changed(self, value=None):
...
+    def get_pinhole_size(self):
+        # Keep it default as the pinhole and beamsize interaction is locked for now
+        return 200
+
+    def get_beam_focus_label(self):
+        value = self.mirror_idx_ch.get_value()
+
+        if value not in self.focus_sizes:
+            value = -1
```
rhfogh, line 143: This line is unnecessary and  can be removed

mxcubecore/HardwareObjects/DESY/P11Beam.py
rhfogh: Looks like the function is still there, and the only difference seems to be the conversion to a numpy array before finding the min. What does that add compared to just doing min(self._beam_size_dict.values())

mxcubecore/HardwareObjects/DESY/P11DetectorDistance.py
```diff
@@ -149,6 +149,16 @@ def _set_value(self, value):
         # Wait until motor is reachiung the actual distance within tolerance.
         tolerance = 1.0  # mm Actual tolerance is within 0.3 range.
         while abs(self.get_value() - value) >= tolerance:
+            _state = self.chan_state.get_value()
+            if _state == "ON":
+                state = self.STATES.READY
```
rhfogh, line 155: Should update_state also be  called when you reach state READY? With this code it does not.

rhfogh (changes requested):
I only went through parts of this, but it looks OK. The outstanding comments should be addressed, though.

mxcubecore/HardwareObjects/DESY/P11Beam.py
beteva: Agree with @rhfogh

mxcubecore/HardwareObjects/DESY/P11Beam.py
beteva: evaluate_beam_info is still there.

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
beteva: @agruzinov , did you look at TangoMachineInfo?

mxcubecore/HardwareObjects/DESY/P11Transmission.py
```diff
@@ -42,7 +42,6 @@ def __init__(self, name):
         self.chan_state = None
 
     def init(self):
-
         limits = self.get_property("limits", None)
```
beteva: You can use the AbstractActuator limits handling and remove it from your HO if you just change in the xml file "limits" to "default_limits".

beteva: You can also remove the value_changed method, as it is already handled in the AbstractActuator.

beteva:
@agruzinov sorry for being fussy. I see that there is a lot of legacy code in the P11 HardwareObjects and you are doing good cleaning and adaptation job.

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: OK, afetr looking closer, if I did not modify it, i'm getting a bunch of errors about the conversion

agruzinov: 2024-09-06 16:09:20,054 |HWR  |ERROR  | Cannot load brick beam_size_brick : initialization failed
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/configuration.py", line 82, in load_brick
    new_instance = class_obj(None, brick_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/bricks/beam_size_brick.py", line 123, in __init__
    self.connect(HWR.beamline.beam, "beamInfoChanged", self.beam_info_changed)
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/base_components.py", line 772, in connect_hwobj
    _sender = emitter(sender)
              ^^^^^^^^^^^^^^^
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/base_components.py", line 66, in emitter
    _emitter_cache[ob] = _QObject(ho=ob)
    ~~~~~~~~~~~~~~^^^^
  File "/home/p11user/.localpython3119/lib/python3.11/weakref.py", line 428, in __setitem__
    self.data[ref(key, self._remove)] = value
              ^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot create weak reference to 'NoneType' object
2024-09-06 16:09:20,095 |HWR.P11Pinhole|DEBUG  |      - motor pinholey is at -235
2024-09-06 16:09:20,095 |HWR.P11Pinhole|DEBUG  |      - motor pinholez is at -13986
2024-09-06 16:09:20,138 |HWR  |ERROR  | Cannot load brick task_toolbox_brick : initialization failed
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/configuration.py", line 82, in load_brick
    new_instance = class_obj(None, brick_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/bricks/task_toolbox_brick.py", line 67, in __init__
    self.task_tool_box_widget = TaskToolBoxWidget(self)
                                ^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/widgets/task_toolbox_widget.py", line 69, in __init__
    self.discrete_page = CreateDiscreteWidget(self.tool_box, "Discrete")
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/widgets/create_discrete_widget.py", line 44, in __init__
    CreateTaskBase.__init__(
  File "/gpfs/local/shared/MXCuBE/mxcubeqt/mxcubeqt/widgets/create_task_base.py", line 105, in __init__
    HWR.beamline.beam.connect(
    ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'connect'

agruzinov: 2024-09-06 16:09:15,038 |HWR  |ERROR  | Cannot initialize Hardware Object "/beam"
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareRepository.py", line 472, in _load_hardware_object
    hwobj_instance.init()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/DESY/P11Beam.py", line 56, in init
    self.mirror_idx_changed()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/DESY/P11Beam.py", line 131, in mirror_idx_changed
    self.evaluate_beam_info()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/abstract/AbstractBeam.py", line 221, in evaluate_beam_info
    _size = min(self._beam_size_dict.values())
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'list' and 'tuple'
2024-09-06 16:09:15,135 |HWR  |ERROR  | Cannot initialize Hardware Object "/flux"
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareRepository.py", line 472, in _load_hardware_object
    hwobj_instance.init()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/DESY/P11Flux.py", line 49, in init
    self.measure_flux()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/DESY/P11Flux.py", line 59, in measure_flux
    beam_size = HWR.beamline.beam.get_beam_size()
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_beam_size'

agruzinov: 2024-09-06 16:09:19,090 |HWR  |ERROR  | Cannot initialize Hardware Object "/mxcollect"
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareRepository.py", line 472, in _load_hardware_object
    hwobj_instance.init()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/DESY/P11Collect.py", line 54, in init
    super().init()
  File "/gpfs/local/shared/MXCuBE/mxcubecore/mxcubecore/HardwareObjects/abstract/AbstractCollect.py", line 98, in init
    beam_div_hor, beam_div_ver = HWR.beamline.beam.get_beam_divergence()
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_beam_divergence'

agruzinov: AttributeError: 'NoneType' object has no attribute 'get_beam_divergence'
+==================================================================================================================+
| role                 | Class                     | file                            | Time (ms)| Comment                   
+==================================================================================================================+
| beamline             | Beamline                  | beamline_config.yml             | 11       | Start loading contents:   
| session              | P11Session                | session.xml                     | 55       |                           
| machine_info         | P11MachineInfo            | mach-info.xml                   | 58       |                           
| transmission         | P11Transmission           | transmission.xml                | 38       |                           
| energy               | P11Energy                 | energy.xml                      | 22       |                           
| beam                 | None                      | beam.xml                        | 27       | No object loaded          
| flux                 | None                      | flux.xml                        | 97       | No object loaded          
| detector             | P11EigerDetector          | eiger.xml                       | 64       |                           
| resolution           | Resolution                | resolution.xml                  | 4        |                           
| hutch_interlock      | DoorInterlockMockup       | door-interlock-mockup.xml       | 0        |                           
| safety_shutter       | P11Shutter                | shutter-eh.xml                  | 9        |                           
| fast_shutter         | P11FastShutter            | fast-shutter.xml                | 7        |                           
| sample_changer       | P11SampleChanger          | sample-changer.xml              | 83       |                           
| diffractometer       | P11NanoDiff               | nanodiff.xml                    | 3306     |                           
| sample_view          | QtGraphicsManager         | sample-view.xml                 | 7        |                           
| lims                 | P11ISPyBClient            | lims.xml                        | 443      |                           
| queue_manager        | QueueManager              | queue.xml                       | 18       |                           
| queue_model          | QueueModel                | queue-model.xml                 | 3        |                           
| collect              | None                      | mxcollect.xml                   | 3        | No object loaded          
| energy_scan          | EnergyScanMockup          | energyscan-mockup.xml           | 3        |                           
| offline_processing   | OfflineProcessingMockup   | offline-processing-mockup.xml   | 7        |                           
| online_processing    | OnlineProcessingMockup    | online-processing-mockup.xml    | 9        |                           
| characterisation     | P11EDNACharacterisation   | characterisation.xml            | 11       |                           
| beamline             | Beamline                  | beamline_config.yml             | 4297     | Done loading contents     
+==================================================================================================================+

agruzinov: after explicit conversion to the list() it goes away.

agruzinov: python 3.11.9, Debian12

agruzinov: if it works for everyone else, I thought there is no need to touch Abstract Class.

beteva: Unfortunately I do not see what needs to be casted to list. Beam divergence is a tuple, which is constructed by two properties (see the init of the AbstractBeam)

beteva: Ok, got it now. It is not the divergence, but the beam size. The problem you have is that you cast to list before converting to numpy array,  while in the AbstractBeam min compares **lists** (and I insist on the plural) of floats. The code beneath takes care to choose what is actually you beam size, in case you have several values (e.g. you have the same beam size values when you have both aperture and definer, using the same aperture).

agruzinov: OK, thanks for the comments. I just needed to convert to list and not the tuple.

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: Added, thanks.

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
agruzinov: Thanks for the comment. It is now re-implemented using TangoMaschineInfo.

mxcubecore/HardwareObjects/DESY/MjpgStreamVideo.py
agruzinov: Fixed.

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
agruzinov: Fixed.

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
agruzinov: Fixed.

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Fixed

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Fixed.

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Fixed.

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
agruzinov: After test this part is still needed apparently. Keeping uncommented. Fixed.

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
agruzinov: Reimplemeted. Fixed.

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: Fixed. Good catch, thanks!

mxcubecore/HardwareObjects/DESY/P11DetectorDistance.py
agruzinov: Thanks for the catch. Fixed.

mxcubecore/HardwareObjects/DESY/P11Transmission.py
agruzinov: Thanks for the comment. Updated including doc strings. Closing.

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Fixed. Thanks.

mxcubecore/HardwareObjects/DESY/P11Collect.py
agruzinov: Fixed.

beteva (approved):

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: Sorry, one more time, I guess I've removed the wrong line :) Which one was that?

agruzinov:
Thanks everyone for the looking into it and nice suggestions. I've renamed it to WIP for now as it requere more testing.

mxcubecore/HardwareObjects/MotorsNPosition.py
agruzinov: Thanks for the proposal. I've updated it now.

mxcubecore/HardwareObjects/MotorsNPosition.py
agruzinov: @beteva Thanks for the suggestion about the NState! I've rewrote the beamstop, collimator, pinhole and diode motors to use this class instead of MultiNState. It required also to add respective changes to the qt code. I've tested it and it seems to be working for now with this updates. Could you please have  a look?

rhfogh (approved):
Approved on a lightning glance. We have been over this before, the non-DESY parts look innocuous, the DESY parts can be left to the DESY people, and I am sure Antonia can vet MotorsNPosition. Let us get this merged, finally

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Returns:
+            float: The current in milliamps (mA). Returns 0.0 in case of an error.
+        """
+        try:
+            return self.current.get_value()
+        except Exception as err:
+            logging.getLogger("HWR").exception(f"Error reading current: {err}")
+            return 0.0
+
```
beteva: You can remove this, as it is already in the inherited class.

mxcubecore/HardwareObjects/TangoMotor.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, name):
         self.step_limits = None
 
     def init(self):
-        """Connects to all Tine channels and commands"""
+        """Connects to all Tango channels and commands"""
```
beteva, line 54: Thanks for cleaning up

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
beteva: Still unnecessary inheritance from HardwareObject

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
agruzinov: Thanks for the comment! I was wondering is machine_state_brick in qt somehow in use. If I add it in the designer mode t is not seemed to be displaying anything and in the mock-up too...

agruzinov: OK, it looks like it is ok for now. Closing.

mxcubecore/HardwareObjects/DESY/P11Beam.py
agruzinov: Seems to work now. Closing.

mxcubecore/HardwareObjects/MotorsNPosition.py
agruzinov: @fabcor-maxiv , I've updated it according to the latest suggestions, but additionally the corresponding hardware objects are rewritten and using different class. And it also do not produce any reccurency errors. Ive tested it, and it seems to be happy, but I would keep this class as an emergency rollback solution if this is ok.

beteva:
@agruzinov I am merging your PR.
Could you, please,, try to make smaller PRs.  This way it is easier to review and thus should not take such long time to merge, as it was with this one. Also the rebase would be easier.
It would be great if you could run black on the PR related files only. People will be less reluctant to merge 😄

agruzinov:
@beteva Thank you and everyone for the suggestions. Unfortunately it is not always possible to make the PR smaller as there is still a development phase that require a lot of fixes in different places that could be done and tested only during sparse commissioning beam times, but I will try to. Cheers!

PR #968, opened 2024-07-09
Rhfogh xml yaml

rhfogh:
Fixed yaml syntax and yaml linting, conf. recent changes in develop branch


rhfogh:
Since these changes are OK'ed on other branches (and I need to sort out some problem with failures in mxcubewe commits), I am merging these myself.

Issue #970, opened 2024-07-11
Dependency on qt view objects and mount_sample


marcus-oscarsson:
We are looking into making the mxcubeweb layer associated with the queue a little bit thinner and push down some code into mxcubecore.  Its something we wanted to do for a while and its going to be a multi-step process. 

One important step in the process would be to get rid of the tight coupling to the qt view objects.  There is a good example in `Sample.mount_sample`:

https://github.com/mxcube/mxcubecore/blob/e78be97ec32d514527c54d0fec64e1d7088a7d48/mxcubecore/queue_entry/base_queue_entry.py#L840

It would already be good if we could start with `Sample.mount_sample` method. I would suggest to remove the `view` argument entirely and replace the `setText` with a signal. The centering method in the example above should be replaced with a setting variable accessible with ie. `HWR.beamline.sample_view` or similar.

This would require some testing on the Qt side, something I cant do with full coverage so we would need some help from you Qt developers here.

PR #971, opened 2024-07-12
[mount_sample] Simplified mount_sample and related logic - Part 1

marcus-oscarsson:
Related to issue #970 

- Removed `SampleChanger.load_sample` and replaced it with `SampleChanger.load`

The `SampleChanger.mount_sample` method uses a method called `load_sample` that are only defined on a few sample changers and not defined in `AbstractSampleChanger`. `SampleChanger.mount_sample` seems to be the only place where `load_sample` is used. Perhaps one of the Qt developers could confirm that its not used directly by the UI or some site specific qt related code.

- Introduced a progress message signal on `AbstractSampleChanger`
- Introduced a centering_method setting that is meant to be application wide on the `SampleView` HardwareObject

### DIFF
--- a/mxcubecore/queue_entry/base_queue_entry.py
+++ b/mxcubecore/queue_entry/base_queue_entry.py
@@ -675,7 +675,6 @@ class SampleQueueEntry(BaseQueueEntry):
                     self.sample_centring_result = gevent.event.AsyncResult()
                     try:
                         mount_sample(
-                            self._view,
                             self._data_model,
                             self.centring_done,
                             self.sample_centring_result,
@@ -766,16 +765,12 @@ class BasketQueueEntry(BaseQueueEntry):
         BaseQueueEntry.__init__(self, view, data_model)
 
 
-def mount_sample(view, data_model, centring_done_cb, async_result):
-    view.setText(1, "Loading sample")
+def mount_sample(data_model, centring_done_cb, async_result):
+    HWR.beamline.sample_changer.trigger_progress_message("Loading sample")
     HWR.beamline.sample_view.clear_all()
-    log = logging.getLogger("queue_exec")
+    log = logging.getLogger("user_level_log")
 
     loc = data_model.location
-    holder_length = data_model.holder_length
-
-    snapshot_before_filename = "/tmp/test_before.png"
-    snapshot_after_filename = "/tmp/test_after.png"
 
     robot_action_dict = {
         "actionType": "LOAD",
@@ -786,34 +781,13 @@ def mount_sample(view, data_model, centring_done_cb, async_result):
         "sessionId": HWR.beamline.session.session_id,
         "startTime": time.strftime("%Y-%m-%d %H:%M:%S"),
     }
-    # "xtalSnapshotBefore": data_model.get_snapshot_filename(prefix="before"),
-    # "xtalSnapshotAfter": data_model.get_snapshot_filename(prefix="after")}
 
-    sample_mount_device = HWR.beamline.sample_changer
-
-    if hasattr(sample_mount_device, "__TYPE__"):
-        if sample_mount_device.__TYPE__ in ["Marvin", "CATS"]:
-            element = "%d:%02d" % tuple(loc)
-            sample_mount_device.load(sample=element, wait=True)
-        elif sample_mount_device.__TYPE__ == "PlateManipulator":
-            sample_mount_device.load_sample(sample_location=loc)
-        else:
-            if (
-                sample_mount_device.load_sample(
-                    holder_length, sample_location=loc, wait=True
-                )
-                is False
-            ):
-                # WARNING: explicit test of False return value.
-                # This is to preserve backward compatibility (load_sample was supposed to return None);
-                # if sample could not be loaded, but no exception is raised, let's skip
-                # the sample
-                raise QueueSkipEntryException(
-                    "Sample changer could not load sample", ""
-                )
+    if not HWR.beamline.sample_changer.load(sample=data_model.loc_str, wait=True):
+        raise QueueSkipEntryException("Sample changer could not load sample", "")
 
     robot_action_dict["endTime"] = time.strftime("%Y-%m-%d %H:%M:%S")
-    if sample_mount_device.has_loaded_sample():
+
+    if HWR.beamline.sample_changer.has_loaded_sample():
         robot_action_dict["status"] = "SUCCESS"
     else:
         robot_action_dict["message"] = "Sample was not loaded"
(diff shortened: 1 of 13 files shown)


mxcubecore/queue_entry/base_queue_entry.py
```diff
@@ -764,16 +763,12 @@ def __init__(self, view=None, data_model=None):
         BaseQueueEntry.__init__(self, view, data_model)
 
 
-def mount_sample(view, data_model, centring_done_cb, async_result):
-    view.setText(1, "Loading sample")
+def mount_sample(data_model, centring_done_cb, async_result):
+    HWR.beamline.sample_changer.trigger_progress_message("Loading sample")
```
marcus-oscarsson, line 769: `view.setText` replaced by `HWR.beamline.sample_changer.trigger_progress_message`

mxcubecore/queue_entry/base_queue_entry.py
```diff
@@ -764,16 +763,12 @@ def __init__(self, view=None, data_model=None):
...
     HWR.beamline.sample_view.clear_all()
-    log = logging.getLogger("queue_exec")
+    log = logging.getLogger("user_level_log")
```
marcus-oscarsson, line 771: `user_level_log` seemed more appropriate

mxcubecore/queue_entry/base_queue_entry.py
```diff
@@ -784,60 +779,32 @@ def mount_sample(view, data_model, centring_done_cb, async_result):
...
         dm = HWR.beamline.diffractometer
-        if dm is not None:
-            if hasattr(sample_mount_device, "__TYPE__"):
-                if sample_mount_device.__TYPE__ in (
-                    "Marvin",
-                    "PlateManipulator",
-                    "Mockup",
-                ):
-                    return
+        centring_method = HWR.beamline.sample_view.centring_method
```
marcus-oscarsson: `centring_method` now on `HWR.beamline.sample_view` and not `view.listView().parent().parent().centring_method`

mxcubecore/queue_entry/base_queue_entry.py
```diff
@@ -784,60 +779,32 @@ def mount_sample(view, data_model, centring_done_cb, async_result):
...
-        view.setText(1, "Sample not loaded")
+    if not HWR.beamline.sample_changer.has_loaded_sample():
+        HWR.beamline.sample_changer.trigger_progress_message("Sample not loaded")
         raise QueueSkipEntryException("Sample not loaded", "")
     else:
-        view.setText(1, "Sample loaded")
+        HWR.beamline.sample_changer.trigger_progress_message("Sample loaded")
         dm = HWR.beamline.diffractometer
-        if dm is not None:
-            if hasattr(sample_mount_device, "__TYPE__"):
```
marcus-oscarsson, line 833: Removed by adding `CENTRING_METHOD.NONE` (which is default), seems a bit odd to not center depending on  `__TYPE__`

mxcubecore/queue_entry/base_queue_entry.py
```diff
@@ -784,60 +779,32 @@ def mount_sample(view, data_model, centring_done_cb, async_result):
...
         robot_action_dict["status"] = "SUCCESS"
     else:
         robot_action_dict["message"] = "Sample was not loaded"
         robot_action_dict["status"] = "ERROR"
 
     HWR.beamline.lims.store_robot_action(robot_action_dict)
 
-    if not sample_mount_device.has_loaded_sample():
-        # Disables all related collections
-        view.setOn(False)
```
marcus-oscarsson, line 826: This is a bit odd, it will disable the view if sample could not be loaded. Is it really necessary ?

marcus-oscarsson:
This code is used when using the Qt UI and running the queue with the automatic option We need to remove the dependency of the Qt view so that also the web version can use this logic. It seems like there is not alot of sites using this piece of code maybe @rhfogh and/or @MartinSavko could confirm ?

rhfogh:
The only use for any of these functions in the qt branch is in plate_navigator_widget, functions navigation_item_double_clicked and navigation_table_double_clicked, which both call HWR.beamline.plate_manipulator.load_sample. Not sure whether it might not be better to keep load_sample for plate manipulators (only), rather than calling a function with '_' prefix from outside the class.

marcus-oscarsson:
I agree that `SampleChanger._load_sample` should not be called from outside of the class but `SampleChanger.load` could perhaps be used instead so that its more consistent with `AbstractSampleChanger`. For the code to work without making various type checks the queue load sample mechanism have to be named the same, its less important from a UI perspective and there we could call it `load_sample` even if it seems a bit unnecessary.

rhfogh:
From the Qt side there are only two calls, both like  
        HWR.beamline.plate_manipulator.load_sample(
            (table_item.row() + 1, table_item.column() * self.num_drops + 1),
            wait=False)

Is it the case that PlateMsnipulator.load_sample 1) is simply renamed to _load_sample, 2) has the same interface?  
It would be simple enough to just rename the two calls from the Qt side, but I am not at all at home with plate manipulators, so someone else ought to make the judgement call.

marcus-oscarsson:
@rhfogh The reason for the renaming  was that `SampleChanger` only defines a `load` method and not a `load_sample` method. As `PlateManipulator` is a sample changer and is used as such we expect it to have `load` method. There was previously a `isinstance` check but I think that we should move away from that as much (and far :) ) as we can.

marcus-oscarsson:
Good to go :)

jbflo:
> Good to go :)

[ Harvester ]
Maybe we should add the logic that were removed from this  PR : https://github.com/mxcube/mxcubeweb/pull/1469
to this one.

jbflo:
> Good to go :)

Good !
As discussed  https://github.com/mxcube/mxcubeweb/pull/1469#discussion_r1816378466
I add the necessary for the Harvester in this pr

marcus-oscarsson:
Nice @jbflo , well done. There seems to be some linting issues, but looks good apart from that !

jbflo:
> Nice @jbflo , well done. There seems to be some linting issues, but looks good apart from that !

Edited
@marcus-oscarsson  got it now.

marcus-oscarsson:
@jbflo, I think you got it with your last commit, everything is passing  now

marcus-oscarsson:
@jbflo, did you do all the changes you wanted to ? are we good to go again ;)

jbflo:
> @jbflo, did you do all the changes you wanted to ? are we good to go again ;)

Yes good to go. Thanks

marcus-oscarsson:
well, big thanks to you for doing the last mile :)

PR #972, opened 2024-07-12
Deprecating Device and Equipment

marcus-oscarsson:
The `Device` and `Equipment` classes are still in the code base but now with a deprecation warning. There are still a few more objects that use Device and Equipment to look at. Please have a look at your site specific and local code, let me know and make the necessary changes. I will follow up with a PR that removes `Equipment` and `Device`

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -1138,6 +1138,12 @@ class Device(HardwareObject):
 
     (NOTREADY, READY) = (0, 1)  # device states
 
+    def __init_subclass__(cls, **kwargs):
+        warnings.warn(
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
I do not know much about making deprecations.

This page says we should put one in `__init_subclass__`: https://dev.to/hckjck/python-deprecation-2mof

marcus-oscarsson:
I guess its in-case the subclass does not call `__init__`

rhfogh (approved):
Looks good. The warnings will not have the same effect as an exception, right?

beteva:
> The warnings will not have the same effect as an exception, right?

The warning only prints the text. It does not raise any exception and does not stop the execution.

PR #974, opened 2024-07-22
MAXIV: update Machine Info HWO

elmjag:
Updated MachInfo hardware object, for use at MAXIV BioMAX and MicroMAX beamlines.

Based on older BioMAX HWO, updated to follow the new AbstractMachineInfo API and refactored.

### DIFF
--- a/mxcubecore/HardwareObjects/MAXIV/MachInfo.py
+++ b/mxcubecore/HardwareObjects/MAXIV/MachInfo.py
@@ -1,153 +1,109 @@
-"""
-[Name] MachInfoMockup
-
-[Description]
-MachInfo hardware objects are used to obtain information from the accelerator
-control system.
-
-This is a mockup hardware object, it simulates the behaviour of an accelerator
-information by :
-
-    - produces a current value that varies with time
-    - simulates a control room message that changes with some condition
-      ()
-    - simulates
-
-[Emited signals]
-machInfoChanged
-   pars:  values (dict)
-
-   mandatory fields:
-     values['current']  type: str; desc: synchrotron radiation current in milli-amps
-     values['message']  type: str; desc: message from control room
-     values['attention'] type: boolean; desc: False (if no special attention is required)
-                                            True (if attention should be raised to the user)
-
-   optional fields:
-      any number of optional fields can be sent over with this signal by adding them in the
-      values dictionary
-
-      for example:
-         values['lifetime']
-         values['topup_remaining']
-"""
-
+import re
+import math
 import logging
 import gevent
-import time
-import PyTango
-from mxcubecore import HardwareRepository as HWR
-from mxcubecore.BaseHardwareObjects import HardwareObject
+from tango import DeviceProxy
+from mxcubecore.utils.units import sec_to_hour, A_to_mA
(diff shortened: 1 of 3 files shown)


beteva (commented):
Hi @elmjag, nice to have provided tests.  
This is your specific HO, but I was wondering, as it calls tango devices, is there a reason you do not use TangoMachineInfo?

elmjag:
Not sure, this implementation is based on old version of MachineInfo HWO from BioMAX.

I did not knew that TangoMachineInfo existed. I'll take a look if we can use that instead.

elmjag:
I have looked at TangoMachineInfo now. The TangoMachineInfo is designed to read all machine info topics from one specified tango device.

Here at MAXIV the topics are provided by multiple tango devices. We also need to do some parsing for a topic, thus we'll need to have a custom MachineInfo HWO anyway.

I feel that trying to use TangoMachineInfo in our case will make the implementation more complicated compared to how it's done in this PR.

mxcubecore/HardwareObjects/MAXIV/MachInfo.py
```diff
@@ -1,153 +1,109 @@
...
 
-   optional fields:
-      any number of optional fields can be sent over with this signal by adding them in the
-      values dictionary
 
-      for example:
-         values['lifetime']
-         values['topup_remaining']
-"""
+def catch_errs(func):
```
marcus-oscarsson: :), I would have simply called it `catch_errors` but its  matter of taste I guess and it is after all site specific code

mxcubecore/HardwareObjects/MAXIV/MachInfo.py
```diff
@@ -1,153 +1,109 @@
...
+import gevent
+from tango import DeviceProxy
+from mxcubecore.utils.units import sec_to_hour, A_to_mA
+from mxcubecore.HardwareObjects.abstract.AbstractMachineInfo import AbstractMachineInfo
 
-[Description]
-MachInfo hardware objects are used to obtain information from the accelerator
-control system.
+# how often we refresh machine info
+REFRESH_PERIOD_SEC = 30
```
marcus-oscarsson, line 10: You could also easily make it configurable via `get_property`  but perhaps not desired in your case.

marcus-oscarsson (approved):

marcus-oscarsson:
Looks good to me, Ill let you rebase and merge :), you (of-course) do as you like with the comments

mxcubecore/HardwareObjects/MAXIV/MachInfo.py
elmjag: good point, I have changed it to `catch_errors`

mxcubecore/HardwareObjects/MAXIV/MachInfo.py
elmjag: I feel it's not worth the hassle in this case. We have not changed this value on a very long time.

elmjag:
> Looks good to me, Ill let you rebase and merge :), you (of-course) do as you like with the comments

I have re-based, but I don't have the commit bit to do the merge. :disappointed:

marcus-oscarsson:
Oh, I think it should be ok now :)

PR #975, opened 2024-07-25
remove import of devtools in test_collection

elmjag:
Don't import `devtools` module for the test collection queue entry. Mxcube-core does not list devtools package as a dependency.  

Also remove unused import of HardwareRepository.

### DIFF
--- a/mxcubecore/queue_entry/test_collection.py
+++ b/mxcubecore/queue_entry/test_collection.py
@@ -1,9 +1,6 @@
 import json
 
 from pydantic import BaseModel, Field
-from devtools import debug
-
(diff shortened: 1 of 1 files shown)


mxcubecore/queue_entry/test_collection.py
```diff
@@ -66,7 +63,6 @@ def __init__(self, view, data_model: TestCollectionQueueModel):
 
     def execute(self):
         super().execute()
-        debug(self._data_model._task_data)
```
marcus-oscarsson, line 69: Oh, yes. `debug` is on the other hand quite handy so we should probably add that dependency. We can do that in another PR though, because I understand that this would break as it is.

PR #976, opened 2024-07-25
CatsMaintMockup: fix call to super __init__()

elmjag:
The CatsMaintMockup does not inherit Equipment class anymore.

This fixes issue with loading `sample_changer_maintenance` HWO when running mxcubeweb with mocked beamline.


beteva:
Thanks @elmjag

PR #978, opened 2024-07-30
support python 3.11

elmjag:
- mark python 3.11.* as supported for this package.
 - require later version of greenlet package, as version 1.1.3 does not work with python 3.11.
 - require later versions of gevent and gipc packages, to statisfy greenlet 3.* requirmets.

I have tested this with mxcube-web in python 3.11 environment. I tested both with mocked beamline and with the MicroMAX set-ups. I did not see any problems.

Mxcube-web depends on this package. We must first enable python 3.11 support here, before we can do that for mxcube-web.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10"]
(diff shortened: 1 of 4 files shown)


marcus-oscarsson:
Thanks !

elmjag:
Gah, you are to fast! I was still scratching my head wondering why it did not run the unit tests action :)

Anyway, I think I have a solution to that now: https://github.com/mxcube/mxcubecore/pull/979

marcus-oscarsson:
:), sorry

PR #979, opened 2024-07-31
CI: run pytests on package version changes

elmjag:
Trigger pytest if project definition or poetry lock is updated.

Make sure that we test the (potentially) new versions of dependency packages.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,8 @@ name: Pytest
     types: [opened, reopened, synchronize]
     paths:
       - '**.py'
+      - 'pyproject.toml'
+      - 'poetry.lock'
 jobs:
   Python:
     runs-on: ubuntu-latest


fabcor-maxiv:
I am not quite sure we should skip tests. Of course, ideally, it would be nice to not run pointless checks. But from my point of view we are not on top of things. We struggle enough with breaking pipelines and PRs that get merged even though the checks failed. Maybe we should keep things simple unless we have a good reason. Was there a specific reason to have this additional rule?

elmjag:
Not sure I understand your comment @fabcor-maxiv. Are you arguing for removing [current job filter ](https://github.com/mxcube/mxcubecore/blob/develop/.github/workflows/tests.yml#L7) altogether?

To clarify, currently pytest runs only when some of the python (*.py) files have changed.

This PR aims to _add_ `pyproject.toml` and `poetry.lock` to the set of files that trigger pytests. The idea that if we change the dependencies for the project, we should run pytests.

fabcor-maxiv (approved):
> Are you arguing for removing [current job filter ](https://github.com/mxcube/mxcubecore/blob/develop/.github/workflows/tests.yml#L7) altogether?

Yes. Filters are good of course... until we realize that we filter too much as was the case here.

> To clarify, currently pytest runs only when some of the python (*.py) files have changed.
> 
> This PR aims to _add_ `pyproject.toml` and `poetry.lock` to the set of files that trigger pytests. The idea that if we change the dependencies for the project, we should run pytests.

Yes, I understand that. Until we realize there are some more cases where we would have wanted the jobs to run but they didn't because we excluded too much in our filters because there is something we had not considered.

I am not sure we win much by filtering and skipping jobs in some particular cases.

But anyway, I don't mind much either way. We can keep the filters, it's fine with me. I guess my suggestion is more like "food for thought".

marcus-oscarsson:
I guess there is a risk that we don't run the tests when we should but it is also quite time consuming and frustrating to wait for the unnecessary tests to finish. It's somewhat less of a concern for this repository because it only contain python code and one could argue that PyTest should always run. 

However, without filters, the tests would run even if we simply update documentation. The current change seems reasonable to me and I suggest that we stick with it for the time being. We could perhaps exclude the `[docs, doc, depreciated]` folders as an alternative solution.

Issue #977, opened 2024-07-30
Moving XML nested properties to YAML


rhfogh:
@elmjag noted that the xml_yaml_conversion branch did not work for nested properties, which were simply not found in get_properties.
Looking closer there is a clear bug. The properties for xml-configured objects in the new shared XML/YAML code are set by getting the properties from HardwareObjectNode.get_properties(). The problem is that nested XML elements, more precisely XML elements that contain other elements are not treated as properties, but as contained objects, and so are not in get_properties. 

Looking through the startup (Qt version, mock mode) there are two different type of case:

These are:

1) Properties that are simple dictionaries. 

- Session.file_info
- Slits.gapH
- Slits.gapV
- Detector.beam 
- SampleChangerMockup.test_sample_names
- Gphl-CollectEmulator.simcal_parameters
- Gphl-CollectEmulator.simcal_options
- Gphl-CollectEmulator.environment_variables

These can be handled very simply by converting teh data to a single property that is a dictionary, and I am starting to code this.

2) Properties that are lists of dictionaries. These are:

- EnergyScan.mad
- EnergyScan.elements
- OfflineProcessing.programs

The structure of these can be seen in this example:
```

<programs>
    <program>
        <executable>path_to_edna_autoproc.sh</executable>
        <event>after</event>
    </program>
    <program>
        <executable>path_to_edna_thumb.sh</executable>
        <event>image</event>
    </program>
</programs>

```

Writing temporary code that read this kind of data structure would be a lot more messy, particularly if you wanted it to be able to deal with arbitrary  nested data structures (that are likely not present in the current MXCuBE). 

I would propose that we do *not* fix the problem in handling this kind of nested data structure, but simply make the necessary changes to the code as we upgrade the relevant config files. Arguments in favour would be 1) that this is a very limited part of the code and so should be tractable. 2) that there is really no obvious mapping, since a dictionary containing a single list of dictionaries is not the kind of data structure you would really want. 

What do people think of this?

marcus-oscarsson:
There is this `xmltodict` project: https://pypi.org/project/xmltodict/, Ive used it abit in the past and it seems to work quite well. Perhaps it can be of help ?

rhfogh:
On second thoughts I shall make a PR where the programs example above is translated into property as

obj.programs = [
  { 
    "executable": "path_to_edna_autoproc.sh",
    "event": "after",
  },
  { 
    "executable": "path_to_edna_thumb.sh",
    "event": "image",
[...]

PR #980, opened 2024-08-02
reshuffle code for populating _config attribute for XML configs

elmjag:
When loading HWOBJ with XML config file, populate its `_config` attribute _before_ calling `init()` method.

This way, the `self.config` attribute can be used inside `init()` method for both YAML and XML configured hardware objects.

Before this change, `self.config` was only available for YAML configured objects. For XML configured objects, the `self.config` was `None`.

With this change, it is possible to write hardware objects that can be configured using both YAML and XML configuration files. For example, here is a version of Session HWO, that supports both YAML and XML config files:

https://github.com/elmjag/mxcubecore/blob/1a58568dd8244db208deb7aa4f6146c35ab128b6/mxcubecore/HardwareObjects/Session.py

Solves: https://github.com/mxcube/mxcubecore/issues/977

### DIFF
--- a/mxcubecore/HardwareRepository.py
+++ b/mxcubecore/HardwareRepository.py
@@ -277,20 +277,11 @@ def _attach_xml_objects(yaml_export_directory: Optional[Path], container, hwobj,
     hwobj._name = role
     container._roles.append(role)
 
-    hwobj._config = hwobj.HOConfig(**hwobj.get_properties())
     setattr(container, role, hwobj)
     objects_by_role = hwobj._objects_by_role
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
Excellent!

It should not break anything, but if it does we can fix it when it happens.

marcus-oscarsson:
I guess you noticed that the tests are failing ?

elmjag:
> I guess you noticed that the tests are failing ?

Yes, the tests are broken on `xml_yaml_conversion` branch even without this commit. I kind of hope I can take a look at them in not to distant future.

PR #981, opened 2024-08-07
Checkbeam in BeamlineActions

beteva:
- Define GenericNState to overload get_value and _set_value when no hardware involved. To be used by BeamlineActions.
- AbstractNState default_value should be an Enum. Impossible to set it as such in the config file, so do it at init().
- BeamlineActions typo fixed.


marcus-oscarsson:
Thanks @beteva

PR #983, opened 2024-08-13
attach objects by role for yaml configs

elmjag:
When loading HWOBJ with YAML config file, attach all it's sub-HWOBJs as attributes. Use sub-HWOBJs role (key name in the config file) as attribute name when attaching.

That is, perform same operation as when loading HWOBJ with XML config file.

This allows to replace the `parent.get_object_by_role(role)` calls with `parent.role` expression. As far as I understand, this the API we are aiming for.

### DIFF
--- a/mxcubecore/HardwareRepository.py
+++ b/mxcubecore/HardwareRepository.py
@@ -194,6 +194,7 @@ def load_from_yaml(
                     _table=_table,
                 )
                 _instance.hardware_objects[f"/{hwobj.load_name}"] = hwobj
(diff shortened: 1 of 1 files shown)


elmjag:
Changing this PR to draft.

I realized that is suppose to work without this patch. I'll need to investigate this more.

elmjag:
Yep, seems to work fine without this change.

Objects already attached here:
https://github.com/mxcube/mxcubecore/blob/3c8f491fb62c545dce2f53906a380e2a4edd0aed/mxcubecore/HardwareRepository.py#L223

PR #985, opened 2024-08-13
make Session HWOBJ work with both YAML and XML configure files

elmjag:
Update Session HWOBJ to access its configuration using 'self.config' attribute.

Drops unused 'template' and 'suffix' attributes.

Converts 'synchrotron_name', 'beamline_name' and 'endstation_name' to be proxy attributes for fields in 'self.config', this way there no need copy values.

Replace self["foo"] calls, they are not supported when YAML config file is used.

With these changes it's possible to configure this HWOBJ using both XML and YAML configure files.

### DIFF
--- a/mxcubecore/HardwareObjects/Session.py
+++ b/mxcubecore/HardwareObjects/Session.py
@@ -22,23 +22,15 @@ class Session(HardwareObject):
     def __init__(self, name):
         HardwareObject.__init__(self, name)
 
-        self.synchrotron_name = None
-        self.beamline_name = None
-
         self.session_id = None
         self.proposal_code = None
         self.proposal_number = None
         self.proposal_id = None
         self.in_house_users = []
-        self.endstation_name = None
         self.session_start_date = None
         self.user_group = ""
         self.email_extension = None
-        self.template = None
-
         self.default_precision = 5
-        self.suffix = None
-
         self.base_directory = None
         self.base_process_directory = None
         self.base_archive_directory = None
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/Session.py
```diff
@@ -22,71 +22,55 @@ class Session(HardwareObject):
...
 
-        base_process_directory = self["file_info"].get_property(
-            "processed_data_base_directory"
-        )
-
-        base_archive_directory = self["file_info"].get_property(
-            "archive_base_directory"
-        )
+    def init(self):
+        file_info = self.config.file_info
```
beteva: @elmjag, are we sure that all these properties are defined? Would it raise  some exception, if not the case?

elmjag: No, there are no checks if these properties exist in the configure file. If they are missing it raises `AttributeError` exception.

My understanding is that the plan to deal with this using Pydantic models in the future. Like each HWOBJ will have a pydantic model, which defines required configuration properties. The Pydantic machinery will do the validation of the configuration file.

However, we decided to do this as a step 2, after we are done with XML to YAML migration. So for now, it will be somewhat user hostile, where we assume that the user provides a correct configuration files to HWOBJs.

elmjag: I have changed the code to treat `inhouse_users` as an optional property. I did run into one instance of session configuration file that did not specify `inhouse_users` property.

PR #988, opened 2024-08-16
XRF centred position

meguiraun:
We have the need to know the centred positions for the XRF and Energy Scans. By default, mxcubecore does not do anything with that although there are some traces of such info in the code.

This MR adds the centred position in the XRF hwobj so our custom hwobj has the info to be used in the method `_execute_spectrum`. Mxcubeweb already loosely links XRF task with a centering point, so this MR closes the gap at hwobj level
 
Update: same for Energy scan

Small addition for mxcubeweb https://github.com/mxcube/mxcubeweb/pull/1349

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractEnergyScan.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractEnergyScan.py
@@ -13,6 +13,7 @@ class AbstractEnergyScan(object):
         self.data_collect_task = None
         self._egyscan_task = None
         self.scanning = False
+        self.cpos = None
 
     def open_safety_shutter(self, timeout):
         """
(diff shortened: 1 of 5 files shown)


marcus-oscarsson:
Its nice if we can keep the additions to the queue layer of mxcube web as small as possible and push as much as possible back to mxcubecore. For instance I guess the only line(s) you would need to add is
```
_shape = params["shape"]
if _shape:
    model.shape = _shape
```

Perhaps even `model.shape = params["shape"]` would suffice and then you handle this part:

```
point = HWR.beamline.sample_view.get_shape(_shape)
model.centred_position = point.get_centred_position()
```

In mxcubecore, what do you think ?

It's otherwise sensible to associate positions with XRF and energy scans and the suggested solution above would indeed work.

meguiraun:
thanks, yup, I think that can work. Will come back!

beteva (approved):
Ok as it is now.

PR #989, opened 2024-08-16
Allow `.yaml` extension for YAML config files

fabcor-maxiv:
The recommended file extension for YAML files is `.yaml`. Both `.yaml` and `.yml` are valid extensions for YAML files.

### DIFF
--- a/mxcubecore/HardwareRepository.py
+++ b/mxcubecore/HardwareRepository.py
@@ -181,7 +181,7 @@ def load_from_yaml(configuration_file, role, _container=None, _table=None):
             msg0 = "Done loading contents"
         for role1, config_file in _objects.items():
             fname, fext = os.path.splitext(config_file)
-            if fext == ".yml":
+            if fext in (".yaml", ".yml"):
                 load_from_yaml(
                     config_file, role=role1, _container=result, _table=_table
                 )


fabcor-maxiv:
> The official recommended filename extension for YAML files has been `.yaml` since 2006.

-- <https://en.wikipedia.org/wiki/YAML>

> FAQ
> 1. Is there an official extension for YAML files?
>
>    Please use ".yaml" when possible.

-- https://yaml.org/faq.html

fabcor-maxiv:
I have not tested this code, I assume it is a trivial change, but maybe I am completely wrong.

fabcor-maxiv:
There is also this, which is a bit more complex to handle if we want to allow both `.yaml` and `.yml`: <https://github.com/mxcube/mxcubecore/blob/1dcd5a49ce2e99597c44a53a58cdc9d0e2fa4804/mxcubecore/HardwareRepository.py#L67>

rhfogh:
If the official extension is '.yaml', should we not just standardise on that? My personal taste is for '.yml' but if there is a standard we ought to use it.

fabcor-maxiv:
> If the official extension is '.yaml', should we not just standardise on that? My personal taste is for '.yml' but if there is a standard we ought to use it.

My recommendation would be:

* wherever possible accept both `.yaml` and`.yml`
  * which in our case it seems like it would be easy for everything but the main `beamline_config` file
  * and I guess it would be a backwards compatible change, so it could be done directly on our main `develop` branch
* for `beamline_config`
  * it seems like it would be a bit more involved change to accept both `.yaml` and `.yml` in order to have it backwards-compatible, not very hard but more than a one-line change
  * if we opt to accept only `beamline_config.yaml` in a non backwards-compatible manner, then it should be done not on the main `develop` branch but on the "migration to yaml"  branch

marcus-oscarsson:
I think it would be easy enough to keep the backward compatibility without adding to much complexity. I think that would worth a try if it does not turn out too complicated or time consuming.

PR #990, opened 2024-08-20
ESRF Implementation of new beam size and definer handling

beteva:
ESRFBeamDefiner,
ID30A3 and ID232 BeamDefiner inheriting from ESRFBeamDefiner


marcus-oscarsson:
@beteva I guess you noticed but there is a formatting issue in: ID30A3BeamDefiner.py

PR #992, opened 2024-08-21
docs: XML to YAML config migration document

elmjag:
Adds a documentation section 'YAML configuration migration', with details on how to migrate to YAML configration.

### DIFF
--- a/docs/source/dev/yaml_conf_migration.md
+++ b/docs/source/dev/yaml_conf_migration.md
@@ -1,0 +1,188 @@
+# YAML configuration migration
+
+Historically, MXCuBE used XML files for configuring hardware objects.
+Now it's possible to use YAML files instead of XML files.
+Currently, it is possible to use both XML and YAML files within the same beamline configuration.
+You can use XML for some hardware objects and YAML for others.
+This allows for a gradual migration to YAML configuration.
+
+When using a YAML file, you may need to update the Python code of your hardware object.
+Some of the methods for accessing hardware object configuration are not supported when using YAML.
+
+This document provides some guidance on needed modification when migrating from XML to YAML.
+
+## `beamline_config.yml` format changes
+
+The format of the entry point configuration file `beamline_config.yml` have changed.
+The old format had the following style:
+
+```yaml
+_initialise_class:
+  class: mxcubecore.HardwareObjects.Beamline.Beamline
+_objects:
+  !!omap
+  - session: session.xml
+  - data_publisher: data_publisher.xml
+  - machine_info: machine_info.xml
+
+# Non-object attributes:
+advanced_methods:
+  - MeshScan
+  - XrayCentering
+tunable_wavelength: true
+```
+
+This format is not supported anymore.
+The new format, corresponding to the example above, looks like this:
+
+```yaml
+class: mxcubecore.HardwareObjects.Beamline.Beamline
+objects:
+  !!omap
+  - session: session.xml
+  - data_publisher: data_publisher.xml
+  - machine_info: machine_info.xml
+configuration:
+  advanced_methods:
+    - MeshScan
+    - XrayCentering
+  tunable_wavelength: true
+```
+
+The changes to the format are outlined below.
+
+### `_initialise_class`
+
+The `_initialise_class` dictionary has been replaced by a `class` key-value pair.
+
+
+### `_objects`
+
+The `_objects` dictionary has been renamed to `objects`.
+Otherwise, the format of the dictionary is the same as before.
+
+### New `configuration` dictionary
+
+All the general configuration parameters of the beamline have moved inside the new `configuration` dictionary.
+
+## Converting XML files to YAML
+
+Note that MXCuBE-Web provides some support for automatically converting XML configuration files to YAML.
+See the [YAML configuration migration](https://mxcubeweb.readthedocs.io/en/latest/dev/yaml_conf_migration.html)
+section in MXCuBE-Web documentation for details.
+For details on the format of the YAML configurations file, see [Yaml-configured objects](configuration_files.md#yaml-configured-objects).
+
+This section provides an example of equivalent configuration in XML and YAML formats.
+Given the following XML configuration file:
+
+```xml
+<object class="Shanxi">
+ <!-- configuration -->
+ <simple_prop>prop_val</simple_prop>
+ <nested>
+   <child_a>uz</child_a>
+   <child_b>ve</child_b>
+ </nested>
+
+ <!-- child objects -->
+ <object role="session" href="/session"/>
+ <object role="lims" href="/lims"/>
+</object>
+```
+
+Gives the following equivalent configuration in YAML:
+
+```yaml
+class: Shanxi.Shanxi
+configuration:
+  simple_prop: prop_val
+  nested:
+    child_a: uz
+    child_b: ve
+objects:
+  session: session.yaml
+  lims: lims.yaml
+```
+
+The hardware object class specified by `<object class="Shanxi">` becomes the `class: Shanxi.Shanxi` key-value.
+In the YAML format, the fully qualified class name must be specified.
+
+All the configuration property XML tags becomes entries in the YAML's `configuration` dictionary.
+
+Each `<object/>` tags becomes an entry in the YAML's `objects` dictionary.
+The tag's `role` attribute is used as the entry's key name.
+The tag's `href` attribute is converted to a config file name and specified as the entry's value.
+
+## `self["prop_name"]` expressions not supported
+
+Using index expressions to access configuration properties is no longer supported.
+Use object's `config` attribute or `get_property()` methods.
+
+For example, following old-style code:
+
+```python
+def init(self):
+    foo = self["foo"]
+```
+
+Needs to be converted to one of the following styles:
+
+```python
+def init(self):
+    # use 'config' attribute to access configuration property
+    foo = self.config.foo
+```
+
+```python
+def init(self):
+    # use 'get_property()' method to access configuration property
+    foo = self.get_property("foo")
+```
+
+## `@property` annotation for child objects
+
+Using `@property` annotated attribute to provide access to a child object is not supported.
+
+Consider this old-style code:
+
+```python
+class Shanxi(HardwareObject):
+    def init(self):
+        self._session = self.get_object_by_role("session")
+
+    @property
+    def session(self):
+        return self._session
+```
+
+This code provides access to its child hardware object `session` via the annotated `session` attribute.
+This style does not work anymore.
+Remove the annotated `session` attribute and assignment to `self._session` attribute.
+Use the following configuration file.
+
+```yaml
+class: Shanxi.Shanxi
+objects:
+  session: session.yml
+```
+
+During the initialization of the `Shanxi` object,
+the child object will be automatically assigned to the `session` attribute.
+See [Accessing child objects](configuration_files.md#accessing-child-objects) for more details.
+
+## `name` parameter in `__init__` method
+
+When a hardware object is loaded using a YAML configuration file, it is created with the following code:
+
+```python
+ClassName(name="role")
+```
+
+Thus, the hardware object's `__init__` method must accept the `name` parameter.
+Below is an example that works with YAML configuration file:
+
+```python
+class Shanxi(HardwareObject):
+    def __init__(self, name):
+        super().__init__(name)
+```


fabcor-maxiv:
Preview here: https://mxcubecore--992.org.readthedocs.build/en/992/dev/yaml_conf_migration.html

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
+# YAML configuration migration
+
+Historically, MXCuBE used XML files for configuring hardware objects (HWO).
+Now it's possible to use YAML files instead of XML files.
+Currently, it is possible to use both XML and YAML files within the same beamline configuration.
+You can use XML for some hardware objects and YAML for others.
+This allows for a gradual migration to YAML configuration.
+
+When using a YAML file, you may need to update the python code of your hardware object.
```
fabcor-maxiv:
```suggestion
When using a YAML file, you may need to update the Python code of your hardware object.
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+Some of the methods for accessing hardware object configuration are not supported when using YAML.
+
+This document provides some guidance on needed modification when migrating from XML to YAML.
+
+## `beamline_config.yml` format changes
+
+The format of the entry point configuration file `beamline_config.yml` have changed.
+The old format had the following style:
+
+```
```
fabcor-maxiv: This will make we have the correct syntax highlighting

```suggestion
```yaml
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+_initialise_class:
+  class: mxcubecore.HardwareObjects.Beamline.Beamline
+_objects:
+  !!omap
+  - session: session.xml
+  - data_publisher: data_publisher.xml
+  - machine_info: machine_info.xml
+  ...
```
fabcor-maxiv: Let's not put three dots like this. They have a particular meaning in YAML documents. They mark the end of the document. <https://yaml.org/spec/1.1/#document%20boundary%20marker/>

```suggestion
  # ...
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+# Non-object attributes:
+advanced_methods:
+  - MeshScan
+  - XrayCentering
+
+tunable_wavelength: true
+...
```
fabcor-maxiv:
```suggestion
# ...
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+```
+class:
+  mxcubecore.HardwareObjects.Beamline.Beamline
+objects:
+  !!omap
+  # Hardware:
+  - session: session.xml
+  - data_publisher: data_publisher.xml
+  - machine_info: machine_info.xml
+  ...
```
fabcor-maxiv:
```suggestion
  # ...
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+configuration:
+  advanced_methods:
+    - MeshScan
+    - XrayCentering
+
+  tunable_wavelength: true
+  ...
```
fabcor-maxiv:
```suggestion
  # ...
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+## `beamline_config.yml` format changes
+
+The format of the entry point configuration file `beamline_config.yml` have changed.
+The old format had the following style:
+
+```
+_initialise_class:
+  class: mxcubecore.HardwareObjects.Beamline.Beamline
+_objects:
+  !!omap
```
fabcor-maxiv, line 23: No idea why we need the `!!omap`. Do we need it? I get that it is unrelated to the YAML configuration, I am curious.

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+
+  tunable_wavelength: true
+  ...
+```
+
+The changes to the format are outlined below.
+
+### `_initialise_class`
+
+The `_initialise_class` dictionary have been replaced by a `class` key-value pair.
```
fabcor-maxiv: Typo `s/have/has`.

But also I do not see any "key-value pair" in the example above.

```suggestion
The `_initialise_class` dictionary has been replaced by a `class` key-value pair.
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+tunable_wavelength: true
+...
+```
+
+This format is not supported anymore.
+The new format, corresponding to the example above, looks like this:
+
+```
+class:
+  mxcubecore.HardwareObjects.Beamline.Beamline
```
fabcor-maxiv: This seems strange... Is it even valid YAML? See also below...

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+The changes to the format are outlined below.
+
+### `_initialise_class`
+
+The `_initialise_class` dictionary have been replaced by a `class` key-value pair.
+
+
+### `_objects`
+
+The `_objects` dictionary have been renamed to `objects`.
```
fabcor-maxiv:
```suggestion
The `_objects` dictionary has been renamed to `objects`.
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+Otherwise, the format of the dictionary is the same as before.
+
+### new `configuration` dictionary
```
fabcor-maxiv:
```suggestion
### New `configuration` dictionary
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+
+All the general configuration parameters of the beamline have moved inside the new `configuration` dictionary.
+
+## Converting XML files to YAML
+
+Note that MXCuBE-Web provides some support for automatically converting XML configuration files to YAML.
+See the [YAML configuration migration](https://mxcubeweb.readthedocs.io/en/latest/dev/yaml_conf_migration.html)
```
fabcor-maxiv, line 71: I guess the document does not exist yet... and now there is a link to the current document (link to itself).

Anyway we should use better notation for the link, some relative link. I never remember what is the correct notation, maybe this:

```suggestion
See the [YAML configuration migration](something.md)
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+section in MXCuBE-Web documentation for details.
+For details on the format of the YAML configurations file, see [Yaml-configured objects](configuration_files.md#yaml-configured-objects).
+
+This section provides an example of equivalent configuration in XML and YAML formats.
+Given the following XML configuration file:
+
+```
```
fabcor-maxiv:
```suggestion
```xml
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+   <child_b>ve</child_b>
+ </nested>
+
+ <!-- child objects -->
+ <object role="session" href="/session"/>
+ <object role="lims" href="/lims"/>
+</object>
+```
+
+Gives the following equivalent configuration in YAML.
```
fabcor-maxiv:
```suggestion
Gives the following equivalent configuration in YAML:
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+```
+class: Shanxi.Shanxi
+configuration:
+  simple_prop: prop_val
+  nested:
+    child_a: uz
+    child_b: ve
+objects:
+  session: session.yml
+  lims: lims.yml
```
fabcor-maxiv:
```suggestion
  session: session.yaml
  lims: lims.yaml
```

Ha! :D

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+The tag's `href` attribute is converted to a config file name and specified as the entry's value.
+
+## `self["prop_name"]` expressions not supported
+
+Using index expressions to access configuration properties is no longer supported.
+Use object's `config` attribute or `get_property()` methods.
+
+For example, following old-style code:
+
+```
```
fabcor-maxiv: In this case the Python syntax seems to be automatically detected, but better safe...

```suggestion
```python
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+Needs to be converted to following style:
+
+```
+def init(self):
+    # 'config' attribute option
+    foo = self.config.foo
+
+    # 'get_property()' method option
+    foo = self.get_property("foo")
+```
```
fabcor-maxiv:
```suggestion
Needs to be converted to one of the following styles:

    ```python
    def init(self):
        # 'config' attribute option
        foo = self.config.foo
     ```

    ```python
    def init(self):
        # 'get_property()' method option
        foo = self.get_property("foo")
     ```

```
Well, we don't need the extra indentation, but otherwise it messes up the formatting of he comment here.

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+
+## `@property` annotation for child objects
+
+Using `@property` annotated attribute to provide access to a child object is not supported.
+
+Consider this old-style code:
+
+```
```
fabcor-maxiv:
```suggestion
```python
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+    def session(self):
+        return self._session
+```
+
+This code provides access to its child hardware object `session` via the annotated `session` attribute.
+This style does not work anymore.
+Remove the annotated `session` attribute and assignment to `self._session` attribute.
+Use the following configuration file.
+
+```
```
fabcor-maxiv:
```suggestion
```yaml
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+During the initialization of the `Shanxi` object,
+the child object will be automatically assigned to the `session` attribute.
+See [Accessing child objects](configuration_files.md#accessing-child-objects) for more details.
+
+## `name` parameter in `__init__` method
+
+When a hardware object is loaded using a YAML configuration file, it is created with the following code:
+
+```
+    ClassName(name="role")
```
fabcor-maxiv:
```suggestion
ClassName(name="role")
```

docs/source/dev/yaml_conf_migration.md
```diff
@@ -0,0 +1,192 @@
...
+```
+
+Thus, the hardware object's `__init__` method must accept the `name` parameter.
+Below is an example that works with YAML configuration file:
+
+```
```
fabcor-maxiv:
```suggestion
```yaml
```

fabcor-maxiv (changes requested):
Good stuff

docs/source/dev/yaml_conf_migration.md
elmjag: I think this syntax to show 'omitted parts' is just confusing. I have removed these lines. I think it's will be obvious for readers that's it's a truncated example.

fabcor-maxiv: Sometimes I write `# [...]` I do not know if it makes it clearer that it is an ellipsis, that stuff is omitted.

fabcor-maxiv (approved):

docs/source/dev/yaml_conf_migration.md
rhfogh: We need it to ensure that the order of objects is maintained and reproducible, so we can control the loading order. `omap` gives an ordered dictionary rather than just a plain dictionary. We need the objects ordered, because the initialisation of certain objects may depend on other objects having been in initialised already.

docs/source/dev/yaml_conf_migration.md
rhfogh: Maybe it should just say:

class:  mxcubecore.HardwareObjects.Beamline.Beamline

on a single line?

rhfogh (approved):
Very nice!. Thanks.

docs/source/dev/yaml_conf_migration.md
fabcor-maxiv: OK, so we write it as a list in the Yaml document (with `!!omap`) and the Python Yaml parser (`ruamel.yaml`) transforms it into a `dict` (which is ordered by default on recent Python versions or `OrderedDict` on older Python versions).

rhfogh: I would not call it a  list. This is not a list of dictionaries (which it would be if you used the same syntax but removed the '!omap'), it is an ordered mapping. But apart from that we agree.

docs/source/dev/yaml_conf_migration.md
elmjag: This is link to MXCube-**web** documentation. This URL will be valid once we merge the `xml_yaml_conversation` branch there.

Here is documentation it refers to:
https://github.com/mxcube/mxcubeweb/blob/xml_yaml_conversion/docs/source/dev/yaml_conf_migration.md

docs/source/dev/yaml_conf_migration.md
elmjag: added!

docs/source/dev/yaml_conf_migration.md
elmjag: added

docs/source/dev/yaml_conf_migration.md
fabcor-maxiv: Ah... right... there is a way to make it cleaner still, but I do not remember, so I guess it is fine for now : )

docs/source/dev/yaml_conf_migration.md
elmjag: I'm not a YAML lawyer, so I have no idea if it's valid YAML or not, but...

Interesting enough, both version works! But, let's go with the one-line version. I like the compactness of that style.

fabcor-maxiv (approved):

PR #996, opened 2024-08-26
'emulate' get_object_by_role() access for YAML HWOBJs

elmjag:
Make it possible to use `get_object_by_role()` method on HWOBJs loaded with YAML config files, at least in some situations.

This hack will make it easier to migrate to using YAML config files, as not all code invoking get_object_by_role() needs to be updated stright away.


rhfogh (approved):
Good idea.

PR #999, opened 2024-08-27
NState: support non-enum types of values

elmjag:
Make it passible to use the 'bare' types of values for AbstractNState derived hardware objects.

For example if an AbstractNState have following valid values:

  self.VALUES = Enum("ValueEnum", {"LEVEL1": 1, "LEVEL2": 2})

This patch makes "LEVEL1" and "LEVEL2" values valid, in addition to self.VALUES.LEVEL1 and self.VALUES.LEVEL2.

This is needed when getting the hardware objects values REST API from the front-end. For example when using BeamDefinerMockup HWOBJ.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractNState.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractNState.py
@@ -61,6 +61,14 @@ class AbstractNState(AbstractActuator):
         Returns:
             (bool): True if within the values.
         """
+        if not isinstance(value, self.VALUES):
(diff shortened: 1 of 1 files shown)


elmjag:
Putting this to draft for now, not sure anymore if this change is needed.

See this related issue: https://github.com/mxcube/mxcubeweb/issues/1370

elmjag:
These two PRs:

https://github.com/mxcube/mxcubeweb/pull/1379
https://github.com/mxcube/mxcubecore/pull/1006

seems to solve this issue.

PR #1001, opened 2024-08-27
Prevent re-loading same YAML config file twice

elmjag:
As discussed here: https://github.com/mxcube/mxcubecore/pull/997#issuecomment-2311657405

When loading hardware objects from YAML configuration file, check if that file have been loaded earlier. Raise an error if we detect that same file is being reloaded.
    
The aim is to enforce strict tree-like structure of hardware objects. This to avoid confusion of the situations when same hardware object is attached at different points.


rhfogh (approved):
Excellent!
Thanks.

PR #1004, opened 2024-08-28
allow update grid result data

meguiraun:
so we can append new data

related to https://github.com/mxcube/mxcubeweb/pull/1378

### DIFF
--- a/mxcubecore/HardwareObjects/SampleView.py
+++ b/mxcubecore/HardwareObjects/SampleView.py
@@ -330,10 +330,26 @@ class SampleView(AbstractSampleView):
         return grid
 
     def set_grid_data(self, sid, result_data, data_file_path):
+        """
+        Sets grid rsult data for a shape with the specified id.
+
+        Args:
+            sid (str): The id of the shape to set grid data for.
+            result_data: The result data to set for the shape. Either a base64 encoded string for PNG/image
+            or a dictionary for RGB (keys are cell number and value RGBa list). Data is only updated if result is RGB based
+            data_file_path (str): The path to the data file associated with the result data.
+
+        Returns:
+            None
+
+        Raises:
+            AttributeError: If no shape with the specified id exists.
+        """
+
         shape = self.get_shape(sid)
 
         if shape:
-            if shape.result:
+            if shape.result and type(shape.result) == dict:
                 # append data
                 shape.result.update(result_data)
             else:
@@ -526,7 +542,9 @@ class Grid(Shape):
         self.num_cols = -1
         self.num_rows = -1
         self.selected = False
-        self.result = {}
+        # result is a base64 encoded string for PNG/image heatmap results
+        # or a dictionary (for RGB number based results)
+        self.result = None
         self.pixels_per_mm = [1, 1]
         self.beam_pos = [1, 1]
         self.beam_width = 0
--- a/mxcubecore/HardwareObjects/SampleView.py
+++ b/mxcubecore/HardwareObjects/SampleView.py
@@ -333,8 +333,12 @@ class SampleView(AbstractSampleView):
         shape = self.get_shape(sid)
 
         if shape:
-            shape.set_result(result_data)
-            shape.result_data_path = data_file_path
+            if shape.result:
+                # append data
+                shape.result.update(result_data)
+            else:
+                shape.set_result(result_data)
+                shape.result_data_path = data_file_path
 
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -333,8 +333,12 @@ def set_grid_data(self, sid, result_data, data_file_path):
         shape = self.get_shape(sid)
 
         if shape:
-            shape.set_result(result_data)
-            shape.result_data_path = data_file_path
+            if shape.result:
+                # append data
+                shape.result.update(result_data)
```
marcus-oscarsson, line 354: Whats the structure of `result_data` ?

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -522,7 +526,7 @@ def __init__(self, mpos_list, screen_coord):
         self.num_cols = -1
         self.num_rows = -1
         self.selected = False
-        self.result = []
+        self.result = {}
```
marcus-oscarsson: Why did you decide to change to `dict` ?

mxcubecore/HardwareObjects/SampleView.py
meguiraun: like this for RGB:
```
{"result": 
   {
    '44': [255, 255, 255, 0.0], 
    '61': [255, 255, 255, 0.0], 
    '56': [255, 255, 255, 0.0],
    '10': [255, 255, 255, 0.0]},
"data_file": "a/path/to/file"}
```

mxcubecore/HardwareObjects/SampleView.py
meguiraun: so it can easily be updated, it seems it does no really matter as we were just assigning data to it, or?

mxcubecore/HardwareObjects/SampleView.py
marcus-oscarsson: I see, It would probably be nice to document. 

We need to be able to pass a list or a string and that still seems supported, so no problem. Just keep in mind that iterating over a dictionary does not guarantee order, you thus have to use the keys to access, not sure how that is dealt with when the heatmap is displayed ?

mxcubecore/HardwareObjects/SampleView.py
marcus-oscarsson: Yes, your right. It would perhaps as mentioned above just be nice to document that its either a base64 encoded string for PNG/image results or a dictionary (for number based results).

mxcubecore/HardwareObjects/SampleView.py
fabcor-maxiv: > Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

-- https://docs.python.org/3/library/stdtypes.html#dict

Is that good enough for your use case here? I do not know.

axelboc: Just a thought for the future that maybe the `Grid` shape class should be subclassed? This might help document and deal with different result formats more safely.

marcus-oscarsson: Hm, yes but what that means if you update a key ? I would not expect to be able to iterate over a dictionary/hash-map in a certain order (its not why I would choose such a structure). Its perhaps not a discussion for this PR. 

Also keep in mind that we pass this structure as JSON to the browser and that is where the results are displayed

meguiraun: Is the UI the one accessing the elements directly if RGB is mode is set

mxcubecore/HardwareObjects/SampleView.py
meguiraun: Yes, I will add some documentation. Very good point

marcus-oscarsson: Let me know if you would like me to merge this now and you follow up with some documentation or if you add documentation later ?

meguiraun: argsss, I guess better wait so I force myself and add some doc on this MR

marcus-oscarsson: Is completely up to you :)

meguiraun: added some documentation

PR #1008, opened 2024-09-02
hwobj-ify abstract energy scan

meguiraun:
small additions to make this object a hardware object


beteva:
Hi @meguiraun I was planning to clean the Energy scan the same way as the XRF, but you were faster 😃

beteva (commented):
Thanks

PR #1020, opened 2024-09-17
Added compatability between pydantic 1 and 2

marcus-oscarsson:
Adding compatibility between Pydantic v1 and v2. We are using [spectree](https://github.com/0b01001001/spectree?tab=readme-ov-file) in `mxcubeweb` which for the moment only supports version 1.x of Pydantic.

See https://github.com/mxcube/mxcubeweb/pull/1398 for more information


pyproject.toml
```diff
@@ -35,7 +35,7 @@ lxml = "^4.9.1"
 matplotlib = "^3.6.2"
 numpy = "^1.23.5"
 Pillow = "^9.3.0"
-pydantic = "^1.10.2"
```
fabcor-maxiv, line 38: Hard pin here seems suspicious to me. I do not see much reason to have a hard pin here, we have the lock file for this.

fabcor-maxiv (changes requested):

fabcor-maxiv (approved):
Seems fine

PR #1021, opened 2024-09-17
actions: notify teams upon new discussion in category

walesch-yan:
This PR aims at making the communication in Work Groups easier. Instead of having one discussion, the idea was to use Github's discussion categories, so that different discussions can be grouped under one topic (i.e. Cybersecurity, Automation, ...). There is an issue however regarding Github's notification system, as you cannot subscribe to a category in the discussion session. 
To fix this issue, this PR adds an action that triggers every time there is a new discussion, it then checks for the category in which the discussion has started and notifies every member in the corresponding Team. By adding the "@team" as a comment in the discussion.
For it to work the category name must be the same as the team name (otherwise no notification will be added). Additionally, a secret should be added as "DISCUSSIONS_NOTIFICATIONS" which allows reading of teams in the organization and the writing of comments in the discussions section.

### DIFF
--- a/.github/workflows/discussions.yml
+++ b/.github/workflows/discussions.yml
@@ -1,0 +1,90 @@
+---
+# This workflow notifies working groups about new discussions in their corresponding
+# categories, by mentioning the Team in a first comment.
+# It is necessary since GitHub doesn't support subsriptions to discussion categories
+# see related feature request: https://github.com/orgs/community/discussions/3951
+# If GitHub implements this feature, this workflow becomes obsolete.
+
+# To work, this workflow requires a GitHub App installed on the repository with the
+# following permissions:
+#   - Write and Read of repository Discussions
+#   - Read of Organization members (needed for reading the teams)
+
+# !! You need to make sure that the app's id and a generated private key are saved
+# as corresponding organization secrets !!
+
+name: Notify Team on New Discussion
+"on":
+  discussion:
+    types: [created]
+
+jobs:
+  notify_team:
+    runs-on: ubuntu-latest
+    steps:
+      # First generate the app token for authentication
+      - name: Get GitHub App token
+        id: github-app-token
+        uses: tibdex/github-app-token@v1
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - name: Notify Team
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.github-app-token.outputs.token }}
+          script: |
+            const discussion = context.payload.discussion;
+            const category = discussion.category.name;
+
+            const query = `
+              query($discussionNumber: Int!) {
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
Related: <https://github.com/orgs/community/discussions/3951>.

fabcor-maxiv:
This is actually possible to get notifications for specific category(ies)... but only on Slack! Argh! This is frustrating...
<https://github.blog/changelog/2023-03-08-introducing-the-ability-to-subscribe-to-specific-discussion-categories-on-slack/>

marcus-oscarsson:
@fabcor-maxiv: Yes, @walesch-yan, did some research before finding this solution which in the end seemed to be the best one. It does add some extra overhead and we would have wished that this feature existed in GitHub. However, using the categories makes it easier to organize each discussion topic in a much better way, instead of having one large discussion.

walesch-yan:
Yes, unfortunately while there has been this mention about work being done on this feature from GitHub, there have been no updates on this for 2 years. I thought that using the action would be a better approach than using an additional communication service like Slack to the collaboration.

.github/workflows/discussions.yml
```diff
@@ -0,0 +1,69 @@
+---
```
fabcor-maxiv, line 1: Maybe explain in this file why this workflow is necessary. At least I would put the link to the feature request (that is still not implemented on GitHub's side): <https://github.com/orgs/community/discussions/3951>.

fabcor-maxiv (approved):
I agree with all that, of course, you don't need to convince me. I always try to write down related links for future reference.

.github/workflows/discussions.yml
walesch-yan: You are right, I will adjust this

fabcor-maxiv (approved):

walesch-yan:
I had to make a small change after I realized that you cannot create PAT's of an organization, instead it will work by creating a blank GitHub App, with permissions to write in a repositories discussion and read the organizations teams. 
This has some more advantages:
- No user has to generate a PAT for this
- The notifications are clearly shown to be from a bot and not a user, making them more distinguishable
- The permissions are more restricted, while PAT that have permissions to write to an repository's discussion section also have access to everything else in a public repository, App permissions can be limited to the discussion section.

.github/workflows/discussions.yml
```diff
@@ -0,0 +1,90 @@
...
+        id: github-app-token
+        uses: tibdex/github-app-token@v1
+        with:
+          app_id: ${{ secrets.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
+
+      - name: Notify Team
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ steps.github-app-token.outputs.token }}
```
walesch-yan, line 36: The changes are limited to this part of the file, which creates the token for authorization, before executing the script

PR #1025, opened 2024-09-19
Bumped Pydantic

marcus-oscarsson:
Bumped Pydantic. 

@oldfielj-ansto (or anyone else for that matter :) ), we are currently using the `pydantic.v1` package. Do you see any immediate issue with that ? We hope that we can get rid of this import relatively soon but we need to stay compatible with  [spectree](https://www.google.com/search?channel=fs&client=ubuntu&q=spectreee+python) which until now requires `pydantic.v1`


fabcor-maxiv:
I hope we find time to test this

marcus-oscarsson:
Initial testing looks, good. There should not be any change since we are using the `pydantic.v1` package. See https://docs.pydantic.dev/latest/migration/#install-pydantic-v2 for more information. We will need to change some model definitions when we are no longer using `pydantic.v1`. I did test that as well and it seems to work but it is indeed a bit of a bigger change that needs even more testing.

PR #1027, opened 2024-09-19
port CentringMath HWOBJ to work with the yaml config API changes

elmjag:
Don't use [] expression to read config properties.

Renamed attributes `self.cameraAxes` and `self.gonioAxes`, to `self._camera_axes` and `self._gonio_axes`, as they now collide with the 'auto created' properties on the ConfiguredObject superclass.

Use calls to `self.get_property()` when populating these attributes.

### DIFF
--- a/mxcubecore/HardwareObjects/CentringMath.py
+++ b/mxcubecore/HardwareObjects/CentringMath.py
@@ -10,36 +10,54 @@ class CentringMath(Procedure):
     CentringMath procedure
     """
 
+    def __init__(self, name: str):
+        super().__init__(name)
+        self._gonio_axes = []
+        self._camera_axes = []
+
     def init(self):
         """
         Ggonio axes definitions are static
         motorHO is expected to have get_value() that returns coordinate in mm
         """
-        self.motorConstraints = []
-        self.gonioAxes = []
-        for axis in self["gonioAxes"]:
-            self.gonioAxes.append(
-                {
-                    "type": axis.type,
-                    "direction": eval(axis.direction),
-                    "motor_name": axis.motorname,
+
+        def get_axis_property(property_name: str) -> list:
+            #
+            # get 'axis' key from the specified config property
+            #
+            # treat the property as optional, if not specified,
+            # return an empty list
+            #
+            return self.get_property(property_name, {}).get("axis", [])
+
+        def get_gonio_axes():
+            for axis in get_axis_property("gonioAxes"):
+                yield {
+                    "type": axis["type"],
+                    "direction": eval(axis["direction"]),
+                    "motor_name": axis["motorname"],
                     "motor_HO": HWR.get_hardware_repository().get_hardware_object(
-                        axis.motorHO
+                        axis["motorHO"]
                     ),
                 }
-            )
+
+        def get_camera_axes():
+            for axis in get_axis_property("cameraAxes"):
+                yield {
+                    "axis_name": axis["axisname"],
+                    "direction": eval(axis["direction"]),
+                }
+
+        self.motorConstraints = []
+        self._gonio_axes = list(get_gonio_axes())
 
         """
         This version is lacking video microscope object. Therefore we model only
         static camera axes directions, but no camera axes scaling or center - which
         are dynamic. Therefore, camera coordinates are relative, in mm.
         """
-        self.cameraAxes = []
-        for axis in self["cameraAxes"]:
-            self.cameraAxes.append(
-                {"axis_name": axis.axisname, "direction": eval(axis.direction)}
-            )
 
+        self._camera_axes = list(get_camera_axes())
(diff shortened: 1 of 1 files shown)


rhfogh (approved):
This really looks very neat. A question: This will clearly work for XML-configured cases, and convert them to a sensible data structure. Is it supposed to also work for yml-configured cases (I cannot quite see how) or is this an intermediate step that will require recoding once you switch to yaml configuration files?

elmjag:
> Is it supposed to also work for yml-configured cases (I cannot quite see how) or is this an intermediate step that will require recoding once you switch to yaml configuration files?

Yes, it works with YAML files as well. I have tested both with XML and YAML configuration files.

It uses `self.get_property()` calls to read gonio and camera axis configurations. On yaml branch `self.get_property()` returns identical data structures for both XML and YAML config files.

rhfogh:
OK, I had missed a few tricks reading it (including the fact that some functions were defined only inside other functions). This is even neater than I had noticed.
To help  me understand it further, could you send me the equivalent yaml configuration file? Also, is it the plan that eventually the data should be defined straightforwardly in a yaml config file and end up inside the CentringMath.config attribute? Meanwhile, how could one access this from outside the object, except by going through the internal (underline-started) attributes?

elmjag:
> To help me understand it further, could you send me the equivalent yaml configuration file? 

This is the XML file I used for testing.

```
<device class="CentringMath">
  <gonioAxes>
     <axis>
        <motorname>phiy</motorname>
        <direction>[0, -1, 0]</direction>
	<type>translation</type>
	<motorHO>udiff_phiy</motorHO>
     </axis>

     <axis>
        <motorname>phiz</motorname>
        <direction>[1, 0, 0]</direction>
        <type>translation</type>
        <motorHO>udiff_phiz</motorHO>
     </axis>

     <axis>
        <motorname>phi</motorname>
        <direction>[0, -1, 0]</direction>
        <type>rotation</type>
        <motorHO>udiff_omega</motorHO>
     </axis>

     <axis>
        <motorname>sampx</motorname>
        <direction>[0,0,1]</direction>
        <type>translation</type>
        <motorHO>udiff_sampx</motorHO>
     </axis>

     <axis>
        <motorname>sampy</motorname>
        <direction>[-1,0,0]</direction>
        <type>translation</type>
        <motorHO>udiff_sampy</motorHO>
     </axis>
  </gonioAxes>
   
  <cameraAxes>
     <axis>
        <axisname>X</axisname>
        <direction>[1, 0, 0]</direction>
     </axis>
     <axis>
        <axisname>Y</axisname>
        <direction>[0, 1, 0]</direction>
     </axis>
  </cameraAxes>
</device>
```

This is the YAML I used in testing. (Generated by the MXCuBE-web migration mode).

```
class: CentringMath.CentringMath
configuration:
  cameraAxes:
    axis:
      - axisname: X
        direction: '[1, 0, 0]'
      - axisname: Y
        direction: '[0, 1, 0]'
  gonioAxes:
    axis:
      - direction: '[0, -1, 0]'
        motorHO: udiff_phiy
        motorname: phiy
        type: translation
      - direction: '[1, 0, 0]'
        motorHO: udiff_phiz
        motorname: phiz
        type: translation
      - direction: '[0, -1, 0]'
        motorHO: udiff_omega
        motorname: phi
        type: rotation
      - direction: '[0,0,1]'
        motorHO: udiff_sampx
        motorname: sampx
        type: translation
      - direction: '[-1,0,0]'
        motorHO: udiff_sampy
        motorname: sampy
        type: translation

```

> Also, is it the plan that eventually the data should be defined straightforwardly in a yaml config file and end up inside the CentringMath.config attribute? Meanwhile, how could one access this from outside the object, except by going through the internal (underline-started) attributes?

I'm not aware of any plans regarding this. But I would suggest to document and implement a dedicated `CenteringMath` hardware object API. Say there should be `get_gonio_axes()` and `get_camera_axes()` methods, that return the axes in some specified format.

rhfogh:
@elmjag

Thanks, I do understand now. 

For the medium term I would propose 

1) to get rid of 'axis', so that the yaml file read
  gonioAxes:
      - direction: '[0, -1, 0]'
        motorHO: udiff_phiy
        motorname: phiy
        type: translation
      - direction: '[1, 0, 0]'
        motorHO: udiff_phiz
        motorname: phiz
        type: translation

etc.
Outside of xml there is no need for the additional nesting level defined by 'axis'.

2) to put gonioAxes and cameraAxes inside .config same as the other properties, and access them directly without the need for special getter functions. It would be simpler that way.

That is in the medium term, though. In the short term your changes makes it possible to work with a mixture of XML and YAML - and I very much approve.

rhfogh (approved):
We might want further simplifications later, but this excellent for now.

PR #1029, opened 2024-09-23
Rhfogh develop

rhfogh:
Fixed bug in GPhL workflow recentring handling.

Also added provisional code for integration with ESRF workflow and LIMS. NB this code is specific to GPhL workflow and only executed when relevant code is present on the main MXCuBE side - so, in practice - a no-op.

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -2022,6 +2022,9 @@ class GphlWorkflow(TaskNode):
         self.acquisition_dose = 0.0
         self.strategy_length = 0.0
 
+        # Workflow attributes - for passing to LIMS (conf Olof Svensson)
+        self.workflow_params = {}
+
         # # Centring handling and MXCuBE-side flow
         self.set_requires_centring(False)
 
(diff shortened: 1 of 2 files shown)


mxcubecore/model/queue_model_objects.py
```diff
@@ -2022,6 +2022,9 @@ def __init__(self):
         self.acquisition_dose = 0.0
         self.strategy_length = 0.0
 
+        # Workflow attributes - for passing to LIMS (conf Olof Svensson)
```
marcus-oscarsson, line 2025: What does "(conf Olof Svensson)" mean ?

rhfogh: This is code that 1) is used on Massif-1 but not anywhere else or by thw main develop version; 2) is still provisional, since the ICAT LIMS and connections ro it (as I understand it) are still not stable or documented. I have coded this so that it will work (but do nothing) in the main develop branch, and put it there to make it easier to track what is going on and avoid having a completely different branch for MASSIF-1.

Anyway, that is what it means.

marcus-oscarsson: Ah, I see

Issue #1030, opened 2024-09-23
Python 3.8 incompatible typehints in mxcubecore/HardwareObjects/Harvester.py


jbflo:
Same as [here](https://github.com/mxcube/mxcubeweb/issues/1414)

PR #1031, opened 2024-09-23
[type hint] fix for Python 3.8 incompatible typehints

jbflo:
@marcus-oscarsson  @woutdenolf As suggest i had the the line to fix the [issue](https://github.com/mxcube/mxcubecore/issues/1030)

### DIFF
--- a/mxcubecore/HardwareObjects/Harvester.py
+++ b/mxcubecore/HardwareObjects/Harvester.py
@@ -40,6 +40,7 @@ It has some functionalities, like Harvest Sample, etc....
     </object>
 -----------------------------------------------------------------
 """
+from __future__ import annotations
 import gevent
 import logging
 from typing import Optional


fabcor-maxiv:
I do not know why this fixes the issue. I must say I never understood how `from __future__ import annotations` works.

If it were me, I think I would have actually fixed the type annotations ( `typing.Dict[str]` I guess). But I am fine with this fix if it means we do not have to change it back again to `dict[str]` later when we get rid of Python 3.8.

fabcor-maxiv:
But bigger issue, I do not know why this issue went undetected by our checks... that is not good.

jbflo:
> But bigger issue, I do not know why this issue went undetected by our checks... that is not good.

Yeah.. In addition of this easy fix, we could maybe investigate why it passed the check .

marcus-oscarsson:
Hm, its more two separate issues I would say. I also prefer to explicitly do `from typing import Dict` but `from __future__ import annotations` would also work especially as we would remove it once we drop support for `Python 3.8` (which I guess could be relatively soon)

Yes, we should investigate why the tests passed even though they should not have. The application should not even have started with Python 3.8, so it is indeed strange.

fabcor-maxiv (approved):

PR #1033, opened 2024-10-02
Rhfogh develop

rhfogh:
Fixed problem with detector distance changing on wavelength interleaving even when detector distance set to be teh same.

Main changes: 

- doing energy,wait_ready() in AbstractCollect.set_resolution
- Removing move_detector (not reliably impelmented) and instead calculationg collect resolution from collect detector distance (when set)
- Removed P11Collect.do_collect as a duplicate of the superclass function.

### DIFF
--- a/mxcubecore/HardwareObjects/DESY/P11Collect.py
+++ b/mxcubecore/HardwareObjects/DESY/P11Collect.py
@@ -63,7 +63,7 @@ class P11Collect(AbstractCollect):
     @task
     def move_motors(self, motor_position_dict):
         """Moves motors to the specified positions.
-        
+
         Args:
             motor_position_dict (dict): Dictionary containing motor positions.
         """
@@ -71,10 +71,10 @@ class P11Collect(AbstractCollect):
 
     def _take_crystal_snapshot(self, filename):
         """Takes a snapshot of the crystal and saves it to the given filename.
-        
+
         Args:
             filename (str): Path to save the crystal snapshot.
-        
+
         Raises:
             RuntimeError: If unable to move to the centring phase.
         """
@@ -95,7 +95,7 @@ class P11Collect(AbstractCollect):
 
     def set_transmission(self, value):
         """Sets the transmission value on the beamline.
-        
+
         Args:
             value (float): Transmission value to set.
         """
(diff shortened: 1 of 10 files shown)


mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -220,32 +220,45 @@ def do_collect(self, owner):
...
 
-            if "wavelength" in self.current_dc_parameters:
+            wavelength = self.current_dc_parameters.get("wavelength")
+            energy = self.current_dc_parameters.get("energy")
+            detector_distance = self.current_dc_parameters.get("detector_distance")
+            dd0 = self.current_dc_parameters.get("resolution")
+            if dd0 and dd0.get('upper'):
+                resolution = dd0["upper"]
+            else:
+                resolution = None
```
beteva: I did not know that the resolution was a dictionary, but I see that it was already the case before. Probably comes from the Qt version.Would it be more elegant to do it like this:
```
try:
   resolution = self.current_dc_parameters.get("resolution").get("upper")
except AttributeError:
   resolution = None
```

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -220,32 +220,45 @@ def do_collect(self, owner):
...
                 )
-                self.set_energy(self.current_dc_parameters["energy"])
-
-            dd = self.current_dc_parameters.get("resolution")
-            if dd and dd.get("upper"):
-                resolution = dd["upper"]
-                log.info("Collection: Setting resolution to %.3f", resolution)
+                wavelength = HWR.beamline.energy.calculate_wavelength(energy)
+            if energy:
+                self.set_energy(energy)
```
beteva, line 245: here i would rather use the energy HO, as you already use it:
```
HWR.beamline.energy.set_value(energy)
or
HWR.beamline.energy.set_wavelength(wavelength)
```
This will also allow the GUI to update the values. The fact that there is no timeout insures that the energy is either set, or there is an error, so separate set_energy and set_resolution methods are not necessary in my opinion.

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -220,32 +220,45 @@ def do_collect(self, owner):
...
-
-            elif "detector_distance" in self.current_dc_parameters:
+            elif resolution:
                 log.info(
-                    "Collection: Moving detector to %.2f",
-                    self.current_dc_parameters["detector_distance"],
+                    "Collection: Setting resolution to %.2f", resolution,
                 )
-                self.move_detector(self.current_dc_parameters["detector_distance"])
+                self.set_resolution(resolution)
```
beteva, line 260: Same here, why not using
`HWR.beamline.resolution.set_value(resolution)`
See the comment about energy.

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
```diff
@@ -425,13 +437,8 @@ def set_resolution(self, value):
...
         """
-        pass
-
-    def move_detector(self, value):
-        """
-        Descript. :
-        """
-        pass
+        HWR.beamline.energy.wait_ready()
+        HWR.beamline.resolution.set_value(value)
```
beteva, line 440: I believe we could remove this method. This would be the first step to simplify the procedure 😄

mxcubecore/HardwareObjects/mockup/CollectMockup.py
```diff
@@ -209,14 +209,8 @@ def set_wavelength(self, wavelength):
     def set_energy(self, energy):
         HWR.beamline.energy.set_value(energy)
```
beteva, line 211: I believe we could remove this method,

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -95,138 +95,28 @@ def _take_crystal_snapshot(self, filename):
...
-            self.update_data_collection_in_lims()
-
-        except RuntimeError as e:
-            failed_msg = "Data collection failed!\n%s" % str(e)
-            self.collection_failed(failed_msg)
-        else:
-            self.collection_finished()
-        finally:
-            self.data_collection_cleanup()
+            super().set_resolution(value)
```
beteva: I would remove this method. If you really want to keep it, than there are two more secure ways to compare floats:
```
1. abs(HWR.beamline.resolution.get_value() - value) < 0.01
2. from math import isclose
   isclose(HWR.beamline.resolution.get_value(), value, abs_tol=0.01)
```

beteva (commented):
I've suggested removing methods like set_energy, set_resolution, but than it requires replacing them in some of the local synchrotron implementations. Not sure it is worth the effort. Do at your convenience.

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
rhfogh: Change made

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
rhfogh: That might be problematical,. At least at P14 (admittedly way behind), Collect.set_energy passes the values to a DetectorServer, so that they are not set at the moment, but only set when collection starts. I suspect this kind of thing is the reason why there are set_energy (etc.) functions in this class in the first place

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
rhfogh: Same reason - at P14 this is passed to a DetectorServer and set later. I would prefer setting things directly here, like you propose, but if (some) beamline owners want special servers and special set methods, we can hardly force to change? ;-)

mxcubecore/HardwareObjects/abstract/AbstractCollect.py
rhfogh: See previous comments. And, one way or the other, we do need a wait_ready or a timeout between setting the energy and setting the resolution. Otherwise the resolution is set before the energy has finished changing, and we get the wrong detector distance.

mxcubecore/HardwareObjects/mockup/CollectMockup.py
rhfogh: Not as long as the superclass function exists, is a no-op, and is used.

mxcubecore/HardwareObjects/DESY/P11Collect.py
rhfogh: Indeed - math,is_close() is better.

As for removal, I gathered from our talk that the overall refactoring should wait till we were anyway refactoring the queue. I would like to get rid of all these functions too (where we can, of course).

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -3062,10 +3063,18 @@ def adjust_transmission(self, values):
...
+                        },
+                        "transmission": {
+                            "value": new_transmission,
+                        },
+                        "use_dose": {
+                            "value": use_dose,
+                        },
+                        "experiment_time": {
+                            "value": new_experiment_time,
+                        },
```
fabcor-maxiv: If I am not mistaken, it should be possible to get black to keep those on one line by removing the trailing comma `,` inside the curly brackets, in case that is something you could want. :)

fabcor-maxiv: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma

rhfogh: @fabcor-maxiv Ah, that was useful. Thanks. I was wondering why the program was so zealous about splitting things up.

Issue #1023, opened 2024-09-17
channel and commands in YAML files


elmjag:
Give the following XML configuration file:

```
<object class="SomeClass">
  <tangoname>some/tango/device</tangoname>
  <command type="tango" name="Open">Open</command>
  <channel type="tango" name="State" polling="1000">State</channel>
</object>

```
How would it be expressed as YAML config file? Did we discuss it at some point?

Currently, when generating YAML files with mxcubeweb-server's `--export-yaml-config` flag, the `channel` and `command` tags are just ignored.

beteva:
I do not recall discussing this and we should have a really good look, as it is not only Tango, but also Exporter, probably Epics and all the rest using the commands/channels mechanism.

rhfogh:
I do not think it was ever discussed, either. And I am pretty sure I never looked at it. The exported yaml config files deal with properties and contained objects, but channels and commands are handled separately in the XML world and are simply not looked at in the conversion.

I think the commands and channels need their own, well-thought-out handling in the yaml world, and that they should be treated as special cases - as they already are in the XML world. 

One possibility might be separate attributes 'channels' and 'commands' in parallel with 'objects' and 'configuration', and putting a list of dictionaries in with a dictionary for every channel and every command. One could write the conversion code once to re-export from XML it was decided how the end result should look. But it would have to be the people who understand commands and channels who look at this.

elmjag:
We did some light discussions on this topic, and came up with following suggestion.

Let's add a custom section for each protocol, that define command and channel bindings. The structure of each protocol section will have following format:

```
class: MegaCommunicator        
<protocol>:
  <end-point>:                   # E.g. tango device / exporter address / EPICS prefix / etc
    commands:
      <command-1-name>:          # MXCuBE name
        <config-prop-1>: <val-1>
        <config-prop-2>: <val-2>        
      <command-2-name>:          # command without any config properties
    channels:
      <channel-1-name>:          # MXCuBE Name
        <config-prop-1>: <val-1>
        <config-prop-2>: <val-2>
      <channel-2-name>:          # channel without any config properties
```

An example how you would configure Tango commands and channels.


```
class: TangoShutter.TangoShutter

configuration:
  values: something

tango:
  b312a-eh/dia/detc-01:    # tango device server name
    commands:
      Open:                # MXCuBE name
        name: Open         # tango command name, optional, default is MXCuBE name
      Close:  
    channels:
      State:               # MXCuBE name
[...]

PR #1034, opened 2024-10-02
adds support for configuring commands and channels using YAML

elmjag:
Adds support for configuring commands and channels using YAML. Currently it's possible to configure `Tango`, `exporter` and `EPICS` commands and channels.

This implements roughly the format discussed in: https://github.com/mxcube/mxcubecore/issues/1023
The actual format implement is documented here: https://mxcubecore--1034.org.readthedocs.build/en/1034/dev/commands_channels.html

This is still work in progress. Still on the todo list:

* write some tests
* do a bit more "manual" testing

I would like to get some feedback, before spending more time on this.

### DIFF
--- a/docs/source/dev/commands_channels.md
+++ b/docs/source/dev/commands_channels.md
@@ -1,0 +1,297 @@
+# Commands and Channels
+
+The mxcubecore provides a hardware object-level abstraction
+for communicating using various flavors of control system protocols.
+A hardware object can utilize instances of
+[`CommandObject`](#mxcubecore.CommandContainer.CommandObject) and
+[`ChannelObject`](#mxcubecore.CommandContainer.ChannelObject) objects.
+These objects provide a uniform API for accessing a specific control system.
+Mxcubecore provides support for using protocols such as
+[_Tango_](https://www.tango-controls.org/),
+[_EPICS_](https://docs.epics-controls.org),
+_exporter_ and more.
+
+The `CommandObject` and `ChannelObject` objects can be created using the `add_command()` and `add_channel()` methods of hardware objects.
+Another option is to specify them in the hardware object's configuration file.
+If `CommandObject` and `ChannelObject` are specified in the configuration file,
+the specified objects will be automatically created during hardware object initialization.
+
+## Configuration files format
+
+The general format for specifying `CommandObject` and `ChannelObject` objects is as follows:
+
+```yaml
+class: <hardware-object-class>
+<protocol>:                      # protocol in use, tango / exporter / epics / etc
+  <end-point>:                   # tango device / exporter address / EPICS prefix / etc
+    commands:
+      <command-1-name>:          # command's MXCuBE name
+        <config-prop-1>: <val-1>
+        <config-prop-2>: <val-2>
+    channels:
+      <channel-1-name>:          # channel's MXCuBE Name
+        <config-prop-1>: <val-1>
+        <config-prop-2>: <val-2>
+```
+
+The `CommandObject` and `ChannelObject` specification are grouped by the protocol they are using.
+Each protocol have its own dedicated section in the configuration file.
+The semantics for the protocol are similar but protocol-specific, see below for details.
+
+Currently, the following protocols can be configured using YAML configuration files:
+
+ - [Tango](#tango-protocol)
+ - [exporter](#exporter-protocol)
+ - [EPICS](#epics-protocol)
+
+## Tango Protocol
+
+The format for specifying _tango_ `CommandObject` and `ChannelObject` objects is as follows:
+
+```yaml
+class: <hardware-object-class>
+tango:
+  <tango-device-name>:
+    commands:
+      <command-1-name>:
+        <config-prop-1>: <val-1>
+      <command-2-name>:
+        <config-prop-1>: <val-1>
+    channels:
+      <channel-1-name>:
+        <config-prop-1>: <val-1>
+        <config-prop-2>: <val-2>
+      <channel-2-name>:
+        <config-prop-1>: <val-1>
+```
+
+`<tango-device-name>` specifies the tango device to use.
+Multiple `<tango-device-name>` sections can be specified, in order to use different tango devices.
+Each `<tango-device-name>` contains optional `commands` and `channels` sections.
+These sections specify `CommandObject` and `ChannelObject` object to create using the `<tango-device-name>` tango device.
+
+### Commands
+
+`commands` is a dictionary where each key specifies a `CommandObject` object.
+The key defines the MXCuBE name for the command.
+The values specify an optional dictionary with configuration properties for the `CommandObject` object.
+The following configuration properties are supported:
+
+| property | purpose            | default             |
+|----------|--------------------|---------------------|
+| name     | tango command name | MXCuBE command name |
+
+### Channels
+
+`channels` is a dictionary where each key specifies a `ChannelObject` object.
+The key defines the MXCuBE name for the channel.
+The values specify an optional dictionary with configuration properties for the `ChannelObject` object.
+The following configuration properties are supported:
+
+| property       | purpose                            | default             |
+|----------------|------------------------------------|---------------------|
+| attribute      | tango attribute name               | MXCuBE channel name |
+| polling_period | polling periodicity, milliseconds  | polling is disabled |
+| timeout        | tango device timeout, milliseconds | 10000               |
+
+By default, a tango `ChannelObject` object will use tango attribute change event, in order to receive new attribute values.
+For this to work, the tango device must send the change events for the attribute.
+For cases where such events are not sent, the attribute polling can be enabled.
+If `polling` property is specified, MXCuBE will poll the tango attribute with specified periodicity.
+
+### Example
+
+Below is an example of a hardware object that specifies Tango commands and channels.
+
+```yaml
+class: MyTango
+tango:
+  some/tango/device:
+    commands:
+      Open:
+      Close:
+      Reset:
+        name: Reboot
+    channels:
+      State:
+      Volume:
+          attribute: currentVolume
+          poll: 1024
+```
+
+In the above example, commands `Open`, `Close` and `Reset` as well as channels `State` and `Volume` are configured.
+All command and channel objects are bound to the commands and attributes of the _some/tango/device_ tango device.
+
+`Open` and `Close` commands are bound to _Open_ and _Close_ Tango commands.
+The `Reset` has a configuration property that binds it to _Reboot_ tango command.
+
+The `State` channel will be mapped to _State_ attribute of the Tango device.
+Its value will be updated via Tango change events.
+
+The `Volume` channel will be mapped to the _currentVolume_ attribute of the tango device.
+The _currentVolume_ attribute's value will be polled every 1024 milliseconds.
+
+## Exporter Protocol
+
+The format for specifying _exporter_ `CommandObject` and `ChannelObject` objects is as follows:
+
+```yaml
+class: <hardware-object-class>
+tango:
+  <exporter-address>:
+    commands:
+      <command-1-name>:
+        <config-prop-1>: <val-1>
+      <command-2-name>:
+        <config-prop-1>: <val-1>
+    channels:
+      <channel-1-name>:
+        <config-prop-1>: <val-1>
+      <channel-2-name>:
+        <config-prop-1>: <val-1>
+```
+
+`<exporter-address>` specifies the exporter address to use.
+Multiple `<exporter-address>` sections can be specified to use devices at different addresses.
+Each `<exporter-address>` contains optional `commands` and `channels` sections.
+These sections specify `CommandObject` and `ChannelObject` objects to create using the `<exporter-address>` tango device.
+
+`<exporter-address>` specifies the exporter's host and port number.
+It has the following format: `<host>:<port>`.
+`<host>` is the host name or IP address to use.
+`<port>` is the TCP port number to use.
+Note that, due to YAML parsing rules, you need to use quotes when specifying the exporter address.
+Below is an example of an exporter address that can be used in a YAML configuration file:
+
+```
+"foo.example.com:9001"
+```
+
+### Commands
+
+`commands` is a dictionary where each key specifies a `CommandObject` object.
+The key defines the MXCuBE name for the command.
+The values specify an optional dictionary with configuration properties for the `CommandObject` object.
+The following configuration properties are supported:
+
+| property | purpose               | default             |
+|----------|-----------------------|---------------------|
+| name     | exporter command name | MXCuBE command name |
+
+### Channels
+
+`channels` is a dictionary where each key specifies a `ChannelObject` object.
+The key defines the MXCuBE name for the channel.
+The values specify an optional dictionary with configuration properties for the `ChannelObject` object.
+The following configuration properties are supported:
+
+| property  | purpose                  | default             |
+|-----------|--------------------------|---------------------|
+| attribute | exporter attribute name  | MXCuBE channel name |
+
+### Example
+
+Below is an example of a hardware object that specifies exporter commands and channels.
+
+```yaml
+class: MyExporter
+exporter:
+  "foo.example.com:9001":
+    commands:
+      Open:
+      Close:
+      Reset:
+        name: Reboot
+    channels:
+      State:
+      Volume:
+          attribute: currentVolume
+```
+
+In the above example, commands `Open`, `Close` and `Reset` as well as `State` and `Volume` channels are configured.
+All command and channel objects are bound to the exporter host _foo.example.com_ at port _9001_.
+
+`Open` and `Close` commands are bound to _Open_ and _Close_ exporter commands.
+The `Reset` has a configuration property that binds it to the _Reboot_ exporter command.
+
+The `State` channel will be mapped to _State_ exporter attribute.
+The `Volume` channel will be mapped to _currentVolume_ exporter attribute.
+
+## EPICS Protocol
+
+The format for specifying _EPICS_ `ChannelObject` objects is as follows:
+
+```yaml
+class: EpicsCommunicator
+epics:
+  <prefix>:
+    channels:
+      <channel-1-name>:
+        <config-prop-1>: <val-1>
+        <config-prop-2>: <val-2>
+      <channel-2-name>:
+        <config-prop-1>: <val-1>
+```
+
+`<prefix>` specifies the EPICS PV prefix to use for that section.
+Multiple `<prefix>` sections can be specified, in case not all channels share a common prefix.
+Each `<prefix>` contains a `channels` section, which specifies `ChannelObject` objects to create.
+
+It is also possible to use the empty string, `""`, as the prefix.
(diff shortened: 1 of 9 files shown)


docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+
+`channels` is a dictionary where each key specifies a `Channel` object.
+The key defines the MXCuBE name for the channel.
+The values specify an optional dictionary with configuration properties for the `Channel` object.
+The following configuration properties are supported:
+
+| property  | purpose                           | default             |
+|-----------|-----------------------------------|---------------------|
+| attribute | tango attribute name              | MXCuBE channel name |
+| poll      | polling periodicity, milliseconds | polling is disabled |
```
fabcor-maxiv:
```suggestion
| polling_period      | polling periodicity, milliseconds | polling is disabled |
```

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
+# Commands and Channels
+
+The mxcubecore provides a hardware object-level abstraction
+for communicating using various flavors of control system protocols.
+A hardware object can utilize instances of `Command` and `Channel` objects.
```
fabcor-maxiv: Link to the documentation for the actual classes, please.
(I always forget how to do it, look up the Myst doc: <https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html>).

Also isn't it `mxcubecore.CommandContainer.ChannelObject` and `mxcubecore.CommandContainer.CommandObject`?

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+These objects provide a uniform API for accessing a specific control system.
+Mxcubecore provides support for using protocols such as _tango_, _EPICS_, _exporter_ and more.
```
fabcor-maxiv: Links, links, links! Documentation should have lots of links. Typically the first occurrence of every term should have a link to its definition.

```suggestion
Mxcubecore provides support for using protocols such as [_tango_](https://pytango.readthedocs.io/), _EPICS_, _exporter_ and more.
```

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+Another option is to specify them in the hardware object's configuration file.
+If `Command` and `Channel` are specified in the configuration file,
+the specified objects will be automatically created during hardware object initialization.
+
+## Configuration files format
+
+The general format for specifying `Command` and `Channel` objects is as follows:
+
+```yaml
+class: MegaCommunicator
```
fabcor-maxiv: To keep consistent with the rest of the example:

```suggestion
class: <hardware-object-class>
```

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+ - [Tango](#tango-protocol)
+ - [exporter](#exporter-protocol)
+ - [EPICS](#epics-protocol)
+
+## Tango Protocol
+
+The format for specifying _tango_ `Command` and `Channel` objects is as follows:
+
+```yaml
+class: TangoCommunicator
```
fabcor-maxiv:
```suggestion
class: <hardware-object-class>
```

docs/source/dev/commands_channels.md
fabcor-maxiv: I guess there should be `timeout` as well, no? And `username`? Well, but only if anyone actually uses those, otherwise better leave them out indeed.

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+For this to work, the tango device must send the change events for the attribute.
+For cases where such events are not sent, the attribute polling can be enabled.
+If `polling` property is specified, MXCuBE will poll the tango attribute with specified periodicity.
+
+### Example
+
+Below is an example of a hardware object that specifies Tango commands and channels.
+
+```yaml
+class: MyTango
```
fabcor-maxiv, line 107:
```suggestion
class: my_package.my_sub_package.MyTangoClass
```

Otherwise that is a good example: complex and complete. That is good.

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+      Reset:
+        name: Reboot
+    channels:
+      State:
+      Volume:
+          attribute: currentVolume
+          poll: 1024
+```
+
+In the above example, commands `Open`, `Close` and `Reset` as well as `State` and `Volume` channels are configured.
```
fabcor-maxiv:
```suggestion
In the above example, commands `Open`, `Close` and `Reset` as well as channels `State` and `Volume` are configured.
```

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+All command and channel objects are bound to the commands and attributes of the _some/tango/device_ tango device.
```
fabcor-maxiv, line 123:
```suggestion
All command and channel objects are bound to the commands and attributes of the `some/tango/device` tango device.
```

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+
+`Open` and `Close` commands are bound to _Open_ and _Close_ Tango commands.
```
fabcor-maxiv, line 125:
```suggestion
`Open` and `Close` hardware object commands are bound to `Open` and `Close` Tango commands.
```

docs/source/dev/configuration_files.md
```diff
@@ -59,18 +59,27 @@ and an `init()` function that is executed after configured parameters and contai
...
-  base_result_url: https://your.limsresults.org
-  login_type: proposal
+  altai: big
+  choir: 52
 objects:
-  lims_rest: lims_rest.yaml
+  darkhan: darkhan.yaml
   session: session.yaml
+tango:
+  "some/tango/device":
```
fabcor-maxiv, line 70:
```suggestion
  some/tango/device:
```

mxcubecore/model/protocols/epics.py
fabcor-maxiv, line 1: We should try to follow the Google style for doctrings:

* <https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings>
* <https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html>

mxcubecore/model/protocols/epics.py
```diff
@@ -0,0 +1,57 @@
...
+Provides an API to read configured EPICS channels.
+"""
+from typing import Optional, Iterable, Tuple, Dict
+from pydantic import BaseModel
+
+
+class Channel(BaseModel):
+    """
+    an EPICS channel configuration
+    """
```
fabcor-maxiv:
```suggestion
    """EPICS channel configuration."""
```

mxcubecore/model/protocols/epics.py
```diff
@@ -0,0 +1,57 @@
+"""
+Models the `epics` section of YAML hardware configuration file.
+
+Provides an API to read configured EPICS channels.
+"""
```
fabcor-maxiv:
```suggestion
"""Models for the `epics` section of YAML hardware configuration file.

Provides an API to read configured EPICS channels.
"""
```

mxcubecore/model/protocols/tango.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    the 'tango' section of the hardware object's YAML configuration file
+    """
+
+    __root__: Dict[str, Device]
+
+    def get_tango_devices(self) -> Iterable[Tuple[str, Device]]:
+        """
+        get all tango devices specified in this section
+        """
```
fabcor-maxiv:
```suggestion
        """Get all Tango devices specified in this section."""
```

mxcubecore/protocols_config.py
```diff
@@ -0,0 +1,135 @@
...
+    get the callable that will set up commands and channels for a specific protocol
+    """
+    return _protocol_handles()[protocol_name]
+
+
+def _setup_protocol(hwobj: HardwareObject, config: dict, protocol: str):
+    """
+    add the Command and Channel objects configured in the specified protocol section
+
+    parameters:
```
fabcor-maxiv, line 113: I think the sections are case-sensitive.

```suggestion
    Parameters:
```

mxcubecore/protocols_config.py
```diff
@@ -0,0 +1,135 @@
...
+        protocol: name of the protocol to handle
+    """
+    protocol_config = config.get(protocol)
+    if protocol_config is None:
+        # no configuration for this protocol
+        return
```
fabcor-maxiv, line 119: Why not raise an exception instead of failing silently?

Or am I misunderstanding something?

fabcor-maxiv (commented):
Good stuff.

A first review pass, a first batch of comments and improvement suggestions.

mxcubecore/protocols_config.py
elmjag: This is not an error.

This is the case when the config does not have a section for that specific protocol. I.e. there are no commands/channels configured using that protocol. This is normal situation.

The algorithm for handling protocol sections is as follows:

```
for protocol in <supported-protocols>:
  if config have protocol section then
    handle that protocol section
  else
   do nothing
```

fabcor-maxiv: Ah I see : )

~I guess I would expect the "for" and "if" in one function, and the "handle" in the other function. But maybe that is not possible.~

docs/source/dev/commands_channels.md
henrique-silva: https://docs.epics-controls.org/en/latest/ for the official EPICS docs (that is being reworked now as far as I know, but at least it's not the APS website from the 80's anymore 😄)

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,288 @@
...
+  "FOO:B:x1.":
+    channels:
+      State:
+        suffix: stat
+      Vol:
+```
+
+We have an `FOO:B:x1.` prefix specified, with two `Channel` objects `State` and `Vol`.
+`State` will use `FOO:B:x1.stat` PV name, specified by section's prefix and the `suffix` configuration property.
+`Vol` will use `FOO:B:x1.Vol` PV name, specified by section's prefix and channels MXCuBE name.
```
henrique-silva: So, I think this might be a bit problematic, because the modeling of PV is usually `$(PREFIX):$(DEVICE).$(PROPERTY)`, so one record - or PV - ( `$(PREFIX):$(DEVICE)`) will have multiple properties, for example an analog out has, .DESC (description), .VAL (value), .ASLO (slope).
But in this case that you're modelling it is way more likely that this will be separated into multiple PVs, so you'd likely have:

```yaml
class: EpicsCommunicator
epics:
  "FOO:B:":
    channels:
      State:
        suffix: pv_1.STAT
      Vol:
        suffix: volume.VAL
```

In the end I think the code will handle just fine if you're just appending, but the prefix is only the part before teh last colon (in most naming conventions of EPICS that I've seen)

fabcor-maxiv: I am not sure I am understanding you correctly, but would <https://github.com/mxcube/mxcubecore/issues/1023#issuecomment-2367904138> answer your question?

henrique-silva: Yes, that's pretty much it.

It's just the non-obvious separation between PV/record and properties in EPICS that is a bit weird for modelling, but the simple concatenation should handle all cases

fabcor-maxiv: @elmjag I think it is worth modifying the epics example to show a more complex and complete example, that shows how to combine all those things.

docs/source/dev/commands_channels.md
elmjag: Yes, it's `mxcubecore.CommandContainer.ChannelObject` and `mxcubecore.CommandContainer.ChannelObject`. Added links to API docs for these objects.

docs/source/dev/commands_channels.md
elmjag: Added links to Tango and EPICS web sites.

docs/source/dev/commands_channels.md
elmjag: changed

docs/source/dev/commands_channels.md
elmjag: changed

docs/source/dev/commands_channels.md
elmjag: fixed

docs/source/dev/commands_channels.md
elmjag: Here the italics mark-up is used to distinguish between Open in the context of the MXCuBE configuration and Open in the context of the tango device.

docs/source/dev/configuration_files.md
elmjag: changed

docs/source/dev/commands_channels.md
elmjag: Will this example make things more clear?

```
class: EpicsCommunicator
epics:
  "FOO:B:":
    channels:
      State:
        suffix: pv_1.STAT
      Vol:
        suffix: volume.VAL
      Freq:
```

Regarding `Freq` channel, I want to include an example channel, without the optional config property `suffix`.

docs/source/dev/commands_channels.md
elmjag: But this is not an example of how to write a hardware configuration file. This is an example how to configure tango commands and channels. This long class name will distract the reader from the main point of the example.

docs/source/dev/commands_channels.md
elmjag: This is to make it consistent with the style below.

https://github.com/mxcube/mxcubecore/pull/1034#discussion_r1798848269

mxcubecore/model/protocols/epics.py
elmjag: but it's so ugly :(

mxcubecore/model/protocols/epics.py
elmjag: reluctantly changed!

mxcubecore/model/protocols/epics.py
elmjag: stupid google!

mxcubecore/model/protocols/tango.py
elmjag: image

mxcubecore/protocols_config.py
elmjag: works fine with lower case `p`:

image

docs/source/dev/commands_channels.md
elmjag: For `username`, I think it's should be deprecated. Looking that the code, I don't see any practical uses of the `username` anywhere. Feels like it's a left over from some older feature, which is not used anymore.

For `timeout`, I'll try to add it. That seems like a useful feature, as sometimes you need to tweak tango timeout.

elmjag: changed to `polling_period`

elmjag: I have added `timeout` config.

docs/source/dev/commands_channels.md
henrique-silva: Yeah, I think it's clearer, but take my opinion with a grain of salt since I don't really know much about MxCube 😅 

But yes, it does show that you can map PVs and fields to the mxcube channels

elmjag: updated example

mxcubecore/model/protocols/epics.py
```diff
@@ -0,0 +1,47 @@
...
+    suffix: Optional[str]
+    poll: Optional[int]
+
+
+class Prefix(BaseModel):
+    """Configuration of an EPICS prefix section."""
+
+    channels: Optional[Dict[str, Optional[Channel]]]
+
+    def get_channels(self) -> Iterable[Tuple[str, Channel]]:
```
fabcor-maxiv, line 21: I think it should be `Iterator` and not `Iterable`.

```suggestion
    def get_channels(self) -> Iterator[Tuple[str, Channel]]:
```

elmjag: It is valid, generators _are_ Iterables:

```
>>> from typing import Iterable
>>> def foo():
...   yield 1
...   yield 3
>>> isinstance(foo(), Iterable)
True
```

Here I want to specify a more generic return type. It is _some_ kind of iterable. Today it's generator, but in the future perhaps it is a list.

fabcor-maxiv: I see, I guess it is fine... I do not know for sure.

---

The rules I had read some time ago were like this:

* When you produce you should use as precise a type you can.
* When you consume you should use as generic a type as you can that still fulfils your usage.

so that is what I always follow and I never had any issue.

So for example, I would do something like this:

```python
def producer() -> list[int]:
    return [1, 2, 3]

def consumer(things: typing.Iterable[int]) -> None
    for thing in things:
        print(thing)

some_things = producer()
consumer(some_things)
```

If later the `producer` function is refactored into a generator (or iterator or any other iterable), I still do not need to change anything in `consumer`.

rhfogh: A question purely for my education (you guys have studied more than I have). Reading you example one might be tempted to think that producer promises to return a list. Would it be OK to write code like 

list2 = list1 + producer() 

Or should one  instead write

list2 = list1 + list(producer())

in case producer was later refactored to return something else? Or is this principle only valid for certain contexts?

(I know you could write something like 

list2 = []; list2.extend(list1); list2.extend(producer()) 

but the question is whether you need to assume that the return type of producer() might be(come) any iterable at some later date.

elmjag: If you have

```
def producer() -> list[int]:
    return [1, 2, 3]
```

Then, it explicitly promises to return a list. Thus you can use return value as a list.

However, if you have:

```
def producer() -> Iterable[int]:
    return [1, 2, 3]
```

Then it does not promise a list. This can any type of iterable value. Then you should explicitly convert this to a list, if you a doing list operations on it.

Here is documentation on Iterable type annotation: https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable


However, bear in mind that none of this is enforced during runtime. Nothing in python stops you from doing this:

```
def producer() -> Iterable[int]:
    return False
```

Python will happily execute this and return a boolean to you.

The only way to find you that your code is not following the type annotations is to run a static checker tool, such as mypy.

rhfogh: @elmjag Thanks, that makes sense. I know that Python does not respect type annotations, but I want to be a good boy and follow our rules, so I was trying to make sure I understood what our rules were.

marcus-oscarsson: I suggest that we try out pyright: https://github.com/microsoft/pyright

elmjag: > I suggest that we try out pyright: https://github.com/microsoft/pyright

Yes, I agree. I wrote a new issue: https://github.com/mxcube/mxcubecore/issues/1053

PR #1035, opened 2024-10-03
export YAML configuration to '.yaml' files

elmjag:
When 'export yaml configuration' option is invoked, write configurations to `.yaml` files, instead of `.yml`.

The `.yaml` is the canonical file extension for YAML files.


fabcor-maxiv (approved):
Indeed. Unless we have a good reason not to (technical limitation?), then we should go for `.yaml`.

PR #1036, opened 2024-10-03
units: add milliseconds (ms) to seconds conversion function

elmjag:
Adds milliseconds to seconds conversion function.

### DIFF
--- a/mxcubecore/utils/units.py
+++ b/mxcubecore/utils/units.py
@@ -14,6 +14,13 @@ def us_to_sec(us: float) -> float:
     return us / 1_000_000.0
 
 
+def ms_to_sec(ms: float) -> float:
+    """
+    convert milliseconds (ms) to seconds
+    """
+    return ms / 1000.0
+
+
 def sec_to_us(sec: float) -> float:
     """
     convert seconds to microseconds (μs)
--- a/test/pytest/test_utils_units.py
+++ b/test/pytest/test_utils_units.py
@@ -1,6 +1,7 @@
 from math import isclose
 from mxcubecore.utils.units import (
     us_to_sec,
+    ms_to_sec,
     sec_to_us,
     sec_to_hour,
     ev_to_kev,
@@ -14,6 +15,11 @@ from mxcubecore.utils.units import (
 def test_us_to_sec():
     assert isclose(us_to_sec(500_000), 0.5)
     assert isclose(us_to_sec(123.4), 0.0001234)
+
+
+def test_ms_to_sec():
+    assert isclose(ms_to_sec(13), 0.013)
+    assert isclose(ms_to_sec(2148.5), 2.1485)
 
 
 def test_sec_to_us():


rhfogh (approved):
Sure, fine. Not that I can see the need,
value /= 1000
is faster, but if anyone wants it ...

fabcor-maxiv:
> Not that I can see the need,

It is about avoiding [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

Just dividing by one thousand (or any other mathematical operation, really) without any explaining comment is not good practice. I would be fine with having a clarifying comment. Having named function that makes the intent very clear, is also a very good solution.

fabcor-maxiv (approved):
I think I said it already, I am not super enthusiastic about the inconsistent naming: `meter_to_mm` `us_to_sec` `A_to_mA`. Maybe `m_to_mm` and `s_to_us` or `meter_to_millimeter` and `second_to_microsecond` (the short ones could be aliases to the long ones). But whatever, it is nitpicking.

rhfogh:
Depends on how you see it.

Personally I do not think m_to_mm() makes any more sense than
multiply_by_two(), multiply_by_three(), multiply_by_ten() etc. Which of course no one uses.
This seems to me to be nitpicking for the sake of it. 

If you want to make the code clearer, you should work on the variable names: value_in_s, value_in_ms, etc. Or you can make it clear from the documentation what are the units of each parameter and input value. 

To avoid magic numbers we have e.g. the constant HC_OVER_E, which serves to make sure 1) that everybody converts between energy and wavelength with the same constants, 2) that it is clear what is being done (instead  of relying on people recognising the value 12.384). 

Once it is clear what the values mean and the unit they should have, the only reason I can see for avoiding the 'magic' number 1000 is to support programmers who are unaware that there are 1000 ms to a s. But is that really a problem we need to solve?

PS. The site you link to talks about replacing naked values with named constants, and I fully support that. What you are doing here is something else.

fabcor-maxiv:
> If you want to make the code clearer, you should work on the variable names: value_in_s, value_in_ms, etc. 

I'd be okay with that. 

> Or you can make it clear from the documentation what are the units of each parameter and input value.

This MUST be done any way.

> This seems to me to be nitpicking for the sake of it.

We should not push for using these functions if the code is already made clear another way. These functions offer a way to make things clear.

elmjag:
> I think I said it already, I am not super enthusiastic about the inconsistent naming: `meter_to_mm` `us_to_sec` `A_to_mA`. Maybe `m_to_mm` and `s_to_us` or `meter_to_millimeter` and `second_to_microsecond` (the short ones could be aliases to the long ones). But whatever, it is nitpicking.

I think it's a good idea to use long names. My idea was to use established unit names, but I realize now that I ended up mixing long and short SI unit names.

But let's do that in a separate PR.

marcus-oscarsson:
I would definitively vote for just having **one** function for each and I would also prefer the **long version.** I also agree with @rhfogh that the context of the code and variable naming often should be enough to understand whats meant. However, extra clarity is never a bad thing, I'm all for this if it makes the code easier to read.

marcus-oscarsson:
> value /= 1000
@rhfogh why would it be faster, because it skips a function call or some other reason ?

rhfogh:
@fabcor-maxiv 
I still think we would be better off without those functions. They bloat the code, they add little clarity, and by their very existence they push people to use them everywhere or feel naughty for refusing. And there is a potentially infinite number of them - one for each pair of scientific units..

If the units are clear from context, using unit-specific functions add very little, If the units are *not* clear from context, well, they should be. And if there are particular cases that still look obscure, you are better off with explicit comments than with this kind of function - like this one:

\# Conversion from kEv to A, wavelength = HC_OVER_E/energy
HC_OVER_E = h * c / e * 10e6

Or like this one from some of my code:
                return (
                    flux_density
                    * flux.get_dose_rate_per_photon_per_mmsq(energy)
                    * 1.0e-6  \# convert to MGy
                )

But, OK, I am not going to insist. I did Approve the PR, after all.

rhfogh:
@marcus-oscarsson 
It would be faster because is skips a function call, yes. Python function calls used to be extremely slow - the most important optimisation you could do was to reduce the number for function calls. I believe they may have improved this now (by how much?), but how much slowing down is it worth to have 49 alternative ways of doing 'value *= 1000'?

marcus-oscarsson:
> @marcus-oscarsson It would be faster because is skips a function call, yes. Python function calls used to be extremely slow - the most important optimisation you could do was to reduce the number for function calls. I believe they may have improved this now (by how much?), but how much slowing down is it worth to have 49 alternative ways of doing 'value *= 1000'?

Yes its true that function calls in Python are expensive and I agree that we should not be doing them if we don't have to. I thought that you perhaps had something else in mind as well ;). 

I'm also not a big fan of these conversion functions as long as the conversion is well understood and that its clear what the resulting value is. I have nothing against if people wants to use these conversion functions but I don't see the necessity (in most cases).

rhfogh:
Just for fun, I found this text on the linked-to article about magic numbers to avoid:

In some contexts, the use of unnamed numerical constants is generally accepted (and arguably "not magic"). While such acceptance is subjective, and often depends on individual coding habits, the following are common examples:

    
[...]

    the use of powers of 10 to convert metric values (e.g. between grams and kilograms) or to calculate percentage and [per mille](https://en.wikipedia.org/wiki/Per_mille) values

marcus-oscarsson:
:), sounds very sensible to me. Its clearly something that is quite subjective. I would not like to see that we enforce the use of these functions for trivial conversions as long as its clear whats meant by using for instance a meaning full variable name, like `distance_in_meters`. I guess these functions are useful when they add clarity and it would be up to each one of us to judge  when to use the functions or a suiting variable name and/or comment.

PR #1038, opened 2024-10-03
Prevent mockup value to be set to exception on abort

### LINKED ISSUE #1037: TypeError on aborting resolution mockup motor
When aborting the change of resolution in the mockup environment, the following error raises
Screenshot from 2024-10-03 14-34-16

I saw this error pop up when trying to abort a change in the web version of mxcube, however sending the abortion request a second time (second click on the web version) seems to fix this issue, i.e. the motor shows the value at the time of abortion and no further error message shows up.

walesch-yan:
This PR closes #1037.
Previously, aborting a value change would set the value of an mockup actuator to an exception, instead of a value. With these changes, the value remains the last actually set value.

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/ActuatorMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/ActuatorMockup.py
@@ -108,7 +108,8 @@ class ActuatorMockup(AbstractActuator.AbstractActuator):
 
     def _callback(self, move_task):
         value = move_task.get()
-        self._set_value(value)
+        if not isinstance(value, gevent.GreenletExit):
+            self._set_value(value)
 
     def _set_value(self, value):
         """


rhfogh (commented):
Could this happen in other cases where a task is spawned (which would need looking at)? 
What is actually going on?
Would it be  better to raise an error than to just keep the previous sate?

marcus-oscarsson:
@rhfogh, in principle yes. But for the case of stop/abort one would quickly notice this issue because the action would *not* stop. 

The issue is that a `GreenletExit exception` is raised in the greenlet when `kill()` or `sig interrupt`  (Ctrl+C) is performed. A greenlet returns the value returned by the greenlet or re-raises the exception that killed it, except for if that exception is `GreenletExit`. `GreenletExit` is not considered and exceptional event and is returned as a value by `get()`. 

The `GreenletExit` exception needs to be handled if communication with a device or other action is performed in a greenlet, otherwise you would have exactly this issue :)

walesch-yan:
Yes, it is exactly as @marcus-oscarsson described it. In regard to your last question, I might add that raising an error here would lead to the previous problem, that every time `abort()` is called, an error would be returned, while exiting and keeping the previous state is actually the expected behavior in that case.

This behavior  was the reason for a UI crash that happens in the mockup version of the web application (see mxcube/mxcubeweb#1435)

rhfogh:
@walesch-yan @marcus-oscarsson Makes sense.

PR #1039, opened 2024-10-08
[Harvester] FIx typo

jbflo:
- FIx EMBLFlexHCD inheritance Object is a module, not a class Error
- remove unnecessary wait command 
- fix Typo, call set method instead


rhfogh (approved):
Sure, why not?

PR #1040, opened 2024-10-13
Added spacegroup synonyms, for interpreting user input names

rhfogh:


rhfogh:
Note
1) 'synonyms' is a new field, so it can break nothing

2) Space group names and synonyms follow CCP4 v9 symop.lib, with or without internal spacings, with the following additions:

SG3   'P 2'
SG4   'P 21'
SG5   'C 2
SG6   'P m'
SG7   'P c'
SG8   'C m'
SG9   'C c'
SG10  'P 2/m'
SG11  'P 21/m'
SG12  'C 2/m'
SG13  'P 2/c'
SG14  'P 21/c'
SG15  'C 2/c'
SG17  'P2212', 'P 2 21 2', 'P2122', 'P 21 2 2'
SG18  'P21221', 'P 21 2 21', 'P22121', 'P 2 21 21'

Note: For SG17 and SG18 these are alternate settins, but they are treated as synonyms

R versus H names:
=================
Space groups 146, 148, 155, 160, 161, 166, and 167
are given names starting with 'R'. Name versions starting with 'H' are treated as synonyms.
No distinction is made at the name/synonym level between rhombohedral and hexagonal settings

marcus-oscarsson:
There seems to be some linting issue with ` mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py`

rhfogh:
@marcus-oscarsson Yeah, I noticed - but is there no way of getting the actual output of black off the test? then I could see what the problems were, and fix them. I have run  black  on the files before committing (through Pycharm), and they pass. Running the precommit locally would require setting up a new environment with Python 3.8 especially to run the precommit test - a bit messy - and my first attempt to install flake8 in Pycharm failed.

The quick solution would be to remove GphlWorkflow.py from testing (the way it was until recently). At some point I probably need to fiddle until I manage to get the tests running locally, but it is not how I want to spend an entire day right now, if I can avoid it.

fabcor-maxiv:
> is there no way of getting the actual output of black off the test? then I could see what the problems were

I was thinking about that recently... If I am not mistaken it should be possible to have this feedback directly in the GitHub pull request. I want to investigate this.

> Running the precommit locally would require setting up a new environment with Python 3.8 especially to run the precommit test

This [requirement was removed on mxcubeweb](https://github.com/mxcube/mxcubeweb/pull/1417/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9). I guess we should do it here as well, then.

As it is now, in principle, simply having installed Python 3.8 on the system should be good enough. What I mean by that is that there should not be no need to create any kind of environment for Python 3.8, and there should be no need to install pre-commit for Python 3.8.

But yes, we should make it consistent with what we have now on mxcubeweb.

marcus-oscarsson:
Agree with @fabcor-maxiv, that we should remove Python 3.8 from pre-commit. You could try to run black on the file from the command line perhaps the pycharm integration prevents some formatting somehow ?

rhfogh:
Thanks.  That did the trick. Running pre-commit in mxcubeweb environment, and removing default_language_version from mxcubecore default, and it all worked.

For some reason the pycharm black did not change 
                round(reslimits[0], resolution_decimals) + 0.1 ** resolution_decimals
to
                round(reslimits[0], resolution_decimals) + 0.1**resolution_decimals
but the pre-commit black did.

But who cares - I can run the tests now.

PR #1043, opened 2024-10-17
Re-run all pre-commit hooks

fabcor-maxiv:


fabcor-maxiv:
I re-ran all the pre-commit hooks.
Somehow, surprisingly, I do not know why, there were things to fix.
This does not make much sense to me, I do not understand how this could happen...

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -2356,64 +2355,64 @@ def process_centring_request(self, payload, correlation_id):
...
+        #                 result = self._return_parameters.get()
+        #                 if result is StopIteration:
+        #                     return StopIteration
+        #             finally:
+        #                 dispatcher.disconnect(
+        #                     self.receive_ok_cancel,
+        #                     self.PARAMETER_RETURN_SIGNAL,
+        #                     dispatcher.Any,
+        #                 )
+        #                 self._return_parameters = None
```
fabcor-maxiv, line 2415: @rhfogh You might want to see this...

This changes the indentation of your commented code. It might be an issue for you later if/when you want to re-enable this code. One solution would be to remove the code, since you can always find it in the history of the git repository,

marcus-oscarsson:
Thanks, Ill let @rhfogh merge once he is ok/reviewed with the changes

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
rhfogh: @fabcor Yeah, I noticed it changed the indentation, but black insisted on doing that. I would have preferred to keep it in (as a useful reminder that there is a TODO here), but since we cannot properly merge changes without doing as black tells us, removing the section is probably the best choice.

rhfogh:
This is done already, now through PR #1040

PR #1045, opened 2024-10-18
Update pre-commit hooks

fabcor-maxiv:


marcus-oscarsson:
Its a bit strange that we need to do all those linting updates, but I guess some configuration must have changed since the first time we ran it.

marcus-oscarsson:
Looks good to me :), thanks !

fabcor-maxiv:
> Its a bit strange that we need to do all those linting updates, but I guess some configuration must have changed since the first time we ran it.

I think that in this pull request, the changes are mostly due to black. Black updates its style every year. And we were using the 2022 style. I quite like the style improvement.

Looks like most of the changes are something like the following, which I believe is indeed much better:

```diff
-        self.energy_scan_parameters[
-            "transmissionFactor"
-        ] = self.transmission.get_value()
+        self.energy_scan_parameters["transmissionFactor"] = (
+            self.transmission.get_value()
+        )
```

marcus-oscarsson:
Yes I find it nicer too

PR #1047, opened 2024-10-18
Enforce usage of isort

fabcor-maxiv:
Update the configuration for isort.

Enforce usage of isort via a pre-commit hook.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,11 +1,4 @@
 [settings]
-; see https://github.com/psf/black
-multi_line_output=3
-include_trailing_comma=True
-force_grid_wrap=0
-combine_as_imports=True
-use_parentheses=True
-line_length=88
-sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
-default_section=THIRDPARTY
-ensure_newline_before_comments = True
+profile=black
+py_version=38
+force_grid_wrap=2
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,6 +26,11 @@ repos:
           - --remove-duplicate-keys
           - --ignore-pass-after-docstring
 
+  - repo: https://github.com/pycqa/isort
+    rev: 5.13.2
+    hooks:
+      - id: isort
+
   - repo: https://github.com/psf/black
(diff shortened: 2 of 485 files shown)


fabcor-maxiv:
I saw that we have `.isort.cfg` but it does not appear to be used. So I updated the config in this file to something that I hope is more modern and fits better with black, and added a pre-commit hook for isort to make sure the isort rules are enforced.

So a lot of files were changed, but it should hopefully not have broken anything...

...unless we have cases where the import order is meaningful...

...and now writing this, it makes me think that we do have the gevent import that might be sensitive to the order, so maybe it does break things. Needs to be investigated.

fabcor-maxiv:
> ...and now writing this, it makes me think that we do have the gevent import that might be sensitive to the order, so maybe it does break things. Needs to be investigated.

Well, actually, no. Regarding gevent what matters is that the very first import in the application (i.e. mxcube-web or mxcube-qt) should be a gevent import. But mxcubecore being a library, it should not matter.

I do not know if there are other cases of meaningful import order besides gevent.

fabcor-maxiv:
What do people think of this?

@marcus-oscarsson @rhfogh @elmjag

rhfogh (approved):

rhfogh:
Fine with me. There was one tricky import set-up in GphlWorkflowConnection, and I checked: that is preserved.

Not sure I always prefer their style (what is wrong with having two imports on the same line?), but it is always good to have things standardised.

fabcor-maxiv:
> Not sure I always prefer their style

It is also not my preferred style. If it were just me I would choose quite different settings. I went with something I thought would be an okay compromise


> what is wrong with having two imports on the same line?

Typically for these things, I think the idea is that it makes diffs easier to read when things are added or removed. Indeed, having only one import per line makes for nicer diffs when adding or removing imports. Just like with dicts, lists, function parameters, and so on. But with these settings we are not pushing this logic too far, so I guess it is an okay compromise for people who are not as extreme as I would be.

There is quite a lot flexibility in the settings. For example see this: <https://pycqa.github.io/isort/docs/configuration/multi_line_output_modes.html>. But there is much more.

rhfogh:
Well, the style we have chosen is certainly better than most of the alternatives.

marcus-oscarsson:
Ok, why not

PR #1048, opened 2024-10-18
Snapshots web

walesch-yan:
This PR includes changes to the snapshot routine, which did not work on the web version before. 
It moves the complexity of the `take_snapshots` functionality from the cameras to the `sampleview` objects. Furthermore it allows an overlay(grids/points) as input, which are put as additional layer on the latest image. Then it returns the combined image.

We further changed a few references of different snapshot calls, so that the correct ones are called (between `get_snapshot` which returns a snapshot, `take_snapshot` which is mainly used to reduce code duplication between the other two and `save_snapshot` for saving the snapshot to the disk instead of returning it).

Furthermore, the routines in `multicollect` object and `minidiff` were adjusted to utilize the new functions

### DIFF
--- a/mxcubecore/HardwareObjects/GenericDiffractometer.py
+++ b/mxcubecore/HardwareObjects/GenericDiffractometer.py
@@ -793,9 +793,10 @@ class GenericDiffractometer(HardwareObject):
 
     def get_snapshot(self):
         """
-        Get sample_view snapshot
+        Get snapshot from sample view
 
-        :returns: bytes object of current camera image
+        Returns:
+            bytes: A bytes object of the current camera image.
         """
         if HWR.beamline.sample_view:
             return HWR.beamline.sample_view.get_snapshot()
--- a/mxcubecore/HardwareObjects/GenericDiffractometer.py
+++ b/mxcubecore/HardwareObjects/GenericDiffractometer.py
@@ -792,8 +792,13 @@ class GenericDiffractometer(HardwareObject):
     #     return self.current_positions_dict.get("phi")
 
     def get_snapshot(self):
+        """
+        Get sample_view snapshot
+
+        :returns: bytes object of current camera image
+        """
         if HWR.beamline.sample_view:
-            return HWR.beamline.sample_view.take_snapshot()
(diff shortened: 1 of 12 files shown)


walesch-yan:
I will leave this as a draft, as we currently still need to do some clean-up of the old routine and some related unused functions

walesch-yan:
We finished the clean-up and removed/replaced all references to the old routine

marcus-oscarsson:
I'm sorry @walesch-yan, could you re-base and resolve the conflicts from the previously merged PR (import order changed)

walesch-yan:
@marcus-oscarsson Yes, no problem

marcus-oscarsson:
Thanks, your a star !

walesch-yan:
Should be fine now

mxcubecore/HardwareObjects/GenericDiffractometer.py
```diff
@@ -793,9 +793,10 @@ def get_motors(self):
 
     def get_snapshot(self):
         """
-        Get sample_view snapshot
+        Get snapshot from sample view
 
-        :returns: bytes object of current camera image
+        Returns:
+            bytes: A bytes object of the current camera image.
```
walesch-yan: Changed the docstring to conform with Google style

mxcubecore/HardwareObjects/abstract/AbstractMultiCollect.py
```diff
@@ -148,7 +150,7 @@ def prepare_oscillation(
     def set_detector_filenames(
         self, is_first_frame, frame_number, start, filename, shutterless
     ):
-        pass
+        passself.number_of_snapshots = 0
```
marcus-oscarsson: Watch out, I don't think this is called :). But it will result in an error. I'm not sure if it was actually added by you, I think its the result of a merge.

walesch-yan: You are absolutely right, very well spotted thanks! :) I fixed it and checked for other possible merge errors, but seems fine by me

marcus-oscarsson:
:+1:

PR #1056, opened 2024-10-25
[WF] - BES and Gphl Workflow integration

marcus-oscarsson:
Changes made after latest test with Gphl workflows, mostly done by @olofsvensson and @rhfogh. @rhfogh Could you have a quick look at the Gphl specific parts ?


rhfogh (approved):
One problem: There is a bug fix missing that I have. But I'll just merge this and make a new PR ASAP to include that fix.

PR #1059, opened 2024-10-25
set correct return type

walesch-yan:
The function `dc_from_output` previously return a `NoneType` object, while a `list` is expected, as it does in the parent class. Hence, I changed it to return an empty list instead.

This type mismatch is at the root of this issue mxcube/mxcubeweb#1309


fabcor-maxiv:
Feels like it is a perfect occasion to add type hints

walesch-yan:
You are right, almost missed a perfect opportunity, I will work on that

walesch-yan:
I added type hints for the return types of every function in `EDNACharacterisationMockup` and `EDNACharacterisation`, where the types were not previously declared by a docstring. 
I'm still considering to add them to all other function as well for the sake of consistency, even when the docstring already hints at the type. Any opinions?

mxcubecore/HardwareObjects/mockup/EDNACharacterisationMockup.py
```diff
@@ -17,31 +17,39 @@
...
+        return XSDataInputMXCuBE.parseString(self.edna_default_input)
 
-    def characterise(self, edna_input):
+    def characterise(self, edna_input) -> XSDataResultMXCuBE:
         return XSDataResultMXCuBE.parseString(edna_test_data.EDNA_RESULT_DATA)
 
-    def is_running(self):
-        return
+    def is_running(self) -> bool:
+        return False
```
walesch-yan, line 45: Here, I also added a return value for the `input_from_params` and `is_running` functions, to be consistent with the output types

fabcor-maxiv:
It is better to avoid having the type in both the annotation and the docstring, in case we update one but forget to update the other.

My preference is for types to be in the type hint annotation rather than in the docstring.

Sphinx can recognize and use the type annotations to generate the documentation, no need to have it in the docstring as well.

walesch-yan:
> It is better to avoid having the type in both the annotation and the docstring, in case we update one but forget to update the other.

That is a good point. 
Personally, I also prefer the type hints annotation, so I will make the changes as long as there are no objections

fabcor-maxiv (approved):

marcus-oscarsson:
Very nice @walesch-yan, great initiative !

walesch-yan:
Thanks!

PR #1062, opened 2024-10-28
adds support for configuring hardware objects with YAML files

elmjag:
Adds support for _new_ YAML format for hardware objects configuration files.

With this changes it's possible configure all of the beamline hardware objects using YAML files. See https://mxcubecore--1062.org.readthedocs.build/en/1062/dev/configuration_files.html#yaml-configured-objects for details on the _new_ YAML config format.

This is a breaking change, see https://mxcubecore--1062.org.readthedocs.build/en/1062/dev/yaml_conf_migration.html for details on known non-backward compatible changes.

A level of backward compatibility with XML hardware configuration files is retained, so it should be possible to mix XML and YAML configuration files. See the above document for details.

Putting this as Draft for now. Let's discuss first discuss these changes at the upcoming developers meeting.


marcus-oscarsson:
@elmjag, nice coincidence, I just sent you an email regarding this. As I mentioned in the mail it looks very well overall

I had some question about the `id` attribute that is perhaps easier to discuss here :)

marcus-oscarsson:
Well done @elmjag and also @rhfogh very nice progress !

deprecated/HardwareObjects/Camera.py
```diff
@@ -482,15 +482,15 @@ def takeSnapshot(self, *args, **kwargs):
                             # img.save(*args)
                         except Exception:
                             logging.getLogger("HWR").exception(
-                                "%s: could not save snapshot", self.name()
+                                "%s: could not save snapshot", self.id
```
marcus-oscarsson, line 485: I was wondering where the `id` attribute gets assigned and I noticed that `.name()` is sometimes replaced by `.name` and sometimes with `id`

marcus-oscarsson:
At the same time as Elmir made this PR I also sent en email regarding the same topic. After an exchange with Elmir we decided that it was better to extract my comments from the mail and share them here:

**Objects and Configuration:**
 
I was wondering if we should avoid assignment at initialization of the roles in the objects section to a "role" attribute of the object. I think that this "implicit" assignment may lead to confusion and cause issues with existing code.
 
https://mxcubecore--1052.org.readthedocs.build/en/1052/dev/yaml_conf_migration.html#property-annotation-for-child-objects
 
We could perhaps add an attribute that contains the objects just like we have `.config`. I would perhaps not suggest that we call this attribute objects because that is a bit too general and have lots of meaning, but how about `.roles` ?
 
In the context of validation both the `config` section and `objects` section  can be specified by a PyDantic (or other if someone prefers another solution) model on the object and in this way be typed and well defined.

**Commands and channels:**

I'm looking through the last changes and its looking good as far as I  can see. I noticed few nitty gritty things noted below.
 
  - "polling_period" in the table should perhaps be "poll" ?
 
    https://mxcubecore--1052.org.readthedocs.build/en/1052/dev/commands_channels.html#example
 
  - For the exporter example, the protocol should be "Exporter" I imagine :) its written Tango,

    https://mxcubecore--1052.org.readthedocs.build/en/1052/dev/commands_channels.html#example

I'm not very familiar with EPICS but what you suggest looks reasonable and it will work nicely with Tango and Exporter.
 
These are just a few details, its looking really good overall !

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,297 @@
...
+The `Volume` channel will be mapped to the _currentVolume_ attribute of the tango device.
+The _currentVolume_ attribute's value will be polled every 1024 milliseconds.
+
+## Exporter Protocol
+
+The format for specifying _exporter_ `CommandObject` and `ChannelObject` objects is as follows:
+
+```yaml
+class: <hardware-object-class>
+tango:
```
marcus-oscarsson: Exporter :)

docs/source/dev/commands_channels.md
```diff
@@ -0,0 +1,297 @@
...
+
+`channels` is a dictionary where each key specifies a `ChannelObject` object.
+The key defines the MXCuBE name for the channel.
+The values specify an optional dictionary with configuration properties for the `ChannelObject` object.
+The following configuration properties are supported:
+
+| property       | purpose                            | default             |
+|----------------|------------------------------------|---------------------|
+| attribute      | tango attribute name               | MXCuBE channel name |
+| polling_period | polling periodicity, milliseconds  | polling is disabled |
```
marcus-oscarsson, line 94: `poll` and not `polling_period` ?

rhfogh:
@marcus-oscarsson 

I do not understand what you mean when you say that
`I was wondering if we should avoid assignment at initialization of the roles in the objects section to a "role" attribute of the object. I think that this "implicit" assignment may lead to confusion and cause issues with existing code.`
In my original draft it was mandatory that roles had to match pre-existing attributes in the classes. You got an error if you tries to set a role that was not in the object already. In the current draft the error message has been replaced by a warning, but I would hope that this is temporary, and that we can go back to enforcing that roles must be pre-existing in the code. 

As I  understand it, the roles are an integral part of the class structure, and defined in the code. That makes them reliable and makes it possible for linters to check for uses of missing roles. Using Pydantic instead would make a lot of sense if you wanted the roles to be variable and locally configurable, or generally if you wanted it to be easier to change them around. But personally, that is exactly what I would not want.  What is the advantage, as you see it, of defining roles in Pydantic instead of  simply hard-coding them in? If it is the ability to have additional roles in site-specific code, could we not instead add those attributes in site-specific subclasses?.

docs/source/dev/commands_channels.md
fabcor-maxiv: I'd rather have `polling_period` everywhere. Unless I misunderstood something.

marcus-oscarsson:
>In my original draft it was mandatory that roles had to match pre-existing attributes in the classes

I would be alright with that as well, its less implicit and the result would be very similar. I don't see an issue with this If these attributes are defined before hand and an error is raised if the attribute does not exist. Both would lead to something fairly well defined. Adding "public" attributes to an HardwareObject automatically as a part of configuration otherwise IMO has to be a very special case. Isolating these these attributes in a object specifically for that purpose would then be a better alternative. We risk to overwrite attributes the author meant to use for other things if its not done without warning or error.

But I'm as I mentioned in the beginning fine with this if the attribute (and preferably its type) is defined on the HardwareObject before the configuration step is done.

docs/source/dev/commands_channels.md
marcus-oscarsson: It does not matter to me if its `poll` or `polling_period`, however the example configuration uses `poll` so in that case the configuration file examples should be changed ?

elmjag: Ah, ohh, the documentation currently uses `poll` and `polling_period` intermixed. 

We agreed in a previous PR that is should be `polling_period`. I'll update documentation so _only_ `polling_period` is used.

docs/source/dev/commands_channels.md
elmjag: fixed

docs/source/dev/commands_channels.md
elmjag: It should be `polling_period` everywhere now.

elmjag:
These changes have been merged via this PR: https://github.com/mxcube/mxcubecore/pull/1177

Issue #1051, opened 2024-10-21
Camera width/height should be numeric


axelboc:
In https://github.com/mxcube/mxcubeweb/pull/1462#discussion_r1808431660, I've noticed that `HWR.beamline.sample_view.camera.get_width/height()` returns strings instead of numbers. Can this be changed by any chance? This would avoid a conversion in the front-end.

walesch-yan:
Did you use a specific camera device to test this? At least for the mockup camera `HWR.beamline.sample_view.camera.get_width/height()` always returns an int when I tried it out.

axelboc:
No, just the default mock environment. This is what I get from the `/sampleview/camera` endpoint (defined in `samplecentring.py`):

![image](https://github.com/user-attachments/assets/43647dd6-1610-42b5-bfbd-17651ab6f36c)

You're right, though, it doesn't come from `get_width/height()` but from `HWR.beamline.sample_view.camera.get_stream_size()`, because the default config has `VIDEO_FORMAT: MPEG1` under `mxcube` in `server.yaml`: https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/core/components/beamline.py#L109-L112

walesch-yan:
Yes, you are right `HWR.beamline.sample_view.camera.get_stream_size()` does return a string, and does so for every type of camera. And I agree that a function called `get_stream_size` should return a tuple of `int/float` rather than a list of `strings`. 

I think the reason behind it is that we set the stream_size as string since it needs to be for the usage of the video_streamer, but we could convert it in the `get_stream_size` (which is never called in `mxcubecore`).

PR #1065, opened 2024-10-29
change stream_size type and add type hints

walesch-yan:
As discussed in #1051, I changed the type of `_current_stream_size` to be a tuple of `int` for `TangoLimaMpegVideo`, `TangoLimaVideoLoopback` and `MDCameraMockup`. We now only convert to a string, when we start the `video_streamer`. 

On the occassion I added type hints for every function in `TangoLimaMpegVideo` and `MDCameraMockup` and removed the unused `port` parameter from the resp. `start_video_stream_process` functions and an unused import from `TangoLimaMpegVideo`.

Additionally, I was wondering if `TangoLimaVideoLoopback` is actually used? The class seems to be outdated compared to the others, it does not make use of the `video-streamer` and while it has `set` and `get` functions for the `current_stream_size`, it never uses the variable.

### DIFF
--- a/mxcubecore/HardwareObjects/TangoLimaMpegVideo.py
+++ b/mxcubecore/HardwareObjects/TangoLimaMpegVideo.py
@@ -13,9 +13,9 @@ Example configuration:
 </object>
 """
 
-import os
 import subprocess
 import uuid
+from typing import List
 
 import psutil
 
@@ -27,7 +27,7 @@ class TangoLimaMpegVideo(TangoLimaVideo):
         super(TangoLimaMpegVideo, self).__init__(name)
         self._format = "MPEG1"
         self._video_stream_process = None
-        self._current_stream_size = "0, 0"
+        self._current_stream_size = (0, 0)
         self.stream_hash = str(uuid.uuid1())
         self._quality_str = "High"
         self._QUALITY_STR_TO_INT = {"High": 4, "Medium": 10, "Low": 20, "Adaptive": -1}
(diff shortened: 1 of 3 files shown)


mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
```diff
@@ -62,42 +63,42 @@ def poll(self):
...
         # return 576 # JN ,20140807,adapt the MD2 screen to mxCuBE2
         return 493
 
-    def set_live(self, state):
+    def set_live(self, state) -> bool:
         self.liveState = state
         return True
 
     def imageType(self):
         return None
```
walesch-yan, line 95: Actually I was not sure, what the purpose of this function is, is this some leftover code?

mxcubecore/HardwareObjects/TangoLimaMpegVideo.py
```diff
@@ -40,26 +40,26 @@ def init(self):
...
-    def set_quality(self, q):
+    def set_quality(self, q) -> None:
         self._quality_str = q
         self._quality = self._QUALITY_STR_TO_INT[q]
         self.restart_streaming()
 
-    def set_stream_size(self, w, h):
-        self._current_stream_size = "%s,%s" % (int(w), int(h))
+    def set_stream_size(self, w, h) -> None:
+        self._current_stream_size = (int(w), int(h))
```
walesch-yan, line 52: From previous discussions, it might make sense to keep `_current_stream_size` as `str` an convert to `int` when using the get function, instead I decided to use a tuple of ints, because we actually did a double conversion before to int and then to str on the input parameters. Hence I only convert to `str` when starting the `video-streamer` process.

mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
marcus-oscarsson: It looks like a left over, you can probably remove it. Just check that its not referenced somewhere. That its camel cased tells me/us that its rather old.

marcus-oscarsson:
`TangoLimaVideoLoopback` Is no longer used we can safely remove it :)

marcus-oscarsson:
Does it mean that we have to update the UI bit  as well ?

walesch-yan:
> `TangoLimaVideoLoopback` Is no longer used we can safely remove it :)

Okay, I will remove it then in a future Pr:)

walesch-yan:
> Does it mean that we have to update the UI bit as well ?

When I tried it on my pc it worked without making any changes to the UI, my guess is that since we did a conversion in the UI, we now basicaly convert from int to int, which will not break anything but can then be removed. Maybe @axelboc knows more about it and if these changes could break parts of the UI.

marcus-oscarsson:
:), working late I see. Sounds good. I suspected that it was the case but just wanted to give a heads up in case an update is needed.

axelboc:
I'll open a PR tomorrow morning.

PR #1066, opened 2024-10-31
Refactoring ISPyBClientMockup for authnentication tests

dominikatrojanowska:
The [PR](https://github.com/mxcube/mxcubeweb/pull/1455) is dependent on this one.
The change provided here allows to perform correctly authentication tests that are using CREDENTIALS_1: https://github.com/mxcube/mxcubeweb/blob/develop/test/test_authn.py#L24

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
@@ -95,9 +95,9 @@ class ISPyBClientMockup(HardwareObject):
 
     def login(self, login_id, psd, ldap_connection=None, create_session=True):
         # to simulate wrong loginID
-        if login_id != "idtest0":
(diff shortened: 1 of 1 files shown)


fabcor-maxiv (approved):

marcus-oscarsson:
Thanks

marcus-oscarsson:
I'm sorry guys but there is an issue with this code:

`get_proposal` at this line https://github.com/mxcube/mxcubecore/pull/1066/files#diff-49fa365d70805e215a94138cdd0f725d690114c71b0c7cf53e1ab6d10d055e5fR125 will always return the same proposal regardless of login used. 

Hence logging-in with idtest1 returns the proposal of idtest0, right ? This means that the tests for `mxcubeweb` will fail. If you agree I will disable this test in `mxcubeweb` for the moment and then you can have a closer look at returning the right proposal (when you get some time)

PR #1068, opened 2024-11-01
queue models: handle omitted 'centred_pos' argument

elmjag:
In the utility function` to_collect_dict()`, take into the account that `centred_pos` argument can be omitted.

Don't call `centred_pos.get_index()` if `centred_pos` is omitted or is None.

For background information, the call to `centred_pos.get_index()` have been recently introduced in this PR: https://github.com/mxcube/mxcubecore/pull/1056

The line in the diff: https://github.com/mxcube/mxcubecore/pull/1056/files#diff-5194ff6aa825f5865aa6953d626e6a91865a9be8ff8538f47613e62c80712f93R2705

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -2702,7 +2702,9 @@ def to_collect_dict(data_collection, session, sample, centred_pos=None):
                 data_collection.experiment_type
             ],
             "skip_images": acq_params.skip_existing_images,
-            "position_name": centred_pos.get_index(),
+            "position_name": (
+                centred_pos.get_index() if centred_pos is not None else None
+            ),
             "motors": centred_pos.as_dict() if centred_pos is not None else {},
             "ispyb_group_data_collections": data_collection.ispyb_group_data_collections,
             "workflow_parameters": data_collection.workflow_parameters,


mxcubecore/model/queue_model_objects.py
```diff
@@ -2702,7 +2702,9 @@ def to_collect_dict(data_collection, session, sample, centred_pos=None):
                 data_collection.experiment_type
             ],
             "skip_images": acq_params.skip_existing_images,
-            "position_name": centred_pos.get_index(),
+            "position_name": (
+                centred_pos.get_index() if centred_pos is not None else None
```
rhfogh, line 2706: Why not just
   centred_pos.get_index() if centred_pos  else None
?
Or do you explicitly what to get an error if centered_pos is 0, or an empty collection?

elmjag: I'm copying the style from the line below:

`"motors": centred_pos.as_dict() if centred_pos is not None else {},`

As far as I understand, in python it's considered a good style to explicitly check for None values with `val is not None`.

marcus-oscarsson: Thanks !

rhfogh: @elmjag It depends on context, I would say. In the general case it is better to check explicitly for None, in case the value unexpectedly turned out to be something else that evaluated False, like 0.0 or an empty dictionary and you got the wrong behaviour. In this case anything that has an as_dict() function will evaluate as True, and anything that evaluates as False should return an empty dictionary (or an error). The  question question is what result you want if centred_pos is, say, a dictionary (empty or not) - an empty dictionary or an error? 

But we are down in the nitpicking here - either will do.

PR #1070, opened 2024-11-04
Use types from typing for Type Hints

walesch-yan:
In #1065, I mistakenly added Type hints that use the built-in `tuple` type, which is introduced in `Python 3.9` and hence not supported in `Python 3.8`. I replaced these and some other instances I found in the code by the corresponding types from `typing`.

This mismatch seems to be a reason for a failing test in the web version mxcube/mxcubeweb#1483


marcus-oscarsson:
Nice, well done :)

rhfogh:
One more reason to upgrade to a newer minimum Python.

PR #1079, opened 2024-11-07
Initial AbstractLims with ICAT integration - Part 1

marcus-oscarsson:
This PR contains the initial work on an LIMS abstraction for MXCuBE making the LIMS system integration easier. This work provides:

  - A light abstraction to facilitate LIMS integration
  - A means to use several LIMS systems at the same time: for instance ISPyB and ICAT

We have tried to document whats done in more detail in: 
- [lims_integration](https://mxcubecore--1079.org.readthedocs.build/en/1079/dev/lims_integration.html)
- [AbstractLims](https://mxcubecore--1079.org.readthedocs.build/en/1079/dev/autosummary/mxcubecore.HardwareObjects.abstract.AbstractLims.html#module-mxcubecore.HardwareObjects.abstract.AbstractLims)

    
It's well understood that the API might need to evolve and that there is a need for more well defined data structures, as outlined in the documentation and discussed within the community in the past. This PR is therefore titled with "Part 1"

This work really contains over 30 commits made by Alex (@antolinos) (and some by me :) ) and have been conducted over the last year. The original commits have been re-based and squashed as it was the easiest way to clearly separate this work from other parallel development. We hope that it does not make things harder to follow. It should really be stressed that most of the credit for this work goes to Alex :) 

During the last 6 months we have also taken a substantial amount of time to test and validate this work on the ESRF beamlines so we know that it works well. So its more than just a nice idea, it also works ! :). 

We will follow up this PR with a corresponding PR for `mxcubeweb`


mxcubecore/HardwareObjects/abstract/ISPyBAbstractLims.py
fabcor-maxiv, line 1: These file names drive me crazy...

I would recommend that if possible more sane names are used. At least for new code.

For example `ISPyBAbstractLims.py` should really be `ispyb_abstract_lims.py`, but if it is too big of a change, then `IspybAbstractLims.py` would be acceptable. In both cases the class name would be `IspybAbstractLims`. I know that the stylized marketing name is `ISPyB`, but we are writing code here not marketing prose. Underscores in "snake case" and upper case letter in "upper camel case" indicate word boundaries.

marcus-oscarsson: I agree that would be really nice with snake cased file names. The issue today is that files containing a hardware object needs to have the same name as the class. The hardware object loader expects this convention. I have really no opinion between `IspybAbstractLims` and `ISPyBAbstractLims` its the same to me. I guess the name `ISPyBAbstractLims` was kept to be consistent with the previous `ISPyBClient`, otherwise `IspybAbstractLims` will be just as good.

docs/source/dev/lims_integration.md
```diff
@@ -0,0 +1,229 @@
...
+  <after_offset_days>20</after_offset_days>
+  <compatible_beamlines>ID30A-1</compatible_beamlines>
+  <!-- the list of sessions will be filtered by the override_beamline_name, this is needed when
+  HWR.beamline.session.beamline_name does not match with the name of the instrument in the catalog -->
+  <override_beamline_name>ID30A-1</override_beamline_name>
+
+  <data_portal_url>https://data2.esrf.fr/investigation/{id}/datasets</data_portal_url>
+  <logbook_url>https://data2.esrf.fr/investigation/{id}/logbook</logbook_url>
+  <user_portal_url>
+    https://smis.esrf.fr/manageAForm.do?action=view&amp;expSessionVO.pk={id}</user_portal_url>
```
axelboc: Are these configuration options maybe a bit too ESRF-centric? Should the list of URLs be more generic?

antolinos: Yes, I think it would be a good idea to anonymize this.

marcus-oscarsson: :), Yes

PR #1086, opened 2024-11-09
don't report expected failure as an error

### LINKED ISSUE #1085: Queue model object: HWR import failure is not an error
```python
try:
    from mxcubecore import HardwareRepository as HWR
except ImportError:
    logging.getLogger("HWR").exception("Could not import HardwareRepository")
```

In BES this appears like an actual failure while it is expected which confuses debugging.

woutdenolf:
The traceback at error level is confusing when debugging other things.

### DIFF
--- a/mxcubecore/HardwareObjects/Native/__init__.py
+++ b/mxcubecore/HardwareObjects/Native/__init__.py
@@ -106,7 +106,7 @@ def queue_update_result(server_hwobj, node_id, html_report):
     return result
 
 
-def queue_get_full_path(server_hwobj, suddir, tag):
+def queue_get_full_path(server_hwobj, subdir, tag):
     """ """
     return HWR.beamline.session.get_full_path(subdir, tag)
 
--- a/mxcubecore/HardwareObjects/Native/__init__.py
+++ b/mxcubecore/HardwareObjects/Native/__init__.py
@@ -137,7 +137,15 @@ def queue_get_model_code(server_hwobj):
     # standard Python modules, so we only need to send over the code for the
     # queue model itself
 
+    queue_model_enumerables_code = inspect.getsource(queue_model_enumerables)
+
+    queue_model_objects_code = inspect.getsource(queue_model_objects)
+    queue_model_objects_code = queue_model_objects_code.replace(
+        "from mxcubecore import HardwareRepository as HWR",
+        "HWR = None  # HardwareRepository not available in this context",
+    )
+
     return [
-        (queue_model_enumerables.__name__, inspect.getsource(queue_model_enumerables)),
-        (queue_model_objects.__name__, inspect.getsource(queue_model_objects)),
+        (queue_model_enumerables.__name__, queue_model_enumerables_code),
+        (queue_model_objects.__name__, queue_model_objects_code),
     ]
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -45,15 +45,7 @@ except Exception:
         "Cannot import dependencies needed for GPHL workflows - GPhL workflows might not work"
     )
 
-# This module is used as a self contained entity by the BES
-# workflows, so we need to make sure that this module can be
-# imported eventhough HardwareRepository is not available.
-try:
-    from mxcubecore import HardwareRepository as HWR
-except ImportError:
-    logging.getLogger("HWR").warning("Could not import HardwareRepository")
-    HWR = NotImplemented
-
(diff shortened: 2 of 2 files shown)


woutdenolf:
Alternatively we could do this

```python
try:
    from mxcubecore import HardwareRepository as HWR
except ImportError:
    if os.getenv("IMPORTING_OVER_RPC") != "1":
        raise
    HWR = None
```

Or just

```python
from mxcubecore import HardwareRepository as HWR
```

and patch in `queue_get_model_code`

```python
def queue_get_model_code(server_hwobj):
    queue_model_enumerables_code = inspect.getsource(queue_model_enumerables)

    queue_model_objects_code = inspect.getsource(queue_model_objects)
    queue_model_objects_code = queue_model_objects_code.replace(
        "from mxcubecore import HardwareRepository as HWR",
        "HWR = None  # HardwareRepository not available in this context",
    )

    return [
        (queue_model_enumerables.__name__, queue_model_enumerables_code),
        (queue_model_objects.__name__, queue_model_objects_code),
    ]
```

I kind of prefer the second one. What do you think?

marcus-oscarsson:
I also prefer the second alternative, what do you think @rhfogh ?

marcus-oscarsson:
@woutdenolf, I think what you proposed in this PR is fine, do you prefer one of your alternative solutions instead?

mxcubecore/HardwareObjects/Native/__init__.py
```diff
@@ -106,7 +106,7 @@ def queue_update_result(server_hwobj, node_id, html_report):
     return result
 
 
-def queue_get_full_path(server_hwobj, suddir, tag):
+def queue_get_full_path(server_hwobj, subdir, tag):
```
woutdenolf, line 109: Unrelated but a fix nevertheless.

mxcubecore/model/queue_model_objects.py
```diff
@@ -42,17 +42,10 @@
...
 
-# This module is used as a self contained entity by the BES
-# workflows, so we need to make sure that this module can be
-# imported eventhough HardwareRepository is not avilable.
-try:
-    from mxcubecore import HardwareRepository as HWR
-except ImportError:
-    logging.getLogger("HWR").exception("Could not import HardwareRepository")
-
+from mxcubecore import HardwareRepository as HWR
```
woutdenolf, line 48: This module does not need to take care about dodgy usage.

mxcubecore/HardwareObjects/Native/__init__.py
```diff
@@ -137,7 +137,15 @@ def queue_get_model_code(server_hwobj):
     # standard Python modules, so we only need to send over the code for the
     # queue model itself
 
+    queue_model_enumerables_code = inspect.getsource(queue_model_enumerables)
+
+    queue_model_objects_code = inspect.getsource(queue_model_objects)
+    queue_model_objects_code = queue_model_objects_code.replace(
+        "from mxcubecore import HardwareRepository as HWR",
+        "HWR = None  # HardwareRepository not available in this context",
```
woutdenolf, line 145: Move the dodgy stuff here.

woutdenolf:
I went for the second alternative (tested with the ID30a2 simulator).

rhfogh (approved):
For now the complex stuff should indeed be in Native, not in QMO.

For the slightly longer term it does not seem tenable to use the QMO when HWR is not available, HWR is all through half the QMO classes. We really need to refactor the queueing system, to get something that 1) is done in a single, consistent way everywhere, 2) lives fully in core, not in the client side code. 3) will work seamlessly when enqueueing things from inside core HWR classes, not just when starting from a UI.

marcus-oscarsson:
I agree, that we need to change the "queue exchange" format/protocol to something other than queue mode objects. It is also something we have been discussing for sometimes and I think we are coming to a moment where that has to be done.

mxcubecore/HardwareObjects/Native/__init__.py
marcus-oscarsson: This is a tricky one :),

mxcubecore/model/queue_model_objects.py
marcus-oscarsson: Thats nice !

marcus-oscarsson:
Looks good, I'm happy if you are. I think the `.replace` stuff might cause us some interesting cases but we will hopefully not be using this way of sending task's to the queue for much longer (much being a very relative term ;) )

PR #1088, opened 2024-11-27
Updates Cats90 and all other PX1 specific classes

HilbertCorsair:


mxcubecore/HardwareObjects/Cats90.py
```diff
@@ -520,6 +541,7 @@ def init(self):
         self._chnLidLoadedSample.connect_signal("update", self.cats_loaded_lid_changed)
         self._chnNumLoadedSample.connect_signal("update", self.cats_loaded_num_changed)
         self._chnSampleBarcode.connect_signal("update", self.cats_barcode_changed)
+        self.SPECIFIC_STATES = CatsStates
```
beteva: Not needed, already assigned few lines above

beteva (commented):
Why defining PX1Resolution and not just use the resolution HO instead. You only need to configure the xml file.
Also could you explain why defining extra PX1DetectorDistance, as it could be just a TangoMotor.
Since you are on it, we use  google style doc strings to describe methods and classes documentation (please, see in the contributing guidelines for more details)

HilbertCorsair:
Hi @beteva I saw your posts after running the pre-commit pushing the files. I will remove that line from Cats90.py I missed that. You are right about the  Resolution and TangoMotor. These are the actual classes that are used and configured in the beamline_config.yml file. Although I updated the PX1Resolution and PX1DetectorDistance these are not currently being used.

HilbertCorsair:
The updates to the PX1DetectorDistance and PX1Resolution were made quite a long time ago and I decided to include them as well in this commit to update the file although they are not in use at the moment since we use the Resolution and TangoMotor classes as you pointed out.

HilbertCorsair:
I can remove those files that are not in use if you think it is best.

mxcubecore/HardwareObjects/Cats90.py
```diff
@@ -141,7 +144,25 @@ def get_sample_address(basket_number, sample_number):
             return ""
 
 
+@unique
+class CatsStates(Enum):
+    """Shutter states definitions."""
+
+    RUNNING = HardwareObjectState.READY, "RUNNING"
+    OF = HardwareObjectState.OFF, "DISABLE"
```
fabcor-maxiv: Maybe this was supposed to be `OFF` instead of `OF`?

```suggestion
    OFF = HardwareObjectState.OFF, "DISABLE"
```

mxcubecore/HardwareObjects/Cats90.py
```diff
@@ -544,6 +565,18 @@ def init(self):
 
         self.update_info()
 
+    def get_state(self):
+        """Get the device state.
+        Returns:
+            (enum 'HardwareObjectState'): Device state.
+        """
```
fabcor-maxiv:
```suggestion
    def get_state(self) -> HardwareObjectState:
        """Get the device state.

        Returns:
            Device state.
        """
```

mxcubecore/HardwareObjects/SOLEIL/PX1/PX1Beamline.py
```diff
@@ -0,0 +1,44 @@
...
+            if self.environment_hwobj is not None:
+                log.debug("PX1Beamline. moving px1environment to default phase ")
+                self.environment_hwobj.gotoDefaultPhase()
+
+            time.sleep(1)  # allow some time on quit to allow actions to be triggered
+        except BaseException as e:
+            import traceback
+
+            log.error("Error while launching beamline close procedure. %s" % str(e))
+            log.error(traceback.format_exc())
```
fabcor-maxiv: If you use `log.exception()` in an `except` block, then the exception with traceback will be automatically printed:

```suggestion
            log.exception("Error while launching beamline close procedure.")
```

<https://docs.python.org/3/library/logging.html#logging.exception>

mxcubecore/HardwareObjects/SOLEIL/PX1/PX1CatsMaint.py
```diff
@@ -1,20 +1,55 @@
...
+from CatsMaint import CatsMaint
 
 from mxcubecore import HardwareRepository as HWR
-from mxcubecore.HardwareObjects.CatsMaint import CatsMaint
+from mxcubecore.BaseHardwareObjects import HardwareObjectState
+
+log = logging.getLogger("HWR")
+
+__copyright__ = """ Copyright © 2020 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
```
fabcor-maxiv: Use the header from <https://github.com/mxcube/mxcubecore/blob/develop/py_file_header.txt>.

mxcubecore/HardwareObjects/TangoMotor.py
```diff
@@ -217,20 +204,14 @@ def start_moving(self):
                 self.cmd_on()
         # ensure that the state is updated at least once after the polling time
         # in case we miss the state update
-        gevent.spawn(self._update_state)
+        #gevent.spawn(self._update_state)
```
fabcor-maxiv: Either remove the line or leave it in, but do not comment it.

beteva:
@hellmig could you, please, check if the changes made in Cats90.py  are OK for you. Thanks

marcus-oscarsson:
@hellmig, did you have an opportunity to check if it looks good on your side ?

hellmig:
Hi @beteva and @marcus-oscarsson, sorry for not noticing the notification earlier! I'll have a look.

marcus-oscarsson:
No problem, thanks for having a look !

PR #1090, opened 2024-11-29
Update P11Collect.py

agruzinov:
Resolution was not changing if the difference was more than 0.01. This one should fix it. Tested at the beamline.


rhfogh (approved):
Yes, good.

beteva (approved):

PR #1091, opened 2024-12-03
Update P11SampleChanger.py

agruzinov:
Make sure SC tree view is loaded even if the SC is disabled. Usefull for the testing of the features that are not require SC like testing of ISPyB.


agruzinov:
Linting is somehow fishy. Need to check, closing.

PR #1093, opened 2024-12-05
GPhL bug fixes and iprovements

rhfogh:
Added 'crystal thickness' input parameter, and use to calculate good transmission values also when beam is much narrower than the crystal.

Workflow now reads also energy, oscillationRange, and exposureTime from the diffractionPlan  when in automation mode.

Bug fixes for running MAD experiments in automation mode (multiple energies), for interleaved experiments with hdf5 files (temporary patch only).


marcus-oscarsson:
Looks like there are some linting errors ;)

marcus-oscarsson:
Looks good as far I can see otherwise

PR #1096, opened 2024-12-16
[DONT MERGE] - Rebase of xml yaml conversion

marcus-oscarsson:
Rebase of xml_yaml_conversion branch on `develop`, This can of course not be merged into `xml_yaml_conversion`, its more for making the work visible.

@elmjag, @rhfogh let me know if it looks alright or @elmjag if you are already doing this and prefer doing it yourself. @rhfogh there where some conflicts with recent changes to `gphl/gphl-workflow.yml` and `/Gphl/GphlWorkflow.py` let me know if those files look alright.


CONTRIBUTING.md
```diff
@@ -111,22 +111,33 @@ The exact routine is described more preceisly in [MEP01](https://github.com/mxcu
...
- * perecent (%) for ratios like attenuation
- * keV for energy
- * K (Kelvin) for temperature
- * Å (Ångström) for resolution
- * Pixels are to be used for beam location (center)
- * Datetime YYYY-MM-DD HH:MM:SS(.ff) ,possibly with hundreds of seconds (ff), and with 24 hour clock.
 
+Units must be documented via docstrings
+for the parameters and return values of functions and methods,
+for variables,
```
rhfogh, line 117: My code has several variables called 'resolution'. If I read you right, I am supposed to rename them all to 'resolution_in_A' - even though there is not a single resolution anywhere in the body of code that is *not* in A. 

Are we serious about this? Are we going to enforce it?

By temperament I would prefer to enforce whatever rules we have - having rules but ignoring them is the worst of both worlds. But is this really a reasonable requirement?

CONTRIBUTING.md
```diff
@@ -111,22 +111,33 @@ The exact routine is described more preceisly in [MEP01](https://github.com/mxcu
...
+and for attributes of classes and objects.
+
+The units of variables within a function or method must be documented as well.
```
rhfogh, line 120: Surely this is overkill. if the function interface defines the units, why would we need to document them separately for every variable - unless, of course, the function has variables in more than one unit?

docs/source/dev/lims_integration.md
```diff
@@ -0,0 +1,230 @@
+# LIMS integration
```
rhfogh, line 1: This is good

rhfogh:
There seems top quite a few instances of 'self.name()' around, even though ConfiguredObject.name is now a property. Is this 1) a known TBD? 2) me overlooking something? 3) something that needs fixing?

rhfogh:
Would I be right in thinking that BaseHardwareObjects.HardwareObjectNode is scheduled for removal, and 9only remains to support XML config files for now?

CONTRIBUTING.md
marcus-oscarsson: This is just a re-base of what is already in develop, I did not actually add anything :)

rhfogh:
Dear MArcus,

The two files you mention look fine. I'll test after this is merged in, 
but nothing to hold up the work until then.

Yours,
Rasmus

rhfogh (approved):
Very nice work

PR #1100, opened 2025-01-13
Adds RedisMpegVideoClass for a redis camera video-streamer

HilbertCorsair:
Adds a new class - RedisMpegVideo which acts as a HardwareObject that starts the vodeostreamer for a redis camera connectrd to a redis pub/sub channel.

### DIFF
--- a/mxcubecore/HardwareObjects/RedisMpegVideo.py
+++ b/mxcubecore/HardwareObjects/RedisMpegVideo.py
@@ -1,0 +1,172 @@
+"""
+Class for streaming MPEG1 or MJPEG video with cameras Redis Pub/Sub server
+Example configuration:
+
+<object class="RedisMpegVideo">
+  <username>Camera redis</username>
+  <host>localhost</host>
+  <uri>redis://localhost:6379</uri>
+  <cam_type>redis</cam_type>
+  <port>8000</port>
+  <width>1360</width>
+  <height>1024</height>
+  <format>MJPEG</format>
+</object>
+"""
+
+import subprocess
+import uuid
+
+import psutil
+
+from mxcubecore.BaseHardwareObjects import HardwareObject
+
+
+class RedisMpegVideo(HardwareObject):
+    def __init__(self, name):
+        super().__init__(name)
+        self._video_stream_process = None
+        self._current_stream_size = "0, 0"
+        self.stream_hash = str(uuid.uuid1())
+        self._quality_str = "High"
+        self._QUALITY_STR_TO_INT = {"High": 4, "Medium": 10, "Low": 20, "Adaptive": -1}
+
+    def init(self):
+        super().init()
+        self._debug = self.get_property("debug", False)
+        self._quality = self.get_property("compression", 10)
+        self._mpeg_scale = self.get_property("mpeg_scale", 1)
+        self._image_size = (self.get_width(), self.get_height())
+        self._uri = self.get_property("uri")
+        self._host = self.get_property("host")
+        self._port = str(self.get_property("port"))
+        self._format = self.get_property("format")
+
+    @property
+    def uri(self):
+        return self._uri
+
+    @property
+    def host(self):
+        return self._host
+
+    @property
+    def format(self):
+        return self._format
+
+    @format.setter
+    def format(self, format):
+        self._format = format
+
+    @property
+    def port(self):
+        return self._port
+
+    @port.setter
+    def port(self, p):
+        self._port = str(p)
+
+    def get_width(self):
+        w = int(self.get_property("width"))
+        return w
+
+    def get_height(self):
+        h = int(self.get_property("height"))
+        return h
+
+    def get_quality(self):
+        return self._quality_str
+
+    def set_quality(self, q):
+        self._quality_str = q
+        self._quality = self._QUALITY_STR_TO_INT[q]
+        self.restart_streaming()
+
+    def set_stream_size(self, w, h):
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/RedisMpegVideo.py
```diff
@@ -0,0 +1,172 @@
...
+                    str(self._quality),
+                    "-s",
+                    self._current_stream_size,
+                    "-of",
+                    self.format,
+                    "-id",
+                    self.stream_hash,
+                    "-irc",
+                    "mxcubeweb",
+                    # "-d",
```
walesch-yan, line 131: I think the redis-channel should not be hardcoded but be configurable through the xml files, as it depends on the actual channel name

mxcubecore/HardwareObjects/RedisMpegVideo.py
```diff
@@ -0,0 +1,172 @@
+"""
+Class for streaming MPEG1 or MJPEG video with cameras Redis Pub/Sub server
+Example configuration:
+
+<object class="RedisMpegVideo">
+  <username>Camera redis</username>
+  <host>localhost</host>
+  <uri>redis://localhost:6379</uri>
+  <cam_type>redis</cam_type>
```
walesch-yan, line 9: I think there is no cam_type parameter in the class, hence we might not need to add it to an example

mxcubecore/HardwareObjects/RedisMpegVideo.py
```diff
@@ -0,0 +1,172 @@
...
+            ]
+
+            for p in ps:
+                p.kill()
+
+            self._video_stream_process = None
+
+    def start_streaming(self, _format=None, size=(0, 0), port=None):
+        _s = size
+
```
walesch-yan, line 151: I guess this can be removed as you are setting it below

walesch-yan (changes requested):
In general it looks good:).
However,  I was wondering, since the code is very similar to the one for the [TangoLimaMPEGVideoClass](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/TangoLimaMpegVideo.py), if we could merge the two classes and just difference between them by a set flag? That way, we could prevent confusion that might come with multiple different camera classes. What do you think about the feasibility of this?

mxcubecore/HardwareObjects/RedisMpegVideo.py
```diff
@@ -0,0 +1,172 @@
...
+import subprocess
+import uuid
+
+import psutil
+
+from mxcubecore.BaseHardwareObjects import HardwareObject
+
+
+class RedisMpegVideo(HardwareObject):
```
marcus-oscarsson, line 25: A solution that would result in less duplication and easier maintenance would be to merge this code with `TangoLimaMpegVideo` so that both functionalities can be achieved via configuration. It looks like the principal difference is the use of `"-irc", "mxcubeweb"` which could be added via configuration. 

The resulting class could arguably be renamed, perhaps to `MpegVideo` , what do you think @walesch-yan

marcus-oscarsson: Seeing the comment above it looks like we had the same thought :), sorry for the repetition

walesch-yan: No problem at all:) 
After some more investigation, one other important difference is that this class would directly inherit from `hardwareobjects`, where `TangoLimaMPEGVideo` inherits from `TangoLimaVideo`, which seems to contain some more properties, similar to the ones in the `RedisMPEGVideoClass`, but also some code for directly pulling images from a camera, instead of using the `video-streamer`, so we might have to check if this is still used

marcus-oscarsson: I'm pretty sure that the inheritance from `TangoLimaVideo` can be removed now, there might be some methods that needs to be moved but I think `TangoLimaVideo` can be retired soon. ;)

mxcubecore/HardwareObjects/RedisMpegVideo.py
HilbertCorsair: Good point ! I missed that one.

mxcubecore/HardwareObjects/RedisMpegVideo.py
HilbertCorsair: Thanks for the inputs. I like the idea of merging the two classes into one MpegVideo class. I am looking into it.

laispc:
Hi @HilbertCorsair , that is a good initiative! Here at ESS we are starting to also use Redis for the sample camera view, and we are doing this in a similar way to this PR. So I can help to review/test it with our case too, if this sounds okay. For now we use a site-specific class, but would be nice to have a common class like this.

The only difference is that, at `start_video_stream_process()` we don't use/pass the arguments `-hs` (host) and `-p` (port). The rest looks okay for our case too. In summary, the video stream process looks like this for us:

```
    self._video_stream_process = subprocess.Popen(
        [
            "video-streamer",
            "-uri",
            "redis://localhost:6379",
            "-irc",
            "ArinaxCamStream",
            "-id",
            self.stream_hash,
        ],
        close_fds=True,
    )
```

So it would be nice to have `-hs` and `-p` (at least) flexible. Let me know what you think.

marcus-oscarsson:
Closing this for the time being as Dan is no longer working at Soleil.

PR #1106, opened 2025-01-17
Exclude files from distribution packages

### LINKED ISSUE #1069: Package size - we are exceeding the total PyPI size limit of 10Gb
We are starting to hit the size limit, total for all packages, on PyPi. I had to remove the really old packages, that I assume that no one is using (or should not be).

Our release strategy is indeed a bit unconventional which creates quite alot of packages. However I have a feeling that our packages are quite big. 

One issue is that the `.hkli` files are quite big. The entire `/mxcubecore/configuration/` directory can probably be moved out of the package which would make the package smaller.

fabcor-maxiv:
Exclude some files from being added to distribution packages. These files are not needed in the distribution packages and they take a lot of disk space.

GitHub: fix https://github.com/mxcube/mxcubecore/issues/1069

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,12 @@ classifiers = [
     "Topic :: Scientific/Engineering",
 ]
 
+# Exclude some files from being added to *sdist* and * wheel* distribution packages
+exclude = [
+    "mxcubecore/configuration",
+]
+
+
 [tool.poetry.dependencies]
 python = ">=3.8,<3.12"
 typing-extensions = "^4.3.0"


fabcor-maxiv:
Before:

```none
total 62M
33M mxcubecore-1.203.0-py3-none-any.whl
30M mxcubecore-1.203.0.tar.gz
```

After:

```none
total 2.4M
1.4M mxcubecore-1.203.0-py3-none-any.whl
1.1M mxcubecore-1.203.0.tar.gz
```

rhfogh:
If you exclude just mxcubecore/configuration/mockup/gphl/test_samples that alone will remove 200Mb and leave the configuration data in for people to look at. Or I could find another way of dealing with that directory, if desired.

It depends whether anybody thinks those configuration data are 1) useful and 2) sufficiently up to date to make sense.

fabcor-maxiv:
OK, this PR now excludes `mxcubecore/configuration/mockup/gphl/test_samples` only. New numbers are:

```none
total 3.5M
2.1M mxcubecore-1.203.0-py3-none-any.whl
1.5M mxcubecore-1.203.0.tar.gz
```

fabcor-maxiv:
We can exclude files at multiple levels:

* git repository level (with `.gitignore`)
* *sdist* level
* *wheel* level (which commands what really gets installed)

But in most cases the distinction from *sdist* to *wheel* is irrelevant.

marcus-oscarsson:
Looks good, thanks alot for having a look.

@rhfogh I think that the configuration files in `mxcubecore/configuration/` really needs a good cleanup and that they probably should be separated from the code. The exclusion is just for the package itself and few people would browse the package to get to these configuration files, the files are still in the repository.

rhfogh:
@marcus-oscarsson I'd agree with your latest comment

fabcor-maxiv:
So? : D

Can we merge this as it is? Should we revert back to my initial version of the pull request? Something else?

Which source directory tree branches do we want to exclude from the distribution packages?

* `mxcubecore/configuration/mockup/gphl/test_samples`?
* `mxcubecore/configuration`?
* Something else?

marcus-oscarsson:
I think that excluding `mxcubecore/configuration` would be even more efficient. Otherwise, what you did is perfectly fine if the package size is in the order of what you wrote above i.e 1.5M instead of 30M

fabcor-maxiv:
OK, so back to first version of the PR.

This is the whole `mxcubecore/configuration` source directory tree branch that is now excluded.

marcus-oscarsson:
Looks good to me :)

marcus-oscarsson:
Thanks

PR #1110, opened 2025-01-28
ICAT integration update

marcus-oscarsson:
Improvements made to the ICAT integration mostly by @antolinos, @woutdenolf.

- Emitting signal when the list of available session changes
- General linting 
- Added metadata for XRF, EnergyScan and Detector


elmjag (approved):
I don't yet understand all the intricacies of LIMS managment, but this looks good to me.

PR #1111, opened 2025-01-28
ISPyB LIMS: move get_lims_name() to parent class

elmjag:
Identical implementation of `get_lims_name()` method can be used by both `ProposalTypeISPyBLims` and `UserTypeISPyBLims` classes.

Move this method to the parent class, to make it available for `UserTypeISPyBLims` class as well.


marcus-oscarsson:
Tanks for the nice and clean PRs, easy to follow. Looks good you can merge after rebase

PR #1112, opened 2025-01-28
ISPyB LIMS: add method for fetching user full name

elmjag:
Adds a method to ISPyB data adapter for fetching information about a user.

This is useful when looking up user's full name.


marcus-oscarsson:
Looks good to me, you can merge when the rebase is ready :)

PR #1113, opened 2025-01-28
ISPyB LIMS: add a hook method for ISPyB authentication

elmjag:
Add an abstract method `ispyb_login()`, which is invoked when auth server is set to `ispyb`.

This allow to implement ISPyB authentication by overriding this method in your custom ISPyB lims class.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/ISPyBAbstractLims.py
+++ b/mxcubecore/HardwareObjects/abstract/ISPyBAbstractLims.py
@@ -98,13 +98,13 @@ class ISPyBAbstractLIMS(AbstractLims):
         ]
 
     def get_user_name(self):
-        raise Exception("Abstract class. Not implemented")
+        raise NotImplementedError
 
     def get_full_user_name(self):
-        raise Exception("Abstract class. Not implemented")
+        raise NotImplementedError
 
     def is_user_login_type(self):
-        raise Exception("Abstract class. Not implemented")
+        raise NotImplementedError
 
     def store_beamline_setup(self, session_id, bl_config):
         self.adapter.store_beamline_setup(session_id, bl_config)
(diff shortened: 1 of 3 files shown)


mxcubecore/HardwareObjects/abstract/ISPyBAbstractLims.py
```diff
@@ -129,6 +129,9 @@ def ldap_login(self, login_name, psd):
 
         return self.ldapConnection.authenticate(login_name, psd)
 
+    def ispyb_login(self, login_name, psd):
+        raise Exception("Abstract class. Not implemented")
```
fabcor-maxiv:
```suggestion
        raise NotImplementedError
```

https://docs.python.org/3/library/exceptions.html#NotImplementedError

elmjag: This is the style used in this class, but I agree with @fabcor-maxiv, it should raise `NotImplementedError`. I'll change it everywhere.

Also, it should be:

`raise NotImplementedError()`

😅

fabcor-maxiv: I read somewhere that `raise SomeException` ends up running faster in some cases but never slower than `raise SomeException()`. There is also this linting rule: <https://docs.astral.sh/ruff/rules/unnecessary-paren-on-raise-exception/>.

Anyway, with or without parenthesises, my point is that we should use `NotImplementedError`.

elmjag: Changed to raise `NotImplementedError` everywhere.

marcus-oscarsson:
Ok, same as the others looks good. You can rebase and merge :)

PR #1115, opened 2025-01-30
User-based login detailed

dominikatrojanowska:
At Max IV we had a meeting with beamline staff and discussed in details user-based login which is used on our site.   
This MR is recap of the meeting and starting point for discussion of desired behavior of MXCuBE when user logs in.


marcus-oscarsson:
Does this describe how it works on the currently deployed installation, how you wish that it would work or how it works with the sources in the current develop, i.e something close to version 1.20x ?

dominikatrojanowska:
How Max IV staff wish that it would work. So we want to discuss here what's possible and what's not in the future mxcube

marcus-oscarsson:
I see, I would suggest that you create an issue instead and that it could be discussed as a specific topic on user based authentication/authorization using ISPyB at MAXIV.  Because this documentation should describe how things works in the current code base.

fabcor-maxiv:
> I see, I would suggest that you create an issue instead

I was the one to suggest Dominika to start this as a pull request so that we can discuss and make changes to the Markdown document, that should act as some kind of a specification/requirements, even if we end up never merging this pull request (very likely scenario). With a GitHub issue we do not have Markdown document we can collaborate on. But whatever process gets the best discussion. :)

fabcor-maxiv:
> > Does this describe how it works on the currently deployed installation, how you wish that it would work or how it works with the sources in the current develop
>
> How Max IV staff wish that it would work. So we want to discuss here what's possible and what's not in the future mxcube

Yes, the idea is that on one hand we have the "wish list" from MaxIV (this document in the pull request), and on the other hand we have the reality of what is already implemented in MXCuBE and/or can/will be implemented in the near future. And we should try to find a compromise in agreement with the other facilities. What can be done in common MXCuBE-Web code, what can be done in facility-specific hardware objects and what can only be done in a fork of MXCuBE-Web?

marcus-oscarsson:
I think it would be good to initiate this discussion with a meeting. The beginning of next week will be a bit tight and we also have the general dev meeting on Tuesday. Id propose to meet either Thursday or Friday next week or sometime the week after. The general developers meeting has already a quite loaded agenda but we can start to discuss this on `mxcube-web` basis and the escalate the topic for next meeting.

fabcor-maxiv:
Oh, I did not realize that we are on mxcubecore. Maybe this should have been on mxcubeweb, no?

marcus-oscarsson:
Its a functionality that is shared between the applications so I think its fine.

marcus-oscarsson:
Converting to draft, its easier to see what is pending to be merged like this.

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
+# User-based login details
+
+When logging in the very first user sees modal dialog `Select session` and must choose an experimental session. As long as one user is logged in, the consecutive users do not see this modal dialog. Pressing  outside the modal or closing with `X` causes  unsuccessful login and landing back on the login page.
+
+## Multiple users:
+
+Any number of users can be logged in at the same time if:
+* they belong to the same proposal,
```
dominikatrojanowska: ESRF: listed are actually experimental sessions, not proposals   
proposal remains the same for several years, session is related to planned beam time (experiments)  
users can belong to many sessions  

MAXI IV: uses proposal in similar meaning

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+* they are **staff**.
+
+Users, who are not part of the proposal, can not login. After unsuccessful login attempt the user should see the message: “Authentication failed”.
+
+The first logged in user automatically becomes an **active user**. The **active user** is **in control** and the other users are **observers**.
+
+The **active user** should be able to force logout **observers**, except **staff**.
```
dominikatrojanowska: ESRF: active user can logout anybody - no distinction between staff and common users

For additional rules it is possible to overwrite relevant UserManager method, but it is not really recommended, since it's better to keep as common solutions as possible.

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Control:
+
+It should be clear who is **in control** (hostname of the machine should be included to know if its remote or on the beamline) and who are the **observers**.
+
+**Observers** can ”ask for control”, **active user** can give or deny control.
+
+There is a timeout so an **observer** does not stay waiting for the control forever. After the timeout the user will get control (if there was no deny signal from **active user**). This feature is enabled by default, but can be disabled.
+
+**Staff** can always “take control” or “ask for control”, even if the “get control after time out” feature is disabled.
```
dominikatrojanowska: anybody can take control whether is staff or not

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+The **active user** can change proposal without signing out. This automatically log out any **observers** that are not part of the new proposal, except **staff**. Users can only choose proposals they belong to.
+
+## Control:
+
+It should be clear who is **in control** (hostname of the machine should be included to know if its remote or on the beamline) and who are the **observers**.
+
+**Observers** can ”ask for control”, **active user** can give or deny control.
+
+There is a timeout so an **observer** does not stay waiting for the control forever. After the timeout the user will get control (if there was no deny signal from **active user**). This feature is enabled by default, but can be disabled.
```
dominikatrojanowska: There is some parameter called: timeout_gives_control = 30 seconds that might be used after "take control", but it's not used anymore (it was in the previous version of mxcube, 5-6 years ago)

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Staff** can always “take control” or “ask for control”, even if the “get control after time out” feature is disabled.
+
+The **active user** should be able to give control to any specific user (**observer** or **staff**).
+
+If the **active user** logs out, any of the **observers** can ”ask for control”, and this user will immediately become the **active user**.
```
dominikatrojanowska: ”ask for control” --> "take control"

if nobody took a control in the meantime,
- after log back in, the user is still in control or
- 1st newly logged in user automatically gets control

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+## New login:
+
+Meaning: login of already logged in user in another browser, browser tab or window, computer, etc.
+
+Already logged users (whether they are **active**, **observers** or **staff**) can always make a new login.
+
+New login inherits all session data such as queues, control status, drawn points etc.
+
+After new login the old one is quit silently and immediately.
```
dominikatrojanowska: There is one active session in the backend
Old login can be quit in case of new login on another **computer** or new browser **window**

It may be more difficult to implement in case of **tabs** because browsers share cookies between tabs, this would require token login
TODO: investigate + open an issue

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+## User with **staff** privileges
+
+They can always:
+* Login - they do not need to be part of any proposal (it does not need to be checked to let them login). After login they see the `Select session` modal with the list of proposals they belong to. They could choose any proposal from it. If they type in a number or keyword they search through full list of proposals (even the ones they do not belong to) and they can select each one.
+* Logout any user (**active** or **observer**).
+* Take and move control from user to another one.
```
dominikatrojanowska: take - agreed
move - two steps action: take from one user (becoming an active) and give it another
this is relevant to all users

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+* Change proposal without logging out.
```
dominikatrojanowska: this is relevant to all users

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+* Restart the server
```
dominikatrojanowska: remove it

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Sessions:
```
dominikatrojanowska: Lims Sessions - Proposal time - meaning experimental time on the beamline, 

this section is specific to MAX IV, remove it since it does not need to be specified in this document, in case of specific implementation document it with something like "Max IV extends this to ..."

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+Users, who are not part of the proposal, can not login. After unsuccessful login attempt the user should see the message: “Authentication failed”.
+
+The first logged in user automatically becomes an **active user**. The **active user** is **in control** and the other users are **observers**.
+
+The **active user** should be able to force logout **observers**, except **staff**.
+
+**Observers** can remain logged in even if the **active user** logs out.
+
+The **active user** can change proposal without signing out. This automatically log out any **observers** that are not part of the new proposal, except **staff**. Users can only choose proposals they belong to.
```
dominikatrojanowska: The **active user** can change proposal / session without signing out. **Active user** can only choose proposals that all of the logged in users belong to (intersection).   
List of sessions displayed is the same everywhere - for every user

Example:
* user 1 has access to sessions A B C  
* user 2 has access to sessions A B  
* user 3 has access to session A  
All of them are logged in and see only session A. If user 3 logs out, user 1 and 2 see sessions A and B.  If only user 1stays logged in all the sessions are visible.  

Staff can move experimental session timewise - reschedule user
Staff can move experimental session beamline-wise

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Control:
+
+It should be clear who is **in control** (hostname of the machine should be included to know if its remote or on the beamline) and who are the **observers**.
+
+**Observers** can ”ask for control”, **active user** can give or deny control.
```
dominikatrojanowska: There is also a "take control" option

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+Already logged users (whether they are **active**, **observers** or **staff**) can always make a new login.
+
+New login inherits all session data such as queues, control status, drawn points etc.
+
+After new login the old one is quit silently and immediately.
+
+## User with **staff** privileges
+
+They can always:
+* Login - they do not need to be part of any proposal (it does not need to be checked to let them login). After login they see the `Select session` modal with the list of proposals they belong to. They could choose any proposal from it. If they type in a number or keyword they search through full list of proposals (even the ones they do not belong to) and they can select each one.
```
dominikatrojanowska: agreed, proposal / session

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,74 @@
...
+
+If user login and there is session MXCuBE will use it – applies to **staff** members and common users.
+
+If user login and there is no sesion the MXCuBE creates new one applies to **staff** members and common users.
+
+If they change proposal what happens with the session is covered above.
+
+## Improper logout (without sign out button):
+
+Closing or crashing browser of last logged in user activates an “idle timeout” (15 minutes for example as default value). If there is any **active user** or **observer** still logged in, the timeout will not be activated.
```
dominikatrojanowska: There is a timeout in meaning web session expiration time.

After closing window, losing internet connection the queue will not be removed or cleared. There might be also an automatic data collection f.e. during the night so only http part is cleared, after login back the queue points etc. need to be back also

TODO: re-discuss at MAX IV site, 
Optionally: prompt the user to logout when closing the browser window to avoid improper logout

ESRF Policy: It has to be an explicit action from users "I'm finished with experiments". It is not allowed to switch users without local interventions. Local intervention means logout user and activate incoming session. 

Below points are not meaningful

dominikatrojanowska (commented):
I wrote down notes from todays meeting, just to keep them somewhere. In next step I will remove and update relevant parts

docs/source/dev/user-based-login-details.md
axelboc: May I suggest sticking with the term "user in control" rather than "active user"?

In a way, an "active user" would fit more closely with the concept of front-end sessions (or "active user session"): a user session is active as long as it has a valid token — it becomes inactive once this token is lost or expires. So this could be the term used in the last section in my opinion.

docs/source/dev/user-based-login-details.md
axelboc: "can give or deny control" => "can accept or deny the request"

... to distinguish this behaviour from the "give control" feature (i.e. observers can "take control", but the user in control can also decide to "give control" to any observer).

docs/source/dev/user-based-login-details.md
axelboc: So if I understand correctly, this section talks about the same human user logging in to MXCuBE once, and then opening MXCuBE again in a separate tab/window/browser.

Two things can happen currently:

1. The user opens MXCuBE again within the **same browser session** => the browser sends the MXCuBE session cookie to the back-end and the user is automatically logged back in within the same MXCuBE "user session".
2. The user opens MXCuBE again in a **separate browser session** => the browser has no MXCuBE session cookie to send, so the user lands on the login screen; if they log in again, they end up creating a new MXCuBE "user session" (the UI shows two "active users": one in control and one observer).

This is the current behaviour. What was suggested during the meeting is to take advantage of user-based login to avoid creating two separate "user sessions" in the second case (i.e. when the same human user opens MXCuBE in two separate browser sessions) — any new user session would effectively close/replace any previous user session opened by the same human user.

---

Some clarification as to what constitutes a "browser session":

A user can typically have two browser sessions running at the same time within the same browser: one for "public browsing" mode and one for "private browsing" mode (= incognito). All tabs/windows within each of these modes share the same browser session.

So a user can currently log in to MXCuBE multiple times with the **same user session** by opening the app in either:
- multiple _public_ tabs/windows in the same browser, or
- multiple _private_ tabs/windows in the same browser

Conversely, a user can currently log in to MXCuBE twice with two **different user sessions** by opening the app in:
- two windows in the same browser but in two different browsing modes (one public and one private);
- two windows in two different browsers (or two different browser profiles, or two different machines of course).

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok, I will change that :)

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok, thanks

docs/source/dev/user-based-login-details.md
marcus-oscarsson: yes, and we can if there are no objections and you are happy with give/take control remove the feature "time out gives control"

docs/source/dev/user-based-login-details.md
marcus-oscarsson: or put slightly differently: "user in control" is not passed to another user on logout. The "user-in-control" becomes unassigned at logout and the next user that logs in becomes the "user in control"

marcus-oscarsson:
Thanks for taking notes and providing comments, looks good as far as I can see

docs/source/dev/user-based-login-details.md
dominikatrojanowska: Added an issue: https://github.com/mxcube/mxcubecore/issues/1139

docs/source/dev/user-based-login-details.md
dominikatrojanowska: thanks for clarification  
I created new issue regarding separate browser session: https://github.com/mxcube/mxcubecore/issues/1140

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
+# User-based login details
+
+When logging in the very first user sees modal dialog `Select session` and must choose an experimental session. As long as one user is logged in, the consecutive users do not see this modal dialog. Pressing outside the modal or closing with `X` causes unsuccessful login and landing back on the login page.
+
+## Multiple users:
+
+Any number of users can be logged in at the same time if:
+* they belong to the same `session` (at MAX IV site `proposal` is used in similar meaning),
```
marcus-oscarsson: I think we said that we would move MAX IV specific documentation to a separate document (or atleast section) if Im not mistaken ?

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
...
+
+Users, who are not part of the proposal, can not log in. After unsuccessful login attempt the user should see the message: “Authentication failed”.
+
+The first logged in user automatically becomes an **user in control**. The other users are **observers**.
+
+The **user in control** is able to force logout other users.
+
+**Observers** can remain logged in even if the **user in control** logs out.
+
+The **user in control** can change session / proposal without signing out. He can only choose between intersection sessions - sessions to which all logged-in users belong to. Hence, the list of available  sessions displayed is the same for every user. Example:
```
marcus-oscarsson: I would probably phrase it something like, "The users can only see the sessions they have in common".(Intersection sounds a bit mathematical)

marcus-oscarsson: Also I would not write session/proposal. Within mxcube the concept sued is experimental session, it may or may not be the same as a proposal (but its not necessary to mention here)

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
...
+    User 1 has access to sessions A B C
+    User 2 has access to sessions A B
+    User 3 has access to session A
+    All of them are logged in and see only session A.
+    If user 3 logs out, user 1 and 2 will see sessions A and B.
+    If only user 1 stays logged in, all the sessions are visible.
+
+## Control:
+
+It should be clear who is **in control** (hostname of the machine should be included tondicate whether it is remote or on the beamline) and who are the **observers**.
```
marcus-oscarsson: ... to indicate ...

marcus-oscarsson: ... and who is observing
or
... who the observers are

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
...
+
+Users can ”ask for control” or "take control". The **user in control** can accept or deny the request.
+
+There is a parameter named: TIMEOUT_GIVES_CONTROL which, if enabled, triggers a timeout (30 seconds) after the control request and. If there has been no deny signal in the meantime, the request is accepted automatically. In the current version of MXCuBE, this feature is disabled by default.
```
marcus-oscarsson, line 34: I think we can remove this phrase, if we all agree that we can remove this feature ? It makes no sense when "anyone" can take control

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
...
+
+The **user in control** is able to "give control" to any other user.
+
+If the **user in control** logs out this role is not passed to any other (logged-in) user. Once logged out the "user in control" is unassigned and the next user that logs in becomes the "user in control". If no new user appears upon re-logging of the former **user in control**, he/she is still in control.
+
+## New login:
+
+A new login occurs when the same user logs into MXCuBE and then opens it again in a separate tab, window, browser, or on another computer.
```
marcus-oscarsson, line 42: ~~Tricky to phrase this well, I would perhaps phrase it more like; There can only be a single login from a given user at a time.~~

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
...
+
+Any user can always initiate a new login.
```
marcus-oscarsson: Not sure that this is relevant ?, of course any user can always login right ?

docs/source/dev/user-based-login-details.md
```diff
@@ -0,0 +1,64 @@
...
+
+Currently, two scenarios can occur:
+
+* If the user opens MXCuBE again within **the same browser session** by opening multiple public tabs/windows or multiple private tabs/windows in the same browser, the browser sends the MXCuBE session cookie to the back-end. As a result, the user is automatically logged in within the same  MXCuBE "user session". In this case, the new login inherits all session data, including queues, drawn points, control state etc.
```
marcus-oscarsson, line 46: Cookies and thus sessions are shared between windows and tabs, so this is inherit to the technology we use. So its the same session ....

docs/source/dev/user-based-login-details.md
marcus-oscarsson: Sorry, so this is the behavior that we said we would like, but its not yet implemented so the original sentence is fine.

marcus-oscarsson:
A few more comments, I'm not sure that you had the time to look at it since the last time

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok

docs/source/dev/user-based-login-details.md
dominikatrojanowska: @fabcor-maxiv @axelboc  
who else should be included in we all?

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok, I can add the sentence of what is being planned, shall I?

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok, from the previous sentence seems that there are no restrictions, so I'm removing that one

docs/source/dev/user-based-login-details.md
dominikatrojanowska: ok, so shall I remove those information? I found it valuable to mention and distinguish this one and the following point

docs/source/dev/user-based-login-details.md
dominikatrojanowska: removed

docs/source/dev/user-based-login-details.md
axelboc: I didn't even know about this feature. :sweat_smile:  As discussed, one can always cancel the request and forcibly take control.

docs/source/dev/user-based-login-details.md
axelboc: I'd recommend not to. We don't know how long it will take and can't predict if it will happen as planned.

marcus-oscarsson: yes, not necessary. We should just stick to the current behaviour

docs/source/dev/user-based-login-details.md
marcus-oscarsson: Its quite explicit and detailed, but its fine as it is.

docs/source/dev/user-based-login-details.md
dominikatrojanowska: should I rise an issue to remove this feature at all then?   
the parameter I mentioned is set here: https://github.com/mxcube/mxcubeweb/blob/develop/mxcubeweb/app.py#L277   
and there is some state and actions related on the frontend: https://github.com/mxcube/mxcubeweb/blob/develop/ui/src/reducers/remoteAccess.js

marcus-oscarsson: Sounds excellent !

dominikatrojanowska: Added: https://github.com/mxcube/mxcubeweb/issues/1588  
I assume, this thread could be resolved, as it fits to the current behavior, and this `md` document should be updated after removal

marcus-oscarsson:
I guess this starts to get ready to be merged, or do you still want to make some more modifications ?

PR #1117, opened 2025-01-31
set_value should be re-entrant adding lock to ensure reentrancy 

marcus-oscarsson:
`set_value` can be called multiple times and as it performs IO can be interrupted by `gevent` meaning that we have to make sure that its re-entrant.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractActuator.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractActuator.py
@@ -30,6 +30,7 @@ import math
 from ast import literal_eval
 
 from mxcubecore.BaseHardwareObjects import HardwareObject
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Looks like the pages build are failing due a depricated `GitHub` action function. We have not look with the documentation build these days !

fabcor-maxiv:
> Looks like the pages build are failing due a depricated `GitHub` action function. We have not look with the documentation build these days !

For this one, it was only yesterday that they stopped supporting it, so we were quite responsive :D

We should still have paid attention to the deprecation notice I guess (where? how?), and acted before it starts breaking, though... :D

fabcor-maxiv (approved):
No idea about the context, but seems fine

PR #1119, opened 2025-01-31
Update Actions in `pages` GitHub workflow

fabcor-maxiv:
In particular because of some actions being deprecated: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


marcus-oscarsson:
Ah, you also created a PR for this, closing mine then :)

PR #1120, opened 2025-01-31
Updated GitHub actions to v4

marcus-oscarsson:
An attempt to update GitHub actions to v4 and fix the deprecation error

### DIFF
--- a/.github/workflows/pages.yaml
+++ b/.github/workflows/pages.yaml
@@ -1,6 +1,4 @@
 ---
-
-
 name: "Pages"
 
 concurrency:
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
Ah, thanks! :D
But yes I was already working on it as well.

fabcor-maxiv:
I think we need to remember to close these branches as well.

When merging a PR, the branch gets automatically deleted. But when closing a PR without merging it, the branch is not deleted.

It seems to me like we have dozens of stale/abandoned branches. Maybe it would be good to do some clean-up action.

_[Also that is one of the reasons why I prefer working with forks whenever possible, so that the main repository does not get "polluted" with stale branches.]_

marcus-oscarsson:
Ill do cleanup run :)

PR #1121, opened 2025-01-31
[EMBL FlexHCD SC] - Added possibility to use custom script for changing phase

marcus-oscarsson:
Added possibility to use custom script for changing phase when using the EMBL Flex HCD sample changer.


elmjag:
@marcus-oscarsson I'm a bit confused why you want a review from me :) I don't really know what this code is about...

marcus-oscarsson:
@elmjag, No real reason I just picked someone, its fine if you feel that your not the person to look at this.

PR #1122, opened 2025-01-31
[COLLECT] - flux threshold configured on flux hardware object

marcus-oscarsson:
The `flux_threshold` is now configured on the `Beam` HardwareObject. and `flux_threshold` on  `AbstractMultiCollect` is no longer used.


fabcor-maxiv (approved):
Seems fine

PR #1126, opened 2025-02-04
ISPyB LIMS: make it possible to use custom ISPyB adapter object

elmjag:
Adds a factory method `_create_data_adapter()`, which creates an instance of data adapter. This allows to supply different implementation of the adapter objects in custom ISPyB LIMS classes.

We use a custom data adapter at MAX IV, to implement our site specific features.


marcus-oscarsson:
Good idea, thanks !

PR #1127, opened 2025-02-04
ISPyB LIMS: store sessions in 'session_manager' for user-type logins

elmjag:
On successful login using user-type logins, store returned sessions in `session_manger` attribute of the `HWR.beamline.lims `object. This way session are avail later on.


marcus-oscarsson:
Looks good, thanks !

PR #1128, opened 2025-02-04
ISPyB LIMS: update 'login_ok' attribute on successful user-type logins

elmjag:
Make sure `self.login_ok` attribute is set to `True` after a successful login. The `self.login_ok` attribute attribute is consulted in `ISPyBAbstractLIMS.is_connected()` method.

The `lims.is_connected()` method is used by the rest of the system to check if LIMS sub-system is in usable state.


marcus-oscarsson:
Ok, thanks !

PR #1129, opened 2025-02-05
ISPyB LIMS: simplify create_session() method

elmjag:
Simplify `ISPyBDataAdapter.create_session()` method. After creating a new proposal session, just return a `Session` object.

Skip loading more sessions and constructing `LimsSessionManager` object. It's not needed and makes it harder to use `create_session()` method.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py
+++ b/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py
@@ -128,9 +128,7 @@ class ISPyBDataAdapter:
     def isEnabled(self) -> object:
         return self._shipping  # type: ignore
 
-    def create_session(
-        self, proposal_id: str, beamline_name: str
-    ) -> LimsSessionManager:
+    def create_session(self, proposal_id: str, beamline_name: str) -> Session:
 
         try:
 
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py
```diff
@@ -163,12 +161,7 @@ def create_session(
                 "Session created. session_id=%s" % session_id
             )
             response = self._collection.service.findSession(session_id)
-            session: Session = self.__to_session(asdict(response))
-            return self.get_sessions_by_code_and_number(
-                session.code,
-                session.number,
-                beamline_name,
-            )
+            return self.__to_session(asdict(response))
```
marcus-oscarsson, line 164: Are you sure about this one, it does not seem to be the same web service calls behind. I think that `get_sessions_by_code_and_number` in the end uses `findSessionsByProposalAndBeamLine` and this change will be using `findSession`.

I myself don't know sufficiently about ISPyB, there is maybe no difference in the end ?

elmjag: Yes, this changes the semantics of `create_session()` method. 

Before the change it was:

 - creating a new session
 - fetching that session by ID
 - fetching _all_ sessions for specified proposal and beamline
 
 This change removes the last step, because IMHO it's wrong place to do this operation. _If_ you need to fetch complete list of session, after creating new session, it's better to explicitly make a call to`get_sessions_by_code_and_number()`.

Now it returns the result of step 2. Only the newly created session is returned.

It does change which actual SOAP call that returns data for the `Session`. However it seems to do the right thing. We have tested this at our beamlines, seems to work fine.

marcus-oscarsson: Ok

fabcor-maxiv (approved):

Issue #1046, opened 2024-10-18
Python 3.8 end of life


fabcor-maxiv:
[Python 3.8 has reached its end of life](https://discuss.python.org/t/python-3-8-is-now-officially-eol/66983).

How do we want to handle this?

fabcor-maxiv:
I happened to be thinking about this now because I wanted to update our tooling, starting with the *pre-commit* hooks, but encountered issues. I realized it is because a bunch of tools have already dropped support for Python 3.8, for example `pre-commit` and `black`. This is manageable, but annoying. We have pinned a whole bunch of things, so that is good, this should limit the amount of cases where things start breaking because a dependency dropped support for Python 3.8.

Anyway, do we have a general plan of how we want to handle Python 3.8's end of life?

fabcor-maxiv:
For example [here](https://github.com/mxcube/mxcubecore/pull/1045), I ran `pre-commit autoupdate` at first, but then had to pin to some earlier versions because of Python 3.8 being dropped.

marcus-oscarsson:
I think we can also drop Python 3.8 but we should have a look at the deployment survey to get an idea of how many sites that are still using Python 3.8 and if it would be a blocking issue if we where to drop it. We should bring it up on our next meeting. We have to, in my opinion, sooner or later drop the Python 3.8 support.

In the mean time, to all of those still using Python 3.8. Please let us know and let us know what you think.

rhfogh:
While we are at it, could we consider moving to a newer version than 3.9? The fewer changes the better.  Ideally 3.13 I guess, but as far as we can go. What would be the blocker?

fabcor-maxiv:
> we should have a look at the deployment survey to get an idea of how many sites that are still using Python 3.8

I just checked and the questionnaire did not ask for Python version explicitly, so most answer sheets do not say it.


> In the mean time, to all of those still using Python 3.8. Please let us know and let us know what you think.

I think the question should be a bit more precise. We want to know which setup is both on Python older than 3.9 AND living on the edge of what is on GitHub. If setups have Python 3.8 or older, but live with an older version of MXCuBE (and no intention to catch up with the upstream `develop` branch) then for those setups it should also not matter if we drop Python 3.8.

fabcor-maxiv:
[...]

PR #1133, opened 2025-02-10
Fix type hint annotation issue for Python 3.8

fabcor-maxiv:
The issue was introduced in that commit:
`4fb849824bf4bcb9b91699dc74aa670370249104`

That commit introduced type hint notations
that are not supported in Python 3.8.

With `from __future__ import annotations`,
Python 3.8 should be able to run the code anyway.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractCollect.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractCollect.py
@@ -20,6 +20,8 @@
 
 """Abstract hardware object for data collection."""
 
+from __future__ import annotations
+
 import abc
 import collections
 import errno


fabcor-maxiv:
@rhfogh Can you try this and see if it fixes the issue you saw earlier? It is a one-line change. If not, can you add more details (console output)?

marcus-oscarsson:
@fabcor-maxiv, @rhfogh, which issue is it ?

fabcor-maxiv:
Running code like this with Python 3.8:

```python
from mxcubecore.HardwareObjects.abstract import AbstractCollect
```

triggers a failure like this:

```none
  File "script.py", line 8, in <module>
    from mxcubecore.HardwareObjects.abstract import AbstractCollect
  File "/home/fabcor/workspace/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/AbstractCollect.py", line 67, in <module>
    class AbstractCollect(HardwareObject, object):
  File "/home/fabcor/workspace/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/AbstractCollect.py", line 596, in AbstractCollect
    def prepare_input_files(self) -> tuple[str | None, str | None, str | None]:
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

rhfogh (approved):
That solves the problem

marcus-oscarsson:
I see, but I could not find the original discussion for this issue. It also makes me wonder, it seems like some site is still running with Python 3.8 then ? Because during the last developers meeting it did not seem to be the case and we decided to drop both Python 3.8 and 3.9. Or was it the tests that where failing ?

fabcor-maxiv:
> we decided to drop both Python 3.8 and 3.9

Confirm we have green light at https://github.com/mxcube/mxcubecore/issues/1046 and I can probably start working on it :)


> I could not find the original discussion for this issue

It was not on GitHub :)

rhfogh:
I am running Python 3.8 on my development tests - because we are still supposed to be supporting down to Python 3.8, so that is the test set-up I am using.  And trying to run with the newest checked-in version my code broke. I mailed it to Fabien directly  - after all, why make a public fuss when the agreement is clear and there should be nothing particular that needs discussing? 

I very much look forward to upgrading to at least Python 3.10 (which is where these new annotations came in that  1) are much nicer, 2) people seem to be using anyway). But I felt an obligation to stick to Python 3.8 in my tests as long as the agreement has not changed.

fabcor-maxiv:
> was it the tests that where failing ?

No GitHub workflow failed. Which is kind of worrying. We still have huge gaps in our code checking toolchain.

marcus-oscarsson:
I think thats because there is not test performing an actual collect or upload to LIMS. We should really look at this :)

rhfogh:
In this case I actually got the error on loading AbstractBeam, before anything was executed (and I was using the Qt version anyway). But yes, something that follows the collection more closely  wopuld be good - at GPhL we have our in-house code for simulating the image generation that we use to go through a mock collection.

PR #1134, opened 2025-02-10
Fixed bug (for Python3.8) in type annotations

rhfogh:
The line
    users: Optional[dict[str, LimsUser]] = {}
raises an error in Python 3.8. Changing it to 
    users: Optional[Dict[str, LimsUser]] = {}
(and importing Dict) fixes the problem


marcus-oscarsson:
Ok looks good. `isort` is complaining but apart from that :)

fabcor-maxiv (approved):

PR #1137, opened 2025-02-11
fixed lack of some parameters in duplicated data collection task

mockoocy:
When duplicating data collection tasks some parameters are missing in the duplicated one - notable example is `take_snapshots` parameter.
This fixes it and makes it less prone to happen in the future (by using all the properties in the `AcquisitionProperties.as_dict()`).

Note: the `parameters.centred_position` is still being passed explicitly. That's because the `AcquisitionProperties.as_dict()` method is expected to be json-serializable - it is sent in `/available_tasks` GET response. Thus I could not just include the `centred_position` property there.


marcus-oscarsson:
Thanks

mxcubecore/model/queue_model_objects.py
```diff
@@ -626,26 +626,11 @@ def as_dict(self):
         parameters = acq.acquisition_parameters
 
         return {
+            **parameters.as_dict(),
+            "centred_position": parameters.centred_position,
```
elmjag, line 631: Can you add comment, explaining the special case for handling `centred_position`?

I.e. add the info here from this PR's description.

mockoocy: done

Issue #1135, opened 2025-02-11
Suport for Python 3.13


marcus-oscarsson:
General issue for discussing whats needed to update to support `Python 3.13`

marcus-oscarsson:
Did anyone try to run with `Python 3.13` lately ?

PR #1138, opened 2025-02-11
Drop support for Python 3.8 and Python 3.9

### LINKED ISSUE #1046: Python 3.8 end of life

fabcor-maxiv:
Drop support for Python 3.8 and Python 3.9

Supported versions are:
* Python 3.10
* Python 3.11

Update all tools that were stuck on older versions because the tools themselves dropped Python 3.8.

Format and remove lint according to newest tool versions.

Update and clean-up the content of `pyproject.toml`, in particular use the `[project]` table now that it is supported by Poetry.

Update more development dependencies.

Remove unused `pylint` development dependency.

GitHub: fix https://github.com/mxcube/mxcubecore/issues/1046

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,13 +1,15 @@
-[tool.poetry]
+[project]
 name = "mxcubecore"
 version = "1.228.0"
 license = "LGPL-3.0-or-later"
 description = "Core libraries for the MXCuBE application"
-authors = ["The MXCuBE collaboration <mxcube@esrf.fr>"]
+authors = [
+    {name = "The MXCuBE collaboration", email = "mxcube@esrf.fr"},
+]
 maintainers = [
-    "Marcus Oscarsson <mxcube@esrf.fr>",
-    "Rasmus H. Fogh <rhfogh@globalphasing.com>",
-    "Antonia Beteva <beteva@esrf.fr>",
+    {name = "Marcus Oscarsson", email = "mxcube@esrf.fr"},
+    {name = "Rasmus H. Fogh", email = "rhfogh@globalphasing.com"},
+    {name = "Antonia Beteva", email = "beteva@esrf.fr"},
 ]
 readme = "README.md"
 homepage = "http://github.com/mxcube/mxcubecore"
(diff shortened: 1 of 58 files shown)


pyproject.toml
```diff
@@ -122,6 +122,12 @@ mxcubecore/HardwareObjects/XSDataMXv1.py,
 
 [tool.black]
 line-length = 88
+required-version = "25"
```
marcus-oscarsson, line 128: What does this `required-version = "25"` mean ?

marcus-oscarsson: ah, its the required version of `black` ?

fabcor-maxiv: > What does this `required-version = "25"` mean ?
> ah, its the required version of `black` ?

Yes: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#required-version

Hopefully this would give an helpful error message in case someone is running an outdated version of black.

rhfogh (commented):
I approve (even the changes to GPhL code ;-) ), but just one question:

Were we not supposed to support Python 3.12 also?

elmjag:
> Were we not supposed to support Python 3.12 also?

MXCuBE have never supported 3.12 before, at least officially.

Maybe it works perfectly fine in 3.12. But we should test it first, and make a PR that adds 3.12 to supported versions.

rhfogh:
@elmjag OK, one version at a time, then

rhfogh (approved):

marcus-oscarsson:
I guess we can remove most of the `from __future__ import` as well ?

fabcor-maxiv:
> I guess we can remove most of the `from __future__ import` as well ?

Yes, I have a plan for this: [_`pyupgrade`_](https://pypi.org/project/pyupgrade/) and maybe even [_`refurb`_](https://pypi.org/project/refurb/) as *pre-commit* hooks. I tried those on other code bases and it went smooth.

fabcor-maxiv:
> > Were we not supposed to support Python 3.12 also?
> 
> MXCuBE have never supported 3.12 before, at least officially.
> 
> Maybe it works perfectly fine in 3.12. But we should test it first, and make a PR that adds 3.12 to supported versions.

We can do both 3.12 and 3.13 in one go: <https://github.com/mxcube/mxcubecore/issues/1135>.

marcus-oscarsson:
> We can do both 3.12 and 3.13 in one go: https://github.com/mxcube/mxcubecore/issues/1135.
We can try but I think 3.12 will be easier than 3.13 at least from a web point of view.

marcus-oscarsson:
> Yes, I have a plan for this: [_`pyupgrade`_](https://pypi.org/project/pyupgrade/) and maybe even [_`refurb`_](https://pypi.org/project/refurb/) as _pre-commit_ hooks. I tried those on other code bases and it went smooth.

Ok, perfect. So you would like to do that in a spereate PR then ?

marcus-oscarsson:
I'm sorry, I would have preferred if we at-least finish the pending discussions before we merge.

fabcor-maxiv:
> Ok, perfect. So you would like to do that in a spereate PR then ?

Yes, that's my plan

fabcor-maxiv:
Ouch! I did not think about `.github/workflows/create-tag.yml`. This workflow generated a `v` tag.

fabcor-maxiv:
Issue ticket to handle this: <https://github.com/mxcube/mxcubecore/issues/1142>.

PR #1141, opened 2025-02-11
MAXIV: ISPyB LIMS hardware object

elmjag:
Adds the LIMS hardware object used at MAX IV facility.

This hardware object implements user-based logins, using ISPyB. This HWOBJ works with the ISPyB deployed at MAX IV. It implements some MAX IV specific features, such as lazily created LIMS sessions.

### DIFF
--- a/test/pytest/test_hwo_maxiv_ispyblims.py
+++ b/test/pytest/test_hwo_maxiv_ispyblims.py
@@ -1,0 +1,88 @@
+from json.decoder import JSONDecodeError
+from unittest.mock import (
+    MagicMock,
+    patch,
+)
+
+import pytest
+
+ISPYB_AUTH_ERROR_MESSSAGE = (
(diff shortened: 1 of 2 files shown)


test/pytest/test_hwo_maxiv_ispyblims.py
```diff
@@ -0,0 +1,88 @@
...
+    response_mock.text = ISPYB_AUTH_ERROR_MESSSAGE
+    response_mock.json.side_effect = json_decode_error
+
+    post_mock.return_value = response_mock
+
+    is_ok, err = ispyb_lims.ispyb_login(USER, PASS)
+
+    # check that we got 'login failed' result
+    assert not is_ok
+    assert err == "invalid credentials"
```
marcus-oscarsson, line 88: Just to make sure that the string "invalid credentials" does not find its way to the user . Its likely that its handled again and then transformed to "Authentication failed". The later is what we said would be the message to the user in then end.

fabcor-maxiv (approved):

PR #1143, opened 2025-02-13
Using poetry version -s instead of parsing toml

marcus-oscarsson:
Using poetry version -s instead of parsing toml


fabcor-maxiv:
@marcus-oscarsson I think what is missing is that Poetry is not installed in the "Read package version" step. In the "Update version" step it is installed, so we can probably copy-paste from that step.

Or is the Poetry installation still available in the subsequent steps? I forget how GitHub Actions workflows behave in that regard...

PR #1144, opened 2025-02-13
Using env variable instead

marcus-oscarsson:


.github/workflows/create-tag.yml
```diff
@@ -40,14 +40,12 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install toml
-          echo {tag_name}={v$(poetry version -s)} >> $GITHUB_OUTPUT
+          echo "{tag_name}={v$(poetry version -s)}" >> $GITHUB_ENV
```
fabcor-maxiv, line 43: @marcus-oscarsson

Documentation is a bit confusing to me, but seems like the curly braces should be removed:

```none
echo "tag_name=v$(poetry version --short)" >> "$GITHUB_ENV"
```

I used this example as reference: <https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#example-of-writing-an-environment-variable-to-github_env>.

```yaml
steps:
  - name: Set the value
    id: step_one
    run: |
      echo "action_state=yellow" >> "$GITHUB_ENV"
  - name: Use the value
    id: step_two
    run: |
      printf '%s\n' "$action_state" # This will output 'yellow'
```

marcus-oscarsson: Oh, yes. Indeed. Well, we are not the only ones writing unclear documentation. However I think GitHub lies a step ahead of us on that matter in general ;)

marcus-oscarsson: Writing documentation that is ...

PR #1150, opened 2025-02-14
Rhfogh develop

rhfogh:
Improved acquisition strategies

Support for two-transmission strategies (final low-transmission sweep)
Adapt strategies to optimise DELPHI blocks and multiprocessor XDS processing
Configurable DELPHI block size


mxcubecore/model/queue_model_objects.py
```diff
@@ -2102,6 +2103,7 @@ def set_pre_strategy_params(
         :param relative_rad_sensitivity (float):
         :param use_cell_for_processing (bool):
         :param crystal_thickness (float):
+        :param reference_reflection_files (list(str)):
         :param unused (dict):
```
beteva, line 2107: @rhfogh, may I suggest, while you are on it, to apply the google style doc.

pyproject.toml
```diff
@@ -56,7 +56,7 @@ suds-py3 = "1.4.5.0"
 psutil = "^5.9.4"
 gipc = "^1.6.0"
 py4j = "0.10.9.7"
-f90nml = "1.4.3"
+f90nml = "1.3.1"
 lucid3 = "^3.0.0"
```
beteva, line 60: @rhfogh Is it for purpose that you downgrade the library?

beteva:
Looks fine for me, except for the queue_model_object documentation. Also the test do not pass.

pyproject.toml
rhfogh: Yes. The later version breaks, unfortunately, some messy detail about how f90nml handles lists of elements. At some point one ought presumably to do the work of finding out exactly what the problem is and getting f90nml fixed (there are alrady two fixes in there attributed to my bug reports), but since this is a GPhL-use-only rather niche library I guess we can leave it for a bit.

mxcubecore/model/queue_model_objects.py
rhfogh: You are right.  I'll put it on the TODO list.

pyproject.toml
beteva: Thanks for the explanation. I was just wondering if this was not something that has been forgotten.

rhfogh:
Yes, the tests do not pass. Looking, I got a message about pyproject.toml having changed so you needed to do a poetry.lock command before you could install the dependencies. I must admit I never understood the point of having an autogenerated version list checked in, and every time I change a dependency (or even just change branches?) I get that stuff with zillions of changes in poetry.lock. Do you have to do a fresh install in an empty directory every time for this to work smoothly, or is there some other procedure that will make it work?

marcus-oscarsson:
@rhfogh, `poetry lock --no-update` should solve the issue for you, if you make sure that the `project.toml` file is merged well. Meaning that it matches upstream except for any potential changes to dependencies that you are making, in your case for `f90nml`.

rhfogh:
@marcus-oscarsson Thanks, I was hoping for some advice. And after that, should I check in both pyproject.toml and poetry.lock? I have refrained from mucking about with the latter, because it is huge, opaque, lots of things change, and I am afraid to break something for somebody else.

rhfogh:
@marcus-oscarsson 
I just came across a good example. I had 20 local changes in my poetry.lock. Following this discussion I thought I should check that in, but first I wanted to rebase on the tip of develop. And checking out develop I found that there were 161 differences and 7 conflicts between the poetry.lock I was downloading and the poetry .lock I had in my own branch. Merging by hand seems impossible, since this file is totally opaque and is not supposed to be maintained by hand.

What is the correct way of handling this situation?

marcus-oscarsson:
I understand your concern, so `poetry lock --no-update` updates the lock file with accordance to `pyproject.toml`. So the `poetry.lock` file generated is the one you want to use as long as your `pyproject.toml` is correct containing your changes and the one from upstream. You can thus ignore your old `poetry.lock` file and use the one you just generated. 

The changes between the upstream `poetry.lock` file and the one you generated should be relatively small. It happens that there are new releases of libraries between that we merge PR's and those would then also appear in your lock file in addition to your changes.

fabcor-maxiv:
I will try to write down what I said in the meeting earlier today...

First of all, one should never manually change the content of `poetry.lock` or try to fix git merge conflicts in that file. This file should only ever be modified via the `poetry` tool itself.

If you get in a situation where there are doubts about the content of `poetry.lock`, typically one can follow these steps:

1. remove all local, uncommitted changes in `poetry.lock`, maybe with a command like this for example:
    ```shell
    git restore poetry.lock
    ```
1. let Poetry regenerate the lock file with the command:
    ```
    poetry lock
    ```

beteva:
> 2\. let Poetry regenerate the lock file with the command:
>         ```
>         poetry lock
>         ```

`poetry lock   --no-update`

rhfogh:
@fabcor-maxiv @beteva 
I tried with 'poetry lock --no-update' as well (as per my old notes, but got the resonse:
'The option "--no-update" does not exist'
It looks like the latest poetry just wants 'poetry lock'.
Is that right, Fabien?

fabcor-maxiv:
> > 2. let Poetry regenerate the lock file with the command:
> > `poetry lock`
> 
> `poetry lock --no-update`

Right I hesitated about talking about this, but decided not to in order to keep things simple... so the thing is:

As far as I know the `--no-update` flag does not exist any more in Poetry 2. Since Poetry 2 the "no update" behaviour is the default behaviour:

* <https://python-poetry.org/blog/announcing-poetry-2.0.0/#poetry-lock-defaults-to---no-update>
* <https://python-poetry.org/docs/cli/#lock>

And we changed to Poetry 2 when we dropped Python 3.8 and 3.9: <https://github.com/mxcube/mxcubecore/pull/1138>.

rhfogh:
Right. Fixed the Ruff errors that made the test fail. Now I am getting
Cannot install open-ldap.
See file below. Any ideas?
This is why I am tempted to simply give up on ever making the tests pass.

[pre-commit.log](https://github.com/user-attachments/files/19229061/pre-commit.log)

fabcor-maxiv:
> Right. Fixed the Ruff errors that made the test fail. Now I am getting Cannot install open-ldap. See file below. Any ideas? This is why I am tempted to simply give up on ever making the tests pass.
> 
> [pre-commit.log](https://github.com/user-attachments/files/19229061/pre-commit.log)

I guess you just got unlucky here. There was probably a glitch out of our hands. I re-triggered the job and it did run until the important part.

rhfogh:
OK, then we have my next help request. I understand that two unspecified files failed some unspecified ruff rule. Does it say anywhere which files and which rule? I have ruff integrated in my editor so I can find  the problems, but there are a lot of files, so it would be kind of helpful to know which ones to look at.

fabcor-maxiv:
> OK, then we have my next help request. I understand that two unspecified files failed some unspecified ruff rule. Does it say anywhere which files and which rule? I have ruff integrated in my editor so I can find the problems, but there are a lot of files, so it would be kind of helpful to know which ones to look at.

```none
ruff.....................................................................Passed
ruff-format..............................................................Failed
- hook id: ruff-format
- files were modified by this hook

warning: The following rule may cause conflicts when used with the formatter: `COM812`. To avoid unexpected behavior, we recommend disabling this rule, either by removing it from the `select` or `extend-select` configuration, or adding it to the `ignore` configuration.
2 files reformatted, 457 files left unchanged

poetry-check.............................................................Passed
```

Yes, it does not say which files and that is somewhat annoying (but there is some kind of a reason for this I think, that I can not recall right now). What is important to see is that it is not the "Ruff linter" that is complaining, but the "Ruff formatter" (which replaces Black). In principle running the `ruff format` command should do what is needed.

What bothers me is the `COM812` bit of this message. I thought I had fixed this already, I will give it another look.

But yes from your side, you should consider running `ruff format`, or it might be possible to configure your IDE to do this.

rhfogh:
Well, it finally worked.

I have to say I am thoroughly unimpressed with 1) a test system that simply fails your code without telling you what or where the problem is. 2) The general idea that the code formatting rules should trigger a failure rather than a warning. Was this deliberately coded to be maximally annoying?

marcus-oscarsson:
Thanks for all this work. It may be for some security reason as the report is public, not sure but I also find it annoying. There seems to be a `--show-fixes` switch that shows more details perhaps that could be used.

fabcor-maxiv:
> without telling you what or where the problem is

> switch that shows more details perhaps that could be used

I guess that the reason is that pre-commit hooks is the wrong place for this. The main idea of pre-commit hooks is to be very binary, you get either a thumb up or a thumb down.

I will investigate and see if we can make it a bit more helpful at the pre-commit hook level anyway. In the mean time, the recommendation is still to try and run all these checks before committing and pushing. When you run Ruff locally on the command line or in the IDE you should get all the useful details.

rhfogh:
@fabcor-maxiv Would it be possible to make changes so that you do not get correct code barred on purely aesthetic judgements, like formatting rules or import order??

fabcor-maxiv:
Sadly, no, I have no way of knowing if the code is correct.

rhfogh:
I think you could make an educated guess. If the problem is unused or uninitiated variables it is likely that there is a problem with the code. If the problem is imports not being alphabetically sorted, lines being too long, or doc strings missing a newline it is highly unlikely that this is the sign of a coding problem.

PR #1152, opened 2025-02-17
Use `ruff` as Python code formatter and linter

fabcor-maxiv:
Use `ruff` to replace:

* `autoflake`
* `black`
* `flake8`
* `isort`

Also reorganize the `pre-commit` hooks in a more logical order.


fabcor-maxiv:
Here we go for MXCuBE-Core...

Same as for MXCuBE-Web in <https://github.com/mxcube/mxcubeweb/pull/1555>, the goal of this PR is only to migrate to ruff, not about fixing lint rules violations (yet). And same as for MXCuBE-Web this did trigger some reformatting, mostly removing empty lines, but also some actual reformatting in particular regarding the `assert`s in the tests. These reformattings are a bit strange to me, but I guess also the code itself for these reformatted lines is a bit strange to begin with.

If you check `ruff.toml` you will see that there was a huge number of violations that needed to be ignored. Ultimate goal would be obviously to get all those violations fixed one way or another.

I also did a bit of cleaning on the pre-commit config file to stop files from being ignored for no reason.

marcus-oscarsson:
Thanks a lot very nice. I think we should go for it, it would as you mentioned be very nice to get rid of those violations at some point. Perhaps each site can check their files and then we can do a common effort for the common files. Lets bring it up during our next developers meeting.

PR #1153, opened 2025-02-17
Fix presence of package initializers

fabcor-maxiv:
Add missing package initializers `__init__.py`.
This fixes the violations of the `INP001` ruff rule.

Side effect of this is that it uncovers more occurrences
of violation of the `N999` ruff rule.
It is not clear what the logic is,
why `N999` was not reported when `INP001` was ignored.


marcus-oscarsson:
Thanks again :)

PR #1157, opened 2025-02-19
drop python 2 imports

elmjag:
Remove the imports that dealt with python2 and python3 compatibility. We don't support python 2 anymore.


fabcor-maxiv (approved):

beteva (approved):
Spring cleaning - good initiative 😄

PR #1158, opened 2025-02-20
TangoShutter: refactor (changes from https://github.com/mxcube/mxcubecore/pull/847/)

mockoocy:
Adds code changes from PR https://github.com/mxcube/mxcubecore/pull/847
And a test case for the TangoShutter Hardware Object.

At MAXIV we use these changes for quite a long time, but it has not been merged in the original for some reason (which is rather unclear looking at the PR's discussion).
The difference between the version at https://github.com/mxcube/mxcubecore/pull/847 and this PR is addition of type annotations and adoption of new linting rules.


test/pytest/test_hwo_isara_maint.py
```diff
@@ -90,7 +90,7 @@ def disable_remote_mode(self):
 
 def _disconnect_channels(maint: ISARAMaint):
     """
-    disconnect signal callbacks from tango attribute pollers
+    disconnect signal callbacks from tango attribute pollersf
```
elmjag: What up with this change?

mockoocy: Tragedy in few parts:
1. I opened this file 
2. I wanted to do `ctrl + f` to search for something.
3. I must have missed the `ctrl` and pressed `fn` :D
4. `git add test/`

mockoocy: Removed the change now.

test/pytest/test_hwo_tango_shutter.py
```diff
@@ -0,0 +1,123 @@
...
+        self._is_open = False
+        super().__init__(*args, **kwargs)
+
+    @command()
+    def Open(self):
+        self._is_open = True
+
+    @command()
+    def Close(self):
+        self._is_open = False
```
elmjag: Instead of disabling N802 check for this whole file in [ruff.toml](https://github.com/mxcube/mxcubecore/pull/1158/files#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989),
disable it only for these two special case methods, like this:

```suggestion
    @command()
    def Open(self):  # noqa: N802
        self._is_open = True

    @command()
    def Close(self):  # noqa: N802
        self._is_open = False
```

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -86,26 +98,18 @@ def init(self):
         self.state_channel.connect_signal("update", self._update_value)
         self.update_state()
 
-        try:
-            self.config_values = json.loads(self.get_property("values"))
-        except:
-            self.config_values = None
-
-    def _update_value(self, value):
+    def _update_value(self, value: DevState) -> None:
```
beteva: Are you sure the value is DevState type. I believe this is the output of the DevState command, which is a str.

beteva (approved):

mxcubecore/HardwareObjects/TangoShutter.py
beteva: Nice with the test.

beteva:
Nice with the test

fabcor-maxiv:
I recommend you check how the docstrings are rendered:

https://mxcubecore--1158.org.readthedocs.build/en/1158/dev/autosummary/mxcubecore.HardwareObjects.TangoShutter.html#module-mxcubecore.HardwareObjects.TangoShutter

marcus-oscarsson:
Oh yes, the documentation looks a bit strange.

marcus-oscarsson:
Thanks, alot. I think its really worth taking a look at how the documentation is rendered, there seems to be a tag missing or something similar.

mxcubecore/HardwareObjects/TangoShutter.py
mockoocy: When I checked the type of value there during runtime I got a `DevState` type. 
I believe that the `state_channel` could be set to return a string (with `TangoChannel.read_as_str` set to `True`).
 I think setting it as `DevState | str` is a safe choice then :)

mockoocy:
Thanks @beteva for fixing the xml code rendering in the docs! :)

I also pushed some changes, information about return rendered wrongly, when there was no space between docstring summary and `Args:, Returns:`, etc. blocks. (`:returns` text appeared in the summary)

fabcor-maxiv:
Looks like there is still some issue on the module docstring here: 
image

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -25,30 +25,44 @@
...
+adds specific tango shutter states.
+When redefining a known state, only the VALUES Enum will be updated.
+When defining a new state (new key), the dictionary value should be a
+list. The new state is added to both the VALUES and the SPECIFIC_STATES Enum.
+Attention:
+    - do not use tuples or the python json parser will fail!
+    - make sure only double quotes are used inside the values dictionary. No single quotes (') are allowed !
+    - the second element of the list should be a standard HardwareObjectState name
+    (UNKNOWN, WARNING, BUSY, READY, FAULT, OFF - see in BaseHardwareObjects.py)!
+The <values> property is optional.
```
fabcor-maxiv:
```suggestion
list. The new state is added to both the ``VALUES`` and the ``SPECIFIC_STATES`` enumeration.

Attention:

- do not use tuples or the Python JSON parser will fail!
- make sure only double quotes (``"``) are used inside the values dictionary,
  no single quotes (``'``) are allowed!
- the second element of the list should be a standard ``HardwareObjectState`` name
  (``UNKNOWN``, ``WARNING``, ``BUSY``, ``READY``, ``FAULT``, or ``OFF``;
  see in ``BaseHardwareObjects``)!
    
The ``<values>`` property is optional.
```

mockoocy: It works! Thanks :)
Adjusted the linting and commited.
Btw, in the future some ruff rules could be implemented for formatting these docstrings. It seems pretty easy to break these.

ruff.toml
```diff
@@ -7118,3 +7118,6 @@ convention = "google"
     "PT022",
     "S108",
 ]
+"test/pytest/test_hwo_tango_shutter.py" = [
+    "N802"
+]
```
fabcor-maxiv: Please, do not add anything in the Ruff ignores. We have to actually fix the rule violation. If it truly is not possible to fix the rule violation, then we can set an inline ignore directly in the Python file.

```suggestion
```

test/pytest/test_hwo_tango_shutter.py
fabcor-maxiv: Yes, I was about to comment on this as well.

Otherwise:

```suggestion
    @command()
    def open(self):
        self._is_open = True

    @command()
    def close(self):
        self._is_open = False
```

The upper case "convention" for Tango commands does not really matter here in a test case, I guess. Then you would need to change in the other places in the test case of course.

mockoocy: Done

mxcubecore/HardwareObjects/TangoShutter.py
```diff
@@ -114,36 +127,43 @@ def _initialise_values(self):
...
         try:
-            if self.config_values:
-                _state = self.config_values[str(self.state_channel.get_value())]
-            else:
-                _state = str(self.state_channel.get_value())
-
+            _state = self.get_value().name
+            return self.SPECIFIC_STATES[_state].value[0]
         except (AttributeError, KeyError):
+            logging.exception("Exception in get_state()")
```
fabcor-maxiv: This is strange to me.
Feels like it should have triggered Ruff rule `LOG015`: <https://docs.astral.sh/ruff/rules/root-logger-call/>.

```suggestion
            hwr_logger.exception("Exception in get_state()")
```

With somewhere towards the top, after the imports, something like this:

```
hwr_logger = logging.getLogger("HWR")
```

fabcor-maxiv: > Feels like it should have triggered Ruff rule `LOG015`: https://docs.astral.sh/ruff/rules/root-logger-call/.

Ah okay, it says right there that it is a "preview" rule that needs the `--preview` flag to be enabled.

Still should be fixed, though.

mockoocy: Well spotted :D Fixed

mxcubecore/HardwareObjects/TangoShutter.py
fabcor-maxiv: In principle there are (at least) two blocks or Ruff rules for docstring:

* https://docs.astral.sh/ruff/rules/#pydoclint-doc
* https://docs.astral.sh/ruff/rules/#pydocstyle-d

I am not sure if there are/were new violations to these rules here in this pull request. It feels to me like there are/were some violations, and I am surprised Ruff did not complain. I do not know how far Ruff can check that the docstrings are valid in terms of:
* reStructuredText
* Sphinx doc
* Google style

fabcor-maxiv (approved):

elmjag (approved):

PR #1161, opened 2025-02-21
1160 session selection based on id instead of time range

### LINKED ISSUE #1160: Session selection based on id instead of time range
Currently, the datasets are sent to the ingester along with the beamline and proposal. 
Based on that information and the time of data collection, the ingester can determine which session the data should be associated with.

https://github.com/mxcube/mxcubecore/blob/94fc0fac4eb33bffede10c4503a51b4ac58609db/mxcubecore/HardwareObjects/ICATLIMS.py#L750

However, our experience tells us that sooner or later, the ingester will not be able to determine the correct session—whether due to parallel sessions for the same proposal or two sessions occurring close in time.

So, we are force to use investigationIds instead and a different way to pick up the sessions

antolinos:
This PR contains a new feature:
- Dataset's metadata contains now a new metadata parameter called `investigationId that will be used downstream to associate unequivocally teh datasets to the session
And one fix:
- Fix problem with beamline_name


marcus-oscarsson:
Thanks, ! well done

PR #1164, opened 2025-02-21
Added metadata to the investigations

### LINKED ISSUE #1163: Improve session identification on the list
The current UI allows to select the session and it is prone to errors. We could improve it by showing the session status (if samples have been collected, shipped, etc...)
For doing so we need to get this data from the LIMS system.

antolinos:
it closes #1163 

Now, the backend associated the number of datasets and samples collected as well as the volume of data produced (in bytes)

Data is exposed via
`/mxcube/api/v0.1/login/login_info`


marcus-oscarsson:
Thanks @antolinos , looks good

marcus-oscarsson (approved):

PR #1166, opened 2025-02-24
Added get_samples function to retrieve the acronym

### LINKED ISSUE #1156: Sample names and protein acronyms are not well managed when retrieving samples from DRAC
Since the endpoint retrieves only a reference to the protein acronym and not the acronym itself, the sample name is currently used instead which is obviously not good

https://github.com/mxcube/mxcubecore/blob/1406ec98784bbd4e7491ad803c73b2a2e8298b5f/mxcubecore/HardwareObjects/ICATLIMS.py#L204
https://github.com/mxcube/mxcubecore/blob/1406ec98784bbd4e7491ad803c73b2a2e8298b5f/mxcubecore/HardwareObjects/ICATLIMS.py#L207

We need a way to retrieve the acronym from the sample Id

antolinos:
The acronym is a mandatory parameter for each sample, as it is frequently used when creating the final target folder.
However, it is not retrieved when downloading the puck information, so a second request is made to obtain the sample (or sample sheets) by investigation.

This PR proposes to retrieve such information when the synchronization is triggered as well as a new model has been added to: `model/lims_session`

I wonder if this file should be renamed (to `lims` if it groups all models for the lims) or refactored in smaller bits

### DIFF
--- a/mxcubecore/model/lims_session.py
+++ b/mxcubecore/model/lims_session.py
@@ -3,6 +3,7 @@ from datetime import (
     timedelta,
 )
 from typing import (
+    Any,
     Dict,
     List,
     Optional,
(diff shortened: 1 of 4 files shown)


mxcubecore/model/lims_session.py
```diff
@@ -70,6 +71,43 @@ class Session(BaseModel):
     logbook_URL: Optional[str] = None
 
 
+class Instrument(BaseModel):
```
marcus-oscarsson: It would in general be very nice with some comments on each field and class, unless its evident what it is. It would make it easier to understand what purpose they have and for others to adopt.

marcus-oscarsson:
It looks like the `poetry.lock` is outdated, it should be taken care of if you run `pre-commit run --all-files`

marcus-oscarsson:
Hm, @antolinos there still seems to be some conflicts here could you have another look please ?

marcus-oscarsson:
@antolinos, I did a re-base and updated the `.lock` file. Let me know if it still looks alright ?

antolinos:
Hi @marcus-oscarsson 

Sorry my late reply. It is like classes are now duplicated under lims_session.py probably from the rebase.
How do you want to proceed?

marcus-oscarsson:
Ah, wops. I think its better now ?

mxcubecore/model/lims_session.py
marcus-oscarsson: Perhaps you could follow up with the documentation in a separate PR, I wont let it hold up the PR

PR #1169, opened 2025-03-03
set atexit in mockupcamera

walesch-yan:
There was an issue regarding the correct shutdown of `video-streamer` processes in the MXCuBE-Web repository (see mxcube/mxcubeweb#1577) . A patch was introduced in the latest version of `video-streamer` that handles a signal catching issue (see mxcube/video-streamer#34). 

Until now, MXCuBE maintains a registry of spawned child processes in `/tmp/mxcube.pid`. However, the `mockupcamera` was never registered, which caused another issue for correct process termination. Additionally the `temp` file solution is violating `ruff` rule [S108](https://docs.astral.sh/ruff/rules/hardcoded-temp-file/). 

To solve this instead of registering the pid of the `video-streamer` process in a separate file, which will later be read by the server, this PR makes use of the `atexit.register` functionality directly.
Additionally, I moved the dependency of the `video-streamer` directly to `mxcubecore`, since it is actually not directly used in the web but in the camera components.

### DIFF
--- a/mxcubecore/HardwareObjects/TangoLimaMpegVideo.py
+++ b/mxcubecore/HardwareObjects/TangoLimaMpegVideo.py
@@ -13,6 +13,10 @@ Example configuration:
 </object>
 """
 
+import atexit
+import logging
+import os
+import signal
 import subprocess
 import uuid
 from typing import (
@@ -71,6 +75,10 @@ class TangoLimaMpegVideo(TangoLimaVideo):
 
         return video_sizes
 
+    def clean_up(self) -> None:
+        logging.getLogger("HWR").info("Shutting down video_stream...")
+        os.kill(self._video_stream_process.pid, signal.SIGTERM)
+
     def start_video_stream_process(self) -> None:
         if (
             not self._video_stream_process
@@ -95,10 +103,10 @@ class TangoLimaMpegVideo(TangoLimaVideo):
                     self.stream_hash,
                 ],
                 close_fds=True,
+                stdout=subprocess.DEVNULL,
             )
 
-            with open("/tmp/mxcube.pid", "a") as f:
-                f.write("%s " % self._video_stream_process.pid)
+            atexit.register(self.clean_up)
 
     def stop_streaming(self) -> None:
         if self._video_stream_process:
--- a/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
@@ -1,6 +1,9 @@
 """Class for cameras connected to framegrabbers run by Taco Device Servers"""
 
+import atexit
 import logging
+import os
+import signal
 import subprocess
 import time
 from typing import (
@@ -115,6 +118,10 @@ class MDCameraMockup(BaseHardwareObjects.HardwareObject):
         scale = float(width) / self.get_width()
         return (width, height, scale)
 
+    def clean_up(self):
+        logging.getLogger("HWR").info("Shutting down video_stream...")
+        os.kill(self._video_stream_process.pid, signal.SIGTERM)
+
     def start_video_stream_process(self) -> None:
         if (
             not self._video_stream_process
@@ -139,7 +146,10 @@ class MDCameraMockup(BaseHardwareObjects.HardwareObject):
                     self.stream_hash,
                 ],
                 close_fds=True,
+                stdout=subprocess.DEVNULL,
             )
+
+            atexit.register(self.clean_up)
 
     def stop_streaming(self) -> None:
         if self._video_stream_process:
--- a/poetry.lock
+++ b/poetry.lock
@@ -25,6 +25,42 @@ files = [
 ]
 
 [[package]]
+name = "anyio"
+version = "4.8.0"
+description = "High level compatibility layer for multiple asynchronous event loop implementations"
+optional = false
+python-versions = ">=3.9"
+groups = ["main"]
+files = [
+    {file = "anyio-4.8.0-py3-none-any.whl", hash = "sha256:b5011f270ab5eb0abf13385f851315585cc37ef330dd88e27ec3d34d651fd47a"},
+    {file = "anyio-4.8.0.tar.gz", hash = "sha256:1d9fe889df5212298c0c0723fa20479d1b94883a2df44bd3897aa91083316f7a"},
+]
+
+[package.dependencies]
+exceptiongroup = {version = ">=1.0.2", markers = "python_version < \"3.11\""}
+idna = ">=2.8"
+sniffio = ">=1.1"
+typing_extensions = {version = ">=4.5", markers = "python_version < \"3.13\""}
+
+[package.extras]
+doc = ["Sphinx (>=7.4,<8.0)", "packaging", "sphinx-autodoc-typehints (>=1.2.0)", "sphinx_rtd_theme"]
+test = ["anyio[trio]", "coverage[toml] (>=7)", "exceptiongroup (>=1.2.0)", "hypothesis (>=4.0)", "psutil (>=5.9)", "pytest (>=7.0)", "trustme", "truststore (>=0.9.1)", "uvloop (>=0.21)"]
+trio = ["trio (>=0.26.1)"]
+
+[[package]]
+name = "async-timeout"
+version = "5.0.1"
+description = "Timeout context manager for asyncio programs"
+optional = false
+python-versions = ">=3.8"
+groups = ["main"]
+markers = "python_full_version <= \"3.11.2\""
+files = [
+    {file = "async_timeout-5.0.1-py3-none-any.whl", hash = "sha256:39e3809566ff85354557ec2398b55e096c8364bacac9405a7a1fa429e77fe76c"},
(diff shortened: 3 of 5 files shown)


walesch-yan:
> Until now, MXCuBE maintains a registry of spawned child processes in /tmp/mxcube.pid. However, the mockupcamera was never registered, which caused another issue for correct process termination. Additionally the temp file solution is violating ruff rule [S108](https://docs.astral.sh/ruff/rules/hardcoded-temp-file/).

As of my knowledge, the only pids that are registered to this file are the `video-streamer` process from `TangoLimaMpegVideo` and the server in MXCuBE-Web. As such I would also advocate in completely removing this and use `atexit` manually in these locations or using another protocol for registrations. (A redis server for example, however this would need a more careful introduction).

mxcubecore/HardwareObjects/TangoLimaMpegVideo.py
```diff
@@ -97,8 +105,7 @@ def start_video_stream_process(self) -> None:
                 close_fds=True,
             )
 
-            with open("/tmp/mxcube.pid", "a") as f:
-                f.write("%s " % self._video_stream_process.pid)
+            atexit.register(self.clean_up)
```
walesch-yan, line 109: I now also modified the registration for the video-streamer in `tangoLimaMPEGVideo`

fabcor-maxiv:
Looks like I still have some `video-streamer` and `ffmpeg` processes left behind with this branch of `mxcubecore` and `mxcube_video_streamer==1.8.1`.

I do see the following message that did not seem to exist before though:

> 2025-03-04 08:34:20,200 |HWR    |INFO   | Shutting down video_stream...

walesch-yan:
> Looks like I still have some `video-streamer` and `ffmpeg` processes left behind with this branch of `mxcubecore` and `mxcube_video_streamer==1.8.1`.
> 
> I do see the following message that did not seem to exist before though:
> 
> > 2025-03-04 08:34:20,200 |HWR    |INFO   | Shutting down video_stream...

Yes, I added the message to make sure that the shutdown process is called.
Could it be a process that is still running from the previous version? I.e. when you run and exit the server after a few seconds, are there new processes that are left behind or the same ones?

I do see that my changes have one issue: if the shutdown is not called when exiting the code, the processes will not be terminated (this is similar as it was before), however different than before it will not clean-up after processes that have been run on previous server runs and not terminated correctly, as there is no information left of old processes/there is no file anymore that keeps track.
This is especially a problem when the server is crashing.
I agree that I could revert back to the old tempfile functionality as it is more practical, but we probably should find another solution security-wise and ruff-wise this one is not sustainable

fabcor-maxiv:
> Could it be a process that is still running from the previous version? I.e. when you run and exit the server after a few seconds, are there new processes that are left behind or the same ones?

I checked again now:

* No `ffmpeg` or `video-streamer` process running
* Start MXCuBE-Web
* There are `ffmpeg` and `video-streamer` processes
* Kill MXCuBE-Web with `CTRL+C`
* Message: "_HWR    |INFO   | Shutting down video_stream..._"
* But still there are `ffmpeg` and `video-streamer` processes running
* Run `pkill ffmpeg`
* Message: "_Received SIGTERM, shutting down gracefully..._"
* `ffmpeg` and `video-streamer` processes are gone

> I agree that I could revert back to the old tempfile functionality as it is more practical, but we probably should find another solution security-wise and ruff-wise this one is not sustainable

The `atexit` solution seems cleaner, I would prefer it, obviously. But if really we do not manage to make it work at all, then we can choose to ignore Ruff's rule.

Is there a way to check if the `os.kill` succeeded?

We could also try to send `SIGKILL` instead of `SIGTERM`.

fabcor-maxiv:
Sending `SIGKILL` seems to do the trick for me:

```diff
diff --git i/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py w/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
index 853e40828..7f45a1311 100644
--- i/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
+++ w/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py
@@ -120,7 +120,7 @@ class MDCameraMockup(BaseHardwareObjects.HardwareObject):
 
     def clean_up(self):
         logging.getLogger("HWR").info("Shutting down video_stream...")
-        os.kill(self._video_stream_process.pid, signal.SIGTERM)
+        os.kill(self._video_stream_process.pid, signal.SIGKILL)
 
     def start_video_stream_process(self) -> None:
         if (
```

I am not saying that is what we should do. In principle that should not be necessary. Sending `SIGTERM` should be good enough. I guess it needs more investigating...

walesch-yan:
Thanks for testing it @fabcor-maxiv :)

You are right, it seems that I incorrectly tested `SIGTERM` handling yesterday. I also tried with `SIGINT`, this seems to work as well as `SIGKILL`. I will investigate this and try to make it work for SIGTERM as well.

> Is there a way to check if the os.kill succeeded?

One could wait a few seconds, then use `psutil` library, otherwise one could do something like

```python
    try:
        os.kill(pid, 0) 
        print("process was still running")
    except OSError:
        print("no process detected")
```

since `os.kill` will raise an error if there is no process with the given `pid` and do nothing else. Also a good way to clean up simultaniouly during testing

walesch-yan:
> I am not saying that is what we should do. In principle that should not be necessary. Sending SIGTERM should be good enough. I guess it needs more investigating...

I agree, however currently `SIGKILL` was actually always send, I thought having `SIGTERM` with graceful shutdown would be a better/more elegant solution.

I will spend a little more time investigating the issue

marcus-oscarsson:
Thanks @walesch-yan :)

walesch-yan:
I changed something in the `video-streamer`, that should ensure the correct handling of `SIGTERM`. In my tests, now the changes from this branch did work fine for me. I did not even need the `atexit` part when using ctrl-c in the terminal since it gets correctly propagated in this case. But, since we have `sys.exit` calls for some crashes, it makes sense to keep it. 

If somebody could test it with the current `main` of [video-streamer](https://github.com/mxcube/video-streamer) to confirm this? (i.e. before i update the version)

marcus-oscarsson:
Excellent, well done @walesch-yan

fabcor-maxiv:
> If somebody could test it with the current `main` of [video-streamer](https://github.com/mxcube/video-streamer) to confirm this? (i.e. before i update the version)

Done. Seems to have worked fine. The `ffmpeg` and `video-streamer` processes are gone after killing MXCuBE-Web with `CTRL+C`.

walesch-yan:
Thanks!

walesch-yan:
Short Recap of the changes after latest push:

- `Video-streamer` dependency is added and fixed to latest patch 1.8.2 which handles signals better than previous versions
-  Replace usage of temporary file to save PID of spawned processes in `MDMockupCamera` and `TangoLimaMPEGVideo` classes, by a cleanup procedure called with `atexit` -> Solves Minor security issue and allows more granulated exit handling
- Suppress console output of `video-streamer` application
- remove ruff's `S108` rule from the ignored list for the above mentioned classes

Beneficial Site-Effect:

- The latest Patch of `Video-Streamer` also updates the version of `Jinja2` to one that introduced a patch for recently found security vulnerability. `mxcubecore` doesn't depend directly on `Jinja2`, but a few other dependencies do. Those do not require a minimum version that would fix the vulnerability. Hence, this update fixes this.

To remember:
Since this removes the tempfile, we assume that crashes on the application are handled correctly and call a system exit. Otherwise, we might end up in a scenario, where zombie processes are called, that are not automatically cleaned by the application next correct shutdown.

fabcor-maxiv:
Nice!

---

> Since this removes the tempfile, we assume that crashes on the application are handled correctly and call a system exit. Otherwise, we might end up in a scenario, where zombie processes are called, that are not automatically cleaned by the application next correct shutdown.

Have you tried this scenario, where MXCuBE-Web crashes? Does this really happen that the `atexit` is not run and processes are left behind in such a case?

walesch-yan:
> Have you tried this scenario, where MXCuBE-Web crashes? Does this really happen that the `atexit` is not run and processes are left behind in such a case?

Yes, I tested with two different approaches (using `SIGKILL` and `SIGSEGV`). In both cases, the `atexit` was not called and three processes where left behind (`video-streamer` server, process to poll images and `ffmpeg` stream). 

However, it behaves a little different than what I thought initially: When restarting the `mxcube` server, the newly created `video-stream` process will not correctly start, as the port is already taken by the orphan process. This will lead to a graceful shutdown of the new streamer process, resulting in only the old process running on that port. However, the front-end will be able to connect to it and show the image from the previous streamer. The fact that this happens is difficult to see, as we will see an Image as we used to it.

This unfortunately still has two main issues:
- If we want to manually change the source of the stream in mxcube, even if done correctly, we will see the old stream
- `MXCuBE` is now unable to shut down the old stream, hence it will continue running even when we stop the `MXCuBE` server correctly (instead, we will see a `ProcessLookupError` message)

P.S. If you want to test out the Scenario I described, you can start the server and then forcefully kill it (`SIGKILL`) or add `import ctypes; ctypes.string_at(0)` after registering the cleanup function to `atexit` in `MDMockupCamera` class. The later tries to read memory from `0x0` which is protected by the os and will call a Segmentation Fault (`SIGSEGV`).

PR #1171, opened 2025-03-11
More flexible  handling of detector cover

beteva:
Choose to use or not the detector cover for a given data collection.
The choice is handles as an NState and added as True/False button in  BeamlineActions.

### DIFF
--- a/mxcubecore/HardwareObjects/Microdiff.py
+++ b/mxcubecore/HardwareObjects/Microdiff.py
@@ -324,6 +324,7 @@ class Microdiff(MiniDiff.MiniDiff):
         self.state.connect_signal("update", self._update_state)
 
         HardwareObject.init(self)
+        self.handle_detector_cover = self.get_object_by_role("handle_detcover")
 
     def _update_value(self, value=None):
         if value is None:
@@ -406,22 +407,46 @@ class Microdiff(MiniDiff.MiniDiff):
                 time.sleep(0.5)
 
     def open_detector_cover(self, timeout=10):
-        try:
-            detcover = self.get_object_by_role("controller").detcover
+        use = True
+        if self.handle_detector_cover:
+            try:
+                use = self.handle_detector_cover.get_value().value
+            except AttributeError:
+                use = False
 
-            if detcover.state == "IN":
-                detcover.set_out(timeout)
-        except AttributeError:
-            logging.getLogger("HWR").exception("No detector cover configured")
+        logging.getLogger("user_level_log").info("Use detector cover open: %s" % use)
+
+        if use:
+            try:
+                detcover = self.get_object_by_role("controller").detcover
+
+                if detcover.state == "IN":
+                    detcover.set_out(timeout)
+            except AttributeError:
+                logging.getLogger("user_level_log").exception(
+                    "No detector cover configured"
+                )
 
     def close_detector_cover(self, timeout=10):
-        try:
-            detcover = self.get_object_by_role("controller").detcover
+        use = True
+        if self.handle_detector_cover:
+            try:
+                use = self.handle_detector_cover.get_value().value
+            except AttributeError:
+                use = False
 
-            if detcover.state == "OUT":
-                detcover.set_in(timeout)
-        except AttributeError:
-            logging.getLogger("HWR").exception("No detector cover configured")
+        logging.getLogger("user_level_log").info("Use detector cover close: %s" % use)
+
+        if use:
+            try:
+                detcover = self.get_object_by_role("controller").detcover
+
+                if detcover.state == "OUT":
+                    detcover.set_in(timeout)
+            except AttributeError:
+                logging.getLogger("user_level_log").exception(
+                    "No detector cover configured"
+                )
 
     def phase_prepare(self, phase):
         if phase == "Centring":


mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -406,22 +407,38 @@ def _wait_ready(self, timeout=None):
...
     def open_detector_cover(self, timeout=10):
-        try:
-            detcover = self.get_object_by_role("controller").detcover
+        use = True
+        if self.handle_detector_cover:
+            try:
+                use = self.handle_detector_cover.get_value().value
+            except AttributeError:
+                use = False
+        print(f"Use detecot cover open: {use}")
```
fabcor-maxiv:
```suggestion
```

mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -406,22 +407,38 @@ def _wait_ready(self, timeout=None):
...
     def close_detector_cover(self, timeout=10):
-        try:
-            detcover = self.get_object_by_role("controller").detcover
+        use = True
+        if self.handle_detector_cover:
+            try:
+                use = self.handle_detector_cover.get_value().value
+            except AttributeError:
+                use = False
+        print(f"Use detecot cover close: {use}")
```
fabcor-maxiv:
```suggestion
```

fabcor-maxiv (commented):
@beteva 

Can you remove ALL `print` statements in this file and then remove `T201` from the ignored rules at the following location in `ruff.toml`?

https://github.com/mxcube/mxcubecore/blob/240212d83d3386346428f88925602558abb2b4a1/ruff.toml#L3854

mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -406,22 +407,46 @@ def _wait_ready(self, timeout=None):
...
+            try:
+                use = self.handle_detector_cover.get_value().value
+            except AttributeError:
+                use = False
 
-            if detcover.state == "IN":
-                detcover.set_out(timeout)
-        except AttributeError:
-            logging.getLogger("HWR").exception("No detector cover configured")
+        logging.getLogger("user_level_log").info("Use detecot cover open: %s" % use)
```
fabcor-maxiv: `detecot`? I guess it is supposed to be `detector`, right?

```suggestion
        logging.getLogger("user_level_log").info("Use detector cover open: %s" % use)
```

mxcubecore/HardwareObjects/Microdiff.py
```diff
@@ -406,22 +407,46 @@ def _wait_ready(self, timeout=None):
...
+            try:
+                use = self.handle_detector_cover.get_value().value
+            except AttributeError:
+                use = False
 
-            if detcover.state == "OUT":
-                detcover.set_in(timeout)
-        except AttributeError:
-            logging.getLogger("HWR").exception("No detector cover configured")
+        logging.getLogger("user_level_log").info("Use detecot cover close: %s" % use)
```
fabcor-maxiv:
```suggestion
        logging.getLogger("user_level_log").info("Use detector cover close: %s" % use)
```

beteva: Thanks for correcting my typos

ruff.toml
```diff
@@ -3851,7 +3851,7 @@ convention = "google"
     "S110",
     "SIM105",
     "SIM118",
-    "T201",
+    "ERA001"
```
fabcor-maxiv: Commenting the `print` statements so that we ignore `ERA001` instead of `T201` is not at all what I had in mind. Please let's not do this. This is basically a step backwards. Let's remove the `print` statements once and for all, replace them with logging entries if/when it makes sense.

fabcor-maxiv (changes requested):
I recommend to squash all these commits now if possible, but for sure before merging

ruff.toml
beteva: I know that commenting out the print is not the best to do and there are even more problems with the Microdiff.py (like missing docstrings), but it will be at some time replaced by the AbstractDiffractometer, so can we leave it as it is for now.

Otherwise what is wrong squashing at merge? Maybe for the reviewer is better to know the different steps, not just having one general commit. I agree, though, that once reviewed, we should squash when merging.

fabcor-maxiv: > I know that commenting out the print is not the best to do and there are even more problems with the Microdiff.py (like missing docstrings), but it will be at some time replaced by the AbstractDiffractometer, so can we leave it as it is for now.

I disagree. Either replace with `logger.debug` or delete the `print`, it is simple. If the class is going to be deleted soon anyway, so why hold onto those prints? What is wrong with `logger.debug`? I do not understand the logic at all.


> Otherwise what is wrong squishing at merge?

I am not against squashing at merge, but do we do it? In my experience we do not, but maybe I am wrong.


> Maybe for the reviewer is better to know the different steps, not just having one general commit.

As far as I know, there is a way in GitHub to see the different versions of the same pull request. In my experience it is not necessary to keep the commits, it is better to amend the one commit and "force push with lease".

beteva: > I disagree. Either replace with `logger.debug` or delete the `print`, it is simple. If the class is going to be deleted soon anyway, so why hold onto those prints? What is wrong with `logger.debug`? I do not understand the logic at all.

Well, I guess the logic behind using prints and not logger is when you do not want to log information , which is still handy to have on the screen. As for these prints, I just did the least work and did not look how to replace them.

fabcor-maxiv: > Well, I guess the logic behind using prints and not logger is when you do not want to log information , which is still handy to have on the screen

Then we need to use a different logger, and configure the loggers properly so that the logging entries go to the right places (terminal, log file, UI, or whatever).

OK, at this point I would recommend to revert the prints (i.e. remove the `# ` comment signs) so that we stay with the `T201` violations that we already had in the beginning, instead of taking a step backwards by adding `ERA001` violations.

We should NOT add anything to the `[lint.per-file-ignores]` list of `ruff.toml`, that is absolutely not the way to go. This list was created only to ease the transition. We should remove violations, not add more violations.

beteva: > Then we need to use a different logger, and configure the loggers properly so that the logging entries go to the right places (terminal, log file, UI, or whatever).

Totally agree with you. 


> We should NOT add anything to the `[lint.per-file-ignores]` list of `ruff.toml`, that is absolutely not the way to go. This list was created only to ease the transition. We should remove violations, not add more violations.

Totally agree with you.

marcus-oscarsson:
Thanks @beteva. Looks good as far as I can see. @fabcor-maxiv are you happy like this ?, if its the case could you approve the PR so that we don;t have the "changes requested" ?

fabcor-maxiv (approved):
You'll say it is nitpicking, but I do not like that there are 3 commits:

1. the actual change advertised in the pull request description
1. unrelated bug fix
1. commit advertised as typo fix
    * but is not a typo fix
    * should be squashed in change 1. but now we can not easily squash the whole pull request because of the unrelated commit change that sneaked in

marcus-oscarsson:
Id say that its by principle right but that we sometime need to be a bit pragmatic. I also prefer when the PR's only contain one fix or feature and that the history is easy to follow. This PR is quite easy to pull apart but there are sometimes larger bodies of work that are not so easy for people to separate or very time consuming. Sometimes, at least in my opinion, we have to favor progress and contribution before how we think/know how things should ideally be done.

fabcor-maxiv:
> Id say that its by principle right but that we sometime need to be a bit pragmatic. I also prefer when the PR's only contain one fix or feature and that the history is easy to follow. This PR is quite easy to pull apart but there are sometimes larger bodies of work that are not so easy for people to separate or very time consuming. Sometimes, at least in my opinion, we have to favor progress and contribution before how we think/know how things should ideally be done.

That is one of the reasons why we have code review, so that the ones who know when it is okay to split or not split can tell the ones who do not know. This one would be very easy to fix. I am also being pragmatic by blocking only when I believe the issues are too big to let through, and by approving even though there is clear room for improvement.

marcus-oscarsson:
Et voila ;), re-based, squashed and split

PR #1172, opened 2025-03-12
[Bug fix] Added forgotten self in snapshot logic

marcus-oscarsson:
- `self` was missing in the snapshot logic when plates are used

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractMultiCollect.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractMultiCollect.py
@@ -318,8 +318,8 @@ class AbstractMultiCollect(object):
         snapshot_directory = dc_params["fileinfo"]["archive_directory"]
 
         if HWR.beamline.diffractometer.in_plate_mode():
-            if number_of_snapshots > 0:
-                number_of_snapshots = 1
+            if self.number_of_snapshots > 0:
+                self.number_of_snapshots = 1
 
         if not os.path.exists(snapshot_directory):
             self.create_directories(snapshot_directory)


mxcubecore/HardwareObjects/abstract/AbstractMultiCollect.py
```diff
@@ -318,8 +318,8 @@ def take_snapshots(self, dc_params):
         snapshot_directory = dc_params["fileinfo"]["archive_directory"]
 
         if HWR.beamline.diffractometer.in_plate_mode():
-            if number_of_snapshots > 0:
-                number_of_snapshots = 1
+            if self.number_of_snapshots > 0:
+                self.number_of_snapshots = 1
```
walesch-yan, line 323: Maybe I'm missing something, but when we do this, `number_of_snapshots` is permanently overwritten, shouldn't we add a local variable, which gets set to 1 if `in_plate_mode` and replace the `self.number_of_snapshots` in the loop below with the local variable, to avoid mismatching of the class attribute later?:)

marcus-oscarsson: Hm there is probably some more elegant handling of this indeed. However when you are in plate mode meaning that you have replaced your goniometer head with a plate gripper. You can't rotate more than a few degrees taking more than one snapshot will not work. There should probably also be a feature that removes the possibility to even select more than 1 snapshot in this case (But I don't think its the case at the moment). So setting `self.number_of_snapshots` is not an issue in this case.

walesch-yan: I see in that case you can ignore my comment
Well spotted anyways!

walesch-yan (approved):

marcus-oscarsson:
:), It was actually @beteva that made the fix. I just split the original PR #1171 into two pieces as discussed. Your the one that did that was attentive, its true that its a strange corner case. You did well to spot it ;)

PR #1176, opened 2025-03-17
[RUFF] - Lint with Ruff

marcus-oscarsson:
Just to visualize the linting diffrences


marcus-oscarsson:
I'm starting to work on re-basing this now. I'm not sure how to proceed with the linting because when I lint this I get 828 errors that can't be automatically fixed (and thats just with the files that are changed in this PR).

Ill proceed with the re-base but its possible that we need to disable the pre-commit lint step for this PR and then go back and deal with this on case per case basis. Unless you have some other good idea of how to handle this.

I've made one commit in addition to the original `xml_yaml_conversion` branch and that is the ruff lint and formatting, you can see the results in this PR.

@elmjag, @fabcor-maxiv, Maybe you have some ideas ?

rhfogh:
This is where it would be useful to have a test system that actually identified the errors that we need to fix. Hopefully somebody has an idea how to do that?

PR #1177, opened 2025-03-17
[XML to YAML] - Rebased branch `xml_yaml_conversion` on `develop`

marcus-oscarsson:


marcus-oscarsson:
Ok - So I made a comment earlier but I don't know where that went,  it somehow disappeared from my first PR.

I made one commit that simply runs the new `pre-commit` with `ruff`. As there where some additions to the rules that we apply we get quite a lot of errors that can't be fixed automatically. There was about ***820*** errors that needs to be fixed by hand.

The first commit `[RUFF] - Lint with Ruff` simply does the re-formatting to avoid basic conflicts later on. The rest are your (@elmjag and @rhfogh) commits. 

So we might need to disable/override the `pre-commit` pipeline for this PR looking at the number of linting errors that needs to be fixed. We should then address each file, as we are doing now, on case by case basis. Maybe you have some other good suggestions @fabcor-maxiv and @elmjag.

Otherwise I would ask the original authors to have a look and see if it all looks good. 

@rhfogh, I tried to do as best as I could with the GPHL related files, I managed to solve the conflicts but I kept all the configuration files as they where in the `xml_yaml_conversion` branch, the indentation differences just made it very complicated to merge properly. I let you have a look at those file and make sure that they are up to date (`gphl_setup.yml`, and `gphl_workflow.yml`). I also noticed some inconsistency in how `settings` are retrieved in `GphlWorkflow.py`, there was also the introduction of the `YamlConfiguredObject` from `develop` and I don't know what consequences that might have. So if you can please have a look at:

- gphl_setup.yml
- gphl_workflow.yml
- GphlWorkflow.py (sometimes, settings are accessed as self.settings and sometimes as self.config.settings)

marcus-oscarsson:
@elmjag, Could you have a look at: 

https://github.com/mxcube/mxcubecore/blob/6aaa91ae1065a5e7b0d9222ab217297c59022129/test/pytest/test_setup_commands_channels.py. 

It looks like a test that is only in this branch, if you are still using it could you update the `mock` to `unittest.mock` ?

marcus-oscarsson:
Not sure why but my `pre-commit` says, 42 errors now which I guess is handable. I let you guys have a look and then we can decide what to do with the linting errors.

fabcor-maxiv:
> Not sure why but my `pre-commit` says, 42 errors now which I guess is handable. I let you guys have a look and then we can decide what to do with the linting errors.

```none
Found 42 errors (5 fixed, 37 remaining).
No fixes available (13 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

OK, 42 is not too bad. Looks like some of them (5 + 13) can be fixed automatically by Ruff.
In my experience so far, the "unsafe" fixes (13) are also quite straightforward and can be reviewed easily.
For the rest, either it can be fixes manually or we disable the rule again (inline or in `ruff.toml`).

marcus-oscarsson:
I tried with the `--unsafe-fixes` option, it fixed some things but it does not seem to have fixed what it said it would fix.

marcus-oscarsson:
I simply updated `pre-commit` to run with `--unsafe-fixes`, but maybe it should be done in some other way.

fabcor-maxiv:
@marcus-oscarsson

You lost `.yamllint.yaml` in commit `6aaa91ae1065a5e7b0d9222ab217297c59022129`.

If it is causing too much pain, it is fair to add the ignores on file-by-file basis in `ruff.toml`. Rebasing such a big branch is already hard enough as it is.

marcus-oscarsson:
> You lost `.yamllint.yaml` in commit `6aaa91ae1065a5e7b0d9222ab217297c59022129`.

Yes I noticed that, I added it back

.pre-commit-config.yaml
```diff
@@ -36,7 +36,7 @@ repos:
     rev: v0.9.6
     hooks:
       - id: ruff
-        args: [--fix]
+        args: [--unsafe-fixes]
```
fabcor-maxiv:
```suggestion
        args: [--fix]
```

I would not add the "unsafe" fixes here. Those should be done manually, in my opinion, not automatically in pre-commit hooks or in CI/CD pipelines.

Anyway, for it to work you would need both flags `ruff check --fix --unsafe-fixes`. But again this should be run manually on the command line and reviewed before committing (and for sure before pushing) because those fixes have a higher chance of introducing errors than the normal fixes. I recommend against adding this to the pre-commmit hooks.

marcus-oscarsson: > I would not add the "unsafe" fixes here. Those should be done manually, in my opinion, not automatically in pre-commit hooks or in CI/CD pipelines.
> 
> Anyway, for it to work you would need both flags `ruff check --fix --unsafe-fixes`. But again this should be run manually on the command line and reviewed before committing (and for sure before pushing) because those fixes have a higher chance of introducing errors than the normal fixes. I recommend against adding this to the pre-commmit hooks.

I know, I know I'm sort for working on this, and I noticed that one needs to run  `ruff check --fix --unsafe-fixes`. Ill let you know when I'm done

marcus-oscarsson:
So, I went over the errors reported by ruff and I fixed most and added some exceptions on some lines, most of them would be quite easy to fix, ill comment those in the code in a moment. Ill let you have a look an decide if you'd like to address them in this PR or if we do that in separate PR's later on.

Yes I was a bit naughty and disabled "G004" globally at-least for now. 

Please have a look and double check the re-based result, especially for @elmjag and @rhfogh that now the branch well.

The linting step should pass now, it would be great if you @elmjag or @fabcor-maxiv could have a look at the `mock` import in the tests.

mxcubecore/HardwareObjectFileParser.py
```diff
@@ -376,13 +376,13 @@ def endElement(self, name):
                             )
                 elif name == self.property:
                     del self.objects[-1]  # remove empty object
-                    self.objects[-1]._set_property(name, self.buffer)
+                    self.objects[-1]._set_property(name, self.buffer)  # noqa: SLF001
```
marcus-oscarsson, line 379: This is the rule that is perhaps more difficult to get rid of as it would require changes in several places in the code base. But in principle not hard to do

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -194,8 +194,8 @@ class HOConfig(ConfiguredObject.HOConfig):
         Required during transition, as long as we do nto have teh fields defined"""
 
         # Defaults - should be replaced by proper Pydantic
-        workflows = {}
-        settings = {}
+        workflows = {}  # noqa: RUF012
```
marcus-oscarsson, line 197: I think this can be solved by adding a type hint

mxcubecore/HardwareRepository.py
```diff
@@ -279,13 +276,13 @@ def write_yaml(data, file_name: str):
...
                 objects[role] = "%s.yaml" % obj.id
-            except:
+            except Exception:
                 logging.getLogger("HWR").exception("")
 
     config = result["configuration"] = {}
     for tag, val in hwobj.config.model_dump().items():
         if tag not in objects_by_role:
-            config[tag] = val
+            config[tag] = val  # noqa: PERF403
```
marcus-oscarsson, line 281: This could be re-written with a dict comprehension

mxcubecore/model/protocols/epics.py
```diff
@@ -39,7 +39,7 @@ def get_channels(self) -> Iterable[Tuple[str, Channel]]:
 
         for channel_name, channel_config in self.channels.items():
             if channel_config is None:
-                channel_config = Channel()
+                channel_config = Channel() # noqa: PLW2901
```
marcus-oscarsson: Variable is changed in iteration

elmjag:
> @elmjag, Could you have a look at:
> 
> https://github.com/mxcube/mxcubecore/blob/6aaa91ae1065a5e7b0d9222ab217297c59022129/test/pytest/test_setup_commands_channels.py.
> 
> It looks like a test that is only in this branch, if you are still using it could you update the `mock` to `unittest.mock` ?

@marcus-oscarsson These are new tests that I have added to check code for configuring commands and channels using YAML configs.

Here a [commit](https://github.com/elmjag/mxcubecore/commit/d22bebbe8be662f8329deecda6f557145c30c3a3) that switches over tests to use `unittest.mock` modules, instead of pytest-mock's `mock`.

elmjag:
> The linting step should pass now, it would be great if you @elmjag or @fabcor-maxiv could have a look at the `mock` import in the tests.

How do you mean we should take a look at `mock` import in the tests?

Do we want to remove `pytest-mock` package as a test dependency? I'm not against it, but let's do it outside of this PR and re-base.

marcus-oscarsson:
I thought that `mock` was replaced by `unittest.mock` and that it was done in a previous PR, but maybe I'm mistaken. `pytest-mock` wraps `unittest-mock` right ?, thanks for the commit Ill cherry-pick it then.

marcus-oscarsson:
> @marcus-oscarsson These are new tests that I have added to check code for configuring commands and channels using YAML configs.

Ahh, yes I realized that later, nicely done !

marcus-oscarsson:
What I meant was that, there is an import error when we run the tests, `ModuleNotFoundError: No module named 'mock'` and for some reason I thought that we replaced `mock` with `unittest.mock`.

marcus-oscarsson:
Ok, thats done so the tests should now run

mxcubecore/model/protocols/epics.py
elmjag: I suggest fixing this warning like this:

https://github.com/elmjag/mxcubecore/commit/a99fbf0336587f9fc015428716a1f0800aba8943

I can make a full commit, that fixes all instances of this warning, if you want to cherry-pick it.

fabcor-maxiv:
Seems like tests are stuck in a loop or something like that...

marcus-oscarsson:
> Seems like tests are stuck in a loop or something like that...

I think it just took very long, they failed in the end. I think @rhfogh sorted out some of the issues in his PR and @beteva and @walesch-yan did some fixes when we tested things here at ESRF that also needs to be ported. We should be more or less set after that.

marcus-oscarsson:
Not entirely sure whats going on with the `MachineInfo` test, maybe you have an idea @fabcor-maxiv or @elmjag

elmjag:
> Not entirely sure whats going on with the `MachineInfo` test, maybe you have an idea @fabcor-maxiv or @elmjag

Ohh ahh, something very fishy is going on. I'm looking into it.

marcus-oscarsson:
You probably already spotted that, but it seems that it goes through in the end and that there then are four other tests that fail,

```
test/pytest/test_hwo_maxiv_ispyblims.py EE                               [ 94%]
test/pytest/test_hwo_maxiv_mach_info.py ..                               [ 95%]
test/pytest/test_hwo_tango_shutter.py EE    
```

The one for `ISPyB` looks quite easy to fix, but the other one is more complicated perhaps

elmjag:
> You probably already spotted that, but it seems that it goes through in the end and that there then are four other tests that fail,
> 
> ```
> test/pytest/test_hwo_maxiv_ispyblims.py EE                               [ 94%]
> test/pytest/test_hwo_maxiv_mach_info.py ..                               [ 95%]
> test/pytest/test_hwo_tango_shutter.py EE    
> ```
> 
> The one for `ISPyB` looks quite easy to fix, but the other one is more complicated perhaps

Yeah, here is a fix for `ISPyB` tests: https://github.com/elmjag/mxcubecore/commit/911230b900d56aa69b61dd7869dba9c5dfb63035

Moving on to tango_shutter and mach_info tests.

marcus-oscarsson:
Ok, could you possible make a PR to this branch when you are done, it will be much simpler like that ?

mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -164,8 +164,8 @@ def cmd_name(self):
 
 
 class BeamlineActions(HardwareObject):
-    def __init__(self, *args):
-        HardwareObject.__init__(self, *args)
+    def __init__(self, name):
+        HardwareObject.__init__(self, name)
```
beteva, line 168: Maybe super().__init__(name)

marcus-oscarsson: I feel that thats maybe part of another PR somehow, we can possibly do that change when transfer the commits you and @walesch-yan did on 30B.

marcus-oscarsson:
Look at that linting and tests pass !, thanks @elmjag for the last PR. 

We have a few more commits to add from the tests we made on the beamline (mostly related to ESRF and BLISS) and we also said that we would look at adding back `get_object_by_role`. 

We will try look at that during the week, and then we can hopefully merge this. In the meantime don't hesitate to test on your respective sites.

elmjag:
> > Not entirely sure whats going on with the `MachineInfo` test, maybe you have an idea @fabcor-maxiv or @elmjag
> 
> Ohh ahh, something very fishy is going on. I'm looking into it.

I did not manage to figure out exactly what was going in with `MachineInfo` tests.

I'm guessing that it was some kind of interaction with `TangoShutter` tests. Maybe when `TangoShutter` tests failed, they somehow screwed up tango internal state, due to bad fixture clean-up. I guess this is a topic for future investigations. Now all the tests are passing.

elmjag:
> We have a few more commits to add from the tests we made on the beamline (mostly related to ESRF and BLISS) and we also said that we would look at adding back `get_object_by_role`.
> 

FYI, the `get_object_by_role` have _not_ been removed. It's still present on the YAML branch:

https://github.com/mxcube/mxcubecore/blob/ede45a987f4f5e5f0106416ceb2982f29ef9f665/mxcubecore/BaseHardwareObjects.py#L515

marcus-oscarsson:
@elmjag, thanks excellent, one less thing to do then ;)

marcus-oscarsson:
I just wanted to give you a short update on this. We have now had the time to test this work a bit again after the rebase. 

The application now runs on the beamline but still needs to be tested further. Most thing seems to behave as they should and the simple oscillation data collection seems to work. We have tested both with generated `yaml` files and with our old `xml` files.

The `xml` files needs some updates to work and the `beamline_config.yaml` of course needs to be on the new format with `class` instead of `initialize_class`.

We will provide some more updates next week, and some additional commits and suggestions.

marcus-oscarsson:
We have now tested a bit more extensively and it all looks pretty good. Our aim would be to make the necessary updates to our current XML configuration files so that we can run the beamlines with XML files and this code during a transition period and then move hardware object by hardware object to YAML. 

The updates to the XML files are quite small, and I believe that all the changes we needed to do are already documented. (For information, the generated YAML files seems to work well in most cases.)

When we tested we realized that having the roles attached directly to the hardware object quickly becomes quite messy. With the experience we now have working with this system we would rather prefer that roles where added to a object called `roles` like is done with the configuration (`config`).

For us it's more consistent when `roles` and `config` are both accessed and contained in their respective objects. It will further be easier to create a model for the roles object and validate that, like for `config`. It also reduces the likelihood of name clashes.

We propose to make the necessary changes to make this possible. We know that we have already discussed this and decided for the current solution. However, having worked with this for some time now, we realized that its at least in
our opinion better to introduce a `roles` object. What do you think ?

rhfogh:
@marcus-oscarsson I can live with the 'roles' idea, but I am not that enthusiastic. It depends a bit on what situation you are targeting. For 'config' there is enough different and potentially changeable options that no one can be expected to keep track of them, so it is good to have them in one place. For the roles there are fewer and they have a fairly clear relation to how beamlines are built. I'd say that going from 'detector.distance' to 'detector.roles.distance' adds nothing and if anything makes the relationship less clear.  Detector.distance you can interpret; adding the extra 'roles 'would suggest you have to go to the documentation to make sure exactly what it might mean. As for the beamline itself (which is where most of the roles are) surely it is not hard to keep track of what attributes it might have? One risk of having the 'roles' is that it might encourage people to use it as a grab bag for generic links, as opposed to limiting it to things that belong in the context.

We shall see what other people say, but to help with the discussion do you have a good use case where you think adding the 'roles' is clearly helpful?

marcus-oscarsson:
You can still write your code so that you can access `distance` from the `detector` object via for instance a property that returns `self.roles.distance`.

The advantages for having a enclosing object, `roles`, for all the role objects are:

  - The hardware object can quickly become messy if there are many roles, and some of those roles are not necessarily part of the "public" API of the hardware object in question. 

  - More explicit and clear from where the object comes, which limits confusion. When the inheritance chain becomes deeper its really nice to know where an object comes from.

  - Easier to apply a model to the entire `roles` object rather than parts of a HardwareObject.

rhfogh:
@marcus-oscarsson 
I suspect that I shall lose this argument, which is OK. But could you give a concrete example of an object that is complex enough to require this treatment?

Near as I can see, there are two approaches:
1) We know what a beamline is, and decide which hardware objects can live where, and this is reflected in the code.
2) We have an abstract net of objects that can be freely configured to have any desired set of roles. That means that you do not know where to look for anything without first checking the models, and that we shall need to watch out for different people hooking up objects in different places at different times if we (i.e. someone other than the person writing it) want to find them and use them.

If we opt for 2) how would we organise so that it remained clear what was the correct location and the correct way of addressing to find a given object?

BTW, adding syntactic sugar to make 'detector.roles.distance' look like 'detector.distance' would not help. The point is not to avoid typing six additional characters , but to have a clear idea of how the objects belong together, reflected in the code. Hiding the underlying structure under a property does not really solve the problem, unless we make that the primary way of addressing objects, and make ._roles' a private attribute.

fabcor-maxiv:
Is `roles` the best name? Maybe I am misunderstanding the context. If I read `foo.roles.bar`, I assume that `foo` has many "roles", and one of the "roles" that `foo` has is: `bar`. Is it really what is happening here?

marcus-oscarsson:
@fabcor-maxiv: Yes you are right `roles` is perhaps not the best name, perhaps `_objects`, `object_by_role`. Not sure what to call this object but perhaps you have some ideas ?

@rhfogh `BeamInfo` is perhaps a good example of an object where the current implementation quickly becomes messy, also because some of the objects are optional. 

Adding the "role" objects in a containing object, perhaps better called `_objects`, would not prevent us from having a well defined object hierarchy it gives us the option to define one but it would not be defined implicitly from configuration. Then it also has the advantages that we list above.

fabcor-maxiv:
> `roles` is perhaps not the best name, perhaps `_objects`, `object_by_role`. Not sure what to call this object but perhaps you have some ideas ?

So what are these things that would be listed under this attribute: items, components, composites, children, parents, parts, objects, dependencies, dependents?

What is it that we are 100% sure that these things always have in common? Are they always all going to be hardware objects and nothing else? If yes, then maybe `hwos`, although I am not a big fan of such condensed loose acronym-ish words. I kind of like `object_by_role`, in principle.

marcus-oscarsson:
@fabcor-maxiv its the objects that are listed in the objects section of the `YAML` configuration. The same objects that today (in XML) are configured as:

```
<object hwrid="/aperture" role="aperture" />
```

So `hwobj` or `hwobj_by_role` is what it is

marcus-oscarsson:
Oh, my last message was cut short, 

>So hwobj or hwobj_by_role is what it is

and those are perhaps better names in that case. I would probably also advocate for adding a underscore i.e `_hwobj` or `_hwobj_by_role`.

rhfogh:
@marcus-oscarsson 
1) Is there an up-to-date configuration file for ESRF BeamInfo somewhere? I did not find one in the repositories, and that made it harder to fully appreciate what the problem with BeamInfo is.
2) It may well be that "nothing would prevent us from having a well defined object hierarchy ", but do we have anything on what we would be doing to make that happen? Since this entire discussion seems to be about avoiding constraints on how you can handle hardware objects, I worry that the fact that we 'have the option' to add one does not necessarily mean that it is going to happen, or be respected if it does.

marcus-oscarsson:
**1:**
I think its  enough to look at:

    https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/BeamInfo.py

And you can see that there are three hardware objects that there are three objects that are retrieved from config:

```
        self.aperture_hwobj = None
        self.beam_definer = None
        self.slits_hwobj = None
```

Those are not necessarily part of the public API of the HardwareObject in question and we don't, in my/our opinion, want to make them so implicitly via configuration. In addition to that it becomes quite messy when we inherit these classes and we would like to know from where a object comes from. 

**2:**
We have to create a well defined a object hierarchy otherwise we can't use mxcube correctly. The interface of a hardware object is defined by its public attributes and that we need to adhere to. But we don't think that the interface should be defined implicitly via configuration. We prefer that the objects are added to for instance to `_hwobj_by_role` which is validated and then necessary objects are exposed via a property (and typed) if its part of the interface.

fabcor-maxiv:
> Those are not necessarily part of the public API of the HardwareObject in question and we don't, in my/our opinion, want to make them so implicitly via configuration.

Yes, for this reason alone, it is worth storing all the "dependency hardware objects" in a private container (prefixed with underscore `_`). Those should not be part of the public API.

beteva:
In addition to this, we would like to use some objects only at the initialisation - something like:
```
        _slits = self.get_property("slits")
        _bliss_obj = self.get_object_by_role("bliss")
        if _slits:
            self._slits = {}
            _definer_type.append("slits")
            for name in _slits.split():
                _key, _val = name.split(":")
                self._slits.update({_key: _bliss_obj.getattribute(_val)}) 
       
```
So in the public space there will be only self._slits and not the _bliss_obj.

rhfogh:
Well, it sounds to like you are saying that there would be basically two complete and separate levels here. One would be the publicly accessible HardwareObjects (virtual HardwareObjects??), which *would* be hardcoded into the classes, so that you would know which ones there were and where they would live. The other one would be a number of configurable underlying hardwareobjects that were not publicly visible and could be defined in _objects, and that were linked somehow to the publicly visible layer. Is that correct? That would certainly satisfy my needs, since I would only ever have to access the public HOs and could do everything necessary there (?). It does sound somewhat complex for those who would have to code it, and it might be hard to set up in some cases - for instance, what might you need to know about the configuration of the various beam definers in order to use the system, and how could you access it without having to go into the private layers? But if that is what you are saying I certainly have no objection.

marcus-oscarsson:
So something like the following is what I/we had in mind:

   ~https://github.com/mxcube/mxcubecore/pull/1177/commits/6329060fe550bdc0123534c3947207a044d5855a#diff-3822e4536634bf372f5d0cc7f0fd470b4a5fc2a2741127801c417464b9480779L116~
    https://github.com/mxcube/mxcubecore/pull/1177/commits/53bdbcbba10d7c8fbb91a01d7e559c7b67f20e89

What do you think ?

marcus-oscarsson:
@elmjag, we should perhaps ask what you think as well. Would this complicate things for you a lot ?

rhfogh:
@marcus-oscarsson 
If I read this right, you are proposing to get rid of addressing contained hwobj as attributes entirely, and replace this with a function call to get_object_by_role. I do notice that you get an error if the hwobj is not present, rather than silently returning None. I see two problems with this:

  1) It means that your linters will not know and warn you when you are trying to address a non-existant hwobj. You would have to check by hand explicitly somewhere else whether this hwobj was defined and in which context: everywhere? ESRF-only? ...

2) It means that you cannot use get_object_by_role as an alternative access where you do get None if the hwobj is not defined, as you could previously. 

As you will know from the long-running discussion on this point, this is not my favourite solution.

marcus-oscarsson:
> @marcus-oscarsson If I read this right, you are proposing to get rid of addressing contained hwobj as attributes entirely, and replace this with a function call to get_object_by_role. I do notice that you get an error if the hwobj is not present, rather than silently returning None. I see two problems with this:
> 
>     1. It means that your linters will not know and warn you when you are trying to address a non-existant hwobj. You would have to check by hand explicitly somewhere else whether this hwobj was defined and in which context: everywhere? ESRF-only? ...
> 
>     2. It means that you cannot use get_object_by_role as an alternative access where you do get None if the hwobj is not defined, as you could previously.
> 
> 
> As you will know from the long-running discussion on this point, this is not my favourite solution.

Sorry yes that was a mistake from my side I tried to make the code a bit more concise and doing so I ignored the fact that `get_object_by_role` needs to return `None` if the object does not exist. So I fixed that.

rhfogh:
OK, that changes the balance a bit. It is still the case that

 1)   It means that your linters will not know and warn you when you are trying to address a non-existant hwobj. You would have to check by hand explicitly somewhere else whether this hwobj was defined and in which context: everywhere? ESRF-only? ...

But now it is also the case that 

  2)  Even at runtime you will not get a clear error message if you try to address a non-existant hwobj (or if, for instance you spell the name wrong). You will just get a null result.

The only way to find out whether a given hwobj is available in a given context seems to be going to either the configuration files or the Pydantic definitions and checking by hand. And even then you would have to consider whether this config file or this Pydantic definition were actually relevant for a particular beamline, and what might or might not be the case at different beamlines. Which is something that I have to consider even if most members do not.

It is possible that I am exaggerating, or that I am missing something obvious. But I would like to know: Under your proposal, if I want to write some code that addresses the hwobjs and that I want to work widely across multiple (or all) beamlines: How do I find out which hwobj I can rely on being available? And how do I tell my linter to warn me if I get it wrong or misspell a name? If the answer is 'You cannot do that' that could still be a legitimate trade-off, if we collectively decide to go that way. But if so I think we should be clear that this is what we are doing.

marcus-oscarsson:
> OK, that changes the balance a bit. It is still the case that
> 
>     1. It means that your linters will not know and warn you when you are trying to address a non-existant hwobj. You would have to check by hand explicitly somewhere else whether this hwobj was defined and in which context: everywhere? ESRF-only? ...
> 
> 
> But now it is also the case that
> 
>     2. Even at runtime you will not get a clear error message if you try to address a non-existant hwobj (or if, for instance you spell the name wrong). You will just get a null result.
> 
> 
> The only way to find out whether a given hwobj is available in a given context seems to be going to either the configuration files or the Pydantic definitions and checking by hand. And even then you would have to consider whether this config file or this Pydantic definition were actually relevant for a particular beamline, and what might or might not be the case at different beamlines. Which is something that I have to consider even if most members do not.
> 
> It is possible that I am exaggerating, or that I am missing something obvious. But I would like to know: Under your proposal, if I want to write some code that addresses the hwobjs and that I want to work widely across multiple (or all) beamlines: How do I find out which hwobj I can rely on being available? And how do I tell my linter to warn me if I get it wrong or misspell a name? If the answer is 'You cannot do that' that could still be a legitimate trade-off, if we collectively decide to go that way. But if so I think we should be clear that this is what we are doing.

**1:** What we propose does not change the semantics of  `get_object_by_role` 

**2:** It does contain the hardware objects referred to by a role in a separate object, that object can be modeled with a Pydantic model and with type hints so you can get linter and editor support (not certain that it can be done very elegantly thought).

**3**: You will need to explicitly define the public API of you hardware object using typehints, your linter will know what those are. Like you would today, you however also need to assign it.

What we suggest only contains the hardware objects configured in the objects section a containing object called `_hwobj_by_role`.  It avoids adding attributes implicitly which can lead to quite alot of confusion in our opinion.

marcus-oscarsson:
A question, how is this meant to work:

https://github.com/mxcube/mxcubecore/blob/53bdbcbba10d7c8fbb91a01d7e559c7b67f20e89/mxcubecore/BaseHardwareObjects.py#L119

https://github.com/mxcube/mxcubecore/pull/1177/files#diff-6b804ead1296063fd15a5ec658398e99d4db0dcb7154e8ad9c33009d15e6ea21R109

It seems a bit odd to me to do ?:
```
    def __getattr__(self, attr):
        return getattr(self.__dict__["_config"], attr)
```

rhfogh:
@marcus-oscarsson 
About __getattr__ I think it translates as:
"If the attribute is not present in the object itself, see if it is present in _config and if so return that"
It would be a way to avoid doing HWO.config.attr and instead doing just HWO.attr. Presumably that is why GphlWorkflow.setting and GphlWorkflow.config.settings both work. 

Arguably we should either get rid of the __getattr__ or get rid of the config property - which in practice would mean getting rid of the __getattr__ I guess, it does come across as rather complicated. The __getattr__ does mean you can switch seamlessly between attributes explicitly defined and attributes defined through configuration, but it is probably better and less confusing to address the config explicitly.

marcus-oscarsson:
I see, I have to think a bit more about that. I'm off this afternoon but Ill take a closer look when I have some time. There is a mix between configuration data and HardwareObjects in the  `HOConfig` object which confuses(ed) me a bit. The HardwreObjects defined in `HOConfig` are never used then, its just a sort of declaration.

marcus-oscarsson:
The confusion stems from this section: https://github.com/mxcube/mxcubecore/pull/1177/files#diff-6b804ead1296063fd15a5ec658398e99d4db0dcb7154e8ad9c33009d15e6ea21R109

```
    class HOConfig(ConfiguredObject.HOConfig):
       ....
      # Contained hardware objects

        machine_info = None
        transmission = None
        cryo = None
        energy = None
        flux = None
        beam = None
        hutch_interlock = None
        safety_shutter = None
        fast_shutter = None
        diffractometer = None
        detector = None
        ...
```
Those values are never set then ?

rhfogh:
@marcus-oscarsson 
The problem is not having  get_object_by_role. The problem is how you define 'the public API of your hardware object'. In the original proposal it was mandatory that contained HWOBJ had to be declared explicitly as attributes, which defined the public API. get_object_by_role was then simply a nifty alternative access mode. If we remove the constraint that contained HWO have to be declared as attributes, then I would argue that get_object_by_role _becomes_ the public API - and it is completely untyped - exactly like dict.get()

My problem is not how to define the public API of my own objects. I own, them, I understand them, no one else ever accesses them. My problem is how to find out the public APIs of other people's HWOBJ. And here I would like to understand what rules I should expect. As I understand it, your  `that object can be modeled with a Pydantic model and with type hints so you can get linter and editor support (not certain that it can be done very elegantly thought)` means the same as my `there is no actual interface spec, go and check the Pydantic definitions by hand.`

marcus-oscarsson:
`get_object_by_role` would not be more part of the public interface than it already is, we could indeed add a `_` to make this more clear. There is no `_` today so I did not add one and we should perhaps take one change at the time.

Its quite simple a HardwareObject has an Abstract class that defines an interface we are expected to respect that interface and that should be done by attributes explicitly defined on the object. That also goes for the objects that fore some reason may not have an abstract class. The interface is is defined type hinted public members, a PyDantic is only an addition to this enables us to be even more specific. I don't see how this would, or why it should, be different from defining any other object using Python or other language.

rhfogh:
@marcus-oscarsson OK that is beginning to make sense to me (sorry it took so long). So, then, for objects that are configured (like detector.distance, the diffractometer motors, beamline .flux, beamline.definer, ...) there would still be a _requirement_ to code them specifically as part of the code, only it would be as properties that referred to _objects? And you could have additional configure HWOs that were _not_ part of the public interface that would go into the _objects but would not have a property to make them public? Did I get it right this time? If so I can have no objections.

marcus-oscarsson:
Yes, thats exactly what I meant. My apologies for not being very clear, I should probably have put it like that to start with :)

rhfogh:
@marcus-oscarsson  And my apologies - again - for reacting so defensively. At least it looks like we are getting this hammered out.

On a could of additional points:

  - Thanks for noticing the definition of \_\_getattr\_\_. That clearly needs to go.

  - Also, the point you raise about the beamline HOConfig: It looks to me like that is out of one of my initial drafts, and it is indeed not consistent with the way we are otherwise handling contained HWOs. I suspect that I put it as a way of specifying 'this is more or less what the configuration file should contain', while not really understanding how the Pydantic would work or how, exactly we were going to do things. Anyway, I am pretty sure that code should go. 

Incidentally (now we clearly agree about the main thrust I am sure I shall be fine with whatever you come up with): Will we have separate Pydantic classes for the configuration and the contained HWOs? That would fit most neatly with having a separate get_object_by_role function? Or would we have a single Pydantic object with some way of distinguishing objects from non-objects? Or something else entirely?

marcus-oscarsson:
>     * Thanks for noticing the definition of __getattr__. That clearly needs to go.
I think that would be really preferable yes, its a very convenient construct but it can be very confusing. And in our case, as far as I can see, it actually caused us some issues that where then hidden.

> 
>     * Also, the point you raise about the beamline HOConfig: It looks to me like that is out of one of my initial drafts, and it is indeed not consistent with the way we are otherwise handling contained HWOs. I suspect that I put it as a way of specifying 'this is more or less what the configuration file should contain', while not really understanding how the Pydantic would work or how, exactly we were going to do things. Anyway, I am pretty sure that code should go.

Ok, that's what I thought but I was not entirely sure

> 
> 
> Incidentally (now we clearly agree about the main thrust I am sure I shall be fine with whatever you come up with): Will we have separate Pydantic classes for the configuration and the contained HWOs? That would fit most neatly with having a separate get_object_by_role function? Or would we have a single Pydantic object with some way of distinguishing objects from non-objects? Or something else entirely?

I think the most easy or perhaps evident solution for the `PyDantic` models is to indeed have one for configuration and one for the objects. It would be easier to validate and handle, but there is perhaps some other solution that we did not consider.

mxcubecore/HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -282,8 +282,8 @@ def get_outer_radius(self, distance=None):
 
         beam_x, beam_y = self.get_beam_position(distance)
         pixel_x, pixel_y = self.get_pixel_size()
-        max_delta_x = max(beam_x, self.width - beam_x) * pixel_x
-        max_delta_y = max(beam_y, self.height - beam_y) * pixel_y
+        max_delta_x = max(beam_x, self._width - beam_x) * pixel_x
```
marcus-oscarsson, line 285: So here is one consequence of the `__getattr__` definition. As `AbstractDetector` is a `ConfiguredObject` the configuration values was added as public members to the object. Adding `width` and `height` to the already existing `get_width` and `get_height` in addition to that there is also `_width` and `_height`. Quite confusing and unexpected

test/pytest/test_detector.py
```diff
@@ -69,8 +69,8 @@ def test_get_radius(self, test_object):
             pixel_x, pixel_y = test_object.get_pixel_size()
             bx, by = test_object.get_beam_position(_d)
 
-            rrx = min(test_object.width - bx, bx) * pixel_x
-            rry = min(test_object.height - by, by) * pixel_y
+            rrx = min(test_object.get_width() - bx, bx) * pixel_x
```
marcus-oscarsson, line 72: Same as mentioned above

marcus-oscarsson:
I managed to get all tests to run again, there where a few failing because I missed the following line in my first attempt:

https://github.com/mxcube/mxcubecore/pull/1177/commits/de94005560c864602d7615a31174289321fdc0d1#diff-e3ba2a0b991b4725ee3e2e98deb77bf4777747fb7cdc071954d85a44297968f5R298

However there are still two "MAXIV LIMS" tests failing, I'm not sure how this works:

https://github.com/mxcube/mxcubecore/blob/dd7f4fb8ce79008801c3d52f23cd6b7fd77e3be2/test/pytest/test_hwo_maxiv_ispyblims.py#L36

mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -106,45 +106,161 @@ class HOConfig(ConfiguredObject.HOConfig):
...
-        gphl_connection = None
-        xray_centring = None
-        online_processing = None
-        offline_processing = None
-        characterisation = None
-        image_tracking = None
-        mock_procedures = None
-        data_publisher = None
+    @property
+    def machine_info(self) -> HardwareObject | None:
```
marcus-oscarsson, line 110: I defined the "public interface" by adding a prototype and added the `HardwareObject` type hint. This can of course be more precise

rhfogh:
The recent commits all look good. It is a lot of work to be taking on for you, but at least it should be clean once it is done.
Would we be going for _get_object_by_role, in the end, or is there a case for having it as a non-private function?

marcus-oscarsson:
Thanks, I'm happy to do a few final touches after all the nice work you and @elmjag did. I would prefer to wait to rename  `get_object_by_role` but we should definitely look at it.

I just wanted to see with @elmjag and @fabcor-maxiv if this change does not create to much complications on their side. I also had a question about this section of the tests:

```
    lims._config = lims.HOConfig(  # noqa: SLF001
        ws_root="http://example.com/tstWS/",
        rest_root=REST_ROOT,
    )
```

I did not find the definition of the `HOConfig` for any '`lims` object so I'm wondering what I'm missing ? (@elmjag, @fabcor-maxiv).

We will run another run of tests on the beamlines here and then we are ready to merge (if there are no objections)

elmjag:
> I also had a question about this section of the tests:
> 
> ```
>     lims._config = lims.HOConfig(  # noqa: SLF001
>         ws_root="http://example.com/tstWS/",
>         rest_root=REST_ROOT,
>     )
> ```
> 
> I did not find the definition of the `HOConfig` for any '`lims` object so I'm wondering what I'm missing ? (@elmjag, @fabcor-maxiv).
> 

The `lims.HOConfig` here points to the generic `ConfiguredObject.HOConfig` class, this one: https://github.com/mxcube/mxcubecore/blob/dd7f4fb8ce79008801c3d52f23cd6b7fd77e3be2/mxcubecore/BaseHardwareObjects.py#L94

This piece of code set-up configuration of the lims hardware object for the tests.

Here is a PR that fixes the failing tests: https://github.com/mxcube/mxcubecore/pull/1185

elmjag:
> I just wanted to see with @elmjag and @fabcor-maxiv if this change does not create to much complications on their side.

Sorry, I did not keep-up with the discussions. But don't worry about any complications for us. Good architecture trumps any problems it may create for us. We'll rewrite our stuff as needed.

marcus-oscarsson:
@elmjag, no problem I figured you had vacation or busy on the beamline or something :) Anyways, thanks for the fix for the tests.

We will try to get some tests done during this or next week just to confirm the last changes works well. That will probably generate a few commits and then we are ready to merge this.

marcus-oscarsson:
The tests are taking forever, it seems like there is some issue with `Tango` and `notifd` but not certain.

marcus-oscarsson:
maybe you have an idea of what this could be @fabcor-maxiv or @elmjag

elmjag:
> The tests are taking forever, it seems like there is some issue with `Tango` and `notifd` but not certain.

The message: `Can't create notifd event supplier. Notifd event not available` is just some kind of warning. Not sure exactly what it means, but it have always been printed. Nothing seems to be broken.

The TangoShutter test needs to be updated, here is the PR https://github.com/mxcube/mxcubecore/pull/1202.

All other tango-related test are passing, as far as I understand it.

However, a lot of `test_command_container.py` tests are failing. Seems that it fails to create channel objects, or something like that.

marcus-oscarsson:
Thanks alot @elmjag I at first thought the `test_command_container` was also caused by the Tango issue but that does not seem to be the case then. Ill have a look.

marcus-oscarsson:
Actually the `test_command_conatiner` tests are a bit hard to follow but what seems to be failing is this:

    https://github.com/mxcube/mxcubecore/blob/fe870a260b332bcd144c8ac8b4efbd32546a49d3/test/pytest/test_command_container.py#L941C17-L941C36

I'm wondering, was this designed to fail in case of the missing `exporter_address` and that does simply not work now because there is a `KeyError` being raised ?

There is a similar case for the Tango command on line `932`.

marcus-oscarsson:
Ok, the tests are passing. The use of  `CommandContainer`, which is a mixin, should probably be looked over a bit. We are using `get_property` in that class which is not a direct method of that class but of the class that is inheriting it. It breaks the encapsulation of `CommandContainer`. 

The way that I solved it in the tests is to pass i.e `exporter_address` to the  `CommandContainer` object, which I guess is also the way it should be done in the code base, rather than using `get_property`. The previous way of accessing this attribute via direct attribute access was also working because the same attribute was defined on the parent class.

marcus-oscarsson:
This is now finally ready to be merged :fireworks: :smile:

PR #1179, opened 2025-03-18
Updates some tests to work with YAML

elmjag:
Updates `MAXIV.ISPyBLims` and `TangoShutter` to work with the new YAML code.

Rreplaces calls to `set_property()` in test fixtures, with `HOConfig` object creation.


marcus-oscarsson:
Thanks @elmjag !

PR #1181, opened 2025-03-27
Added inline documentation

### LINKED ISSUE #1180: Create inline documentation about lims sessions classes and fields
Following:
https://github.com/mxcube/mxcubecore/pull/1166#discussion_r2002658583

antolinos:
This MR proposes adding inline documentation to the lims_session file. This class contains the models primarily used for LIMS. It was created by inferring the fields and classes used in ISPyB and has been extended with additional necessary classes.

This should hopefully make sense for most facilities using a proposal system for experiments, but it may require some work to better generalize the concepts and refine them.

### DIFF
--- a/mxcubecore/model/lims_session.py
+++ b/mxcubecore/model/lims_session.py
@@ -16,97 +16,115 @@ from pydantic.v1 import (
 
 
 class Proposal(BaseModel):
-    """Represents a proposal with key identifying details."""
+    """Proposal with key identifying details.
 
-    proposal_id: str = ""  # Unique identifier for the proposal
-    person_id: str = ""  # Identifier of the person associated
-    type: str = ""  # Type/category of the proposal (e.g., "MX")
-    code: str = ""  # Specific code associated with the proposal e.g., "IH-LS"
-    # Proposal number; may be sequential or unique (uniqueness may depend on
-    # the combination of `code` and `number`)
+    Attributes:
+        proposal_id: Unique identifier for the proposal.
+        person_id: Identifier of the person associated.
+        type: Type/category of the proposal (e.g., "MX").
+        code: Specific code associated with the proposal e.g., "IH-LS".
+        number: Proposal number; may be sequential or unique
+            (uniqueness may depend on the combination of code and number)
+    """
+
+    proposal_id: str = ""
+    person_id: str = ""
+    type: str = ""
+    code: str = ""
     number: str = ""
     title: str = ""  # Title of the proposal
 
 
 class Lims(BaseModel):
-    """Represents a LIMS system"""
+    """Represents a LIMS system.
 
-    name: str = ""  # Unique identifier for the LIMS (e.g: ispyb, icat, ...)
-    description: str = ""  # Free text with a short description
+    Attributes:
+        name: Unique identifier for the LIMS (e.g., "ispyb", "icat").
+        description: Free text providing a short description of the LIMS.
+    """
+
+    name: str = ""
+    description: str = ""
 
 
 class Session(BaseModel):
-    """Represents a experiment in a specific time slot in a beamline
-    for a proposal
-    This class has been kept more or less as-it-is for legacy purposes
-    because ISPyB rely on it
+    """Represents an experiment in a specific time slot on a beamline for a proposal.
+
+    This class is maintained for legacy purposes because ISPyB relies on it.
+
+    Attributes:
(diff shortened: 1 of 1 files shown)


mxcubecore/model/lims_session.py
```diff
@@ -16,28 +16,43 @@
...
-    type: str = ""
-    code: str = ""
+    """Represents a proposal with key identifying details."""
+
+    proposal_id: str = ""  # Unique identifier for the proposal
+    person_id: str = ""  # Identifier of the person associated
+    type: str = ""  # Type/category of the proposal (e.g., "MX")
+    code: str = ""  # Specific code associated with the proposal e.g., "IH-LS"
+    # Proposal number; may be sequential or unique (uniqueness may depend on
+    # the combination of `code` and `number`)
```
fabcor-maxiv: There are multiple ways to document class attributes.

This suggestion below is probably the "best" one.
Or maybe it is not the "best", but that is one that I know should work.

There are other ways we can explore if this one is not satisfying.

```suggestion
    """Proposal with key identifying details.
    
    Attributes:
        proposal_id: Unique identifier for the proposal.
        person_id: Identifier of the person associated.
        type: Type/category of the proposal (e.g., "MX").
        code: Specific code associated with the proposal e.g., "IH-LS".
        number: Proposal number; may be sequential or unique
            (uniqueness may depend on the combination of ``code`` and ``number``).
    """

    proposal_id: str = ""
    person_id: str = ""
    type: str = ""
    code: str = ""
    number: str = ""
```

fabcor-maxiv (changes requested):
This does not seem to follow our docstring formatting guidelines.
https://mxcubecore.readthedocs.io/en/stable/dev/contributing.html#style-guidlines

Make sure to check what is the result rendered by Sphinx:
https://mxcubecore--1181.org.readthedocs.build/en/1181/dev/autosummary/mxcubecore.model.lims_session.html

marcus-oscarsson:
Thanks, looks good !

fabcor-maxiv:
Well... when these Readthedocs builds get stuck, they are really stuck!
Let's wait, they will give up eventually. I will keep an eye on it.
But it seems better to me already.

antolinos:
Thanks @fabcor-maxiv and @marcus-oscarsson  for the comments and sorry, I am still getting familiar with this. Let me know if there is anything I can do.

fabcor-maxiv:
> sorry, I am still getting familiar with this

No worries, that is why we have this code review process.


> Let me know if there is anything I can do

Once things are settled (when Readthedocs builds finally decide to get unstuck and we are all happy about the result), maybe I will ask you to squash the commits into a single commit. Or maybe Marcus or I will do the squash operation, if you are not sure how to do it.

fabcor-maxiv:
OK, Readthedocs finally gave up on the old builds that were stuck, so I was able to trigger a fresh build that did finish successfully.

So the rendered result is here:
<https://mxcubecore--1181.org.readthedocs.build/en/1181/dev/autosummary/mxcubecore.model.lims_session.html>

... and it is not great :D

I will try to check why it renders like this. Maybe it is a Sphinx configuration setting, I do not know yet.

fabcor-maxiv:
OK, now I see that these classes are Pydantic models. We have to check how to handle such classes in Sphinx.

marcus-oscarsson:
The source looks good, we can in my opinion merge this now and then look at the sphinx configuration

fabcor-maxiv:
> The source looks good, we can in my opinion merge this now and then look at the sphinx configuration

It does look good to me indeed. And yes, I am tempted to agree that we can merge as it is now.

But maybe the fix is not in changing the Sphinx configuration settings. But maybe the fix is in modifying the docstrings somehow.

mxcubecore/model/lims_session.py
```diff
@@ -16,6 +16,18 @@
 
 
 class Proposal(BaseModel):
+    """Proposal with key identifying details.
+
+    Attributes:
```
fabcor-maxiv:
```suggestion
    Parameters:
```

Alright, seems like using `Parameters` as docstring section title is a simple and easy fix. It will avoid having everything in double.

I am still not 100% convinced it is a very good solution, but more than good enough for now. : )

antolinos: I think it should be fine now

marcus-oscarsson (approved):

PR #1186, opened 2025-04-07
Adds `create_point` attribute to data collections

mockoocy:
This parameter is optional.
It's purpose is to allow generic tasks to create a 2d point when it's ran with no shape selected. Similarily to the Data Collection task. Creating the point will be relegated to frontend, as screen coordinates are required to create a 2D point.

Default value of `False` does not do anything.

This PR does not change any present behaviour (other than few more bytes in `/available_tasks` request :smiley_cat: ).
It will do something when [this frontend PR](https://github.com/mxcube/mxcubeweb/pull/1619) is merged :)

I am not really fond of how it's solved there, but I have no better idea  :shrug:

### DIFF
--- a/docs/source/dev/queue.md
+++ b/docs/source/dev/queue.md
@@ -149,7 +149,7 @@ As with any queue entry in MXCuBE the "dynamic" queue entry also has to inherit 
 # Using BaseQueueEntry and DataCollection making the "Dynmaic" queue entries behave as
 # native queue entries
 
-from mxcubecore.queue_entry.base_queue_entry import BaseQueueEntry
+from mxcubecore.queue_entry.base_queue_entry import BaseQueueEntry, TaskPrerequisite
 from mxcubecore.model.queue_model_objects import (
     DataCollection,
 )
@@ -167,7 +167,13 @@ class TestCollectionQueueEntry(BaseQueueEntry):
     QMO = TestCollectionQueueModel
     DATA_MODEL = TestCollectionTaskParameters
     NAME = "Test collection"
-    REQUIRES = ["point", "line", "no_shape", "chip", "mesh"]
+    REQUIRES = [
+        TaskPrerequisite.POINT,
+        TaskPrerequisite.LINE,
+        TaskPrerequisite.CHIP,
+        TaskPrerequisite.MESH,
+        TaskPrerequisite.NO_SHAPE_2D,
+    ]
 
     def __init__(self, view, data_model: TestCollectionQueueModel):
         super().__init__(view=view, data_model=data_model)
--- a/mxcubecore/queue_entry/base_queue_entry.py
+++ b/mxcubecore/queue_entry/base_queue_entry.py
@@ -28,6 +28,7 @@ import sys
 import time
 import traceback
 from collections import namedtuple
+from enum import Enum
 
 import gevent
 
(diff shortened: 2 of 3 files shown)


mxcubecore/model/common.py
```diff
@@ -47,6 +47,8 @@ class StandardCollectionParameters(BaseModel):
     shutterless: bool
     selection: list = Field([])
     shape: str = ""
+    # When set to True, frontend will create a point when starting task without a shape selected.
+    create_point: bool = False
```
marcus-oscarsson: So what you would like to do is to collect the current position without necessarily having a point before ? 

That could probably be part of your collect logic, `if no point: add ...` would that work ?

marcus-oscarsson: For doing that you could look at `MiniDiff.save_current_position()`

mockoocy: Not really, the thing that I want to achieve is to create a point at location where user right-clicked to open context menu.

In other words, I want to have something that works like so:
- user right clicks anywhere on the screen
- selects the custom task on the context menu
- a 2d point is created there
- the point is passed to the task as a parameter.

marcus-oscarsson: Alright I see. I think that feature already exists, if I'm not mistaken. At least for 2D-points. Its very rarely used, but it used to work.

How is the `create_point` parameter meant to be used, I am not sure its needed. You can do what you describe above in anyways its just the the data collection option needs to be available in the context menu without having a point.

mockoocy: At MAXIV we have some task, that always sets `create_point = True` in the parameters.
On the frontend, the `create_point` parameter is recognized, and if set to `true`, then first 2d point is created, then it's passed to the task (you can see it at the PR linked in the description of this PR - it's very similar to how it's done in normal Data Collection OSC task).

The thing is, there are two pieces of information that must we must know:
- should we create a point for the particular generic task
- where to create it (location of the point on the sample view)

So I created this attribute, to have information if a point should be created for a particular task or not.
And then frontend would pass the `screenCoords` of the point that has been clicked on the sample view.

The other way of doing it, I can think of is to pass the screen coordinates to the task parameters, when no point is selected. Then based on them, a point could be created 'inside' the task.

marcus-oscarsson: It perhaps helps if we discuss this in quick meeting, I think I'm missing something. Perhaps what could help is to compare the functionality you would like with the current 2D point behavior and for instance the `Data Collection (Limited OSC)` collection

image

marcus-oscarsson: Does this feature already do something similar to what you need ?

mockoocy: `Data Collection (Limited OSC)` indeed does something similar. I am down for some quick meeting to explain it further :)

marcus-oscarsson: Cool :), How about tomorrow ?

mockoocy: Okay, I'm free anywhere between 9:00 and 15:00

marcus-oscarsson: Ok, Ill send a mail to the others as well should we say 9.30h preliminary then ?

mockoocy: Sure

mockoocy: As we discussed this morning, 
I removed the `create_point` attribute there.  Instead it's all done on the frontend.
Also I created an enum `TaskPrerequisites` to serve as an alternative to writing string literals in the `BaseQueueEntry.REQUIRES`;
i.e. `REQUIRES = ['point'] ` would be replaced by `REQUIRES = [TaskPrerequisites.POINT]`.
Added some docs for it too.

walesch-yan (approved):
Looks definitely better with the enums. Thanks!:)

PR #1187, opened 2025-04-09
MAXIV MicroMAX: add beamline actions

elmjag:
Adds beamline actions implementation for MAXIV MicroMAX beamline.

### DIFF
--- a/mxcubecore/HardwareObjects/MAXIV/MicroMAX/beamline_actions.py
+++ b/mxcubecore/HardwareObjects/MAXIV/MicroMAX/beamline_actions.py
@@ -1,0 +1,67 @@
+import logging
+
+from mxcubecore import HardwareRepository as HWR  # noqa: N814
+from mxcubecore.HardwareObjects.MAXIV.MAXIVMD3 import NoPositionBookmarkedError
+
+log = logging.getLogger("user_level_log")
+
+
+class PrepareOpenHutch:
+    """
+    Prepare beamline for opening the hutch door.
+
+    - close safety shutter
+    - close detector cover
+    - move detector to a safe position
+    - put MD3 into 'Transfer' phase
+    - if jungfrau is used, take pedestal
+    """
+
+    def __call__(self):
+        try:
+            collect = HWR.beamline.collect
+            diffractometer = HWR.beamline.diffractometer
+            detector = HWR.beamline.detector
+
+            log.info("Preparing experimental hutch for door opening.")
+
+            collect.close_safety_shutter()
+            collect.close_detector_cover()
+
+            log.info("Setting diffractometer to transfer phase.")
+            diffractometer.wait_device_ready()
+            diffractometer.set_phase("Transfer")
+
+            log.info("Moving detector to safe position.")
+            collect.move_detector_to_safe_position()
+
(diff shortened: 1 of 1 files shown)


fabcor-maxiv (approved):

mxcubecore/HardwareObjects/MAXIV/MicroMAX/beamline_actions.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        calculate flux at sample position
+        """
+        flux_at_sample = HWR.beamline.collect.get_instant_flux()
+        log.info("Flux at sample position is %.2e ph/s", flux_at_sample)
+
+
+class SaveMD3Position:
+    def __call__(self):
+        HWR.beamline.diffractometer.bookmark_position()
```
marcus-oscarsson, line 59: I guess we should just check if this is consistently named across the diffractometer implementations. Which base class are you using ?

elmjag: Ah, this is a local, 'temporary' hack. I'm planning to get rid of this, in not to distant future.

We use `GenericDiffractometer` as a base class. But this bookmarking code currently is only in our fork. I can push it to https://github.com/elmjag/mxcubecore if you want to look. It's not very impressive, just saves some motor values in memory and sends them back to MD3.

marcus-oscarsson: Oh, I see. well we can perhaps keep it mind and generalize it if needed. I think there is already something similar called `save_current_position` that uses the MD command `save_centring_positions`.

elmjag: I looked at using MD3 command `saveCentringPositions`. The problem is that there is no way to recall this position, except for moving to centering phase, as far as I understand.

My understanding is that these `SaveMD3Position` and `MoveToMD3SavedPosition` are basically a work-around for an issue where MD3 unexpectedly moves around. My current plan is to solve the issue with these unexpected moves and remove this two beamline actions.

marcus-oscarsson: Ok, yes that is indeed an issue :) Sounds good. It does also seem like something that potentially could be of general interest. 

@beteva is still working on the AbstractDiffractomter perhaps this could be added to that interface ?

beteva: @elmjag,  I guess you need this because of the "feature" of the MD (both 2S and 3) which moves the sample when doing a scan to the last saved position - the one set with the saveCentringPositions command. And indeed, there is no way to recall the position. Maybe we should ask Arinax to add it, as it is really MD specific. What we do at ESRF is to send again  saveCentringPositions to the MD, just before sending the scan command, when we know that the sample is where we want it to be for the acquisition.
Otherwise, some time ago, when the refactoring of the diffractometer abstract object has started, it was decided that all methods, concerning the centring should belong to SampleView, rather than the diffractometer. This is under test at ESRF and very soon I will make the PR.

PR #1188, opened 2025-04-10
Session: don't add extra trailing / in paths

elmjag:
Handle the case where `sub_dir` path is specified with a trailing `/`.

Using `Path()` to concatenate paths always give you a path _without_ a trailing `/`, regardless of it's input parameters. Add `/` 'manually', this way we will always get a path with exact one trailing `/`.


marcus-oscarsson:
thanks @elmjag

PR #1189, opened 2025-04-11
Provide empty method `prepare_directories` in  class `Session`

dominikatrojanowska:
Related PR:
* https://github.com/mxcube/mxcubeweb/pull/1615

### DIFF
--- a/mxcubecore/HardwareObjects/Session.py
+++ b/mxcubecore/HardwareObjects/Session.py
@@ -192,7 +192,8 @@ class Session(HardwareObject):
             _session: The session object containing session-specific
                     information.
         """
-        pass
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/Session.py
```diff
@@ -179,6 +179,10 @@ def get_base_data_directory(self):
 
         return directory
 
+    def prepare_directories(self, _session):
```
marcus-oscarsson, line 182: Looks good, you could perhaps add a doc string ?

dominikatrojanowska: I added a docstring, probably it could be merged

PR #1190, opened 2025-04-14
adds AutoPowerSardanaMotor hardware object

elmjag:
Adds new `AutoPowerSardanaMotor` hardware object. This HWO extends `SardanaMotor` with the 'auto power-on' feature.

We use this at MAX IV for detector distance motors. These motors are powered off, whenever the hutch search is broken. We need a way to transparently power them back on.

### DIFF
--- a/mxcubecore/HardwareObjects/AutoPowerSardanaMotor.py
+++ b/mxcubecore/HardwareObjects/AutoPowerSardanaMotor.py
@@ -1,0 +1,45 @@
+#
+# Disable 'Invalid module name' check.
+#
+# We can possibly make module name ruff compliant once we migrated to
+# YAML config files. With YAML configs we get more flexibility with module
+# names.
+#
+# ruff: noqa: N999
+#
+
+from mxcubecore.HardwareObjects.SardanaMotor import SardanaMotor
+
+
+class AutoPowerSardanaMotor(SardanaMotor):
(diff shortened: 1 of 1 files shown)


beteva:
Hi @elmjag, this looks like an useful feature.. Out of curiosity, why not adding this to the SardanaMotor? Or maybe even add a placeholder in AbstractMotor

elmjag:
> Hi @elmjag, this looks like an useful feature.. Out of curiosity, why not adding this to the SardanaMotor? Or maybe even add a placeholder in AbstractMotor

Regarding `AbstractMotor`, I'm not sure how it will fit in there.

For `SardanaMotor`, I have considered that option. The implementation will be a bit more complex, as it need to be configurable. The 'auto power-on' feature will need to be off by default.

For example, we can add `auto_power_attribute` config property to `SardanaMotor`.

```
<object class="SardanaMotor">
  <actuator_name>tabd_zi</actuator_name>
  <auto_power_attribute>PowerOn</auto_power_attribute>
</object>
```

If `auto_power_attribute` is specified, the 'auto power-on' feature will be activated. The specified attribute will be used to power on the motor.

This version will be a bit more complicated then the simple one in `AutoPowerSardanaMotor`. We can certainly do that instead. I'm a bit undecided which option is the best one.

PR #1194, opened 2025-04-25
Updates to YAML branch after testing

marcus-oscarsson:
Dear all,

Here are finally the updates to the YAML work that we did in order to get things to run properly here at ESRF. We have all, Antonia, Yan and I, spent some time to test on one of our beamlines. The good news is that this is all working as intended. We are running this with XML files for the moment (except for `beamline-config`) and our intention is to gradually switch to `YAML` and make the complete switch after the summer shutdown.

Its, currently, Easter vacation time here so Antonia is away this week and I'm away next week. I'm making this PR on the behalf of all of us so the others have to chip in if I forgot something or made some mistake :)

These changes are mostly to update the access to XML configuration data so that its done in a way that is "YAML compatible". As noted in the [documentation]( https://mxcubecore--1177.org.readthedocs.build/en/1177/dev/yaml_conf_migration.html), there are a few changes the needs to be done to the XML configuration file structure.

The corresponding adaptations also needs to be done in the code that uses the effected configuration data.

There are also a few changes related to the removal of the `getattr` from the config object. That allowed for access to configuration data via direct attribute access to the `HardwareObject` in question.


rhfogh (approved):
Looks fine to a quick squint - but anyway this has been well tested beforehand

PR #1204, opened 2025-05-12
Fixing attribute lookup/retreival and setting for HardwareObjects

marcus-oscarsson:
While testing the mockups with the merged YAML code I noticed an issue with the HardwareObjects that have attribute names that are the same as a configuration property name. `MDCameraMokcup` is a good example that as a internal attribute `image_name`:

    https://github.com/mxcube/mxcubecore/blob/613505a24909edbffde6877881f65ead16327a40/mxcubecore/HardwareObjects/mockup/MDCameraMockup.py#L36

And the configuration file property is also called, reasonably, `image_name`. 

The problem with this is that `__setattr__` and `__getattr_` have been defined "asymmetrically" so that the configuration file property is set via `__setattr__ ->self._set_property(attr, value)` and retrieved by `__getattr_->self.__dict__[attr]`.

    https://github.com/mxcube/mxcubecore/blob/613505a24909edbffde6877881f65ead16327a40/mxcubecore/BaseHardwareObjects.py#L335

    https://github.com/mxcube/mxcubecore/blob/613505a24909edbffde6877881f65ead16327a40/mxcubecore/BaseHardwareObjects.py#L1076

This PR removes the custom definitions of `__setattr__` and `__getattr__` from `HardwareObject` and `HardwareObjectNode`. The behavior above is is a consequence of the removal of `PropertySet` (https://github.com/mxcube/mxcubecore/pull/1177/files#diff-3822e4536634bf372f5d0cc7f0fd470b4a5fc2a2741127801c417464b9480779L302).

The custom definition of `__setattr__` and `__getattr__` was also hiding another important other issue, which was the the python property concept did not work as expected. The situation before the fix proposed in this PR allowed for setting a  property value without setter. However the value retrieved would not be the expected value.

We where very lucky to not bump into this issue on the beamline while we where testing and we did not re-test with the mockups which is our mistake this time. Our apologies for that.

We should really be very careful when re-defining `__setattr__` and `__getattr__`. Actually I would recommend never doing so unless its a very well documented and exceptional case. I looked through the code base and there are still a few custom `__getattr__` for a few specific hardware objects that we should look at, but they are less problematic.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -61,7 +61,6 @@ if TYPE_CHECKING:
 
     from pydantic.v1 import BaseModel
 
-    from .CommandContainer import CommandObject
 
 __copyright__ = """ Copyright © 2010-2020 by the MXCuBE collaboration """
 __license__ = "LGPLv3+"
@@ -149,9 +148,16 @@ class ConfiguredObject:
     def get_by_id(self, _id: str) -> "ConfiguredObject":
         result = self
         for name in _id.split("."):
-            result = getattr(result, name)
+            # Roles are no longer added to HardwareObjects as attributes so
+            # we have to retreive them by role
+            result = result.get_object_by_role(name)
+
+            if not result:
+                result = getattr(result, name)
+
             if result is None:
                 break
+
         return result
 
     @property
@@ -331,19 +337,6 @@ class HardwareObjectNode:
     def __len__(self) -> int:
         warnings.warn("%s.__len__ is Deprecated. Avoid" % self.__class__.__name__)
         return sum(map(len, self.__objects))
-
-    def __setattr__(self, attr: str, value: Any) -> None:
-        try:
-            config = self.__dict__.get("_config")
-            if attr not in self.__dict__ and config and attr in config.model_dump():
-                warnings.warn(
-                    "%s.__setattr__ is Deprecated. Avoid" % self.__class__.__name__
-                )
-                self._set_property(attr, value)
-            else:
-                self.__dict__[attr] = value
-        except AttributeError:
-            self.__dict__[attr] = value
 
     def __getitem__(
         self,
@@ -1072,18 +1065,6 @@ class HardwareObject(ConfiguredObject, HardwareObjectNode, HardwareObjectMixin):
 
         obj = get_hardware_repository().get_hardware_object(name)
         self.__dict__.update(obj.__dict__)
-
-    def __getattr__(self, attr: str) -> Union["CommandObject", Any]:
-        if attr.startswith("__"):
-            raise AttributeError(attr)
-
-        try:
-            return CommandContainer.__getattr__(self, attr)
-        except AttributeError:
-            try:
-                return super().__getattr__(attr)
-            except AttributeError:
-                raise AttributeError(attr)
 
     def commit_changes(self) -> None:
         """Commit last changes back to configuration."""
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -52,7 +52,7 @@ from typing_extensions import (
     Literal,
     Self,
 )
-
+from types import MappingProxyType
 from mxcubecore.CommandContainer import CommandContainer
 from mxcubecore.dispatcher import dispatcher
 
@@ -112,7 +112,6 @@ class ConfiguredObject:
         self._name = name
         self._config: Optional["ConfiguredObject.HOConfig"] = None
         self._hwobj_container: Optional[ConfiguredObject] = hwobj_container
-        self._roles = []
         self._hwobj_by_role = {}
 
     @property
@@ -169,15 +168,7 @@ class ConfiguredObject:
         Returns:
             OrderedDict[str, Union[Self, None]]: Contained objects mapped by role.
         """
-        result = {}
-        for tag in self._roles:
-            if hasattr(self._hwobj_by_role, tag):
-                result[tag] = getattr(self._hwobj_by_role, tag)
-            else:
-                raise ValueError(
-                    "%s object has no attribute %s" % (self.__class__.__name__, tag)
-                )
-        return result
+        return MappingProxyType(self._hwobj_by_role)
 
     def get_properties(self) -> Dict[str, Any]:
         """Get configured properties (not roles)"""
@@ -212,7 +203,7 @@ class ConfiguredObject:
         Returns:
             List[str]: List of hardware object roles.
         """
-        return list(self._roles)
+        return list(self._hwobj_by_role.keys())
 
     def print_log(
         self,
@@ -678,7 +669,6 @@ class HardwareObjectMixin(CommandContainer):
         # Add methods that are exported programatically
         for attr_name in dir(self):
             _attr = getattr(self, attr_name)
-
             if getattr(_attr, "__exported__", False):
                 self._exports[attr_name] = []
 
--- a/mxcubecore/HardwareObjects/abstract/sample_changer/Component.py
+++ b/mxcubecore/HardwareObjects/abstract/sample_changer/Component.py
@@ -7,7 +7,7 @@ class Component(object):
         self.container = container
         self.address = address
         self.scannable = scannable
-        self.id = None
+        self._id = None
         self.present = False
         self.selected = False
         self.scanned = False
@@ -26,7 +26,7 @@ class Component(object):
         Can be None if sample is unknown or not present
         :rtype: str
         """
-        return self.id
+        return self._id
 
     def get_address(self):
         """
@@ -123,8 +123,8 @@ class Component(object):
         Clears all sample info (also in components if object is a container)
         """
         changed = False
-        if self.id is not None:
-            self.id = None
+        if self._id is not None:
+            self._id = None
             changed = True
         if self.present:
             self.present = False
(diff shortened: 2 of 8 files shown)


mxcubecore/BaseHardwareObjects.py
```diff
@@ -332,19 +339,6 @@ def __len__(self) -> int:
         warnings.warn("%s.__len__ is Deprecated. Avoid" % self.__class__.__name__)
         return sum(map(len, self.__objects))
 
-    def __setattr__(self, attr: str, value: Any) -> None:
```
marcus-oscarsson, line 335: This is the naughty code in question

mxcubecore/BaseHardwareObjects.py
```diff
@@ -149,9 +149,16 @@ def hwobj_container(self):
     def get_by_id(self, _id: str) -> "ConfiguredObject":
         result = self
         for name in _id.split("."):
-            result = getattr(result, name)
+            # Roles are no longer added to HardwareObjects as attributes so
```
marcus-oscarsson, line 150: I'm not a fan of the method name `get_by_id` but it is what is at the moment. I would suggest renaming it to `get_full_name` or `full_object_path` or something, but in a separate PR.

As the comment points out we need to also check the "roles" of the object as these are no longer added as attributes directly on the `HardwareObject` in question.

mxcubecore/HardwareObjects/abstract/sample_changer/Component.py
```diff
@@ -7,7 +7,7 @@ def __init__(self, container, address, scannable):
         self.container = container
         self.address = address
         self.scannable = scannable
-        self.id = None
```
marcus-oscarsson, line 10: The fact that this was a property was previously effectively hidden by `__setattr__` , so we need to change the name.

mxcubecore/HardwareObjects/mockup/ISPyBClientMockup.py
```diff
@@ -67,9 +69,7 @@ def init(self):
             "Laboratory": {"laboratoryId": 1, "name": "TEST eh1"},
         }
 
-    @property
-    def loginType(self):
```
marcus-oscarsson, line 71: Same thing as above here, this was defined as a read only property and could not be set.

marcus-oscarsson:
For some reason, the `test_hwo_isara_maint.py` gives the error below and core dumps with exit code -139:

```
2025-05-12 11:38:44,350 |WARNING| error polling Device(test/nodb/_isara) Powered attribute, retrying.
  File "/home/oscarsso/projects/mxcube/mxcubecore/mxcubecore/Command/Tango.py", line 274, in poll
    return read_attr()
           ^^^^^^^^^^^
  File "/home/oscarsso/projects/mxcube/mxcubecore/mxcubecore/Command/Tango.py", line 268, in read_attr
    value = self.raw_device.read_attribute(self.attribute_name).value
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oscarsso/micromamba/envs/mxcubeweb/lib/python3.11/site-packages/tango/green.py", line 208, in greener
    return executor.run(fn, args, kwargs, wait=wait, timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/oscarsso/micromamba/envs/mxcubeweb/lib/python3.11/site-packages/tango/green.py", line 116, in run
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/oscarsso/micromamba/envs/mxcubeweb/lib/python3.11/site-packages/tango/device_proxy.py", line 550, in __DeviceProxy__read_attribute
    return __check_read_attribute(self._read_attribute(value, extract_as))
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
PyTango.DevFailed: DevFailed[
DevError[
    desc = Failed to connect to device test/nodb/_isara
           The connection request was delayed.
           The last connection request was done less than 1000 ms ago
  origin = virtual void Tango::Connection::reconnect(bool) at (/src/cppTango/cppapi/client/devapi_base.cpp:712)
  reason = API_CantConnectToDevice
severity = ERR]

DevError[
    desc = Failed to read_attribute on device test/nodb/_isara, attribute Powered
  origin = virtual Tango::DeviceAttribute Tango::DeviceProxy::read_attribute(const string&) at (/src/cppTango/cppapi/client/devapi_base.cpp:5589)
  reason = API_AttributeFailed
severity = ERR]
]
Segmentation fault (core dumped)
```

@elmjag, @fabcor-maxiv do you see any reason for why this would happen ?

mxcubecore/BaseHardwareObjects.py
rhfogh: @marcus-oscarsson I agree that get_by_id is not very good, but `get_full_name` is misleading. Something with that name would be expected to return a string - the full name , in fact. This has to be called `get_by_<something>`

elmjag:
> For some reason, the `test_hwo_isara_maint.py` gives the error below and core dumps with exit code -139:
> 
> ]
> Segmentation fault (core dumped)
> ```
> 
> @elmjag, @fabcor-maxiv do you see any reason for why this would happen ?

I suspect that it's a problem with tearing down the test-fixture for some of Tango based tests. I'll look into this.

marcus-oscarsson:
I see, it does not seem to happen for the other tango tests. The problem is that it returns exit code -139 which then fails the tests. Even if all the tests passes.

Thant being said, I found quite alot of issues in the tests when running with the `-s` flag that would be worth looking into.

mxcubecore/BaseHardwareObjects.py
```diff
@@ -163,15 +168,7 @@ def objects_by_role(self) -> Dict[str, Union[Self, None]]:
         Returns:
             OrderedDict[str, Union[Self, None]]: Contained objects mapped by role.
         """
-        result = {}
-        for tag in self._roles:
```
marcus-oscarsson, line 167: I had another lesson in the depths of python ;). This is also a quite nasty one. Any `Exception` raised in a property is translated/changed into an `AttributeError` and the original exception gets hidden. This gets especially confusing when one have been debugging `getattr` and `setattr` for some time.

I guess one should never explicitly raise anything from a property ?

mxcubecore/BaseHardwareObjects.py
marcus-oscarsson: Sorry yes, I meant `get_object_by_full_name` or `get_object_from_path` ...

mxcubecore/HardwareRepository.py
```diff
@@ -169,6 +169,10 @@ def load_from_yaml(
 
     if not msg0:
         try:
+            config = configuration.pop("configuration", {})
```
marcus-oscarsson, line 172: Also making sure that configuration data is available in the `_init` method

marcus-oscarsson:
We've again tested this with the YAML PR (https://github.com/mxcube/mxcubeweb/pull/1595) on `mxcubeweb`. Together with this PR the mockups and the tests runs/passes again.

We have to keep an eye an attribute access because this is the bit that changed since we tested on the beamline. It would in general be a good idea to look over the use of the internal python `__` methods, perhaps an exercise for the code camp cleanup session ?

mxcubecore/BaseHardwareObjects.py
```diff
@@ -169,15 +168,7 @@ def objects_by_role(self) -> Dict[str, Union[Self, None]]:
...
-            if hasattr(self._hwobj_by_role, tag):
-                result[tag] = getattr(self._hwobj_by_role, tag)
-            else:
-                raise ValueError(
-                    "%s object has no attribute %s" % (self.__class__.__name__, tag)
-                )
-        return result
+        return self._hwobj_by_role
```
rhfogh: @marcus-oscarsson Would it not be better to return a copy? You can get some tricky bugs if people get hold of the internal dictionary ans start modifying it.

elmjag:
> Segmentation fault (core dumped)
> I see, it does not seem to happen for the other tango tests. The problem is that it returns exit code -139 which then fails the tests. Even if all the tests passes.
> 

The exit code 139 means that the process was terminated by SIGSEGV signal.

I think the issue is that the `IsaraMain` and `MachineInfo` tests leave behind active tango pollers. It's all fine, until pytest process is about to wrap-up the tests run. It seems to trigger a bug(?) somewhere in gevent/greenlet/pytango code. At least the core dump file show stack traces that end up there.

I have tried to use latest versions of gevent and pytango, but it turns out does not solve the issue.

Anyhow, it's bad to leave behind active pollers, after tests are finished. They should be stopped, when no longer needed. I'll see if I can figure out how to properly clean-up my tango pollers.

marcus-oscarsson:
Thanks alot for looking into this, it does not seem to be something that is easy to identify and debug. The tests are passing now so, so its back to how it was before. I guess we agree on that it would be nice if the tests don't segfault ;)

mxcubecore/BaseHardwareObjects.py
marcus-oscarsson: Hm, yes. Perhaps even a immutable copy even ;)

elmjag:
> Thanks alot for looking into this, it does not seem to be something that is easy to identify and debug. The tests are passing now so, so its back to how it was before. I guess we agree on that it would be nice if the tests don't segfault ;)

I have added code that stops the tango poller in ISARAMain tests: https://github.com/mxcube/mxcubecore/pull/1209
Seems like it solves the issue, but of cause it's hard to know for sure with suspected memory corruption bugs.

Regardless, doing proper clean-up after the tests is still a good practice.

PR #1206, opened 2025-05-12
use gevent version 24.10.3

elmjag:
The test are triggering a ref-count bug in the old version of gevent. This was causing tests to fail with:

`
 Segmentation fault (core dumped)
`

Use later version of gevent, it seems to fix the issue.

### DIFF
--- a/mxcubecore/Poller.py
+++ b/mxcubecore/Poller.py
@@ -152,7 +152,7 @@ class _Poller:
                     gevent.spawn(cb, res)
 
     def run(self):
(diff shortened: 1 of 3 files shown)


elmjag:
Gah... it still core dumps with python 3.10. Back to drawing board!

fabcor-maxiv:
> Back to drawing board!

I converted to "draft"

elmjag:
Upgrading gevent does not help at all. Solved differently here: https://github.com/mxcube/mxcubecore/pull/1209

PR #1207, opened 2025-05-12
Update Poetry to version 2.1.3

fabcor-maxiv:


fabcor-maxiv:
Looks like Dependabot uses the latest version of Poetry to update our lockfile:

* <https://github.com/mxcube/mxcubecore/blob/613505a24909edbffde6877881f65ead16327a40/poetry.lock#L1>

So we need to catch up, otherwise we will constantly revert Dependabot's changes and Dependabot will revert our changes. And in this case the changes that are meant are not meaningful changes, but changes like in the latest update that Dependabot made:

* <https://github.com/mxcube/mxcubecore/commit/73819cf7f23c2e0c8fcef18d04ce5f14c1358db3>

A lot of changes are just about being more precise in the lockfile.

---

Ideally we would have a way to tell Dependabot which version of Poetry it should use. But I do not know if it is possible. Maybe the added `requires-poetry` in `pyproject.toml` will do the trick.

elmjag (approved):

marcus-oscarsson:
ok

PR #1209, opened 2025-05-13
ISARA Main tests: stop polling, when tests are finished

elmjag:
Make sure that we stop the Tango attribute polling, when cleaning up the tests fixture.
    
Leaving behind an active poller thread to trigger some kind of memory curruption bug in gevent/greenlet/pytango code. This make    the pytest process to terminate with 'segmentation fault' in some situations.


fabcor-maxiv (approved):
Seems fine to me

marcus-oscarsson:
Well spotted, thanks !

PR #1114, opened 2025-01-30
WIP Added MXLIMS v. 0.3.0 exporter code

rhfogh:
MXLIMS json file is exported automatically to raw data directory fo stdr DataCollection and GPhL workflow

PR #1212, opened 2025-05-13
[WIP] Prototype mxlims integration v. 0.6.1

rhfogh:
This WIP branch shows how you can use the MXLIMS 0.6.1 version to generate and store MXLIMS messages.


mxcubecore/queue_entry/data_collection.py
```diff
@@ -130,11 +136,55 @@ def pre_execute(self):
...
+            )
+            tracking_data.characterisation_id = workflow_parameters.get(
+                "characterisation_id"
+            )
+            self._mxlims_job, mxlims_sample = mxutils.make_mx_experiment(
+                sample=data_model.get_sample_node(),
+                tracking_data=tracking_data,
+                measured_flux = HWR.beamline.flux.get_value()
+            )
+
```
beteva, line 162: Hi @rhfogh,
is it sure that each data collection has the tracking_data and workflow_parameters. Maybe I did not get the whole picture, but this affects all data collections (not just workflows), so should not it be  more conditional (like try/except).

rhfogh: Good you are looking, but it should be OK. tracking_data are added in TaskNode.\_\_init\_\_, and populated in DataCollectionQueueEntry.pre_execute. This should indeed export a MXLIMS file for every DataCollection (if not part of a GPhL workflow) - just that I could not easily test that at home. 

Note  that this is very much WIP. It shows what you can do, but before we could merge it in we would have to agree how we would want this to work outside the GPhL workflow. That is not something I can come up with on my own.

marcus-oscarsson:
Is this related to https://github.com/mxcube/mxcubecore/pull/1114 ?

rhfogh:
Superseded

PR #1218, opened 2025-05-16
Handle Tango exceptions during attribute polling

alessandro-olivo:
Restart attribute polling after a tango exception and trigger a value update after recovering from a tango exception.

### The Issues (code few months ago):

1) When a Tango read_attribute(...) call raises an exception, that attribute stops being polled. As a result, the `update` signal of the related `TangoChannel` is no longer emitted.

2) Now, consider a scenario where an attribute read fails in one poll cycle, but succeeds in the next one. If the value hasn't changed since the last successful read (before the Exception), the `Poller` won't emit an  `update` signal.
    This behavior can cause issues, especially for HWO states. I'll try to clarify with an example.
  Example Scenario:
   >1) The Omega motor's state is successfully read from the Tango device, it returned `ON` state (`READY` in mxcube).
   >image
   >2) A Tango exception occurs during `read_attribute`, so the `update` slot of the motor HWO receives `None`. In this case, the state is interpreted as `UNKNOWN`.
   >image
   >3) The attribute is successfully read again. At this point, I'd expect a signal to be emitted to indicate the motor is back to `READY`. But since from the `Poller` point of view the value hasn’t changed (still `ON`), no  `update` signal is sent, and the state remains `UNKNOWN`
   >image

#### Notes:

- Fix for the first issue:
  In `.../command/Tango.py`
  ```diff
      def poll_failed(self, e, poller_id):
          self.emit("update", None)
  +       poller = Poller.get_poller(poller_id)
  +       if poller is not None:
  +         poller.restart(1000)
  ```
  I want also to highlight that I found this lines in a commented block of `poll_failed(...)`

- Fix for the second issue:
  In `Poller.py`
  ```diff
                  error_cb = self.error_callback_ref()
                  if error_cb is not None:
                      self.queue.put(PollingException(e, self.get_id()))
  +               self.old_res = NotInitializedValue
                  break
  
              del polled_call
  ```

### Current Code
If I understood correctly, the current fix adds a retry loop to handle Tango exceptions and in case of others Exception polling would just stop. This loop prevents to return new value to the `Poller` until the attribute is read successfully. 

However, with this fix, the HWO is no longer notified when a Tango exception happens, in the previous version, the `Poller` emitted a `None`.  Is this change intentional? Or am I missing something?

### DIFF
--- a/mxcubecore/Command/Tango.py
+++ b/mxcubecore/Command/Tango.py
@@ -272,60 +272,20 @@ class TangoChannel(ChannelObject):
         TangoChannel._tangoEventsProcessingTimer.send()
 
     def poll(self):
-        def read_attr():
-            if self.read_as_str:
-                value = self.raw_device.read_attribute(
-                    self.attribute_name, PyTango.DeviceAttribute.ExtractAs.String
-                ).value
-            else:
-                value = self.raw_device.read_attribute(self.attribute_name).value
+        if self.read_as_str:
+            value = self.raw_device.read_attribute(
+                self.attribute_name, PyTango.DeviceAttribute.ExtractAs.String
+            ).value
+        else:
+            value = self.raw_device.read_attribute(self.attribute_name).value
 
-            return value
-
-        while True:
-            try:  # in case of tango communication errors, retry reading the attribute
-                return read_attr()
-            except PyTango.DevFailed:
-                log.warning(
-                    f"error polling {self.raw_device} {self.attribute_name} attribute, retrying.",
-                    exc_info=True,
-                )
-                gevent.sleep(0.1)
-            except Exception:
-                log.exception(
-                    "unexpected exception polling %s %s attribute",
-                    self.raw_device,
-                    self.attribute_name,
-                )
-                raise
+        return value
 
     def poll_failed(self, e, poller_id):
         self.emit("update", None)
-        """
-        emit_update = True
-        if self.value is None:
-          emit_update = False
-        else:
-          self.value = None
-
-        try:
-            self.init_device()
-        except:
-            pass
-
         poller = Poller.get_poller(poller_id)
         if poller is not None:
             poller.restart(1000)
-
-        try:
-          raise e
-        except:
-          logging.exception("%s: Exception happened while polling %s", self.name(), self.attribute_name)
-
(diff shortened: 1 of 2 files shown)


fabcor-maxiv:
That is something @elmjag might have some insight on.

elmjag:
The current version of tango poller is my attempt to handle the scenario where poller would fail to read attribute from the tango device sometimes. For us this can happen if tango device becomes slow (time-out errors) or it is temporary offline (being restarted).

I guess it's true that the Poller won't emit an update signal until it successful in reading the attribute. I have not considered that aspect. I don't quite follow how it can cause problems you are describing. I don't understand how HWO's can receive `None` value, if the poller is stuck in a retry-loop.

Having said that, I think this PR looks promising. This approach looks more elegant. But, I would like to test this change, to check if polling manages to successfully recover from attribute read errors. Not sure when I will have time to test, as I will be traveling for MXCuBE meeting next week. Hopefully the week after that.

alessandro-olivo:
> I don't quite follow how it can cause problems you are describing. I don't understand how HWO's can receive None value, if the poller is stuck in a retry-loop.

My fault, I wasn’t very clear in the PR.

Initially, I explained the issues and the fix I applied a few months ago, before your changes. That fix seems to be working well on our side, but I’d be really happy if you could give it a try whenever you have time.

Your fix works as well, I tested it with exception raising and server restarts, the only thing I’m a bit concerned about is that when `read_attribute` fails, this information is not propagated to the HWO.

See you in Hamburg! :)

elmjag:
I have tested this PR, looks good. It does successfully recover polling on errors. 

Note that this PR changes current behavior. With this PR the channel will emit an `update` signal with value `None`, when reading attribute fails. Without this PR, `update` signal is not send at all on attribute read errors.

I think the new behavior is better, as it allows channel callbacks to be aware that we have lost connection to the underlying tango device. However, this potentially breaks any existing callbacks, that don't handle `None` values.

elmjag (approved):

marcus-oscarsson:
Great guys !

marcus-oscarsson:
You can rebase and merge :)

alessandro-olivo:
Thanks for the reviews and the tests! :)

PR #1222, opened 2025-05-16
Make the user log messages match the UI terms ("resume" instead of "continue")

alessandro-olivo:
If you add a DC to a sample and then click on "Run Queue", the message shown in the user log says:
"Please select, or center on a new position and press continue."
Should we replace that "continue" with "Resume" ?

### DIFF
--- a/mxcubecore/queue_entry/sample_centring.py
+++ b/mxcubecore/queue_entry/sample_centring.py
@@ -83,9 +83,7 @@ class SampleCentringQueueEntry(BaseQueueEntry):
         if motor_positions:
             HWR.beamline.diffractometer.move_motors(motor_positions)
 
-        log.warning(
-            "Please center a new or select an existing point and press resume."
-        )
+        log.warning("Please center a new or select an existing point and press resume.")
         self.get_queue_controller().pause(True)
 
         shapes = list(HWR.beamline.sample_view.get_selected_shapes())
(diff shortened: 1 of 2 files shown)


alessandro-olivo:
I don't know why it fails, I tried to execute ruff but it seems fine 
```shell
$ ruff format mxcubecore/HardwareObjects/sample_centring.py
1 file left unchanged
$ ruff format mxcubecore/queue_entry/base_queue_entry.py
1 file left unchanged
```

mxcubecore/queue_entry/sample_centring.py
```diff
@@ -84,7 +84,7 @@ def execute(self):
             HWR.beamline.diffractometer.move_motors(motor_positions)
 
         log.warning(
-            "Please center a new or select an existing point and press continue."
+            "Please center a new or select an existing point and press resume."
         )
```
walesch-yan: I tried it locally and ruff seems to reformat this into one line as it is now smaller. This would probably be the reason why it failed on the CI Pipeline

fabcor-maxiv: @alessandro-olivo, maybe you are running a different version of Ruff than the one written in `.pre-commit-config.yaml`.

alessandro-olivo: Yep, it must be the version! I'll correct
Thanks!

marcus-oscarsson:
Thanks, could you rebase and then merge :)

alessandro-olivo:
Just rebased, but I can't merge (no auth for mxcubecore too I think )

marcus-oscarsson:
You should have the right to merge now, could you try again please :)

alessandro-olivo:
Now I can merge, just a question, should the commits be squashed before merge?

Issue #1230, opened 2025-05-21
[CC-DESY] Format and lint all Python files with Ruff


fabcor-maxiv:
Currently there are still some Python files that are excluded from the formatting by Ruff. Maybe it is time to format them.

And of course we need to continue linting the Python code with Ruff.

PR #1235, opened 2025-05-26
Draft: Apply ruff rules for tests

dominikatrojanowska:

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -6584,7 +6584,6 @@ convention = "google"
 ]
 "test/test_command_container.py" = [
     "SIM102",
-    "SIM910",
     "SLF001",
 ]
 "test/test_flux.py" = [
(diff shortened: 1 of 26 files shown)


fabcor-maxiv:
Thanks for working on this :)

I wish this pull request would focus on one single Ruff rule only (or group of Ruff rules). For example I would have recommended to do a pull request that only fixes the ["useless yield" `PT022` rule](https://docs.astral.sh/ruff/rules/pytest-useless-yield-fixture/), that would have been more manageable in terms of review.

The problem with big pull requests like this is that if there is disagreement on something, it will be difficult to undo only a part of all the changes that are done here...

dominikatrojanowska:
ok, I will provide bunch of small PRs then, but I will keep also this one for while to store changes :)

PR #1237, opened 2025-05-27
Added type annotations and exception safety on Crims.py 

jbflo:
This PR introduces a partial  refactor of Crims.py

-  Replaced deprecated xml.etree.cElementTree with defusedxml.ElementTree.
 this imply adding "**defusedxml**" to the project 
- Removed old keys that no longer part of the API

- Replaced all except Exception: blocks with specific exception (pylint W0703)

- Added type annotations throughout for better clarity

Consolidated duplicated functions (get_image_size)

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
+++ b/mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
@@ -1,21 +1,34 @@
-import urllib
-import xml.etree.cElementTree as et
+import logging
 from io import BytesIO
+from typing import List, Optional, Tuple
+from urllib.error import HTTPError, URLError
+from urllib.request import Request, urlopen
 
 import requests
-from PIL import Image
+from defusedxml import ElementTree
+from PIL import Image, UnidentifiedImageError
 
 
-def get_image(url):
-    f = urllib.request.urlopen(url)
-    img = f.read()
-    return img
+def get_image(url: str) -> Optional[bytes]:
+    try:
+        return urlopen(url).read()
+    except (URLError, HTTPError) as e:
+        logging.getLogger("user_level_log").warning(
+            "Failed to fetch image from %s: %s", url, e
+        )
+        return None
 
 
-def get_image_size(url):
-    img_data = requests.get(url).content
-    im = Image.open(BytesIO(img_data))
-    return (im.size)
+def get_image_size(url: str) -> Tuple[int, int]:
+    try:
+        img_data = requests.get(url, timeout=900).content
+        with Image.open(BytesIO(img_data)) as im:
+            return im.size
+    except (requests.RequestException, UnidentifiedImageError, OSError) as e:
+        logging.getLogger("user_level_log").warning(
+            "Failed to get image size from %s: %s", url, e
+        )
+        return (0, 0)
 
 
 class CrimsXtal:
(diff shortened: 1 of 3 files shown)


jbflo:
pre-commit passed on my local repo, but failed on here, i'll have to check why.

EDITED : Now OK

fabcor-maxiv:
Why is `defusedxml` not added to `pyproject.toml`, then? Potentially as optional dependency if that does not make sense to have is as mandatory dependency.

marcus-oscarsson:
Thanks @jbflo. We are transitioning to `YAML` but I think that a library like `defusedxml` would actually be quite nice to add to the project globally. This PR should not replace all the usages of `xml.etree.cElementTree` but I think we can add the dependency and perhaps follow up with a PR that replaces `xml.etree.cElementTree` ?

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -134,26 +157,15 @@ def send_data_collection_info_to_crims(
...
         response = requests.get(url, timeout=900)
-        # response = post(url, data=data, timeout=900)
-        print(response.text)
         return response.text
-    except Exception as ex:
-        msg = "POST to %s failed reason %s" % (url, str(ex))
-        return msg
+    except requests.RequestException as e:
+        logging.getLogger("user_level_log").warning("Request to %s failed: %s", url, e)
+        return str(e)
```
marcus-oscarsson: Interesting that you also return the string, how is it used ?

marcus-oscarsson:
Could you resolve the conflict please, and then we can merge :)

jbflo:
> Thanks @jbflo. We are transitioning to `YAML` but I think that a library like `defusedxml` would actually be quite nice to add to the project globally. This PR should not replace all the usages of `xml.etree.cElementTree` but I think we can add the dependency and perhaps follow up with a PR that replaces `xml.etree.cElementTree` ?

Agreed @marcus-oscarsson.
In this particular case, we're receiving the XML data from an external API, so I've added defusedxml globally to handle it safely. For other cases — such as reading local config files — we may not need to enforce defusedxml, especially since we're transitioning to YAML.

marcus-oscarsson:
Ok, I see. The transition to XML could take some time and we should not entirely trust the data just because it comes from local disk (or seems to come from local disk). The data can easily be manipulated or the system tricked so that it seems to come from disk but it does not.

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
jbflo: > Interesting that you also return the string, how is it used ?

my intention was to return it in the  http respond error message. :smiley:

jbflo:
> Could you resolve the conflict please, and then we can merge :)

Done

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
marcus-oscarsson: I see, but its no longer the plan I guess/hope ?

jbflo: I return a boolean instead. a bit more explicit.

marcus-oscarsson (approved):

PR #1239, opened 2025-05-28
Remove the `/deprecated/ `directory

### LINKED ISSUE #1228: [CC-DESY] Get rid of the `/deprecated` directory
Probably the `/deprecated` directory can be removed...

Ask beforehand to all facilities if they still use it. But likely they also use an old version of MXCuBE-Core that will still have the files anyway.

fabcor-maxiv:
The content of the `/deprecated/` directory is not used any longer and is safe to be removed.

GitHub: close https://github.com/mxcube/mxcubecore/issues/1228


marcus-oscarsson:
Well done, thanks !

Issue #1220, opened 2025-05-16
[CC-DESY] Development and collaboration tools


fabcor-maxiv:
Maybe there is some time and interest to discuss and work on our development and collaboration tools: GitHub, the CI/CD workflows, the test runners, pre-commit, Ruff, and so on.

For example, I think I have seen some comments on some pull requests mentioning that some collaborators might have difficulties setting up their own local development environment to correctly take advantage of the helper tools that are available. So we could discuss this and try to help each other out setting things up.

alessandro-olivo:
This could be very helpful (at least for me :) ) I'm quite new to these tools

rhfogh:
_Very_ good idea. I _think_ I _may_ have got to the point where I can get ruff's actual messages by running pre-commit on the command line as a separate step, and part of ruff seems to be integrated in my IDE. But there are several very dodgy aspects to ruff:

  - failing your commits and refusing merging **without informing you what the problem is**

  - Enforcing over 800 rules (more than anyone can easily learn) where many of the things banned are perfectly legal, safe Python that are simply refused as a matter of aesthetics
  - Changing files for you during check-in without showing you what is happening.
  - If I understand it correctly 'fixing' imports by adding some random import statements, when the actual problem is something like accessing an undefined variable.
  - Ruff comes dangerously close to taking control away from the programmer and giving it to an algorithm. That does not sound like it is worth it just to get some slightly neater-looking code.

fabcor-maxiv:
@rhfogh, it would be helpful if you can prepare some concrete examples of those situations you encountered and show them at the meeting so that we can discuss on solid basis.


> _Very_ good idea. I _think_ I _may_ have got to the point where I can get ruff's actual messages by running pre-commit on the command line as a separate step, and part of ruff seems to be integrated in my IDE.

Good, that is exactly what I had in mind. I hope others will join and help because I do not use any IDE, I use vim only.


> But there are several very dodgy aspects to ruff:
> 
> * failing your commits and refusing merging **without informing you what the problem is**

Yes, I understand that. I think there are ways to improve this so that it becomes more visible what Ruff is complaining about or what it is changing. My hope is that we have time to try something during the code camp.


> * Enforcing over 800 rules (more than anyone can easily learn) where many of the things banned are perfectly legal, safe Python that are simply refused as a matter of aesthetics

One of the main points of such tools is that developers should not need to learn all those Python points[^1] and that we can delegate this to a tool instead.

[...]

PR #1241, opened 2025-05-28
Show diff on Pre-commit failure in GitHub workflow

fabcor-maxiv:
Use the `--show-diff-on-failure` option when running `pre-commit` in GitHub Actions workflow.

Some pre-commit hooks might modify files to fix issues, but not all hooks explicitly describe what they modified. When pre-commit is being run locally it is easy to check `git diff`. But if it is running in something like a GitHub Actions workflow, there is no access to the modified files and `git diff`, so it is difficult to understand where the issues are.

GitHub: https://github.com/mxcube/mxcubecore/issues/1220

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -34,6 +34,6 @@ jobs:
       - name: Run Pre-Commit hooks
         run: >
           ${MAMBA_EXE} run --name mxcubecore
-          pre-commit run --all-files
+          pre-commit run --all-files --show-diff-on-failure
 
 ...


fabcor-maxiv:
@rhfogh, here is a different approach. In this case the only change is that we let pre-commit show the `git diff` explicitly when it is run in the GitHub Actions workflow.

Here is an example of a failing job with some errors to see what it could look like: <https://github.com/mxcube/mxcubecore/actions/runs/15301858285/job/43044241913>.

So in this case, the diff shows what Ruff has "fixed".

rhfogh:
Again, that would work, but I actually prefer either of the other two ptions. This version does not tell you which ruff rule is being invoked, which is very useful to know, and I find this kind of diff a lot harder to read than the side-by-side ones I get in my IDE (Pycharm).
It could be a question of what kinds of changes can happen outside ruff. Are there other steps that change the files, and that you cannot catch otherwise?

fabcor-maxiv:
> Are there other steps that change the files, and that you cannot catch otherwise?

Yes, some of the hooks in the `repo: https://github.com/pre-commit/pre-commit-hooks` block do make modifications to files (I believe removing trailing empty spaces, and changing line endings from Windows style to Linux style for example).

https://github.com/mxcube/mxcubecore/blob/e22b794e58aa027081d94e3d1b2f95de733cc547/.pre-commit-config.yaml#L9-L24

That is why I think it makes sense to merge this change in any case, whether or not we change the Ruff hooks. Also it seems to me like this change has no negative impact or side-effect.

rhfogh (approved):

marcus-oscarsson:
Ok, thanks !

PR #1243, opened 2025-05-28
Remove the `--no-fix` option of the `ruff` hook

fabcor-maxiv:
When `ruff check --fix` makes modifications to files, it is not immediately clear where and why the files are modified. When running locally, one can see the changes with `git diff`, but it still does not clarify why (which Ruff rule was broken).

Without the `--fix` option, Ruff clearly shows where the violations are. Then the developer is free to decide how to fix the violations:
* The developer can choose to do the fixes manually.
* The developer can choose to run `poetry run ruff check --fix`.
* The developer can choose to add exceptions.
* The developer can choose to do a combination of the above.

GitHub: https://github.com/mxcube/mxcubecore/issues/1220

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,10 +33,12 @@ repos:
 
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.9.6
+    rev: v0.9.10  # Keep in sync with `poetry.lock` and `pyproject.toml`
     hooks:
+      # For the `ruff` hook (which runs `ruff check`),
+      # we purposefully skip the `--fix` option,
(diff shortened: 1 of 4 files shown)


rhfogh (approved):
Yes, please do

marcus-oscarsson:
Good for me as well, as mentioned previously :)

marcus-oscarsson (approved):

fabcor-maxiv:
OK, I updated this pull request to include `ruff` as a "dev dependency" in Poetry and with more comments and documentation as well.

rhfogh (approved):

marcus-oscarsson:
Well done !

PR #1246, opened 2025-06-02
MAXIV auto-proc: clean-up doc strings

elmjag:
Minor clean-up of the doc strings in MAXIV auto processing hardware object.

- remove doc strings that does not document anything
- follow google style for remaining doc strings

### DIFF
--- a/mxcubecore/HardwareObjects/MAXIV/MAXIVAutoProcessing.py
+++ b/mxcubecore/HardwareObjects/MAXIV/MAXIVAutoProcessing.py
@@ -17,10 +17,6 @@
 #  You should have received a copy of the GNU Lesser General Public License
 #  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
 
-"""
-MAXIVAutoProcessing
-"""
-
 import logging
 import os
 import time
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/MAXIV/MAXIVAutoProcessing.py
```diff
@@ -17,10 +17,6 @@
 #  You should have received a copy of the GNU Lesser General Public License
 #  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
 
-"""
-MAXIVAutoProcessing
-"""
```
fabcor-maxiv, line 22:
```
"""Auto processing for MAX IV."""
```

mxcubecore/HardwareObjects/MAXIV/MAXIVAutoProcessing.py
```diff
@@ -38,42 +34,29 @@
 
 
 class MAXIVAutoProcessing(HardwareObject):
-    """
-    Descript. :
-    """
-
```
fabcor-maxiv, line 44:
```
"""Auto processing for MAX IV."""
```

fabcor-maxiv (approved):
I believe that it is better to have a docstring with obvious content rather than no docstring, as long as it is valid content. On the other hand I agree that docstrings that are not valid (e.g. `Descript. :`) should be deleted.

PR #1247, opened 2025-06-02
Format Markdown documents with `mdformat`

### LINKED ISSUE #1244: Checker, linter, formatter for Markdown documents?
Maybe it would be good for consistency to introduce some kind of checker, linter, formatter for Markdown documents.

There is no obvious candidate that I can think of. Apparently, [Prettier](https://prettier.io/) which we already use in MXCuBE-Web for the JavaScript code, can also handle Markdown, so maybe that would be an acceptable candidate.

For context: <https://github.com/mxcube/mxcubecore/pull/1236#issuecomment-2915541437>.

fabcor-maxiv:
GitHub: close https://github.com/mxcube/mxcubecore/issues/1244


marcus-oscarsson:
I this still draft, or are you happy to review and merge it ?

fabcor-maxiv:
> I this still draft, or are you happy to review and merge it ?

Still draft. There are outstanding issues: <https://github.com/mxcube/mxcubecore/issues/1244#issuecomment-2931132739>.

fabcor-maxiv:
I recreated this pull request from scratch.

This time around:
* we do not use the [`mdformat-myst`](https://pypi.org/project/mdformat_myst/) plugin, because it causes some issues.
* we use the [`mdformat-tables`](https://pypi.org/project/mdformat-tables/) plugin, which seems to give good results
* we do only the changes from `mdformat` with as little manual editing as necessary, for example we do not add the unlinked pages to the index (to be done in separate pull request)

marcus-oscarsson:
~So this basically only formats tables ?~

marcus-oscarsson:
I see `mdformat-tables` is a plugin to `mdformat` that formats tables and the rest is handled by `mdformat`

marcus-oscarsson:
I'm fine with this, just a comment or perhaps more a question. I've picked up somewhere that **CommonMark** and **Markdown** are not exactly the same. `mdformat` enforces **CommonMark** and not **Markdown**, would that be an issue ?

marcus-oscarsson (approved):

fabcor-maxiv:
Yes, [`mdformat`](https://mdformat.readthedocs.io/) is a tool to help us having consistent formatting for "Markdown" documents. It was suggested by @axelboc in the associated issue ticket <https://github.com/mxcube/mxcubecore/issues/1244#issuecomment-2929542958>.

`mdformat` has a bunch of [plugins](https://mdformat.readthedocs.io/en/stable/users/plugins.html):

* Since our use of Markdown is essentially in the context of Sphinx via [MyST](https://myst-parser.readthedocs.io/), I tried the [`mdformat-myst`](https://pypi.org/project/mdformat_myst/) plug-in but it did not give good results, so I got rid of it.

* I added the [`mdformat-ruff`](https://pypi.org/project/mdformat-ruff/) plug-in, because it allows to format Python code examples embedded in Markdown documents according to the same rules we use in actual code. This plug-in seems to work as expected.

* I also added the [`mdformat-tables`](https://pypi.org/project/mdformat-ruff/) plug-in, because tables are a pain to format correctly by hand, so it is helpful to have such a tool. This plug-in seems to work as expected.

---

***Markdown vs. CommonMark***

Let's quote [Wikipedia](https://en.wikipedia.org/wiki/Markdown):

> The initial description of Markdown[[10]](https://en.wikipedia.org/wiki/Markdown#cite_note-10) contained ambiguities and raised unanswered questions, causing implementations to both intentionally and accidentally diverge from the original version. This was addressed in 2014 when long-standing Markdown contributors released [CommonMark](https://en.wikipedia.org/wiki/Markdown#Standardization), an unambiguous specification and test suite for Markdown.

The way I understand it, the "real Markdown" does not have a clear specification. In short, in the beginning Markdown was basically just one Perl script made by one person. Many things were ambiguous. Many things were inconsistent.

Then people started creating their own Markdown-ish syntax: GitHub flavoured Markdown, GitLab flavoured Markdown, and so on. All of those had their own specificities and were slightly incompatible with each other.

So some other people decided to write an actual standardised specification for a Markdown-ish syntax, which is called "[CommonMark](https://commonmark.org/)".

As far as I know, nowadays, all the things called "Markdown" that we see everywhere and use all the time are actually based on and compatible with CommonMark. All platforms (GitHub, GitLab, etc.) add their own stuff on top of CommonMark, but they usually keep it compatible with CommonMark. Nowadays CommonMark is the common ground for all Markdown-ish syntax we usually find. These platforms add stuff because the standardised CommonMark specification does not have many features. One of the most striking example is that CommonMark does not have tables. But even without tables being in the standard specification, most parsers and renderers (GitHub, GitLab, MyST, etc.) do understand tables in a way that is compatible.

> I've picked up somewhere that **CommonMark** and **Markdown** are not exactly the same. `mdformat` enforces **CommonMark** and not **Markdown**, would that be an issue ?

So, long story short, no it is not an issue at all to work with "CommonMark" instead of "Markdown". Because "Markdown" is not really a thing (and has not been for quite a long time), while "CommonMark" is a real, solid thing. Everywhere we see "Markdown", it is quite safe to assume it is actually a "CommonMark" syntax plus some things like tables.

PR #1249, opened 2025-06-03
Ignore Ruff rule `N818`

fabcor-maxiv:
The Ruff rule `N818:error-suffix-on-exception-name` comes from PEP8 and does not seem to make sense outside of Python itself (the interpreter and the standard library).
In general, user code has way more "exceptions" than "errors".

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -96,6 +96,7 @@ ignore = [
     "ANN",  # flake8-annotations: ignored, too many hits (21744)
     "COM812",  # Ruff recommends to ignore this linter rule when using its formatter
     "D",  # pydocstyle: ignored, too many hits (16292)
+    "N818", # error-suffix-on-exception-name: rather pointless outside of Python itself
     "PLR2004",  # magic-value-comparison: ignored, too many hits (249)
     "UP",  # pyupgrade: ignored for now, can be fixed automatically by ruff itself
     "G004", # Allow uses of f-strings to format logging messages
@@ -4627,7 +4628,6 @@ convention = "google"
     "ERA001",
     "F841",
     "N814",
-    "N818",
     "N999",
     "PLW0603",
     "PT009",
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
PEP8 was meant for the Python code base (the interpreter and its standard library), not for user code. So some of the PEP8 rules do not make sense for user code. In this case, if I understood everything correctly, the (main) distinction between an "error" and an "exception" is that an error is a "coding mistake", so that there is no reason for user code to raise "errors", only the Python interpreter (or the standard library) should raise "errors".

marcus-oscarsson:
Ok, sounds reasonable.

marcus-oscarsson (approved):

marcus-oscarsson:
Actually thinking about it. I guess it could make sense in frameworks, like mxcube. However I don't see any need for being that formal. I'm completely fine with ignoring this rule. 

As far as it concerns me, you can re-base and merge :) unless anyone else has some other input

fabcor-maxiv:
Yes, you might be right... For example if we detect that the configuration (YAML/XML) for an hardware object is not valid, we might want to raise an "error", right? I do not know if we actually do that now, but it sounds plausible.

In any case, I do not get the point of the rule of adding the `Error` suffix to all the exception classes. I don't get it. I read the pages many times, I still don't get it. Maybe someone smarter than me understands it.

* <https://docs.astral.sh/ruff/rules/error-suffix-on-exception-name/>
* <https://peps.python.org/pep-0008/#exception-names>

fabcor-maxiv:
I do not see the case happening where the majority of our exceptions are "errors". But maybe I am misunderstanding something.

marcus-oscarsson:
> Yes, you might be right... For example if we detect that the configuration (YAML/XML) for an hardware object is not valid, we might want to raise an "error", right? I do not know if we actually do that now, but it sounds plausible.

That's what I thought. But you are right we can ignore this rule it seems unnecessarily procedural. I don't see that we would have any use for it.

rhfogh:
The simple approach to this would be to look at the consequences. Do we like the code better if it follows the rule? If yes, adopt it. If no, ignore it - Ruff is there to help us, not the other way around.

Issue #944, opened 2024-06-18
Configuration files setup


beteva:
Find a way to share the mockup configuration files between mxcubecore and mxcubeweb repositories,

Mockup configuration files are needed both for testing the hardware objects in the mxcubecore and the implementation in mxcubeweb. The files are now in both repositories and very often differ from each other. Even worse - they are sometimes inconsistent (e.g. udiff_omega.xml in the mxcubeweb erepository does not correspond to the MotorMockup class in mxcubecore)

marcus-oscarsson:
I think we should cleanup the configuration file directory in `mxcubecore`.

    https://github.com/mxcube/mxcubecore/tree/develop/mxcubecore/configuration

We should also in my opinion move it out of the `mxcubecore` module as its not python code
or a resource used directly by the code. The configuration files could become their own
repository but its perhaps more practical to have them at the root of the `mxcubecore` repository.

Like this we can download these files from the UI's and use them for the tests, its a bit more
cumbersome but it removes the duplication that @beteva mentions above.

Its perhaps a good occasion to look at this now as we are also dealing with #931 .

rhfogh:
How do you see this working across the Web and Qt implementations? Would you (still) be able to share mock configuration for hardware objects across the two?

marcus-oscarsson:
Oh yes, actually I see this also in the perspective of #931. Its probably sensible to first cleanup https://github.com/mxcube/mxcubecore/tree/develop/mxcubecore/configuration before also converting those files. It was also the reason why I thought it would be good to tackle #902.

To share the exact same configuration file between web and qt versions would be very nice and I think the XML to YAML is a good step in that direction but it will of course require more work than only that.

Looking at the configuration files with @beteva we realized that keeping them in the `mxcubecore` module is in our opinion not the most sensible thing to do. 

 - Its not python code or a resource directly used by the code so it should probably not be in the source code. 
 
 - Its also a bit odd to download part of the internals of mxcubecore to get the configuration files

So we would be in favor of moving the to the root of the repository like this they are at least not part of the python module, I guess it would even be more elegant if they are kept in their own repository.

marcus-oscarsson:
As a side note our ambition is to start to try out the YAML configuration file system as soon a s possible so that we can fix any remaining bugs. We would of-course very much appreciate if someone else did the same :)

elmjag:
I think the config files used for mxcubecore tests should be kept separate from the ones in mxcubeweb. They are used for different purposes.

[...]

PR #1250, opened 2025-06-04
WIP Yaml configuration update - and adapted to also work for mxcubeqt 

rhfogh:
1) Fixed some bugs so that mockup mode works also with mxcubeqt

2) Made new mockup.yaml directory, porting demo.yaml configs from mxcubeweb, with additional fixes to w0rk also with Qt branch.

3) removed the bulky gphl test_samples directories in the new config tree. When this is accepted we can delet the old one.

NB: having the mockup.yaml in mxcubecore instead of mxcubeweb
1) is tested and works (in mock mode)
2) would make much more sense. 
Almost all the files are shared between the web and qt versions so it is bad to have two duplicates.
Also the configs are adapted to the (mxcubecore) hardware objects so they more naturally belong in the same repository. 
Differences between web and qt configs are handled simply by using a two-directory lookup path for configuration files, so that web- and qt-specific files are picked up from separate directories.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "mxcubecore"
-version = "1.324.0"
+version = "1.325.0"
 license = "LGPL-3.0-or-later"
 description = "Core libraries for the MXCuBE application"
 authors = [


beteva:
Nice work @rhfogh,

Below are some comments concerning the configuration files:

> 2\. Made new mockup.yaml directory, porting demo.yaml configs from mxcubeweb, with additional fixes to work also with Qt branch.

Although a lot of the configuration files are the same, we have two different configuration directories - one in mxcubecore, another in mxcubeweb as they serve two different purposes.

1. The one in mxcubecore is used for the tests only. It has by purpose also some wrong configurations, just to provoke an error for the tests. I guess we should make the copy of this one for the yaml, not the mxcubeweb/demo one.

2. The mxcubeweb demo directory is used to run the mock application with the mockup classes. It is only used for that and I believe this is what is needed for the mxcubeqt as well. The corresponding yaml config files shouid stay in the GUI (mxcubeweb and mxcubeqt) repository.

> 2\. would make much more sense.
>         Almost all the files are shared between the web and qt versions so it is bad to have two duplicates.
>         Also the configs are adapted to the (mxcubecore) hardware objects so they more naturally belong in the same repository.
>         Differences between web and qt configs are handled simply by using a two-directory lookup path for configuration files, so that web- and qt-specific files are picked up from separate directories.

This has already been discussed in #944 and I thought we had agreed on that.

mxcubecore/configuration/mockup.yaml/aperture.yaml
```diff
@@ -0,0 +1,5 @@
+%YAML 1.2
+---
+class: ApertureMockup.ApertureMockup
+configuration:
+  diameter_size_list: '[10, 20, 50, 100, 150]'
```
beteva: This is not the correct file. The ApertureMockup does not expect diameter_size. It has some predefined size list in the file itself: 5, 10, 20, 30, 50, 100. If you need to change them, please use the values (as for any NState HO).

mxcubecore/configuration/mockup.yaml/beamstop.yaml
```diff
@@ -0,0 +1,7 @@
+%YAML 1.2
+---
+class: ExporterNStateMockup.ExporterNStateMockup
+configuration:
+  use_hwstate: true
+  username: beamstop
+  values: '{"OUT": "OFF", "IN": "BEAM"}'
```
beteva: This also looks like an outdated version. There is no need of use_hwstate property

mxcubecore/configuration/mockup.yaml/beamstop_alignment_x.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  GUIstep: 0.1
+  exporter_address: 130.235.94.124:9001
+  motor_name: BeamstopAlignmentX
+  unit: 0.001
+  username: Beamstop Alignemnt X
```
beteva: I guess we do not need this file at all. It also seems that it is a configuration for a MD2 motor and not the MotorMockup

mxcubecore/configuration/mockup.yaml/capillary.yaml
```diff
@@ -0,0 +1,7 @@
+%YAML 1.2
+---
+class: ExporterNStateMockup.ExporterNStateMockup
+configuration:
+  use_hwstate: true
+  username: capillary
+  values: '{"IN": "BEAM", "OUT": "OFF"}'
```
beteva: same as for beamstop.yml

mxcubecore/configuration/mockup.yaml/characterisation.yaml
```diff
@@ -0,0 +1,6 @@
+%YAML 1.2
+---
+class: EDNACharacterisationMockup.EDNACharacterisationMockup
+configuration:
+  edna_command: /users/blissadm/local/bin/edna_script_bes.py
+  edna_default_file: edna_defaults.xml
```
beteva: Is mixing yaml and xml files possible?

mxcubecore/configuration/mockup.yaml/energy.yaml
```diff
@@ -0,0 +1,13 @@
...
+class: EnergyMockup.EnergyMockup
+configuration:
+  mad:
+    defaultenergy:
+      - directory: pk
+        energy: 12.6595
+      - directory: ip
+        energy: 12.6595
+      - directory: rm
+        energy: 12.708
```
beteva: This is not part of the EnergyMockup.

mxcubecore/configuration/mockup.yaml/slits.yaml
```diff
@@ -0,0 +1,14 @@
...
+  gapH:
+    maxGap: 0.3
+    minGap: 0.005
+    stepSize: 0.005
+    updateTolerance: 0.0005
+  gapV:
+    maxGap: 0.3
+    minGap: 0.005
+    stepSize: 0.005
+    updateTolerance: 0.0005
```
beteva: Not sure this is used by the SlitsMockup, nor AbstractSlits

rhfogh:
@beteva OK, so you are saying that we should have three complete, parallel copies of the configuration files: One in mxcubeweb, for running the web GUI in mock mode, one in mxcubeqt, for running the qt gui in mock mode, and one in mxcubecore, for running automated tests. And, of course maintain and upgrade them in parallel.

Personally I would prefer to avoid the additional work required - particularly since it has not been clear until now that those configuration trees were being kept up to date, or by whom. Let us see what the consensus is.

mxcubecore/configuration/mockup.yaml/udiff_backlight.yaml
```diff
@@ -0,0 +1,7 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  exporter_address: 192.168.96.43:9001
+  motor_name: backlight
+  username: backlight
```
beteva: This also looks like an MD and not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_kappaphi.yaml
```diff
@@ -0,0 +1,7 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  exporter_address: 130.235.94.124:9001
+  motor_name: Phi
+  username: Kappa Phi
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_kappa.yaml
```diff
@@ -0,0 +1,8 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  default_limits: (0, 225)
+  exporter_address: 130.235.94.124:9001
+  motor_name: Kappa
+  username: Kappa
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_omega.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  GUIstep: 90
+  exporter_address: 130.235.94.124:9001
+  motor_name: Omega
+  unit: 0.001
+  username: Omega
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_phix.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  GUIstep: 0.1
+  exporter_address: 130.235.94.124:9001
+  motor_name: AlignmentX
+  unit: 0.001
+  username: PhiX
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_phiy.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  direction: -1
+  exporter_address: 130.235.94.124:9001
+  motor_name: AlignmentY
+  unit: 1e-05
+  username: PhiY
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_phiz.yaml
```diff
@@ -0,0 +1,8 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  exporter_address: 130.235.94.124:9001
+  motor_name: AlignmentZ
+  unit: 1e-05
+  username: PhiZ
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_sampx.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  GUIstep: 0.01
+  exporter_address: 130.235.94.124:9001
+  motor_name: CentringX
+  unit: 1e-05
+  username: sampx
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/udiff_sampy.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.2
+---
+class: MotorMockup.MotorMockup
+configuration:
+  GUIstep: 0.01
+  exporter_address: 130.235.94.124:9001
+  motor_name: CentringY
+  unit: 1e-05
+  username: sampy
```
beteva: MD not Mockup motor configuration

mxcubecore/configuration/mockup.yaml/xrf.yaml
```diff
@@ -0,0 +1,4 @@
+%YAML 1.2
+---
+class: XRFMockup.XRFMockup
+configuration: {}
```
beteva: I do not think we have ever used this.

beteva (changes requested):
Dear @rhfogh, you’ve probably just made a copy of some files, already present elsewhere, but I would not like that we keep irrelevant/old/wrong/unused files. Also I think if these are the files, copied from mxcubeweb, that they should not be here.

rhfogh:
@beteva Sorry, but these, or something similar __do__ need to be here. I cannot run tests on a qt implementation using a set of configuration files that only live in the mxcubeweb repository. As you have seen from the relevant issue, the idea was that these should replace the set in mxcubeweb. If we cannot replace them we need to duplicate them somewhere.

For the rest you are quite right that these are a copy - to wit of the demo.yaml directory in mxcubeweb, My need was for a set of mockup config files that would allow me to run the Qt implementation with current mxcubecore and yaml configuration. And as the quickest route I assumed that the files in demo.yaml were correct - seeing that they actually worked when running mock mxcubeweb. And seeing that I could hardly propose having a shared config directory at the same time as I made lots of changes that might break the mxcubeweb version. 

If we are going to have a single, shared mockup config directory it does not matter if we copy what we have to the new location, imperfections and all, and then clean it afterwards. If we are going to continue with separate and parallel config directories then we clearly need to finish cleaning up demo.yaml before we make any copies. Only I submit that the people best placed to do that would be those who have been setting up the yaml configuration, changing the beam definer organisation etc. - and that is not me. Since the overriding demand on the config files is that they must match the hardware objects - which are the same for both the Web and Qt version - it does not make much sense to have two separate groups of people do the fixing in two separate directories. 

If we do not agree on changing the overall organisation of mockup configs it might be better to change this PR to WIP. I can then keep using it for my own personal testing, and we can wait till the mxcubeweb configuration is stable and clean before we try to make a set of configs that are usable for the qt version.

beteva:
> @beteva OK, so you are saying that we should have three complete, parallel copies of the configuration files: One in mxcubeweb, for running the web GUI in mock mode, one in mxcubeqt, for running the qt gui in mock mode, and one in mxcubecore, for running automated tests. And, of course maintain and upgrade them in parallel.
> 
> Personally I would prefer to avoid the additional work required - particularly since it has not been clear until now that those configuration trees were being kept up to date, or by whom. Let us see what the consensus is.


This can be seen from another perspective. I think we have three different needs:
1. Testing and CI
2. Running mock GUI (web and Qt)
3. Having templates for each HO

- The templates are not necessarily used in neither test, nor mock GUI, but can help a lot when doing real beamline configuration.
- The mock GUI files should correspond to the templates for the Mockup object. There is one exception, though. For the Qt there are properties which are used by the GUI only, not the HO (namely unit and GUIstep), This is not the case for the web.
- The test files require maybe more that running mock GUI and may require some error producing parameters (as it is the case for the machine_info).
- For both testing and running GUI, there are configuration files which do not have Mockup classes (like for the queue), but these are the easiest to generalise, as they are the same in the three cases.

So we need to see how to answer all the thee needs with a minimum of duplication of information and risk of diverging in case of refurbishment. Personally I think we should provide template files at least for the Mockup classes and keep them for both testing and GUI and only have them duplicated if an extra configuration is needed.

rhfogh:
@beteva That makes excellent sense. The question is just how. My proposal would be to have a directory with templates for everything,  separate directories for the (limited number of) classes that have to be different for mxcubeweb, mxcubeqt, and core testing, and a setup with two directories in the lookup path, so that you could override the general templates from the specific directories. I'd say that the templates and the test-specific changes would have to be in mxcubecore; the rest could be either in mxcubecore or in the web/qt directory. 

Whatever we agree on, the next question is the best way of moving towards it. Here there are several approaches.

beteva:
> it might be better to change this PR to WIP. I can then keep using it for my own personal testing, and we can wait till the mxcubeweb configuration is stable and clean before we try to make a set of configs that are usable for the qt version.

Could be a good idea to change the PR to WIP or at least split it in two and put the configuration files on a different one. Maybe the people who has set the yaml files in the demo directory could tell what the "clean" files are.
Sorry to fuss with all this, but as we said we are cleaning up, we need to do it properly 😄 🧹

rhfogh:
@beteva  You are quite right. Will do.

marcus-oscarsson:
I guess this is waiting for a discussion in the developers meeting ?. Are there parts of this, like the bug fixes, that can be merged separately ?

rhfogh:
@marcus-oscarsson Yes, this is waiting for a discussion.

I can see it would be nice to get some bug fixes done faster, but I do not think it is a sensible thing to do now. The bug fixes are all about how the code interacts with the configuration, and there was not a Qt branch with functional mock configuration before this PR. I think it makes more sense to decide what to do and then make the permanent solution, rather than spend  time on fixing and testing a broken mock Qt branch, when the whole thing will shortly have to be redone anyway.

PR #1252, opened 2025-06-05
Ruff: EXE001 shebang-non-executable

dominikatrojanowska:
Getting rid of ruff EXE001 rule.

### DIFF
--- a/mxcubecore/HardwareObjects/Beamline.py
+++ b/mxcubecore/HardwareObjects/Beamline.py
@@ -1,4 +1,3 @@
-#! /usr/bin/env python
 # encoding: utf-8
 #
 #  Project: MXCuBE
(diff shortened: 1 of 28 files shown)


rhfogh (approved):
Fine with me - these lines were always boilerplate anyway.

One question, though: Why is this PR removing links like
"# encoding: utf-8"?
Is that a different ruff rule? Not that I shall miss them much, but it is always good to know what the rules are, and why.

dominikatrojanowska:
Thanks for noticing @rhfogh :) it's different ruff rule, don't know how I removed them, but for sure accidentally. I will provide new PR for that.

marcus-oscarsson:
OK, there is some linting warning it seems

dominikatrojanowska:
now, everything passes :) I guess I have some slightly different configuration of pre-commit or version of some tool, that some of linting rules are not showing up on my pc...

PR #1253, opened 2025-06-05
Ruff: PT011 pytest-raises-too-broad

dominikatrojanowska:
Getting rid of ruff PT011 rule.
Related issue: https://github.com/mxcube/mxcubecore/issues/1230

PR #1254, opened 2025-06-05
Ruff: PT012 pytest-raises-with-multiple-statements

dominikatrojanowska:
Getting rid of ruff PT012 rule.

### DIFF
--- a/test/TestAbstractNStateBase.py
+++ b/test/TestAbstractNStateBase.py
@@ -84,5 +84,5 @@ class TestAbstractNStateBase(TestAbstractActuatorBase.TestAbstractActuatorBase):
         # Must be set first so the next command causes a change
         test_object.set_value(val2, timeout=None)
         test_object.set_value(val1, timeout=0)
-        with pytest.raises(RuntimeError)):
+        with pytest.raises(RuntimeError):
             test_object.wait_ready(timeout=1.0e-6)
--- a/test/TestAbstractNStateBase.py
+++ b/test/TestAbstractNStateBase.py
@@ -83,6 +83,6 @@ class TestAbstractNStateBase(TestAbstractActuatorBase.TestAbstractActuatorBase):
 
         # Must be set first so the next command causes a change
         test_object.set_value(val2, timeout=None)
-        with pytest.raises(RuntimeError):
-            test_object.set_value(val1, timeout=0)
+        test_object.set_value(val1, timeout=0)
+        with pytest.raises(RuntimeError)):
             test_object.wait_ready(timeout=1.0e-6)
--- a/test/test_flux.py
+++ b/test/test_flux.py
@@ -64,6 +64,5 @@ class TestFlux(TestAbstractActuatorBase.TestAbstractActuatorBase):
     def test_flux_methods(self, test_object):
         """Test the methods"""
         # Test timeout - expecting to have RuntimeError
-        with pytest.raises(RuntimeError) as info:
+        with pytest.raises(RuntimeError):
             test_object.wait_for_beam(0)
-        print(f"------> Flux: {info}")
--- a/test/test_flux.py
+++ b/test/test_flux.py
@@ -65,5 +65,5 @@ class TestFlux(TestAbstractActuatorBase.TestAbstractActuatorBase):
         """Test the methods"""
         # Test timeout - expecting to have RuntimeError
         with pytest.raises(RuntimeError) as info:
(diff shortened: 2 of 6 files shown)


test/TestAbstractNStateBase.py
```diff
@@ -84,6 +84,6 @@ def test_setting_timeouts_2(self, test_object):
 
         # Must be set first so the next command causes a change
         test_object.set_value(val2, timeout=None)
+        test_object.set_value(val1, timeout=0)
         with pytest.raises(Exception):
-            test_object.set_value(val1, timeout=0)
```
marcus-oscarsson, line 87: I think its more the `set_value` rather than the `wait_ready` that would raise and I guess the wait_ready would raise `TimeoutException` perhaps ?

fabcor-maxiv:
Note that the companion best practice of this is that a `try` block should ideally contain one single statement as well. `try` blocks that contain dozens of lines of code are a very bad idea and a perfect footgun. And we have an awful amount of those in MXCuBE.

test/TestAbstractNStateBase.py
dominikatrojanowska: hm, actually it is `wait_ready` that rises `RuntimeError` --> https://github.com/mxcube/mxcubecore/pull/1253/files#diff-67027594b38337001d4142cdd914e175e482f044010005274682becb7348af58R86

dominikatrojanowska:
> Note that the companion best practice of this is that a `try` block should ideally contain one single statement as well. `try` blocks that contain dozens of lines of code are a very bad idea and a perfect footgun. And we have an awful amount of those in MXCuBE.

So what is you suggestion? Add another issue? Maybe there is some ruff rule excluded for now that complains about it 🤔

fabcor-maxiv:
> > Note that the companion best practice of this is that a `try` block should ideally contain one single statement as well. `try` blocks that contain dozens of lines of code are a very bad idea and a perfect footgun. And we have an awful amount of those in MXCuBE.
> 
> So what is you suggestion? Add another issue? Maybe there is some ruff rule excluded for now that complains about it 🤔

No suggestion, besides a nudge to think about this when writing code or doing code review in the future.

And also a reminder that in many cases we have exceptions that are caught but it is very difficult to know where the exception was raised since the `try` block is so big, so if it ever happens to you in a troubleshooting session (as it has happened to me many times with MXCuBE) that this might be the reason why.

I do not know that there is a Ruff rule for this. I guess I wish there was one.

test/TestAbstractNStateBase.py
marcus-oscarsson: Hm, not sure that it tells me where the exception comes from. I find it a bit odd that `wait_ready` would raise a `RuntimeError`. But thats perhaps another issue. Its fine if its actually the code actually raises `RuntimeError` and that is what is tested

marcus-oscarsson: ~Its unclear from the test if the timeout is supposed to come from `set_value` or `wait_ready` or even both. I guess `set_value` would also raise this exception. So perhaps adding another `try` would be a good idea (it needs to be confirmed if its the case though)~

marcus-oscarsson: I think its fine, you can ignore the last comment

beteva:
To clear this - wait_ready indeed raises RuntimeError (as from BaseHardwareObjects).
set_value can also raise RuntimeError, as it calls wait_ready if the timeout is not 0. It can also raise ValueError, but this is not what is tested in the code, changed in this PR. The same is true for all the three Test*Base.py, as set_value is defined in the AbstractActuator.

test/test_flux.py
```diff
@@ -65,5 +65,5 @@ def test_flux_methods(self, test_object):
         """Test the methods"""
         # Test timeout - expecting to have RuntimeError
         with pytest.raises(RuntimeError) as info:
-            print(f"------> Flux: {info}")
             test_object.wait_for_beam(0)
+        print(f"------> Flux: {info}")
```
beteva: This can be removed

beteva (changes requested):
I am fine with this PR, but will not approve it, as there is a need to resolve #1258  first, before doing any other changes to the code.

marcus-oscarsson:
We will have a look together with @beteva next week. It looks good as far as I can see now. If you rebase we can be ready to merge it on Tuesday (yes Monday is a another holiday)

PR #1255, opened 2025-06-05
Draft: Ruff: ARG002 unused-method-argument

dominikatrojanowska:
Getting rid of ruff ARG002 rule.

### DIFF
--- a/mxcubecore/HardwareObjects/BlissActuator.py
+++ b/mxcubecore/HardwareObjects/BlissActuator.py
@@ -42,11 +42,11 @@ class BlissActuator(AbstractActuator):
         return self.actuator_state
 
     @task
-    def actuator_in(self, wait=True, timeout=None):
+    def actuator_in(self, _wait=True, _timeout=None):
         self._actuator.set_in()
         self.value_changed(self._actuator.state())
 
-    def actuator_out(self, wait=True, timeout=3):
+    def actuator_out(self, _wait=True, _timeout=3):
         self._actuator.set_out()
         self.value_changed(self._actuator.state())
 
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -1318,11 +1318,11 @@ class Null:
     provided here.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *_args, **_kwargs):
         "Ignore parameters."
         return None
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self, *_args, **_kwargs):
         "Ignore method calls."
         return self
 
--- a/mxcubecore/Command/Epics.py
+++ b/mxcubecore/Command/Epics.py
@@ -146,7 +146,7 @@ class EpicsCommand(CommandObject):
             if callback is not None:
                 callback(value)
 
-    def on_polling_error(self, exception, poller_id):
+    def on_polling_error(self, _exception, poller_id):
         # try to reconnect the pv
         self.pv.connect()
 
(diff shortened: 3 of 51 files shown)


marcus-oscarsson:
We have to be careful with this one, a argument may not be used in the method definition but by calling code. The code would break if it was the case.

dominikatrojanowska:
ok, I will create bunch of smaller PRs for general purpose files and for each facility separately to have discussions there

fabcor-maxiv:
Yeah, this is a tough one. I have no idea, currently, on how we should address this rule. We need to spend more time and think this through.

rhfogh:
@fabcor-maxiv How about just ignoring the rule? Surely Ruff is supposed to help us improve bad code, not to tie us into knots to comply with ruff? In most cases those unused arguments are there for a reason: we need a number of different functions to fit into the same slot (e.g. subtyping), and some of them make use of those parameters. Renaming them to start with underscore does not really make a difference either way, and might actually be more confusing for those cases where the additional parameters are in use.

My own IDE already warns for unused parameters, so there is warning also without ruff. If we want to keep the rule to catch the few cases where this is a genuine error, we could simply make an explicit exception (possibly with a comment) in those cases where we actually mean it.

fabcor-maxiv:
> In most cases those unused arguments are there for a reason: we need a number of different functions to fit into the same slot (e.g. subtyping), and some of them make use of those parameters.

For now, me personally, I am not convinced that all of those arguments are here for a good reason.

Let's take one of the first examples...

On one hand:

https://github.com/mxcube/mxcubecore/blob/e3fe71c59b4ab7240e6d8d0586c79b02d17036db/mxcubecore/CommandContainer.py#L52-L55

On the other hand:

https://github.com/mxcube/mxcubecore/blob/e3fe71c59b4ab7240e6d8d0586c79b02d17036db/mxcubecore/Command/Mockup.py#L40-L41

Where does the `list_args` come from? It is not documented, but that is another issue... I can not find any other occurrence of `list_args` in the whole code base. I have no idea that it is needed. Maybe Ruff is right.

mxcubecore/HardwareObjects/BlissActuator.py
```diff
@@ -42,11 +42,11 @@ def get_actuator_state(self, read=False):
         return self.actuator_state
 
     @task
-    def actuator_in(self, wait=True, timeout=None):
+    def actuator_in(self, _wait=True, _timeout=None):
         self._actuator.set_in()
         self.value_changed(self._actuator.state())
 
-    def actuator_out(self, wait=True, timeout=3):
+    def actuator_out(self, _wait=True, _timeout=3):
```
beteva, line 49: One should not change the parameters simply because ruff is complaining. The classes, calling these methods, rely on the name of the parameter.

rhfogh:
@fabcor-maxiv Indeed, maybe Ruff is right. But in those cases the thing to do would be to analyse the code, confirm that Ruff is right, remove the additional parameter, and make sure that nothing breaks. Prepending an underscore and leaving the parameter in does not sound like a solution.

One possible complicating factor is that in some cases  additional parameters could have been put there to support possible future extensions - which may or may not still be relevant. list_args, for instance, was clearly put there to allow the (completely) generic MockupCommand to take a list of additional parameters. Might we still need that? Is anyone using it? Could/should it be replaced by *args, or removed altogether? Someone would have to check and decide.

In the specific case I'd say that it looks very much like an error to have a different signature on MockupCommand and CommandObject (regrettably, Mock objects are often not updated when they should). But, again, the solution would be to remove the pareameter, not rename it.

PR #1236, opened 2025-05-27
Documents DATA_MODEL ui-related methods

mockoocy:
The documentation section, about `QueueEntry.DATA_MODEL` was missing information about `update_dependent_fields` and `ui_schema` methods. Also it gives insight about `ui:schema.col` option that is introduced in https://github.com/mxcube/mxcubeweb/pull/1657

PR #1257, opened 2025-06-05
Remove exception in `tests` GitHub Action workflow

fabcor-maxiv:
Since the jobs in this workflow are "required" for merge requests, the workflow should never be skipped.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,8 +4,6 @@ name: Pytest
 
 "on":
   pull_request:
-    paths-ignore:
-      - docs/**
     types: [opened, reopened, synchronize]
 
 jobs:


mockoocy (approved):

marcus-oscarsson:
Could you rebase and merge please ?

marcus-oscarsson (approved):

marcus-oscarsson:
We are not running any tests for the documentation but I guess it does not hurt

fabcor-maxiv:
> We are not running any tests for the documentation but I guess it does not hurt

In case it was not clear, and I should have mentioned it from the start, this change is necessary because of cases like this: <https://github.com/mxcube/mxcubecore/pull/1236#issuecomment-2943354364>

Issue #1226, opened 2025-05-21
[CC-DESY] Clean up `mxcubecore/configuration`


fabcor-maxiv:
The status of the content of `mxcubecore/configuration` is inconsistent.

A lot of its content seems to be completely outdated. Some facility/beamline does have up-to-date content in this directory.

One worry, is that in principle these files might contain secrets.

One use is that some facilities use the configuration of other facilities as reference / example for their own configuration.

One possibility is to move the content of this directory to <https://github.com/mxcube/mxcube_configuration> which is a "private" GitHub repository created specifically for the purpose of containing such configuration files.

---

For tracking:

* [ ] `alba_xaloc13`
* [ ] `desy_p11`
* [ ] `embl_hh_p13`
* [ ] `embl_hh_p14`
* [ ] `embl_hh_pe2`
* [x] `esrf_id13`
  * "deprecated and can be removed from here" <https://github.com/mxcube/mxcubecore/pull/1256>
* [x] `esrf_id231`
  * "deprecated and can be removed from here" <https://github.com/mxcube/mxcubecore/pull/1256>
* [x] `esrf_id29`
  * "deprecated and can be removed from here" <https://github.com/mxcube/mxcubecore/pull/1256>
* [x] `esrf_id30a2`
  * "deprecated and can be removed from here" <https://github.com/mxcube/mxcubecore/pull/1256>
* [x] `esrf_id30b`
  * "deprecated and can be removed from here" <https://github.com/mxcube/mxcubecore/pull/1256>
* [x] `lnls_manaca`
  * <https://github.com/mxcube/mxcubecore/pull/1422> 
* [x] `lnls_sol`
  * <https://github.com/mxcube/mxcubecore/pull/1399>
* [ ] `mockup`
  * is needed for tests, maybe could be moved to `/tests` folder <https://github.com/mxcube/mxcubecore/pull/1263>
* [ ] `soleil`
* [ ] `soleil_px1`
* [ ] `soleil_px2`
* [ ] `test`

alessandro-olivo:
!!! `mxcubecore/configuration/mockup` is needed for the tests, maybe we can move it in the `/tests` folder

fabcor-maxiv:
For MAX IV, it does not matter, we store all our configuration in a completely different git repository on a completely different git server.

beteva:
ESRF configuration files are very much deprecated and can be removed from here.

fabcor-maxiv:
[...]

Issue #1251, opened 2025-06-04
Move mockup yaml configuration to mxcubecore


rhfogh:
Almost the entire configuration is shared between the web and qt versions; now we are moving to yaml configuration would be a good time to move over to a single, shared config directory. The small differences between web and qt configs  can be easily handled by a two-directory config lookup path. PR #1250 (q.v.) shows the new config location already implemented and tested in both qt and web versions. All we would need to do is to agree and adopt it.

marcus-oscarsson:
That would be great !

marcus-oscarsson:
I guess you would be the best person to hint to what needs to be done as you test the workflows regularly on both versions.  I assume that it would mean that we need a base directory with shared configuration files and then two separate folders with the specific configuration files for web/qt. I would also urge that we should try to curate those and try to keep those differences as small as possible.

rhfogh:
@marcus-oscarsson 
The PR #1250 actually does it already - almost all the config files are shared and there are only two small directories, mxcube-qt and mxcube-web with the separate stuff. Both web and qt versions work with that set-up when I test them. What is missing would be for other people to agree that this should be done and test that it also works for them, then we can adopt it and start removing all the alternative copies.

beteva:
Sorry to repeat what I've said in #1250, it rather belongs here.
Before discussing directories, we should consider what the use cases are. I think we have three different needs:

1. Testing and CI
2. Running mock GUI (web and Qt)
3. Having templates for each HO

- The templates are not necessarily used in neither test, nor mock GUI, but can help a lot when doing real beamline configuration.
- The mock GUI files should correspond to the templates for the Mockup object. There is one exception, though. For the Qt there are properties which are used by the GUI only, not the HO (namely unit and GUIstep), This is not the case for the web.
- The test files require maybe more that running mock GUI and may require some error producing parameters (as it is the case for the machine_info).
- For both testing and running GUI, there are configuration files which do not have Mockup classes (like for the queue), but these are the easiest to generalise, as they are the same in the three cases.


So we need to see how to answer all the thee needs with a minimum of duplication of information and risk of diverging in case of refurbishment. Personally I think we should provide template files at least for the Mockup classes and keep them for both testing and GUI and only have them duplicated if an extra configuration is needed.

rhfogh:
I'll repeat my answer too, then, for completeness.

[...]

PR #1263, opened 2025-06-09
move 'mockup' beamline configs to 'test' directory

elmjag:
The 'mockup' and 'mockup/test' directories are used by the tests only. The don't need to be included into the shiped package.

Move out these files from 'mxcubecore' folder, put then under 'test' instead.

This partially solves: https://github.com/mxcube/mxcubecore/issues/1226


test/conftest.py
```diff
@@ -30,29 +30,32 @@
 monkey.patch_all(thread=False)
 
 
-TESTS_DIR = os.path.abspath(os.path.dirname(__file__))
-ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+TESTS_DIR = Path(__file__).parent
+ROOT_DIR = TESTS_DIR.parent
```
fabcor-maxiv:
```suggestion
```

Seems to me like `ROOT_DIR` is not necessary any longer.

rhfogh (dismissed):
It is NOT CORRECT that the mockup directories are used only for tests. See Issue #1251 for a discussion. There are four use cases for mock configuration files: tests, templates, mxcubeweb demo beamline, and mxcubeqt demo beamline (both the latter being used for manual testing). Can we please close this PR and leave off with changing the mock config directories until we have agreed on a solution that takes care of all the use cases - again see #1251 for a discussion?

marcus-oscarsson:
I think this PR moves things in the right direction looking at #1251. its perhaps a bit to soon, should we split #1251 into a few action points. Would you rather see that we move the necessary files to the `mxcube_configuration` repository or some alternative solution ?

rhfogh:
@marcus-oscarsson Were you asking me, or @elmjag?
I would do things this way:
1) Agree where everything should live
2) Copy them there, and test
3) Delete the no longer necessary copies
Until we have done 1) (taking in account all the needs, not just some of them) we should not start on 2), let alone 3)

marcus-oscarsson:
I was asking anyone interested :)

marcus-oscarsson:
That seems like a good approach to me, by the way does GitHub being a collaborative tool supports polls in issues ?

rhfogh:
@marcus-oscarsson Could we take it up at a developers' meeting? When are we having one? Sounds like it might benefit from a discussion.

marcus-oscarsson:
Yes its perhaps better indeed. I was planing on sending out an invitation for the end of the month, but I can suggest some earlier dates as well.

test/conftest.py
elmjag: It's used just below, line 36.

https://github.com/mxcube/mxcubecore/pull/1263/files#diff-7dae78a00d8d69422a3bfbde96c7c5d794eede504a0df763853896a1ba1b5ff2R36

fabcor-maxiv: Indeed :D

fabcor-maxiv: Is this `sys.path.insert(0, ROOT_DIR)` necessary, though? It seems strange to me that such a thing would be necessary, but I have not tried.

rhfogh: Changing this would probably require refactoring imports. The result would be an improvement, but it is *not* a small job. This trick is used in about ten different places, BTW, though arguably we should get rid of it everywhere. The problem is: how do you make sure that mxcubecore (or other required directories) is actually on the PYTHONPATH, so you can import relative to it? IMHO the best solution would be to have a fixed, agreed directory structure, so that mxcubecore (etc.) had to be on the PYTHONPATH and write your imports with that in mind, but it seems that things have been written under different assumptions.

fabcor-maxiv: `pytest` seems to go through all the tests just fine without this line. I just tried. But maybe other people have different setups.

Maybe it is a remnant of some old workaround for an issue that does not exist any longer (either because pytest has evolved or because we fixed the directory structure)

rhfogh: Maybe - but would it not make sense to specify clearly what directory structures and PYTHONPATHS we are mandating, check that all imports support them, and get rid of this trick everywhere, instead of just cutting it off piecemeal? After all this like of code eats no bread - and for a proper cleanup we would need the bigger operation. Meanwhile I have no objection to just removing it from the test code - by definition it will be OK as long as the test still pass - it just does not accomplish that much by itself.

marcus-oscarsson:
As pointed out earlier we need to clarify/discuss a few bits on the next developers meeting. I think this moves things in the right direction and that we should not have configuration files (other than the ones for the tests) in the `mxcbecore` module. To be discussed of-course.

test/conftest.py
elmjag: > `pytest` seems to go through all the tests just fine without this line. I just tried. But maybe other people have different setups.
> 
> Maybe it is a remnant of some old workaround for an issue that does not exist any longer (either because pytest has evolved or because we fixed the directory structure)

Tests seem to run fine without `sys.path` manipulations, both locally for me and in github CI. I'm guessing this is "remnant of some old workaround".

I added a [commit](https://github.com/mxcube/mxcubecore/pull/1263/commits/229a620e4cf1fe882f2336f2cb2ae52e4e13766a) that removes this code.

elmjag:
As we agreed during developers meeting, I re-based this PR on top of latest version of `develop` branch. The tests run fine.

Let's merge it?

fabcor-maxiv (approved):

rhfogh:
As I remember it, we agreed during the developers' meeting to have three different configuration directories:

 - One in test, for the sole use of testing.
 - One somewhere else (core??) for example config files for abstract classes
 - One in a separate repository to hold the demo config files for web and qt.

Am I wrong here? 
 
As it stands, this PR does NOT conform to that agreement. It moves the mockup files into test, lock stock, and barrel, without adding any of the other locations. And in doing so it breaks, for instance, my test set-up.  The obvious way to proceed would be to make a proposal that conformed to our agreement, all parts of it. If this is too much work and you need a repository available in test in a great hurry, what about ADDING it, and leaving in place the stuff that other people are using until it is replaced?

Just what is going on here?

elmjag:
>  And in doing so it breaks, for instance, my test set-up. 

I was not aware that you had a test set-up that relied on files in `mxcubecore/mxcubecore/configuration/mockup`.

I don't want to make a copy of these configure files, as it will just the situation worse.

I'm not in a hurry, let's wait with this move until we have sorted out other types of configure files. I'll change this PR into draft for now.

rhfogh:
@elmjag Well, my standard test is to run through MXCuBE in Demo mode on my laptop (we have a image simulation program attached) and check if any changes to the workflow still work. Smoke test, if you like, but it does mean I check through the mock mode fairly regularly. I do that for either Web or Qt versions. That does require what amounts to a demo version for both Web and Qt, which is something we agreed we needed to have anyway. There is no populated configuration directory in Qt, so I am using the (old, established) one in mxcubecore. It would be preferable (though not absolutely mandatory) if the Web and Qt demo configuration shared most of their files (anyway we cannot run the Qt demo off the Web configuration directory). I would prefer to keep something checked-in to work on until we get to the final solution, but it does not matter too much what. If you think it is better to postpone the PR till we sort it all out that would be fine with me (we just need to sort out who has to do the work, I guess ;-) ).

marcus-oscarsson:
We can now start to make use of https://github.com/mxcube/mxcube_configuration. This PR should perhaps be remade to against the later ?. We can then remove these files from this repository.

marcus-oscarsson:
@elmjag, I think it would nice to merge this PR. Could you please rebase and solve the conflicts.

elmjag:
> We can now start to make use of https://github.com/mxcube/mxcube_configuration. This PR should perhaps be remade to against the later ?. We can then remove these files from this repository.

Sorry, I'm a bit out of the loop here. How do we want to remake this PR? How does the https://github.com/mxcube/mxcube_configuration comes into picture, regarding this PR?

elmjag:
> @elmjag, I think it would nice to merge this PR. Could you please rebase and solve the conflicts.

I have re-based, I did not run into any conflicts :sunglasses:

marcus-oscarsson:
Cool, I change it to ready for review and merge it then :)

marcus-oscarsson:
> Sorry, I'm a bit out of the loop here. How do we want to remake this PR? How does the https://github.com/mxcube/mxcube_configuration comes into picture, regarding this PR?

That's my mistake, `mxcube_configuration` does not concern this PR. We said that we keep the configuration files for the tests with the tests.

marcus-oscarsson (approved):

marcus-oscarsson:
@rhfogh requested changes in the past so he needs to re-approve. I otherwise need to force merge.

beteva:
I am a bit puzzled, as for me there should not be neither qt, nor web directories.
From the previous discussions I understood that in the mxcubecore repository there were only configuration files for the tests and for the mockup classes. It this true?

test/conftest.py
```diff
@@ -30,13 +29,14 @@
...
 
-TESTS_DIR = os.path.abspath(os.path.dirname(__file__))
-ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+TESTS_DIR = Path(__file__).parent
 
-sys.path.insert(0, ROOT_DIR)
 
-print("DEBUG TESTS")
-print(sys.path)
+def _get_hwr_paths():
```
marcus-oscarsson, line 35: @beteva: Its simply that as we said that we would remove the configuration files from the `mxcubecore` module and move the to the `mxcube_configuration` repository instead. Except for the tests, that would contain their own set of configuration files. Which makes sense as its test data and not the same as example or demo configuration. Its also more practical for the sake for running and maintaining the tests

marcus-oscarsson (approved):

PR #1276, opened 2025-06-11
Fixed issue in `get_value` and improved FluxMockup with simulated flux value 

marcus-oscarsson:
There was a small issue in `set_value`  in `FluxMockup`. The method `get_value` called `measure_flux` which in turn emitted a `valueChanged` signal/event. A problem arises when a handler listens to this event and also calls `get_value` (unnecessary one may think, but still an issue). The end result is a endless recursion. 

Removed the call to `measure_flux` from `get_value` and replaced it by a co-routine that updates the flux every 3 seconds.


fabcor-maxiv (approved):
I guess it's fine

Issue #1275, opened 2025-06-10
Potential issue in `lnls_sol/session.xml`


fabcor-maxiv:
https://github.com/mxcube/mxcubecore/blob/d35ffd76f7eeda402230a13c5329a7009d6fc85d/mxcubecore/configuration/lnls_sol/session.xml#L55

Maybe this should be `precision` instead of `precission`.

marcus-oscarsson:
Is this issue now solved ? 

Was it made by some script or in some other automated way ? Could that script be configured to group issues ?

fabcor-maxiv:
> Is this issue now solved ?

I do not know who to ping and assign to for LNLS.


> Was it made by some script or in some other automated way ? Could that script be configured to group issues ?

This is issue was opened manually.

Issue #1274, opened 2025-06-10
Potential issue in `lnls_sol/beamstop_alignment_x.xml`


fabcor-maxiv:
https://github.com/mxcube/mxcubecore/blob/d35ffd76f7eeda402230a13c5329a7009d6fc85d/mxcubecore/configuration/lnls_sol/beamstop_alignment_x.xml#L2

Maybe this should be `Alignment` instead of `Alignemnt`.

Issue #1273, opened 2025-06-10
Potential issue in `lnls_manaca/session.xml`


fabcor-maxiv:
https://github.com/mxcube/mxcubecore/blob/d35ffd76f7eeda402230a13c5329a7009d6fc85d/mxcubecore/configuration/lnls_manaca/session.xml#L55

Maybe this should be `precision` instead of `precission`.

Issue #1271, opened 2025-06-10
Possible bug in `BIOMAXEigerMockup`


fabcor-maxiv:
https://github.com/mxcube/mxcubecore/blob/a6fa608e74cde42c6d30a884f96bcca7792c21ee/mxcubecore/HardwareObjects/mockup/BIOMAXEigerMockup.py#L57

Maybe this should be `disabled` instead of `dsiabled`.

elmjag:
Wait, what is `BIOMAXEigerMockup.py` ? What is it used for? 

I don't see any references to that class anywhere in mxcubecore nor mxcubeweb. I think we should remove that file.

fabcor-maxiv:
> Wait, what is `BIOMAXEigerMockup.py` ? What is it used for?

No idea.


> I think we should remove that file.

You are probably right. Let's get rid of it!

elmjag:
Here is PR https://github.com/mxcube/mxcubecore/pull/1280 that removes `BIOMAXEigerMockup.py`.

elmjag:
The file have been removed.

PR #1279, opened 2025-06-11
Fix some spelling errors

fabcor-maxiv:
Fix some spelling errors in documentation, docstrings, and comments.


fabcor-maxiv:
I used this tool: <https://pypi.org/project/codespell/>.

As the commit message says, I tried to do only things that are not actual code: documentation, comments, and docstrings. Because I did not want to run the risk of breaking anything. In particular there are a lot of strings like log messages, and exception messages that have spelling errors but I did not fix them. Who knows?... maybe someone relies on the spelling errors in the messages for grepping or something like this...

fabcor-maxiv:
I am considering adding the tool to our toolchain somehow (pre-commit or not), but the exclusion list is very long for now, so in a way it might not be worth it just yet. I will think about it...

fabcor-maxiv:
I also ignored the `.nml` files completely because I do not understand the syntax and I do not know what is a comment or not.

rhfogh:
@fabcor-maxiv This is good work, and appreciated, but it was probably a wise decision not to spend too much time on things that might break when all you are dealing with the spelling in comments and messages.  Probably wise, too, to ignore the nml file - anyway they are too obscure and GPhL specific to be worth the trouble.

rhfogh (approved):
I only looked at my code, but this is so obviously correct that I am sure it generalises to the rest of the changes

fabcor-maxiv:
Another note, the tool does not catch all spelling errors. I did see spelling errors that the tool did not catch and managed to fix some of those manually.

fabcor-maxiv:
Ah and yes, running this tool helped spot some potential issues (in code and/or configuration):

* https://github.com/mxcube/mxcubecore/issues/1267
* https://github.com/mxcube/mxcubecore/issues/1268
* https://github.com/mxcube/mxcubecore/issues/1269
* https://github.com/mxcube/mxcubecore/issues/1270
* https://github.com/mxcube/mxcubecore/issues/1271
* https://github.com/mxcube/mxcubecore/issues/1272
* https://github.com/mxcube/mxcubecore/issues/1273
* https://github.com/mxcube/mxcubecore/issues/1274
* https://github.com/mxcube/mxcubecore/issues/1275

marcus-oscarsson:
Thanks, for me the most useful would be if there was some nice tool that integrated well with the editor, ie vscode.

marcus-oscarsson:
I would probably advice to not add this sort of tool the pipeline unless it functions very well

fabcor-maxiv:
> Thanks, for me the most useful would be if there was some nice tool that integrated well with the editor, ie vscode.

I do not use IDEs, so I am not the best one to give such recommendations. You can try to check if [codespell](https://github.com/codespell-project/codespell) has the integration you need (this is the tool I used for this pull request, I picked it because it is a Python tool). But I also recommend having a look at [typos](https://github.com/crate-ci/typos) which I have tried a bit and seems good as well, it seems more full-featured than codespell.


> I would probably advice to not add this sort of tool the pipeline unless it functions very well

Any particular reason?

PR #1282, opened 2025-06-11
Documents new `ui:options.group` property

mockoocy:
The `TestCollection` queue entry is enhanced to both use `ui:options.col` and `ui:options.group` settings to provide a reference for other implementations of custom tasks.

This property is added in: https://github.com/mxcube/mxcubeweb/pull/1678 - I mark this as draft until that one is merged.


marcus-oscarsson:
Cool !

marcus-oscarsson (approved):

PR #1287, opened 2025-06-16
Remove unused imports

fabcor-maxiv:
These were uncovered by Ruff rule `F401:unused-import`.

### DIFF
--- a/mxcubecore/HardwareObjects/QtGraphicsManager.py
+++ b/mxcubecore/HardwareObjects/QtGraphicsManager.py
@@ -41,12 +41,6 @@ from __future__ import print_function
 import logging
 import math
 import os
-
-try:
-    import cPickle as pickle
-except Exception:
-    import _pickle as pickle
-
 from copy import deepcopy
 from datetime import datetime
 
@@ -59,14 +53,6 @@ from scipy import (
     ndimage,
     signal,
 )
-
-try:
-    import lucid2 as lucid
-except ImportError:
-    try:
-        import lucid
-    except ImportError:
-        pass
 
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore.HardwareObjects import QtGraphicsLib as GraphicsLib
--- a/mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Qt4_LimaVideo.py
+++ b/mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Qt4_LimaVideo.py
@@ -44,26 +44,11 @@ Example Hardware Object XML file :
 </object>
 """
 
-import logging
-import os
-import struct
-import time
-
 from mxcubecore.HardwareObjects.QtLimaVideo import QtLimaVideo
 
 try:
     from Lima import Core
 except Exception:
-    pass
-
-try:
-    from Lima import Prosilica
-except ImportError as e:
-    pass
-
-try:
-    from Lima import Basler
-except ImportError as e:
     pass
 
 
--- a/mxcubecore/dispatcher.py
+++ b/mxcubecore/dispatcher.py
@@ -5,7 +5,10 @@ try:
 
     louie = 1
 except ImportError:
-    from pydispatch import dispatcher  # pyright: ignore[reportMissingImports]
+    # `dispatcher` is imported here but unused, on purpose.
+    # It is here so that one can `from mxcubecore.dispatcher import dispatcher`.
+    # Maybe this can be solved differently, though.
+    from pydispatch import dispatcher  # pyright: ignore[reportMissingImports]  # noqa: F401
     from pydispatch import robustapply  # pyright: ignore[reportMissingImports]
     from pydispatch import saferef  # pyright: ignore[reportMissingImports]
 
--- a/mxcubecore/utils/qt_import.py
+++ b/mxcubecore/utils/qt_import.py
@@ -1,4 +1,6 @@
 # pylint: skip-file
+# ruff: noqa: F401  # This file is full of "unused imports" by design
+
 """
 
 Description
--- a/mxcubecore/Command/Sardana.py
+++ b/mxcubecore/Command/Sardana.py
@@ -58,7 +58,6 @@ except Exception:
 
 try:
     import taurus
-    from sardana.taurus.core.tango.sardana import registerExtensions
     from taurus import (
         Attribute,
         Device,
--- a/mxcubecore/HardwareObjects/ALBA/ALBAAutoProcessing.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBAAutoProcessing.py
@@ -6,7 +6,6 @@ from datetime import datetime
 
 from ALBAClusterJob import ALBAEdnaProcJob
 from PyTango import DeviceProxy
-from xaloc import XalocJob
 from XSDataAutoprocv1_0 import XSDataAutoprocInput
 from XSDataCommon import (
     XSDataFile,
--- a/mxcubecore/HardwareObjects/ALBA/ALBABackLight.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBABackLight.py
@@ -3,7 +3,6 @@ import time
 
 import gevent
 
-from mxcubecore import HardwareRepository as HWR
 from mxcubecore.BaseHardwareObjects import HardwareObject
 
 
--- a/mxcubecore/HardwareObjects/ALBA/ALBADataAnalysis.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBADataAnalysis.py
@@ -3,7 +3,6 @@ import os
 import sys
 import time
 
-from PyTango import DeviceProxy
 from xaloc import XalocJob
 from XSDataCommon import (
     XSDataFile,
--- a/mxcubecore/HardwareObjects/ALBA/ALBAEpsActuator.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBAEpsActuator.py
@@ -45,7 +45,6 @@ Public Interface:
 import logging
 
 from mxcubecore import BaseHardwareObjects
-from mxcubecore import HardwareRepository as HWR
 
 STATE_OUT, STATE_IN, STATE_MOVING, STATE_FAULT, STATE_ALARM, STATE_UNKNOWN = (
     0,
--- a/mxcubecore/HardwareObjects/ALBA/ALBAFastShutter.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBAFastShutter.py
@@ -43,10 +43,8 @@ Public Interface:
 """
 
 import logging
-import time
 
 from mxcubecore import BaseHardwareObjects
-from mxcubecore import HardwareRepository as HWR
 
 STATE_OUT, STATE_IN, STATE_MOVING, STATE_FAULT, STATE_ALARM, STATE_UNKNOWN = (
     0,
--- a/mxcubecore/HardwareObjects/ALBA/ALBAFlux.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBAFlux.py
@@ -2,7 +2,6 @@ import logging
 
 import taurus
 
-from mxcubecore.BaseHardwareObjects import HardwareObject
 from mxcubecore.HardwareObjects.abstract import AbstractFlux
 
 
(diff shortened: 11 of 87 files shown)


mxcubecore/HardwareObjects/QtGraphicsManager.py
```diff
@@ -41,12 +41,6 @@
 import logging
 import math
 import os
-
-try:
-    import cPickle as pickle
-except Exception:
-    import _pickle as pickle
-
```
fabcor-maxiv, line 49: Not 100% sure about this

mxcubecore/utils/qt_import.py
fabcor-maxiv, line 1: The situation in this file is unusual. I guess the imports have to stay even though they are not used in this file itself.

Is it really necessary to keep all those imports? Or is there a chance that it is an old workaround that we can get rid of now?

mxcubecore/dispatcher.py
```diff
@@ -5,7 +5,6 @@
 
     louie = 1
 except ImportError:
-    from pydispatch import dispatcher  # pyright: ignore[reportMissingImports]
```
fabcor-maxiv, line 8: I am not too sure about this one.

mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Qt4_LimaVideo.py
```diff
@@ -44,28 +44,13 @@
...
-try:
-    from Lima import Prosilica
-except ImportError as e:
-    pass
-
-try:
-    from Lima import Basler
-except ImportError as e:
-    pass
-
```
fabcor-maxiv, line 68: I don't know... is this okay?

rhfogh:
Looks good (and thanks for doing all the work). I only looked at bits relevant to me, but they all look good

Just two questions:

- in qt_import, did you either test or make sure that the resulting imports are the same (I assume the latter).

- Wrt. dispatcher how did you determine that dispatcher could eb removed from e.g. mxcubecore/dispatcher.py (despite being marked as 'ignore') while keeping - here and in other places - saferef and robustapply? Is it not dangerous to remove the alternative import from dispatcher, when other files import from this one? Is there something special going o here? It looked to me like even more could be removed (from e.g. GphlWorkflowConnection) and I thought of offering o do it, but it would be good to know first if there are any traps here.

fabcor-maxiv:
@rhfogh, you were a bit too fast on the review :D

I updated the pull request with some review comments on one hand. And on the other hand I also updated the change itself with things I am even less sure about, and made it so that Ruff complains about them (see the log of the failing `pre-commit` job).

mxcubecore/utils/qt_import.py
rhfogh: Looks like we noticed the same things and our comments crossed. With Qt I would be tempted to leave bad enough alone. This all depends on which Qt version people have, and those tend to live in a number of places that are not being regularly updated. The time to clean this up would be after we had a clear agreement on which Qt versions were supported and which were no longer being used (or would never reconnect to develop, at least). And the way developments are going I just cannot see that this is likely to happen any time soon. If there was an actual problem we might consider it, but it is not worth it just to keep the  linter happy, IMHO.

fabcor-maxiv:
The `dispatcher` stuff is among those things that looked suspicious to me from the beginning. And indeed the tests fail because `from mxcubecore.dispatcher import dispatcher` does not work any longer.

fabcor-maxiv:
> * in qt_import, did you either test or make sure that the resulting imports are the same (I assume the latter).

I did not test anything. I assumed those imports should stay. ~I told Ruff to add the `noqa` instructions automatically, so that we know those are false positive cases of "unused import".~ The Ruff rule `F401` is now ignored for the whole `qt_import.py` file.

mxcubecore/utils/qt_import.py
fabcor-maxiv: OK. I might change this to ignore `F401` on the whole file, instead of the `noqa` on each line. Since the whole point of the file is to deal with import shenanigans anyway, we might as well ignore the whole file.

rhfogh:
I think that both dispatcher and qt_import should be left alone. Both are cases where there are alternative import situations for different availabilities, and it just does not make sense to clean them up partially,. Either we make sure that they are no longer needed and remove them (or at least prune them, for qt_import) and that is a big job, that requires analysis of the code and of what is available where in which versions.  Or we leave it till later, possibly adding noqa (as you did) - or even better do it for the whole file, as you proposed.

Interesting that we both zeroed i on the same cases as problematic, BTW.

mxcubecore/utils/qt_import.py
fabcor-maxiv: Done, `F401` is ignored for the whole `qt_import.py` file now.

mxcubecore/dispatcher.py
fabcor-maxiv: Change reverted and `F401` is explicitly ignored for that line. The tests pass again.

fabcor-maxiv:
I need advice for these two:

```none
mxcubecore/HardwareObjects/Gphl/GphlWorkflowConnection.py:78:27: F401 `louie.dispatcher` imported but unused; consider using `importlib.util.find_spec` to test for availability
   |
76 | except ImportError:
77 |     try:
78 |         from louie import dispatcher
   |                           ^^^^^^^^^^ F401
79 |     except ImportError:
80 |         from pydispatch import (
   |
   = help: Remove unused import: `louie.dispatcher`

mxcubecore/HardwareObjects/QtGraphicsManager.py:61:16: F401 `lucid` imported but unused; consider using `importlib.util.find_spec` to test for availability
   |
59 | except ImportError:
60 |     try:
61 |         import lucid
   |                ^^^^^ F401
62 |     except ImportError:
63 |         pass
   |
   = help: Remove unused import: `lucid`
```

I now guess that `from louie import dispatcher` should be kept as-is.

What about `import lucid`?

rhfogh:
For the lucid imports, short-term, I would move it down to line 1936, in the commented-out part of the auto_grid_procedure function.  That is the only place where it is used, and it gets it out of the purview of ruff. Longer term that code section should probably be removed, but I do not think there is anyone who is sufficiently on top of the Qt version and has the time to do a thoroughgoing refactoring and clean-up. I do not know who would know anything about the auto_grid_procedure (Martin Savko??), but personally I do not even know what it does. 

For the GphlWorkflowConnection I think you should just put in enough noqa statements to keep ruff quiet - and then leave it to me. I suspect that all the dispatcher imports can probably be removed from that file, and when my latest PR with the yaml configuration files is sorted out one way or the other, I promise I shall  remove the dispatcher stuff, test the result, and end up with a cleaner file one way or the other. Thanks for pointing it out, BTW - I do need to keep my code clean.

mxcubecore/HardwareObjects/QtGraphicsManager.py
walesch-yan: I don't know anything about the Qt part but looking at the file I would suggest to do the same as @rhfogh suggested with the lucid part, as it seems to only be used in commented out code

fabcor-maxiv: The code that imports `pickle` seems to have been commented out for quite some time, 5 years as far as I can tell. That commented out code should be removed. The import as well. This is not code that anyone should have to maintain, we should not inflict ourselves more pain than necessary.

Or am I misunderstanding something here? It is quite hard for me to understand the history of that file.

rhfogh: It is hard for me too, which is why I was not eager to get into cleaning it up, and preferred to leave it to someone who at least understood what the function was supposed to be doing. But you are probably right. For sure no one can complain about losing five-years-old commented-out code.

fabcor-maxiv:
Ready for review, from my side.

@marcus-oscarsson, you might want to take a look as well (including the "resolved" comment threads).

rhfogh (approved):
Looks good - though I still looked only at the ones we have been discussing.

mxcubecore/HardwareObjects/QtGraphicsManager.py
fabcor-maxiv: As far as I can tell, this is the commit that commented out the code that used to import those things: <https://github.com/mxcube/mxcubecore/commit/2c2454456a10b9db8f9fec7626471bd68dab3ec9>.

rhfogh: It does look pretty much like ancient history, yes.

marcus-oscarsson:
I will have a look tomorrow :)

mxcubecore/dispatcher.py
```diff
@@ -5,7 +5,10 @@
...
```
marcus-oscarsson, line 8: The try except is no longer needed. `louie` is no longer used we entirely rely on `pydispatch`.

There is an entire subject/issue to look at `pydispatch`, and the use of `safeRef` and the definition of `__my_robust_apply` that can give a better error message. (not related to this PR)

mxcubecore/HardwareObjects/Gphl/GphlWorkflowConnection.py
```diff
@@ -78,7 +78,7 @@
         from louie import dispatcher
     except ImportError:
         from pydispatch import (
-            dispatcher,
+            dispatcher,  # noqa: F401  # Unclear if safe to remove
```
marcus-oscarsson: This code block looks strange, 

`from mxcubecore.dispatcher import dispatcher` already does `from pydispatch import dispatcher` so it would be weird if `ImportError` is raised here. Also the code below duplicates whats already done in `mxcubecore.dispatcher`.

I think the entire `try ... except`  should be replaced with from mxcubecore.dispatcher import dispatcher. 

It should otherwise, If that for some reason does not work, be investigated/commented why its not the case

rhfogh: @marcus-oscarsson I think the entire import can probably be removed. As I suggested to Fabien I shall try and remove it myself, only I would rather wait till the yaml configuration PR can be resolved, one way or the other, so I can merge my local branch into the main branch again (my mock tests are not working with the main branch for now). So could we just leave this alone for the moment? I promise this code block will not survive long.

marcus-oscarsson: Ok

mxcubecore/utils/qt_import.py
marcus-oscarsson: As a general remark. We have said in the past that we would like move all code with dependencies on qt from `mxcubecore` to `mxcubeqt`.

marcus-oscarsson:
Looks good as far as I can see. I can not give any guarantees on the site specific bits especially the `lima` imports for PX2.

mxcubecore/dispatcher.py
fabcor-maxiv: OK, I guess it is okay to do this in a separate pull request then.

marcus-oscarsson:
Could you please rebase and merge  @fabcor-maxiv  ?

marcus-oscarsson:
@fabcor-maxiv, Could you please rebase and merge this one ?

marcus-oscarsson:
This PR has been around for some time and there are now some conflicts. @fabcor-maxiv, Ill let you have a look or close it if you prefer to do in some other way ?

fabcor-maxiv:
Yes, I will try to re-do it, today or early next week, hopefully.

marcus-oscarsson:
Thanks, thats great (no urgency though :) ) !

mxcubecore/HardwareObjects/Gphl/GphlWorkflowConnection.py
rhfogh: You can skip this change now. I have removed the entire block of imports as unnecessary in my latest PR.

fabcor-maxiv: OK, thanks, done

fabcor-maxiv:
@marcus-oscarsson I recreated the pull request from scratch. I kept the exclusions we had decided on.

The "Bandit and SonarQube" workflow fails, likely because I created this pull request from my fork.

marcus-oscarsson:
Thanks, sorry for taking some time to review. I've been on vacation. just back ...

PR #1291, opened 2025-06-18
Make phase control optional

alessandro-olivo:
This change is needed to use the feature from this other PR https://github.com/mxcube/mxcubeweb/pull/1699 in `mxcubeweb`


marcus-oscarsson:
So I guess this change is not longer needed ?

alessandro-olivo:
Just a few thoughts about this PR here:  [#1699](https://github.com/mxcube/mxcubeweb/pull/1699#issuecomment-3007970012)

marcus-oscarsson:
Ok, feel free to reopen if still needed :). I guess we are going for a UI based solution right ?

alessandro-olivo:
Yes, I think the UI based solution is the better option

PR #1296, opened 2025-06-24
ISPyB: handle `loginTranslate` property correctly

elmjag:
Fix an issue with loading `loginTranslate` configuration property of ISPyB LIMS hardware objects.

Load boolean `False` value correctly from config file.  Without this change, if the config file specified boolean false value, the `self.loginTranslate` was set to `True`.

For example, following config file did not work correctly:

    class: MAXIV.ISPyBLims.ISPyBLims
    configuration:
      base_url: https://ispyb.maxiv.lu.se
      loginTranslate: false


dominikatrojanowska:
LGTM :)

mockoocy (approved):

PR #1297, opened 2025-06-25
Make "detector mode" in the task dialogs configurable from the backend

alessandro-olivo:
This PR is required for PR [#1723](https://github.com/mxcube/mxcubeweb/pull/1723). You can find more details in that one

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -24,6 +24,7 @@ Any object that inherits from TaskNode can be added to and handled by
 the QueueModel.
 """
 
+import ast
 import copy
 import logging
 import os
@@ -1705,6 +1706,7 @@ class AcquisitionParameters(object):
         self.skip_existing_images = False
         self.detector_binning_mode = str()
         self.detector_roi_mode = str()
+        self.detector_mode_list = self.get_detector_mode_list()
         self.induce_burn = False
         self.mesh_range = ()
         self.cell_counting = "zig-zag"
@@ -1753,6 +1755,7 @@ class AcquisitionParameters(object):
             "skip_existing_images": self.skip_existing_images,
             "detector_binning_mode": self.detector_binning_mode,
             "detector_roi_mode": self.detector_roi_mode,
+            "detector_mode_list": self.detector_mode_list,
             "induce_burn": self.induce_burn,
             "mesh_range": self.mesh_range,
             "mesh_snapshot": self.mesh_snapshot,
@@ -1770,6 +1773,11 @@ class AcquisitionParameters(object):
 
     def copy(self):
         return copy.deepcopy(self)
+
+    def get_detector_mode_list(self):
+        return ast.literal_eval(
+            HWR.beamline.detector.get_property("roi_mode_list", "[]")
+        )
 
 
 class XrayImagingParameters(object):


marcus-oscarsson:
I see, so this is for the detector ROI modes ? How are you planning on using this and did you notice that there is a `detector_roi_mode` attributes ?

alessandro-olivo:
@marcus-oscarsson The thing was, the beamline scientist didn’t want the detector mode to be displayed at all, so I made this change that allows handling that field from the backend

beteva:
Simple curiosity - what does this detector modes correspond to? Is the naming a particular detector specific or is something generic for all the detectors?

alessandro-olivo:
@beteva I added this `detector_modes` just to move this [list](https://github.com/mxcube/mxcubeweb/blob/14e5cb2273e1c6f7758a96526ec6fd8e15091603/ui/src/components/Tasks/DataCollection.jsx#L287) hardcoded in the frontend to the backend. 
This just to have the possibility to hide the below dropdown, which is meaningless for us.
<br>
image

Unfortunately, I have no clue what these codes `['0', 'C18', 'C12', 'C2']` stand for

More details here: https://github.com/mxcube/mxcubeweb/pull/1723

mxcubecore/model/queue_model_objects.py
```diff
@@ -1715,7 +1715,7 @@ def __init__(self):
         self.in_queue = False
         self.in_interleave = None
         self.sub_wedge_size = 10
-
+        self.detector_mode_list = ["0", "C18", "C12", "C2"]
```
marcus-oscarsson: I think this could be `HWR.beamline.detector.get_property("roi_mode_list")` Like this we read the available ROI modes from the detector config file ?

alessandro-olivo: It makes sense, :) I'll change it!

alessandro-olivo: @marcus-oscarsson  I'm getting back to this, in case the "roi_mode_list" is not defined in the configuration file should we put this list `["0", "C18", "C12", "C2"]` as default or an empty list ?

I'm getting back to this, in case "roi_mode_list" is not defined in the configuration file, should we use `["0", "C18", "C12", "C2"]` or an empty list as the default value?

1) `HWR.beamline.detector.get_property("roi_mode_list", ["0", "C18", "C12", "C2"])`
2) `HWR.beamline.detector.get_property("roi_mode_list", [])`

marcus-oscarsson: I think that `["0", "C18", "C12", "C2"]` us used by most Dectris detectors so I think that is somehow a sensible default ?

alessandro-olivo: Furthermore, the list ["0", "C18", "C12", "C2"] was hardcoded in the UI before these changes, so I don't think anyone will complain :)

mxcubecore/model/queue_model_objects.py
```diff
@@ -1705,6 +1705,9 @@ def __init__(self):
         self.skip_existing_images = False
         self.detector_binning_mode = str()
         self.detector_roi_mode = str()
+        self.detector_mode_list = HWR.beamline.detector.get_property(
+            "roi_mode_list", ["0", "C18", "C12", "C2"]
+        )
```
beteva: I think it is better to give an empty list as a default value and not have Dectris specific values in the queue. The above values should be set as a default valuer somewhere in the detector hardware object.

> Furthermore, the list ["0", "C18", "C12", "C2"] was hardcoded in the UI before these changes, so I don't think anyone will complain :)

 Certainly, but as we change the code, why not make it cleaner 😃

alessandro-olivo: That's a good point too! 😄
My concern was about any sites (if there are any) that might expect that list to automatically show up in the frontend without needing any configuration.
That said, I also updated the demo files to make it clear (hopefully) how to configure the `detector_mode_list`.
1) https://github.com/ELETTRA-SincrotroneTrieste/mxcubeweb/blob/809f61236377ee6f34ec1649ba5667c42ab8e93d/demo/detector.xml#L16
2) https://github.com/ELETTRA-SincrotroneTrieste/mxcubeweb/blob/809f61236377ee6f34ec1649ba5667c42ab8e93d/demo/beamline_config.yml#L126

marcus-oscarsson:
Great !, thanks ! Please feel free to rebase and merge

mxcubecore/model/queue_model_objects.py
alessandro-olivo: @marcus-oscarsson About this ? Should I rebase and merge as it is, or should I use "[]" as default ?
No problem for me...

alessandro-olivo:
So we go with ['0', 'C18', 'C12', 'C2'] instead of an empty list? Sorry to bother you, @marcus-oscarsson, I just want to make sure I understood correctly. :)

marcus-oscarsson:
No problem at all, always better to double check :) 

It seems like @beteva was more for the empty list [] and thats a more neutral position so lets go with empty list then :)

marcus-oscarsson:
Wonderful, it seems like there is a conflict (I guess you noticed). Feel free to rebase and merge once the conflicts is resolved

alessandro-olivo:
@marcus-oscarsson Fixed it! Just need a review before I can merge :)

alessandro-olivo:
@marcus-oscarsson Just rebased ;)

marcus-oscarsson (approved):

Issue #1298, opened 2025-06-26
Outdated docs section `Accessing child objects`


elmjag:
The documentation section [Accessing child objects](https://mxcubecore.readthedocs.io/en/stable/dev/configuration_files.html#accessing-child-objects) claims that:

> "child objects can be accessed via the parent object’s `role_name` attribute"

As far as I understand, it does not work like that anymore. The documentation needs to be updated, to reflect how access to child objects works now.

elmjag:
I can make a PR to update the documentation.

However, I must admit that I have lost the plot on how it works now. If @marcus-oscarsson, @rhfogh or someone else who knows can tell me, I'll work on updating documentation.

marcus-oscarsson:
You can use either `self.get_object_by_role` or get the object from the dictionary `self._hwobj_by_role`

marcus-oscarsson:
The title **Accessing child objects** is perhaps a bit confusing, its not really what it is but I guess that its somehow close enough ?

rhfogh:
@marcus-oscarsson 
PLEASE!!! You are saying that the only official ways to get an object are through functions that 1) always return something, even if it is only None. 2) does not give you ANY way to see in  the code (or your linter) if the object you are trying to access is legal, misspelled or undefined in this context. Is that really all we have got???

At some point we were talking about defining child objects as properties, and having get_object_by_role as an additional, useful function. When was that dropped by the wayside? Considering the great emphasis lately on testability and rigour I am surprised at this decision.

marcus-oscarsson:
Well, the only way to access the "role" objects are via `self.get_object_by_role` or via `self._hwobj_by_role`. 

In addition to that we said that a property should be added to access that object if its is part of the public API. It means that all "roles" that are part of the outward facing API should be made available via a property.

But I interpreted this question how do we access "children"/"roles" of the current object, i.e. via self dot (`self.`) something.

**Edit:** This should perhaps also be added to the documentation.

marcus-oscarsson:
We have not really decided how to handle internal "role" attributes. The idea was that the entire self._hwobj_by_role could be a pydantic model and hence typed.

rhfogh:
[...]

PR #1299, opened 2025-06-26
docs: update 'Configuration files' section

elmjag:
Update the text on how to access child hardware objects.

Remove text on using `role_name` attribute, as it is no longer supported. Add text on using `get_object_by_role()` method instead.

Solves: https://github.com/mxcube/mxcubecore/issues/1298

### DIFF
--- a/docs/source/dev/configuration_files.md
+++ b/docs/source/dev/configuration_files.md
@@ -106,7 +106,6 @@ If access is required by outside code, then it should be provided by the object'
 ### Accessing child objects
 
 The `objects` dictionary in the YAML configuration file specifies the child objects.
-These child objects can be accessed via the parent object's `role_name` attribute.
 For example, if a hardware object is configured with this configuration file:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


docs/source/dev/configuration_files.md
```diff
@@ -117,7 +116,13 @@ objects:
 
 An instance of the `Foo` class will be created.
 This instance will have a child object with `bar` role, with configuration from `gazonk.yaml` file.
-That child object will be accessible with `foo.bar` python expression, where `foo` is the parent object.
+That child object is accessible using `get_object_by_role()` method of the parent object.
+The `get_object_by_role()` method requires `role` argument.
+For example, if `foo` is an instance of `Foo` class, it's `bar` object can be fetched as follows:
```
fabcor-maxiv: 1. Typo:

    ```suggestion
    For example, if `foo` is an instance of `Foo` class, its `bar` object can be fetched as follows:
    ```

3. Is it relevant here that "_`foo` is an instance of `Foo` class_"?

elmjag: Fixed typo.

Regarding point 2. I feel the example will be confusing, if it's not clear that `foo` is an `Foo` hardware object.

marcus-oscarsson:
Ok, This section probably then needs some more clarification regarding a "property" convention (as @rhfogh  mentioned) and possibly a warning that `get_object_by_role` will become "protected" . This is a clear improvement over how it was before as what was documented before was not working. So I think we can merge this as it is and then follow up with another PR.

marcus-oscarsson (approved):

docs/source/dev/configuration_files.md
fabcor-maxiv: Then say that `foo` is an hardware object.

PR #1301, opened 2025-06-30
Remove __getattr__ from `CommandContainer`

marcus-oscarsson:
Removing `__getattr__` from`CommandContainer`. The method `get_command_object` does the same thing and does not alter the standard python function `getattr`.


fabcor-maxiv (approved):
I guess it is fine

PR #1303, opened 2025-07-01
Read exposure time limits in `AbstractDetector.__init__`

marcus-oscarsson:
Read exposure time limits in `AbstractDetector.__init__`

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractDetector.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractDetector.py
@@ -50,6 +50,7 @@ Hardware objects used: energy
 
 import abc
 import ast
+import logging
 import math
 
 from mxcubecore.BaseHardwareObjects import HardwareObject
@@ -70,7 +71,6 @@ class AbstractDetector(HardwareObject):
         self._temperature = None
         self._humidity = None
         self._actual_frame_rate = None
-        self._exposure_time_limits = (None, None)
         self._pixel_size = (None, None)
         self._binning_mode = None
         self._roi_mode = 0
@@ -79,6 +79,29 @@ class AbstractDetector(HardwareObject):
 
         self._threshold_energy = None
         self._distance_motor_hwobj = None
+
+        min_exp_time = self.get_property("minimum_exposure_time", None)
+        max_exp_time = self.get_property("maximum_exposure_time", None)
+
+        if min_exp_time is None:
+            logging.getLogger("HWR").warning(
+                "Minimum exposure time not set for detector, using None"
+            )
+        else:
+            min_exp_time = float(min_exp_time)
+
+        if max_exp_time is None:
+            logging.getLogger("HWR").warning(
+                "Maximum exposure time not set for detector, using None"
+            )
+        else:
+            max_exp_time = float(max_exp_time)
+
+        self._exposure_time_limits = (
+            min_exp_time,
+            max_exp_time,
+        )
+
         self._width = None  # [pixel]
         self._height = None  # [pixel]
         self._metadata = {}


mxcubecore/HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -79,6 +79,11 @@ def __init__(self, name):
 
         self._threshold_energy = None
         self._distance_motor_hwobj = None
+        self._exposure_time_limits = (
+            float(self.get_property("minimum_exposure_time", 0.005)),
+            None,
+        )
```
walesch-yan: Should as well remove [line 73](https://github.com/mxcube/mxcubecore/blob/ea87df86d034851e8138f9a7ef2ce0242fb35c67/mxcubecore/HardwareObjects/abstract/AbstractDetector.py#L73) then :)

walesch-yan (approved):

rhfogh (changes requested):
Is it not dangerous to hardcode a fixed minimum exposure time in an abstract class - and particularly to use a float literal to do it? How do we know which detectors have 5ms as the correct lower limit? Should this not be always in configuration anyway?

rhfogh:
Sorry to be difficult, but what is the logic behind this? If the lower limit is None it is clear that it is not set and you need to deal with the problem yourself. If it is set to 0.0 the obvious interpretation is that there is no lower limit, and any positive float value is acceptable. The sensible solution would be to set the value in configuration, if you need a value. What are we trying to achieve here?

marcus-oscarsson:
@rhfogh, yes you are right it was a bit arbitrarily chosen. I'm setting it to 0 if not configured now. I agree that it should probably be even raised as a validation error but at the moment I'm logging it as a warning as it was not handled as an exception previously.

marcus-oscarsson:
@rhfogh, I think you are commenting between commits somehow. We can default the minimum exposure time to None if its not configured instead of 0 if that makes more sense to you ?

rhfogh:
Yes, I'd say that None was a better default value. That forces people to think about what to do when the value is missing, instead of just doing `if exposure_time > exposure_time_limits[0]` and assuming that would be OK. Unless there is some specific reason not to do that, of course.. 
`

mxcubecore/HardwareObjects/abstract/AbstractDetector.py
```diff
@@ -79,6 +79,21 @@ def __init__(self, name):
 
         self._threshold_energy = None
         self._distance_motor_hwobj = None
+
```
rhfogh, line 82: This looks fine. But - is there no maximum_exposure_time any more (there used to be)? Or if there is, how is it supposed to be handled or configured?

marcus-oscarsson: Well as you understand this code is used to retrieve the lowest possible exposure time for for instance strategy calculation. There has never been any need to get the longest possible exposure time (for understandable reasons). 

Ill add the maximum exposure time for completeness but its was never there as far as I know

rhfogh: @marcus-oscarsson As I understood it, this was also used to validate exposure time settings coming in from the GUI. I note that you will find both upper and lowe limites in the current mxcubecore in 
SOLEIL/PX!/PX!Pilatus.py line 64
mxcubecore/conriguration/soleil_px1/pilatus.xml line 20
ALBA/ALBAPilatus.py liine 75
mxcubecore/conriguration/alba_xaloc13/pilatus.xml line 18

It may not be in use any more (in  which case we should maybe no  longer have a tuple of limits), but if it still supported I thought that should be reflected in the abstract class.

PR #1264, opened 2025-06-09
Ruff: C408 unnecessary-collection-call

dominikatrojanowska:
Getting rid of ruff C408 rule.
Related issue: https://github.com/mxcube/mxcubecore/issues/1230

PR #1304, opened 2025-07-01
Reverted previous change from dict to {}

marcus-oscarsson:
Reverted previous change from `dict` to `{}`. This was a change made in https://github.com/mxcube/mxcubecore/pull/1264, fixing/satisfying C408. 

The automatic fix proposed by Ruff does not seem to take into account that the `dict` function can take a tuple and create the dictionary from that. While `{}` won't do this.

### DIFF
--- a/mxcubecore/HardwareObjects/sample_centring.py
+++ b/mxcubecore/HardwareObjects/sample_centring.py
@@ -91,7 +91,7 @@ def prepare(centring_motors_dict):
     move_motors(motors_to_move)
 
     SAVED_INITIAL_POSITIONS = {
-        [(m.motor, m.motor.get_value()) for m in centring_motors_dict.values()]
+        m.motor: m.motor.get_value() for m in centring_motors_dict.values()
     }
 
     phi = centring_motors_dict["phi"]
--- a/mxcubecore/HardwareObjects/EdnaWorkflow.py
+++ b/mxcubecore/HardwareObjects/EdnaWorkflow.py
@@ -146,7 +146,7 @@ class EdnaWorkflow(HardwareObject):
         workflow_list = []
 
         for _wf in self.get_property("workflow"):
-            wf = {_wf}
+            wf = dict(_wf)
             workflow_list.append(wf)
             wf["requires"] = [r.strip() for r in wf.get("requires", "").split(",")]
             wf["doc"] = ""


mxcubecore/HardwareObjects/sample_centring.py
```diff
@@ -90,9 +90,9 @@ def prepare(centring_motors_dict):
             motors_to_move[m.motor] = m.reference_position
     move_motors(motors_to_move)
 
-    SAVED_INITIAL_POSITIONS = {
+    SAVED_INITIAL_POSITIONS = dict(  # noqa: C404
         [(m.motor, m.motor.get_value()) for m in centring_motors_dict.values()]
-    }
+    )
```
walesch-yan: Could we rewrite this as dictionary comprehension it would satisfy the ruff rule(see [rule](https://docs.astral.sh/ruff/rules/unnecessary-list-comprehension-dict/)) and have the same effect as `dict()`,

so probably something like 
```python
{m.motor: m.motor.get_value() for m in centring_motors_dict.values()}
```

rhfogh:
Why not simply disable the ruff rule? What is the advantage of avoiding dict - it is perfectly good Python?

fabcor-maxiv:
> What is the advantage of avoiding dict - it is perfectly good Python?

The rationale is here: <https://docs.astral.sh/ruff/rules/unnecessary-collection-call/>.

> [Why is this bad?](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/#why-is-this-bad)
> 
> It's unnecessary to call, e.g., `dict()` as opposed to using an empty literal (`{}`). The former is slower because the name `dict` must be looked up in the global scope in case it has been rebound.

For me it is also about consistency. Some have argued that MXCuBE contributors come from different backgrounds, may have learned Python in different contexts. Linters are here exactly to help in such situations by helping bring consistency to code written by many different contributors, so that the code base is easier to read as a whole. So that not every module or function is written in a different style.

rhfogh:
@fabcor-maxiv This is about style, not the speed of our code. Except for a very few crucial loops this particular difference will make less than one part per billion difference to program run times - that is not how you should organise your code. If speed was this important we would not be using Python at all. We would rewrite the whole thing in C or Assembler. As  for consistency - are you honestly claiming that it is hard for you to understand the code because it uses `dict()` instead of `{}`? Do you really experience this as an unacceptable cognitive load? Can you  point to anyone who does?

Code needs to fulfil several criteria: Speed, consistency, clarity in reflecting what is happening, and, yes ease of writing (not just reading) for people from various backgrounds. Before you force everybody to work according to your preferred way of doing things you need to establish some reasonable trade-off between the gain from the individual rules and the hassle  everybody has to go through. 'All choice must be removed because that is more consistent' does not cut it. Especially not when there is lots of work still outstanding that would make a difference to the functionality and actual readability of the code, rather than just promoting your personal choice of style guide.

elmjag:
> Do you really experience this as an unacceptable cognitive load? Can you point to anyone who does?

I'm that anyone. I don't want to waste to my few remaining brain cells on deciding if I should write `dict()` or `{}`.  Let the computer do the thinking.

mxcubecore/HardwareObjects/sample_centring.py
dominikatrojanowska: I would opt for @walesch-yan proposition :) it doesn't make sense to write list comprehension and convert it to dict - I think this may cause the cognitive overload

marcus-oscarsson:
In this case its rather simple we simply can't use `{}` it breaks the code. 

We can either use the dict() constructor/function that takes a collection of tuples or use dict comprehension. To me it does not matter which one we use as long as both ways are fairly similar in terms of speed. 

I agree that we should avoid cognitive overhead and simplify our lives as much as possible.  There is a fine line between quite opinionated rules that are enforced by a tool, that itself also comes with a certain cognitive overhead, and using valid syntax. 

Don't get me wrong I'm all for using tools like Ruff and I think that they do provide means to write better code but think that some rules are overly opinionated. I somehow feel that the motivation of variable look up of 'dict' seems rather far fetched. Id be more worried about the statements in the actual comprehension and the parsing of those statements.

Ill try to change this to use dict comprehension instead. I'm a bit reluctant to do so because I did test this change on the beamline and I know that it works, but Ill make the effort to test again.

dominikatrojanowska (approved):

rhfogh:
Well, @elmjag, if you want to follow Ruff rules in what you *write* that is perfectly reasonable, and no one will prevent you. But the question is another one: Do you find it mentally difficult to *read* code if you have to deal with understanding both `dict()` and `{}`? If the answer is 'yes', then I should at least consider doing a certain amount of extra work to help you deal with your limitations, much as I would be surprised that any professional programmer should find it a genuine hardship to deal with such a minute amount of variation. .

The issue is another one, though. Personally I do not want to waste my few remaining brain cells on carefully adapting to the aesthetic preferences of other people when it makes absolutely no difference to the quality of the final product. I will be a good boy and try to work within whatever style guidelines we *agree* on (rather than unilaterally taking myself out of the discussion and my  code out of Ruff) but if you want me to do the extra work you really ought to come up with a better argument than 'I prefer it that way'.

mxcubecore/HardwareObjects/sample_centring.py
rhfogh: @walesch-yan Yes, that would work fine as well (and Ruff rule C404 actually makes sense - it makes the code more readable).

marcus-oscarsson:
Thanks for the input guys. In this case I simply wanted to revert things to the previous working state and not risk breaking things. I do agree that the dict comprehension reads better and its an obvious improvement. 

There is clearly different opinions on how to use Ruff that we need to sort out and find a way of using this tool (or these tools) in such a way that does not cause frustration. I think we need to expect that it will cause some annoyance sometimes but it seems to be that is much more than that at the moment.  You are all very good programmers so I think we ought to be able to find someway of doing this. I think @beteva had a good idea with an issue list specific Ruff rules that we want to exclude project wide was a good start. Feel free to get started and keep the healthy and constructive discussions going.

fabcor-maxiv:
So I had a closer look at the code changes in this pull request, which I had not done so far, I had only read the comments...

As far as I understand these two lines did NOT infringe the C408 Ruff rule to begin with and did NOT need to be changed in the context of <https://github.com/mxcube/mxcubecore/pull/1264>. The proof is that it was not necessary to re-add an exception when reverting the change in this here pull request, right?

I think that all the pull requests for Ruff-related fixes I have submitted so far, were made by Ruff's automatic fixer. The changes in this pull request were made [by hand](https://github.com/mxcube/mxcubecore/pull/1264#issuecomment-2958461517). It was risky to do so, and we know now that it did not (fully) pay off. Additionally Ruff is perfectly capable of automatically fixing infractions to this one particular rule, so the manual editing was NOT necessary.

I strongly recommend that such "large scale" changes be only made by Ruff's automatic fixer. And I will likely reject such "large scale hand-made" pull requests in the future, because the risk of breaking things is too high as proven here. I had reviewed and approved the original pull request, I had not spotted the issue.

fabcor-maxiv:
> Code needs to fulfil several criteria: Speed, consistency, clarity in reflecting what is happening, and, yes ease of writing (not just reading) for people from various backgrounds.

The way I understand it, the authors of this rule, for code consistency purposes, wanted to make a choice between two different styles (`dict()` or `{}`) and they chose the notation that has the best speed performance. Seems like an acceptable choice to me and in line with what others are saying in this discussion.


> In this case its rather simple we simply can't use `{}` it breaks the code.

Exactly![^1] And as far as I understand, Ruff did the right thing and did NOT detect this as an issue. I did check out the code before the original pull request and this is what Ruff had to say about C408 for those two files:

[^1]: Well not "exactly" because there was something in there, it was not a matter of `dict()` vs. `{}`, it was a matter of `dict(_wf)` vs. `{_wf}`, but I hope everyone understood this by now.

```
mxcubecore/HardwareObjects/EdnaWorkflow.py:121:28: C408 Unnecessary `dict()` call (rewrite as a literal)
    |
119 |         if not self.gevent_event.is_set():
120 |             self.gevent_event.set()
121 |         self.params_dict = dict()
    |                            ^^^^^^ C408
122 |         if "reviewData" in dict_dialog and "inputMap" in dict_dialog:
123 |             review_data = dict_dialog["reviewData"]
    |
    = help: Rewrite as a literal

mxcubecore/HardwareObjects/sample_centring.py:87:22: C408 Unnecessary `dict()` call (rewrite as a literal)
   |
85 |     READY_FOR_NEXT_POINT = gevent.event.Event()
86 |
87 |     motors_to_move = dict()
   |                      ^^^^^^ C408
88 |     for m in centring_motors_dict.values():
89 |         if m.reference_position is not None:
   |
   = help: Rewrite as a literal

Found 2 errors.
No fixes available (2 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

Again, this seems fine to me and in line with what everyone else is saying. Ruff does not recommend to change the problematic lines of code (the ones that had to be reverted here). I am not sure I understand where the debate is, I am slightly confused. But maybe I misunderstood people's arguments.


> list specific Ruff rules that we want to exclude project wide

Yes, I believe anyone should feel free to make a pull request to disable a Ruff rule globally and I assume it will be discussed fairly as we discuss any other pull request.


> keep the healthy and constructive discussions going

I try. My apologies, if what I wrote was unhealthy or not constructive.

rhfogh:
> The way I understand it, the authors of this rule, for code consistency purposes, wanted to make a choice between two different styles (dict() or {}) and they chose the notation that has the best speed performance.

That sums up the problem pretty well. The authors of this rule, having decided that it is not acceptable to have more than one alternative way of writing anything, picked `{}` as being slightly better. It may well be slightly better than the alternative, if you are forced to choose between them, but that entirely sidesteps the real question: WHY is it not acceptable to the authors that there can be more than one way of writing anything? And who and why have decided to impose this constraint on the MXCuBE collaboration? Does this particular rule bring any measurable gains by itself (I submit that it does not)? And how does the infinitesimal gain by having a slightly greater uniformity of code compare with the additional work and stress required to force everybody to code to a micro-managed style? 

Ruff may not have proposed to change `dict(_wf)` to `{_wf}`, but the reason we got that problem in the first place is that we have made it clear to people 1) that the use of `dict`  is _bad_ and means that your code _fails_, 2) that this problem is important enough that we need to comply with all Ruff rules on a high priority, without necessarily looking too deep into what the code is actually doing. If you tell people `dict` is an _error_ (as the output says it is), you should expect them to remove it when they see it - you can hardly expect them to just ignore what looks like errors and say 'if there was anything wrong, Ruff would already have changed it'. If this kind of perfectly valid code had been treated as a non-blocking warning (like you would normally expect from a linter) we would not be having these discussions - and we could still replace the specific case of `dict()` with `{}` as we go along.

Personally I think that the discussion about individual Ruff rules should have been made when we *introduced* them, instead of accepting all 800 sight unseen and putting the onus on people who want them removed to do all the work. But that is of course too late now.

rhfogh:
@marcus-oscarsson If I read this correctly, in `dict(_wf)`, _wf is actually a dictionary rather than a tuple. Not that it makes a difference to any of the rest, though

PR #1260, opened 2025-06-06
1259 download hkl and pdb files when samples are sycnhronized

antolinos:
### Sample information send to EDNA 

MXCube sends now the information about the collected sample. This includes the diffraction (renamed to experiment) plan and the processing plan. 
Check the code  [here](https://github.com/mxcube/mxcubecore/pull/1260/files#diff-6896bc377f09530e309fc634151c94bbc06d48c9f8a9f51935464d7a98aa2855R234)

### PDB and HKL
This pull request adds the experiment and processing plans for each sample. When the processing plan includes reference or PDB files, these files will be automatically downloaded. Please note that the files are associated with the protein, not with each individual sample (crystal).

## Configuration
Added a new configuration parameter `only_staff_session_selection`  that is boolean.
If set to True, only beamline managers can select a session in MXCuBE.
The default value is False.

Full example:

```
``<object class="ICATLIMS">
  <ws_root>https://icat.esrf.fr</ws_root>
  <loginType>user</loginType>
  <object role="session" href="/session" />

  <queue_urls>ingester1:61613</queue_urls>
  <before_offset_days>20</before_offset_days>
  <after_offset_days>10</after_offset_days>
  <compatible_beamlines>ID23-2,ID30A-3</compatible_beamlines>
  <!-- the list of sessions will be filtered by the override_beamline_name, this is needed when
  HWR.beamline.session.beamline_name does not match with the name of the instrument in the catalog -->
  <override_beamline_name>ID23-2</override_beamline_name>

  <filter>participant</filter>
  <only_staff_session_selection>False</only_staff_session_selection>
  <data_portal_url>https://data.esrf.fr/investigation/{id}/datasets</data_portal_url>
  <logbook_url>https://data.esrf.fr/investigation/{id}/logbook</logbook_url>
  <user_portal_url>
    https://smis.esrf.fr/misapps/SMISWebClient/protected/aform/manageAForm.do
  </user_portal_url>
</object>
```
     

`## Metadata

Below is an example of the metadata produced and sent to the pipelines:

```
{
    "MX_beamShape": "ELLIPTICAL",
    "sampleId": 323054,
    "MX_beamSizeAtSampleX": 0.003,
    "MX_beamSizeAtSampleY": 0.004,
    "MX_dataCollectionId": 3535355,
    "MX_detectorDistance": 242.29999999999998,
[...]

PR #1305, opened 2025-07-01
Timezone not in date passed by ICAT

marcus-oscarsson:
Removing the `%z` as timezone not in date passed by ICAT

### DIFF
--- a/mxcubecore/HardwareObjects/ICATLIMS.py
+++ b/mxcubecore/HardwareObjects/ICATLIMS.py
@@ -903,9 +903,9 @@ class ICATLIMS(AbstractLims):
                     dataset_name = fileinfo["prefix"]
 
             try:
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -903,9 +903,9 @@ def finalize_data_collection(self, collection_parameters):
                     dataset_name = fileinfo["prefix"]
 
             try:
-                dt_naive = datetime.strptime(
+                dt_naive = datetime.strptime(  # noqa: DTZ007
                     collection_parameters.get("collection_start_time"),
-                    "%Y-%m-%d %H:%M:%S%z",
+                    "%Y-%m-%d %H:%M:%S",
```
walesch-yan, line 908: The same change is also suggested in #1260

PR #1306, opened 2025-07-01
Update beam position when the zoom level changes

marcus-oscarsson:
The diffractometer supports configuring/calibrating the beam center as a function of the zoom. This updates the beam position when the zoom changes.


walesch-yan (approved):
LGTM

PR #1310, opened 2025-07-02
Add support for configuring Sardana commands with YAMLs

elmjag:
Make it possible to configure Sardana commands when using YAML hardware configuration files.

Adds support for `sardana` section in hardware object configuration files. Using the `sardana` section, it is possible to specify Sardana commands.

Note, this only adds support for _commands_ for now. Currently we at MAXIV are not using Sardana channels. Possibly in the future we can add support for channels, if the need arises.


marcus-oscarsson:
Ok, looks good. 

If you can please also have a look the PRs I made the other day. It would be nice to merge those as well :)

marcus-oscarsson (approved):

PR #1311, opened 2025-07-03
Updated child object access documentation to reflect previous discussion

rhfogh:
It looks like it falls to me to update the recent documentation change to reflect the discussion we had in PR #1299. So here it is.


docs/source/dev/configuration_files.md
```diff
@@ -120,6 +120,11 @@ That child object is accessible using `get_object_by_role()` method of the paren
 The `get_object_by_role()` method requires `role` argument.
 For example, if `foo` is an instance of `Foo` class, its `bar` object can be fetched as follows:
 
+The `get_object_by_role` function is intended for private child objects; it is scheduled
+to be renamed to `_get_object_by_role` to reflect this. Child objects that are part of a class public interface -
+particularly that of abstract classes - must be defined as a property
+of the class, so they can (and should) be accessed as, in the example given above, `myFoo.bar`.
+
```
elmjag, line 131: Can you move this section below this part:

> \```python
>      bar = foo.get_object_by_role("foo")
>´```

Right now, it cuts in between the:

> For example, if `foo` is an instance of `Foo` class, its `bar` object can be fetched as follows:

and the example provided:

> \```python
>      bar = foo.get_object_by_role("foo")
>´```

elmjag:
Nice clarification.

docs/source/dev/configuration_files.md
rhfogh: Ooops! Sorry.

Fixed now.

elmjag (approved):

Issue #1224, opened 2025-05-21
[CC-DESY] Clean up banner in files


fabcor-maxiv:
Maybe we do not need to have the banner in all the files.

Maybe we do not need to have the license in all the files, it might be enough to have it a the root of the project directory.

The `Project: MXCuBE` line triggers [Ruff `ERA001`](https://docs.astral.sh/ruff/rules/commented-out-code/). So, if we clean up this, we should remember to also remove the `ERA001` exclusions in `ruff.toml`.

fabcor-maxiv:
@rhfogh, can you please check if we need a banner including the license in all the files of the project, or if it is enough to have it at the project level?

alessandro-olivo:
Just a comment, if you remove the ":"  ruff will pass

```diff
...
- #  Project: MXCuBE 
+ #  MXCuBE Project
 #  https://github.com/mxcube
...
```

rhfogh:
This is still on my todo list. I need to check it with Peter Keller, though, who will not be back from holiday till next week.

dominikatrojanowska:
Thanks for raising discussion :) 

> Just a comment, if you remove the ":" ruff will pass
> ```
> ...
> - #  Project: MXCuBE 
> + #  MXCuBE Project
>  #  https://github.com/mxcube
> ...
I proposed something similar before I saw this discussion, I added a 'name':
```
#  Project name: MXCuBE 
```
so `noqa` could be omitted. For now I will leave it as it is, until @rhfogh will let us know what is the banner decision :)

fabcor-maxiv:
> I proposed something similar before I saw this discussion, I added a 'name':
> 
> ```
> #  Project name: MXCuBE 
> ```

I like this. This is fine for me.

beteva:
> This is still on my todo list. I need to check it with Peter Keller, though, who will not be back from holiday till next week.

I think there is no rush. We could wait until there is a clear decision and make it for the whole project, not little by little.

rhfogh:
Unfortunately we DO have to have banners in all the files - see https://www.gnu.org/licenses/gpl-howto.html
So, we need to add them where they are missing. That would mean also in mxcubeweb

What is more, we also need to have the proper years in the copyright notices. We could do the years with a global replace, but could anyone say what the starting year should be? 

Finally, the COPYING file should be the GPL, not the LGPL, and the LGPL license file should be called COPYING.LESSER (you need both). I shall make a PR to fix that

PR #1313, opened 2025-07-08
Applying ruff rule: commented-out-code (ERA001) for tests

dominikatrojanowska:
Commented out code in test tiles was relatively non-problem-prone to be removed.

Related issues: 
* https://github.com/mxcube/mxcubecore/issues/1230
* https://github.com/mxcube/mxcubecore/issues/1224
* https://github.com/mxcube/mxcubecore/issues/1316

Related PRs: 
* https://github.com/mxcube/mxcubecore/pull/1320 (to be merged after 1320)
* https://github.com/mxcube/mxcubecore/pull/1315

### DIFF
--- a/test/test_beam.py
+++ b/test/test_beam.py
@@ -207,7 +207,6 @@ class TestBeam(TestHardwareObjectBase.TestHardwareObjectBase):
         # slit size in mm, aperture diameters are in microns
         target_width = target_height = max_diameter / 2000.0
         test_object.set_value([target_width, target_height])
-        # beam_width, beam_height, beam_shape, _ = test_object.get_value()
         beam_width, beam_height = test_object.get_beam_size()
         assert target_width == beam_width
         assert target_height == beam_height
--- a/ruff.toml
+++ b/ruff.toml
@@ -5809,7 +5809,6 @@ convention = "google"
     "SLF001",
 ]
 "test/conftest.py" = [
-    "ERA001",
     "N814",
     "PT003",
     "PTH100",
@@ -5819,7 +5818,6 @@ convention = "google"
     "T201",
 ]
 "test/test_aperture.py" = [
-    "ERA001",
     "SLF001",
 ]
 "test/test_base_hardware_objects.py" = [
@@ -5828,7 +5826,6 @@ convention = "google"
     "C419",
     "E501",
     "E713",
-    "ERA001",
     "PLR0913",
     "PT003",
     "PT007",
@@ -5838,12 +5835,10 @@ convention = "google"
     "SLF001",
 ]
 "test/test_beam.py" = [
-    "ERA001",
     "PERF401",
     "SLF001",
 ]
 "test/test_beam_definer.py" = [
-    "ERA001",
 ]
 "test/test_beamline_ho_id.py" = [
     "SIM300",
@@ -5853,7 +5848,6 @@ convention = "google"
     "E501",
     "E712",
     "E713",
-    "ERA001",
     "F821",
     "FBT001",
     "PLR0913",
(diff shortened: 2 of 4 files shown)


fabcor-maxiv (changes requested):
We have a discussion opened on this topic: <https://github.com/mxcube/mxcubecore/issues/1224>. I believe the approach in this pull request is not a good one. Removing the colon `:` seems like a better approach, not sure if it is the best but for sure it is better.

dominikatrojanowska:
> We have a discussion opened on this topic: #1224. I believe the approach in this pull request is not a good one. Removing the colon `:` seems like a better approach, not sure if it is the best but for sure it is better.

Thanks for raising the discussion! Before I saw it I thought that, indeed the previous proposition with `noqa` is not the best one, since there will be comments visible in api documentation (?)

For now I will leave it as it is, until @rhfogh will let us know what is the banner decision :)

marcus-oscarsson:
Can we close this PR then ?

dominikatrojanowska:
> Can we close this PR then ?

you mean close without merging? I prefer not, unless there is a reason to drop the ruff polishing. Could it hang as draft for the decision? or could we merge it, and if decision will come, just create a new PR clearing redundant banners?

test/test_beam.py
```diff
@@ -1,6 +1,6 @@
 # encoding: utf-8
 #
-#  Project: MXCuBE
+#  Project: MXCuBE  # noqa: ERA001
```
fabcor-maxiv:
```suggestion
#  Project name: MXCuBE
```

fabcor-maxiv (approved):
I think this version of the change is fine, besides the one thing I commented on.

I do not know that we want to wait until there is a decision on removing the whole file heading banners, this might be a long while still...

fabcor-maxiv:
Also we could do the same fix in `py_file_header.txt`:

https://github.com/mxcube/mxcubecore/blob/202876c144c38973a8d21bbf856ccc7a4054d0b9/py_file_header.txt#L3

But maybe in a different pull request? I do not know...

test/test_beam.py
dominikatrojanowska: yep, I missed it, thanks :)

dominikatrojanowska:
> Also we could do the same fix in `py_file_header.txt`:
> 
> https://github.com/mxcube/mxcubecore/blob/202876c144c38973a8d21bbf856ccc7a4054d0b9/py_file_header.txt#L3
> 
> But maybe in a different pull request? I do not know...

I will put this change here: https://github.com/mxcube/mxcubecore/pull/1315

beteva:
Didn't we said that for this we wait for @rhfogh to see if we do need the banners? Even if it takes time, I think there is no rush for this type of cleaning.

dominikatrojanowska:
This PR provides also other changes. For me it does no matter whether it will be merged sooner or later.

marcus-oscarsson:
> you mean close without merging? I prefer not, unless there is a reason to drop the ruff polishing. Could it hang as draft for the decision? or could we merge it, and if decision will come, just create a new PR clearing redundant banners?

No, that was me who misread the PR titles, I though it was a duplicate. 

We can remove the commented code but I'd prefer the the "project name change" is made separately. I guess that's what you also have in mind right ?

dominikatrojanowska:
> We can remove the commented code but I'd prefer the the "project name change" is made separately. I guess that's what you also have in mind right ?

Here it is: https://github.com/mxcube/mxcubecore/pull/1320  
I hope, `ruff.toml` updates can stay within this PR :)

test/conftest.py
```diff
@@ -1,6 +1,6 @@
 # encoding: utf-8
 #
-#  Project: MXCuBE
+#  Project name: MXCuBE
```
marcus-oscarsson: So these changes are still here and in the other PR ERA001 ?

dominikatrojanowska: they disappeared after rebasing

marcus-oscarsson:
Ok, could you please re-base and merge please ?

PR #1315, opened 2025-07-08
Draft: Applying ruff rule: commented-out-code (ERA001) for general purpose files

dominikatrojanowska:
This PR takes care only for general purpose files. Changes for site specific files will be proposed in separate PRs to divide the eventual discussion. Most of the removed comments were no used for 5 years or more. But not all. For most of those "quite fresh" I'm adding additional comments. Anyway 

Related issues: 
* https://github.com/mxcube/mxcubecore/issues/1230
* https://github.com/mxcube/mxcubecore/issues/1224

Related PRs: 
* https://github.com/mxcube/mxcubecore/pull/1320 (to be merged after 1320)
* https://github.com/mxcube/mxcubecore/pull/1313

After final "banner decision" the concerned files will be adjusted

### DIFF
--- a/mxcubecore/Command/Epics.py
+++ b/mxcubecore/Command/Epics.py
@@ -173,7 +173,6 @@ class EpicsCommand(CommandObject):
         self.__value_changed_callback_ref = saferef.safe_ref(value_changed_callback)
 
         # store the call to get as a function object
-        # poll_cmd = self.pv.get
         poll_cmd = self.get_pv_value
 
         Poller.poll(
--- a/mxcubecore/HardwareObjects/GenericDiffractometer.py
+++ b/mxcubecore/HardwareObjects/GenericDiffractometer.py
@@ -334,15 +334,6 @@ class GenericDiffractometer(HardwareObject):
         self.capillary = self.get_object_by_role("capillary")
         self.cryo = self.get_object_by_role("cryo")
 
-        # Hardware objects ----------------------------------------------------
-        # if HWR.beamline.sample_view.camera is not None:
-        #     self.image_height = HWR.beamline.sample_view.camera.get_height()
-        #     self.image_width = HWR.beamline.sample_view.camera.get_width()
-        # else:
-        #     logging.getLogger("HWR").debug(
-        #         "Diffractometer: " + "Camera hwobj is not defined"
-        #     )
-
         if HWR.beamline.beam is not None:
             self.beam_position = HWR.beamline.beam.get_beam_position_on_screen()
             self.connect(
@@ -616,11 +607,9 @@ class GenericDiffractometer(HardwareObject):
         :param timeout: timeout in seconds
         :type timeout: seconds
         """
-        # self.ready_event.clear()
         self.current_state = DiffractometerState.tostring(DiffractometerState.Busy)
         method(*args)
         time.sleep(5)
-        # gevent.sleep(2)
         self.wait_device_ready(timeout)
         self.ready_event.set()
 
@@ -725,9 +714,6 @@ class GenericDiffractometer(HardwareObject):
         """
         self.beam_position = list(value)
 
-    # def get_motor_positions(self):
-    #    return
-
     # TODO rename to get_motor_positions
     def get_positions(self):
         """
@@ -819,16 +805,9 @@ class GenericDiffractometer(HardwareObject):
                     "Diffractometer: problem aborting the centring method"
                 )
             try:
-                # TODO... do we need this at all?
-                # fun = self.cancel_centring_methods[self.current_centring_method]
-                pass
-            except KeyError:
+                fun()
+            except Exception:
                 self.emit_centring_failed()
-            else:
-                try:
-                    fun()
-                except Exception:
-                    self.emit_centring_failed()
         else:
             self.emit_centring_failed()
         self.emit_progress_message("")
@@ -949,15 +928,6 @@ class GenericDiffractometer(HardwareObject):
             except Exception:
                 logging.exception("Could not move to centred position")
                 self.emit_centring_failed()
-            else:
-                # if 3 click centring move -180. well. dont, in principle the calculated
-                # centred positions include omega to initial position
-                pass
-                # if not self.in_plate_mode():
-                #    logging.getLogger("HWR").debug("Centring finished. Moving omega back to initial position")
-                #    self.motor_hwobj_dict['phi'].set_value_relative(-180, timeout=None)
-                #    logging.getLogger("HWR").debug("         Moving omega done")
-
             if (
                 self.current_centring_method
                 == GenericDiffractometer.CENTRING_METHOD_AUTO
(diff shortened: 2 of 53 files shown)


mxcubecore/Command/Epics.py
```diff
@@ -126,7 +126,7 @@ def __call__(self, *args, **kwargs):
                     value = args[0]
                     wait = kwargs.get("wait", False)
                     self.pv.put(value, wait=wait)
-                except:
+                except Exception as e:
```
fabcor-maxiv: Unrelated change here.

ruff.toml
```diff
@@ -70,8 +69,6 @@ convention = "google"
     "BLE001",
     "C901",
     "E501",
-    "E722",
-    "ERA001",
```
fabcor-maxiv: Unrelated change here.

fabcor-maxiv (changes requested):
There is some unrelated change in this pull request.

I recommend being extremely careful with this kind of things, strictly focus on one and only one thing, follow carefully the hints from the linter, and use the linter's auto-fixing feature when it is available (in this case, it seems like there is no auto-fixing available). The risk of breaking things by accident is not zero as we have seen recently.

fabcor-maxiv (approved):
This seems fine to me.

mxcubecore/HardwareObjects/GenericDiffractometer.py
```diff
@@ -819,16 +805,9 @@ def cancel_centring_method(self, reject=False):
...
-                pass
-            except KeyError:
+                fun()
+            except Exception:
                 self.emit_centring_failed()
-            else:
-                try:
-                    fun()
-                except Exception:
-                    self.emit_centring_failed()
```
fabcor-maxiv, line 831: Unrelated change!

mxcubecore/HardwareObjects/GenericDiffractometer.py
```diff
@@ -949,15 +928,6 @@ def centring_done(self, centring_procedure):
...
                 self.emit_centring_failed()
-            else:
-                # if 3 click centring move -180. well. dont, in principle the calculated
-                # centred positions include omega to initial position
-                pass
-                # if not self.in_plate_mode():
-                #    logging.getLogger("HWR").debug("Centring finished. Moving omega back to initial position")
-                #    self.motor_hwobj_dict['phi'].set_value_relative(-180, timeout=None)
-                #    logging.getLogger("HWR").debug("         Moving omega done")
-
```
fabcor-maxiv, line 960: This removes a comment that is not code!

mxcubecore/HardwareObjects/LimaEigerDetector.py
```diff
@@ -167,10 +166,6 @@ def prepare_acquisition(
             "omega_increment=%0.4f" % osc_range,
             "wavelength=%s" % HWR.beamline.energy.get_wavelength(),
         ]
-        # either we set the wavelength, or we set the energy_threshold
-        # up to now we were doing both (lost of time)
-        # "wavelength=%s" % HWR.beamline.energy.get_wavelength(),
-        # self.set_energy_threshold(HWR.beamline.energy.get_value())
```
fabcor-maxiv, line 173: This is not code!

mxcubecore/HardwareObjects/LimaPilatusDetector.py
```diff
@@ -140,14 +140,6 @@ def roi_mode_changed(self, mode):
...
         self.emit("detectorRoiModeChanged", (self.roi_mode,))
 
-    def set_roi_mode(self, mode):
-        """Sets roi mode
-
-        :param mode: roi mode
-        :type mode: str
-        """
-        # self.chan_roi_mode.set_value(self.roi_modes_list[mode])
-
```
fabcor-maxiv, line 150: Unrelated change!

fabcor-maxiv (changes requested):
From my point of view this is a bad pull request. There are changes that are not about removing commented code so they do not belong in this pull request. In some cases it looked to me like maybe the commented code could actually be useful info, but it is very hard to tell without knowing the context.

I did not even go halfway through this pull request, so there might be more issues in this pull request.

As [I said the other day](https://github.com/mxcube/mxcubecore/pull/1304#issuecomment-3036627075), I wonder if we should continue with these large scale hand-made pull requests. Maybe automatic fixes (on a clean repository to make sure one and only one thing is fixed in a consistent manner) could be okay, but this one clearly seems hand made and mixes multiple kinds of fixes. Maybe large scale hand made pull requests can be okay but we have to be extremely careful and consistent, it is not easy.

Anyway, thanks @dominikatrojanowska for working on this, this is tedious work. For the future, I recommend considering the following:

* experiment on the things that can be fix automatically by the tool
* do smaller batches
* trying on MXCuBE-Web because the code base is much smaller so pull requests would be easier to review and much less chances of issues going through

mxcubecore/HardwareObjects/LimaPilatusDetector.py
dominikatrojanowska: hm, ok, I found pointless to keep the method that consists only this commented line of code and the parent implements it already. 
https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/abstract/AbstractDetector.py#L182

dominikatrojanowska:
I wanted to write a bunch of comments explaining _what?_ and _why?_ but didn't managed yet.
Should I try to explain or close this PR and eventually in the future provide smaller ones? 
or maybe, to keep some already reviewed changes, I can drop all the commits after: https://github.com/mxcube/mxcubecore/pull/1315#pullrequestreview-2997961455 ?

mxcubecore/HardwareObjects/LimaPilatusDetector.py
fabcor-maxiv: I see, this should be its own pull request then from my point of view.

fabcor-maxiv:
> I wanted to write a bunch of comments explaining _what?_ and _why?_ but didn't managed yet. Should I try to explain or close this PR and eventually in the future provide smaller ones? or maybe, to keep some already reviewed changes, I can drop all the commits after: [#1315 (review)](https://github.com/mxcube/mxcubecore/pull/1315#pullrequestreview-2997961455) ?

I think smaller pull requests is better, of course.

I do not know that it makes sense to revert back to an older version of this pull request. I forgot what I already reviewed. How would we know that the revert to the old version went well?

Maybe I would do like this:

* multiple "code comments removal" across multiple files is fine as long as:
  1. there is absolutely no doubt that this is dead code that has no value any longer (and well this is hard to evaluate, so stay on the safe side)
  1. the pull request is easy to review, so maybe a handful of files at once (maybe around ~5 files maximum as a random rule of thumb)
* if it is not just 100% safe "code comments removal" but there are also some other code cleanups then try to do 1 file at a time (or whatever the minimum number files is required to get a small contained pull request).

PR #1317, opened 2025-07-08
Fixed return value and call to super().__getattr__ in `CommandContainer`

marcus-oscarsson:
Fixed return value and call to `super().__getattr__`  for `CommandContainer`. return statement was missing and object does  not have `__getattr__` but `__getattribute__`


fabcor-maxiv (approved):
OK I guess

PR #1319, opened 2025-07-09
1259 donwload hkl and pdb files when samples are sycnhronized

antolinos:
This bug was introduced when incorrectly fixed FBT003:
```
mxcubecore/HardwareObjects/ICATLIMS.py:414:71: FBT003 Boolean positional value in function call
    |
412 |     @property
413 |     def only_staff_session_selection(self):
414 |         return bool(self.get_property("only_staff_session_selection", False))
    |                                                                       ^^^^^ FBT003


Found 1 error.

### CODE REVIEW

File: mxcubecore/HardwareObjects/ICATLIMS.py [hunk 1]
```diff
@@ -411,7 +411,7 @@ def _store_data_collection_group(self, group_data):
 
     @property
     def only_staff_session_selection(self):
-        return bool(self.get_property("only_staff_session_selection", default=False))
+        return bool(self.get_property("only_staff_session_selection", False))
```
beteva:
```suggestion
        return bool(self.get_property("only_staff_session_selection", default_value=False))
```

marcus-oscarsson (approved):

marcus-oscarsson:
@antolinos, its a bit odd, but there seems to be a conflict. Can you have a look ?

PR #1321, opened 2025-07-10
Replaced erroneous LGPL copy with correct GPL license copy

rhfogh:
You need both license files, and they must be different (they were identical)

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,165 +1,674 @@
-                   GNU LESSER GENERAL PUBLIC LICENSE
+                   GNU GENERAL PUBLIC LICENSE
                        Version 3, 29 June 2007
 
  Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
  Everyone is permitted to copy and distribute verbatim copies
  of this license document, but changing it is not allowed.
 
+                            Preamble
 
-  This version of the GNU Lesser General Public License incorporates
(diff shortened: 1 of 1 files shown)


fabcor-maxiv (approved):
I am not a lawyer.

beteva:
We have to check this, as I remember there were some problems using GPL and not LGPL. I remember that it was somehow linked to the Global Phasing and Qt issues.

rhfogh:
@beteva, @fabcor-maxiv 
I am not a lawyer either, but the GPL site has made it pretty clear. If you look at
https://www.gnu.org/licenses/gpl-howto.html they tell you that  you need a license notice in each file (saying you use LGPL), and copyright statements with years on them. They also say that since the LGPL is a set of additional permissions added to the GPL, you need both of the license files present if you use LGPL.

The main thing is that if **any** of MXCuBE is GPL, **all** of it must be GPL - that is how the GPL is written. And that would mean that non-GPL code, like that from the Global Phasing workflow, cannot link to it.

Qt is a separate problem. The license conditions mean that any group coding to PyQt must have a commercial license, and that the program can only use the PyQt versions for which there was  a commercial license. More precisely, if  you have a PyQt commercial license in 2022 your program can use the Qt versions that were current in 2022 (and that you therefore have as license for). It cannot later download and use a non-commercial 2024 version

PR #1322, opened 2025-07-11
Applying ruff rule ERA001 commented-out-code for LimaPilatusDetector.py and LimaEigerDetector.py

dominikatrojanowska:

### DIFF
--- a/mxcubecore/HardwareObjects/LimaEigerDetector.py
+++ b/mxcubecore/HardwareObjects/LimaEigerDetector.py
@@ -96,7 +96,6 @@ class LimaEigerDetector(AbstractDetector):
         self.update_state(self.STATES.READY)
 
     def last_image_saved(self):
-        # return 0
         return self.get_channel_object("last_image_saved").get_value() + 1
 
     def get_deadtime(self):
@@ -167,10 +166,10 @@ class LimaEigerDetector(AbstractDetector):
             "omega_increment=%0.4f" % osc_range,
             "wavelength=%s" % HWR.beamline.energy.get_wavelength(),
         ]
-        # either we set the wavelength, or we set the energy_threshold
-        # up to now we were doing both (lost of time)
-        # "wavelength=%s" % HWR.beamline.energy.get_wavelength(),
-        # self.set_energy_threshold(HWR.beamline.energy.get_value())
+        # Either we set the wavelength or we set the energy_threshold.
+        # Up to now both ways are possible:
+        # self.set_energy_threshold(HWR.beamline.energy.get_value())  # noqa: ERA001
+        # "wavelength=%s" % HWR.beamline.energy.get_wavelength()  # noqa: ERA001
 
         self.get_channel_object("saving_common_header").set_value(header_info)
 
@@ -179,7 +178,6 @@ class LimaEigerDetector(AbstractDetector):
             self.get_channel_object("acq_trigger_mode").set_value("EXTERNAL_TRIGGER_SEQUENCES")
             self.get_channel_object("acq_nb_sequences").set_value(mesh_num_lines)
             """
-            # self.get_channel_object("acq_trigger_mode").set_value("EXTERNAL_GATE")
             self.get_channel_object("acq_trigger_mode").set_value(
                 "EXTERNAL_TRIGGER_MULTI"
             )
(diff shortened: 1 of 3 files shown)


mxcubecore/HardwareObjects/LimaEigerDetector.py
```diff
@@ -96,7 +96,6 @@ def wait_ready(self, timeout=30):
         self.update_state(self.STATES.READY)
 
     def last_image_saved(self):
-        # return 0
```
dominikatrojanowska, line 99: at least 4 year old code

mxcubecore/HardwareObjects/LimaEigerDetector.py
```diff
@@ -166,11 +165,10 @@ def prepare_acquisition(
             "omega_start=%0.4f" % start,
             "omega_increment=%0.4f" % osc_range,
             "wavelength=%s" % HWR.beamline.energy.get_wavelength(),
+            # either we set the wavelength, or we set the energy_threshold
+            # up to now we were doing both (lots of time)
+            # self.set_energy_threshold(HWR.beamline.energy.get_value())  # noqa: ERA001
         ]
-        # either we set the wavelength, or we set the energy_threshold
-        # up to now we were doing both (lost of time)
-        # "wavelength=%s" % HWR.beamline.energy.get_wavelength(),
```
dominikatrojanowska, line 172: this line is a copy paste from above line so I just removed the comment and move directly under it

mxcubecore/HardwareObjects/LimaEigerDetector.py
```diff
@@ -166,11 +165,10 @@ def prepare_acquisition(
...
-        # self.set_energy_threshold(HWR.beamline.energy.get_value())
```
dominikatrojanowska, line 173: @marcus-oscarsson is this whole comment still relevant? for me it seems like note left in a code base in big PR... Isn't it better to start discussion on this topic instead?

mxcubecore/HardwareObjects/LimaEigerDetector.py
```diff
@@ -179,7 +177,6 @@ def prepare_acquisition(
             self.get_channel_object("acq_trigger_mode").set_value("EXTERNAL_TRIGGER_SEQUENCES")
             self.get_channel_object("acq_nb_sequences").set_value(mesh_num_lines)
             """
-            # self.get_channel_object("acq_trigger_mode").set_value("EXTERNAL_GATE")
```
dominikatrojanowska, line 182: at least 4 year ago the `acq_trigger_mode`  was changed to `EXTERNAL_TRIGGER_MULTI`

mxcubecore/HardwareObjects/LimaPilatusDetector.py
```diff
@@ -140,14 +140,6 @@ def roi_mode_changed(self, mode):
...
         self.emit("detectorRoiModeChanged", (self.roi_mode,))
 
-    def set_roi_mode(self, mode):
-        """Sets roi mode
-
-        :param mode: roi mode
-        :type mode: str
-        """
-        # self.chan_roi_mode.set_value(self.roi_modes_list[mode])
-
```
dominikatrojanowska, line 150: it is pointless to keep the method that consists only this commented line of code and the parent implements it already:
https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/HardwareObjects/abstract/AbstractDetector.py#L182

mxcubecore/HardwareObjects/LimaEigerDetector.py
marcus-oscarsson: We better ask @beteva for confirmation but I think the comment can be removed

beteva: The comment explains that with some version of the API there is a duplication of the actions in the.  We've chosen to use the
`"wavelength=%s" % HWR.beamline.energy.get_wavelength(),`, because we use the filewriter and set the header info.
Others may want to set it with the
`self.set_energy_threshold(HWR.beamline.energy.get_value())`

Please, leave the comment as it is.

mxcubecore/HardwareObjects/LimaEigerDetector.py
beteva: Please do not do this. The copy/paste is deliberate to show the two ways for setting the wavelength.

mxcubecore/HardwareObjects/LimaPilatusDetector.py
beteva: This is actually going to be used as soon as we have some time for testing.

mxcubecore/HardwareObjects/LimaEigerDetector.py
dominikatrojanowska: Ok, I didn't understood from the comment that both options are possible... Rather that both are still in use somewhere in the code base but should not be actually.  I reverted more less to how it was with small change in the comment, I think this way it is more clear. 
Re-review please :)

marcus-oscarsson:
LGTM

marcus-oscarsson (approved):

PR #1326, opened 2025-07-11
Rhfogh fix loglevel

rhfogh:
Change name 'user_log_level' (only place found in code) to the common spelling  'user_level_log'


rhfogh:
Wrong starting point. Withdrawn

PR #1331, opened 2025-07-15
Hide the processing section of "Tasks" dialogs based on backend configuration

alessandro-olivo:
This PR is required for the PR: https://github.com/mxcube/mxcubeweb/pull/1782   (follow the link for details)

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -1715,7 +1715,7 @@ class AcquisitionParameters(object):
         self.in_queue = False
         self.in_interleave = None
         self.sub_wedge_size = 10
-
+        self.disable_processing = False
         self.num_triggers = int()
         self.num_images_per_trigger = int()
         self.hare_num = 1
@@ -1765,6 +1765,7 @@ class AcquisitionParameters(object):
             "mesh_center": self.mesh_center,
             "cell_spacing": self.cell_spacing,
             "sub_wedge_size": self.sub_wedge_size,
+            "disable_processing": self.disable_processing,
         }
 
     def copy(self):


mxcubecore/model/queue_model_objects.py
```diff
@@ -1715,7 +1715,7 @@ def __init__(self):
         self.in_queue = False
         self.in_interleave = None
         self.sub_wedge_size = 10
-
+        self.disable_processing = False
```
marcus-oscarsson, line 1718: Sure, why not. Is this a UI option or more a feature that you also which to use on the backend. Do you actually want to disable processing or always call the same processing ?

alessandro-olivo: Right now, it’s just used to hide something in the UI that isn’t needed, but I can imagine a use case where this information could be used by the backend to decide whether to start processing

marcus-oscarsson: Ok, sounds good. Should we make it configurable then so that its read from the `beamline_config.yaml` ? There is already a section for similair settings

alessandro-olivo: If you follow this link <https://github.com/mxcube/mxcubeweb/pull/1782#issue-3231201171>, you’ll see the implementation in mxcubeweb.
You can already configure it from the `beamline_config.yaml`. I chose to place it under the  `default_acquisition_parameters` section, but there might be a better place.
I chose that section so it could be disabled depending on the method.

Example:
```diff
   ...
  default_acquisition_parameters:
    default:
      ...
      take_snapshots: true
      sub_wedge_size: 10
+     disable_processing: true
      ...
```

marcus-oscarsson: Ok, lets keep it like that :)

marcus-oscarsson:
Looks good to me, please rebase and merge :)

mxcubecore/model/queue_model_objects.py
alessandro-olivo: To make it clearer, I could update the `beamline_config.yaml` file by adding the line `disable_processing: false`. It's not really necessary since "false" is the default, but it helps show how to configure the setting.

alessandro-olivo:
@marcus-oscarsson A review is needed before I can merge

marcus-oscarsson (approved):

PR #1333, opened 2025-07-18
LIMS tests: mock suds in more places

elmjag:
When testing ISPyBLims code, make sure we mock suds client used in `mxcubecore.HardwareObjects.UserTypeISPyBLims` module as well.

Otherwise a real suds Client will be used, which will try to make HTTP requests. The tests would (sometimes!?) wait until request timed out, thus slowing down the test suite.

This makes the tests run like 1 minuter faster!


walesch-yan (approved):
Thanks @elmjag . Could you rebase and merge this?

PR #1334, opened 2025-07-18
ISARAMaint: explicitly read `tangoname` config property

elmjag:
With the new hardware object configuration system, all configuration properties need to be explicitly read with for example `get_property().`

Earlier, it seems that it worked automagically.

### DIFF
--- a/mxcubecore/HardwareObjects/ISARAMaint.py
+++ b/mxcubecore/HardwareObjects/ISARAMaint.py
@@ -91,13 +91,16 @@ class ISARAMaint(HardwareObject):
         self._message = None
 
     def init(self):
-        self.isara_dev = DeviceProxy(self.tangoname)
+        tangoname = self.get_property("tangoname")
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
> Earlier, it seems that it worked automagically.

Yes the overridden `__getattr__` was providing this interesting feature ... 

So, we now need to be explicit and use `get_property`

elmjag:
> > Earlier, it seems that it worked automagically.
> 
> Yes the overridden `__getattr__` was providing this interesting feature ...
> 
> So, we now need to be explicit and use `get_property`

Aha!

I could never wrap my head around all this magic methods. Nice that we got rid of that.

marcus-oscarsson:
Oh, yes it was very confusing and could have been the source for some really nasty bugs !

marcus-oscarsson:
Looks good to me, could you please rebase and merge

marcus-oscarsson (approved):

PR #1335, opened 2025-07-18
AbstractDetector - Fixed reading of configuration for min and max exposure time

marcus-oscarsson:
The configuration data is not set on the object until after the execution of `__init__` and hence can't be read in `__init__`.


beteva (approved):
Thanks for fixing this. I believe get_property returns None if not in the configuration, but indeed it is better to be more explicit.

PR #1336, opened 2025-07-18
MAXIV MachInfo: make it possible to use YAML conf file

elmjag:
Make it possible to create MachInfo objects with `MachInfo(name='foo')` expression.

This is the expression used when loading hardware objects from a YAML configuration file.


beteva (approved):
Looks fine/

PR #1337, opened 2025-07-22
Add permissions to workflows

Phantominumm:
This pull request updates GitHub Actions workflows to explicitly define permissions. The changes ensure each workflow has the minimum required permissions to perform its tasks.

Tell me if I missed a needed permissions.

### DIFF
--- a/.github/workflows/coverage_report.yml
+++ b/.github/workflows/coverage_report.yml
@@ -15,6 +15,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     permissions:
+      contents: read
       pull-requests: write
 
     steps:
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -5,12 +5,13 @@ name: Create/Update Tag
   push:
     branches:
       - develop
-permissions:
-  contents: write
-  packages: write
+
 jobs:
   create-version-tag:
     runs-on: ubuntu-latest
+    permissions:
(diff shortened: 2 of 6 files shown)


fabcor-maxiv:
Same as for MXCuBE-Core, I would encourage you to set permissions at the job level instead of at the workflow level. Otherwise, seems fine.

fabcor-maxiv (approved):

Phantominumm:
If there are multiple jobs, I set the basic necessary permissions at the workflow level, and then the specific ones at the job level. Otherwise, there would be a lot of repetition across jobs.
The cases where there are  only one job in the workflow, it doesn’t make much difference from my perspective whether the permissions are set at the job or workflow level.
If needed, I can always update it to set all permissions at the job level.

fabcor-maxiv:
> it doesn’t make much difference from my perspective whether the permissions are set at the job or workflow level

Because over time, it makes it easier to keep adherence to the "principle of least privilege".

For example, if we modify/remove jobs from a workflow we do not necessarily think about updating the permissions at the workflow level. And if we add a new job to a workflow we might be tempted to add the permissions at the workflow level even though the other jobs do not need the new permissions.

Phantominumm:
Ok I understand it now, I will change it so

fabcor-maxiv (approved):
Seems fine to me.

walesch-yan (approved):

PR #1338, opened 2025-07-22
Add bandit and sonar check and report

Phantominumm:
This is the same workflow than in mxcubeWeb. It introduces a new GitHub Actions workflow to integrate Bandit and SonarQube scans into the CI/CD pipeline. The workflow aims to enhance security and code quality by automating static analysis and vulnerability detection.

### New CI/CD Workflow for Static Analysis and Security Scans:

* **Added Bandit Scan Job**: Configured a Bandit scan to detect security issues in Python code. The job dynamically generates skip rules using Ruff, executes Bandit with specified confidence and severity levels, and uploads the results in SARIF format to GitHub.

* **Added SonarCloud Scan Job**: Configured SonarCloud to perform code quality and duplication analysis. The job sets up Java, caches SonarCloud dependencies, and runs the scan with specific exclusions and settings for the project.
* (`.github/workflows/bandit_sonar.yaml`)

### DIFF
--- a/.github/workflows/bandit_sonar.yaml
+++ b/.github/workflows/bandit_sonar.yaml
@@ -1,0 +1,78 @@
+---
+
+name: Bandit and SonarQube
+
+"on":
+  pull_request: null
+  push: null
+  workflow_dispatch: null
+
+permissions:
+  contents: read
+  security-events: write
+  actions: read
+
+jobs:
+  bandit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Generate Bandit skips from Ruff
+        id: bandit_skips
+        run: |
+          pip install ruff jq
+          SKIPS=$(ruff rule --all --output-format json \
+            | jq -r '.[] | select(.code | test("^S[0-9]{3}$")) | .code' \
+            | sed 's/S/B/' \
+            | paste -sd, -)
+          echo "skips=$SKIPS" >> $GITHUB_OUTPUT
+
+      - name: Bandit Scan
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
+        with:
+          exit_zero: true
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          excluded_paths: ./test/
+          level: MEDIUM
+          confidence: HIGH
+          skips: ${{ steps.bandit_skips.outputs.skips }}
+
+      - name: Upload SARIF file to GitHub
+        uses: github/codeql-action/upload-sarif@v3
+        with:
+          sarif_file: results.sarif
+          category: bandit
+
+  sonarcloud:
+    name: SonarCloud Scan
+    runs-on: ubuntu-latest
+    if: >
+      github.event.pull_request.head.repo.full_name == github.repository ||
+      github.event_name != 'pull_request'
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+
+      - name: Cache SonarCloud packages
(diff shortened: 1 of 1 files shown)


Phantominumm:
Also I remove the comments from sonarCloud seen in precedent PRs

.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Generate Bandit skips from Ruff
+        id: bandit_skips
+        run: |
+          pip install ruff jq
+          SKIPS=$(ruff rule --all --output-format json \
+            | jq -r '.[] | select(.code | test("^S[0-9]{3}$")) | .code' \
+            | sed 's/S/B/' \
+            | paste -sd, -)
+          echo "skips=$SKIPS" >> $GITHUB_OUTPUT
```
fabcor-maxiv: Why? Is that necessary? Is there an issue if we do not filter those out?

.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+"on":
+  pull_request: null
+  push: null
+  workflow_dispatch: null
+
+permissions:
+  contents: read
+  security-events: write
+  actions: read
```
fabcor-maxiv: Consider moving those to the job level for finer-grained permissions.

.github/workflows/bandit_sonar.yaml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          pip install ruff jq
+          SKIPS=$(ruff rule --all --output-format json \
+            | jq -r '.[] | select(.code | test("^S[0-9]{3}$")) | .code' \
+            | sed 's/S/B/' \
+            | paste -sd, -)
+          echo "skips=$SKIPS" >> $GITHUB_OUTPUT
+
+      - name: Bandit Scan
+        uses: shundor/python-bandit-scan@ab1d87dfccc5a0ffab88be3aaac6ffe35c10d6cd
```
fabcor-maxiv: Do we want to trust `shundor` and that they will keep maintaining this action? Would it be hard to write the job ourselves (`pip install bandit` and run `bandit`)?

.github/workflows/bandit_sonar.yaml
Phantominumm: That was discussed in the PR from mxcubeweb, Ruff is already checking these rules so double checking it with bandit is not useful and we decided to skip it to save time.

.github/workflows/bandit_sonar.yaml
Phantominumm: Yes it's a good and easy solution, I will do it thanks.

.github/workflows/bandit_sonar.yaml
fabcor-maxiv: I would argue that it is not useful to filter out and that it is not useful to maintain this step with such a complex command.

But I guess that now that the filter is here, we can keep it, it does bother.

walesch-yan: I think that one idea of this filter is that if we have a specific rule that is ignored in ruff for that matter, it will also be ignored/skipped here, otherwise we would have to maintain many different lists and keep them all up to date. 

One thing that we could maybe do is to add a comment to explain what this command does

fabcor-maxiv: Aaaah... okay... Now with the comment I understand the point. And that is a good point! Thanks! I agree we need that filter.

walesch-yan (approved):

fabcor-maxiv (approved):
OK, let's try this

marcus-oscarsson:
Thanks alot for having a look guys and a big thanks for @Phantominumm for all the nice work on `mxcubeweb` and `mxcubecore` !

@Phantominumm Have now finished his summer internship. We hope to see more of him in the future for another internship :)

PR #1339, opened 2025-07-22
Update Starlette because of vulnerabilities

Phantominumm:
Dependabot alerts about vulnerability from Starlette version so I update it.
It need fastapi update from video-streamer, so it's waiting the update from there

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -558,24 +558,25 @@ yaml = ["PyYAML"]
 
 [[package]]
 name = "fastapi"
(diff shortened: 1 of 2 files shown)


Phantominumm:
Update video-streamer version that include fastapi 0.116.1 and typing-extensions 4.14.1 and removing websockets.
Update also starlette version to 0.47.2 to remove the vulnerability of precedent one.

walesch-yan (approved):

PR #1340, opened 2025-07-29
ISARAMaint: check if it's possible to read attributes

elmjag:
When setting up polling of the tango attributes, check that it possible to read them. If there are issues, log it and run in a 'fault' mode.

This way the hardware object can handle situations when ISARA's tango device is unavailable or reports issues. For example if the sample changer robot is turned off.

Without this checks, it was problematic to run MXCuBE, if the robot is unavailable.

### DIFF
--- a/mxcubecore/HardwareObjects/ISARAMaint.py
+++ b/mxcubecore/HardwareObjects/ISARAMaint.py
@@ -6,6 +6,7 @@ from tango import (
 )
 
 from mxcubecore.BaseHardwareObjects import HardwareObject
+from mxcubecore.utils.tango import TangoAttributeReadError, add_attribute_channel
 
 """
 The sample changer maintenance hardware object for ISARA2 robot.
@@ -93,14 +94,7 @@ class ISARAMaint(HardwareObject):
     def init(self):
         tangoname = self.get_property("tangoname")
         self.isara_dev = DeviceProxy(tangoname)
-
-        polling = self._get_polling()
-
-        self._poll_attribute(tangoname, "Powered", polling, self._powered_updated)
-        self._poll_attribute(
-            tangoname, "PositionName", polling, self._position_name_updated
-        )
-        self._poll_attribute(tangoname, "Message", polling, self._message_updated)
+        self._setup_attribute_polling(tangoname)
 
     def _get_polling(self):
         """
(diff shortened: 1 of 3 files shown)


mxcubecore/HardwareObjects/ISARAMaint.py
```diff
@@ -6,6 +7,9 @@
 )
 
 from mxcubecore.BaseHardwareObjects import HardwareObject
+from mxcubecore.utils.tango import AttrConnectionError, add_attribute_channel
+
+log = logging.getLogger("HWR")
```
marcus-oscarsson: The logger is already available on any `HardwareObject` as `self.log`, see: https://github.com/mxcube/mxcubecore/blob/6209f2c82c75ffe1f96f019035922a23580db36c/mxcubecore/BaseHardwareObjects.py#L644C9-L644C17

This was added quite some time ago and the idea was to use `self.log` or `self.user_log` for logging. Old habits die hard I guess, and many of us simple kept using `logging.getLogger("HWR")`

`self.log` also adds the class name to the log which can be quite useful. 

I would propose to avoid adding a `log` global in each module and use either `self.log` or   `logging.getLogger("HWR")`

mxcubecore/HardwareObjects/ISARAMaint.py
```diff
@@ -113,18 +110,36 @@ def _get_polling(self):
...
+                self._position_name_updated,
+            )
+            add_attribute_channel(
+                self,
+                tangoname,
+                "Message",
+                polling,
+                self._message_updated,
+            )
+        except AttrConnectionError as ex:
```
marcus-oscarsson: Looks good to me. A question that comes to mind seeing this and remembering some discussions we've had in the past on the possibility to re-connect to devices without restarting the server;
 
- Should we handle this in a more general way perhaps even on the command/channel level
 
That could be the start to adding a `reconnect` that simply tries to re connect to i.e  Tango device ?

Just a thought, this otherwise looks fine !

mxcubecore/utils/tango.py
```diff
@@ -0,0 +1,71 @@
+from tango import DevFailed
+
+from mxcubecore.BaseHardwareObjects import HardwareObject
+from mxcubecore.Command.Tango import TangoChannel
+
+
+class AttrConnectionError(Exception):
```
marcus-oscarsson: I personally find the class name a bit confusing, I understand by reading the code that its the exception that is raised when a certain attribute does not exist or can't be read.  The thing that is a bit confusing is that python also has the concept of `attributes`.

Maybe something more specific, like `TangoAttributeReadError` would be more informative ?

marcus-oscarsson: Another question is, should we not handle this in `hwo.add_channel` ?

marcus-oscarsson:
Thanks @elmjag some really nice ideas. We have been discussing some aspects of  being able to handle "unavailable" devices and reconnecting in the past.  I think there are parts of this PR that takes us in that direction. 

The PR looks fine as it is and we can merge it. 

There are one point that would be interesting to develop further and thats `add_attribute_channel` https://github.com/mxcube/mxcubecore/pull/1340/files#diff-7ef488b719ba1750038618cb461cbbe8c3d7a69347c867264c7f9a76ef2759c5R13. That I think could be implemented in `HardwareObject.add_channel`. I think that this would have the advantage of:

- Keeping the logic to channel connection in one place
- Starting to make it possible/easier to implement something like a `reconnect`

elmjag:
Thanks for reviewing.

I'm on vacation now, I'll address comments one I'm back, around 18 Aug.

mxcubecore/HardwareObjects/ISARAMaint.py
elmjag: Aha, I was not aware of `self.log`, changed!

mxcubecore/utils/tango.py
elmjag: Agree, it's a bit cryptic name. Changed it to `TangoAttributeReadError`.

elmjag: Probably, but that feels like a larger task. Adding this logic will change the behavior of `hwo.add_channel()`, so we'll need to carefully consider changes.

This current implementation is safer version, as it's an add-on function, that you explicitly need to call.

I agree that we should think about adding this feature to `add_channel()` in the future.

elmjag:
Not sure what is going on with SonarQube check.

fabcor-maxiv:
> Not sure what is going on with SonarQube check.

You can ignore. This has been seen before for pull requests made from a fork.

This should have been fixed by <https://github.com/mxcube/mxcubecore/pull/1348>, though. So I wonder... Either the branch for this pull request does not have the fix in it, or the fix does not do what we were expecting.

marcus-oscarsson (approved):

mxcubecore/utils/tango.py
marcus-oscarsson: Ok, should we create an issue for that or do you keep it in the back of your head :) ?

PR #1344, opened 2025-08-05
fix the poetry version in the create-tag.yml file

walesch-yan:
There seems to be a mismatch between the poetry version used in the `create-tag` workflow and the one we fix with conda. This will cause an error, when the workflow tries to bump the version. 

This PR is fixing the version to the same value as in conda, lock and pre-commit files.

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -13,39 +13,45 @@ jobs:
       contents: write
       packages: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           token: ${{ secrets.CI_TOKEN }}
-      - name: Set up Python 3.10
-        uses: actions/setup-python@v4
+
+      - name: Set up Conda environment with Micromamba
+        uses: mamba-org/setup-micromamba@v2
         with:
-          python-version: "3.10"
+          cache-environment: true
+          create-args: >-
+            python=3.10
+          environment-file: conda-environment-dev.yml
+          post-cleanup: all
 
       - name: Update version
         id: update-version
         run: |
-            pip install --upgrade pip
-            pip install poetry
-            poetry version minor
+            ${MAMBA_EXE} run --name mxcubecore poetry version minor
             git config --global user.email "oscarsso@esrf.fr"
             git config --global user.name "Marcus Oskarsson"
             git add -A
             git commit -m "[skip ci] Bumped minor version"
             git push -f
-            poetry build
+            ${MAMBA_EXE} run --name mxcubecore poetry build
+
       - name: Publish package to PyPI
         id: publish-pacakge
+        # yamllint disable rule:line-length
         run: |
-          poetry config pypi-token.pypi ${{ secrets.PYPI }}
-          poetry publish
+          ${MAMBA_EXE} run --name mxcubecore poetry config pypi-token.pypi ${{ secrets.PYPI }}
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
Alternative (and maybe better?) solution would be to use the Conda YAML environment file in this GitHub workflow (like it is done in other workflows).

walesch-yan:
Oh definitely better, keeps us from having to change it on multiple places, thanks @fabcor-maxiv

.github/workflows/create-tag.yml
```diff
@@ -13,38 +13,42 @@ jobs:
       contents: write
       packages: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
walesch-yan, line 16: While at it, might as well update to the latest version here

.github/workflows/create-tag.yml
```diff
@@ -13,38 +13,42 @@ jobs:
...
+            python=3.10
+          environment-file: conda-environment-dev.yml
+          post-cleanup: all
 
       - name: Update version
         id: update-version
         run: |
-            pip install --upgrade pip
-            pip install poetry
             poetry version minor
```
fabcor-maxiv: We might need something like this to use the Poetry command installed within the Conda environment:

```suggestion
            ${MAMBA_EXE} run --name mxcubecore poetry version minor
```

Or maybe it is possible to modify the `PATH` environment variable. I do not recall if I tried this.

This workflow is near impossible to try in isolation (without actually creating tags)... so maybe it is simpler to revert back to your first suggestion... :/

.github/workflows/create-tag.yml
```diff
@@ -13,38 +13,42 @@ jobs:
...
           poetry publish
+
       - name: Read package version
         id: set-tag
         # yamllint disable rule:line-length
         run: |
           pip install --upgrade pip
           pip install toml
-          pip install poetry
           echo "tag_name=v$(poetry version -s)" >> $GITHUB_ENV
```
fabcor-maxiv:
```suggestion
          echo "tag_name=v$(${MAMBA_EXE} run --name mxcubecore poetry version -s)" >> $GITHUB_ENV
```

walesch-yan:
So as suggested, I changed the workflow to setup the conda enironment. This will additionally install openldap (and potentially other packages, we add to the environment file) during the workflow, which is not really needed for it, but on the other hand it will keep us from updating the version tag everywhere

fabcor-maxiv (approved):
It looks fine to me.

Did you manage to test somehow?

.github/workflows/create-tag.yml
walesch-yan: You are right, for some reason, I thought that setup-micromamba would automatically activate the environment. So I added this before every poetry command.

I also tested the latest change here on a fork, where it seems to work including the version tag creation on github and the package build. I did not test the part that publishes to PyPI, as it would require some more setup

walesch-yan:
> It looks fine to me.
> 
> Did you manage to test somehow?

Yes, I tested it on a seperate fork after setting up some secrets similar to the ones used here, I did just omit the publish to PyPI part during the test

.github/workflows/create-tag.yml
fabcor-maxiv: I think `setup-micromamba` does not activate the environment. I recall it being a bit weird to activate the environment, but it was quite a long time ago, maybe things are better now.

Issue #1345, opened 2025-08-06
About `take_snapshots`


fabcor-maxiv:
Is [`take_snapshots` an integer value](https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/model/common.py#L12)?

* https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/model/common.py#L12

Or is [`take_snapshots` a boolean value](https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/model/queue_model_objects.py#L1703)?

* https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/model/queue_model_objects.py#L1703
* https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/configuration/mockup/web/beamline_config.yml#L87

If `take_snapshots` is a boolean value, why is [`number_of_snapshots` set to the value of `take_snapshots`](https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/HardwareObjects/abstract/AbstractCollect.py#L809) or [the other way around](https://github.com/mxcube/mxcubeweb/blob/d764ee8932c67587c86df7f3e565a114b7565b6f/mxcubeweb/core/components/queue.py#L1317-L1319)?

* https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/HardwareObjects/abstract/AbstractCollect.py#L809
* ```python
  dc_model.take_snapshots = HWR.beamline.collect.get_property(
      "num_snapshots", self.app.DEFAULT_NUM_SNAPSHOTS
  )
  ```
  https://github.com/mxcube/mxcubeweb/blob/d764ee8932c67587c86df7f3e565a114b7565b6f/mxcubeweb/core/components/queue.py#L1317-L1319

What is [`enable_take_snapshots`](https://github.com/mxcube/mxcubecore/blob/4947e54ce1fab8a16908fa27beb0c809c20980d5/mxcubecore/queue_entry/data_collection.py#L53)?

[...]

PR #1351, opened 2025-08-19
Remove DataCollectionQueueEntry unused attributes

fabcor-maxiv:
These two attributes seem completely unused.

Looking through the related commits in the git log, it seems like these were added with some specific purpose in mind, but was never fully implemented and then never fully cleaned up either.

GitHub: https://github.com/mxcube/mxcubecore/issues/1345


marcus-oscarsson:
Thanks, those attributes can indeed be removed. Well spotted !

PR #1352, opened 2025-08-20
Prevent SonarQube to run om PRs made from forks

marcus-oscarsson:
Prevent SonarQube to run om PRs made from forks

I think we need to check `head` and not `base`. I think the base is always `mxcubecore` and head is whatever repo the PR is made from.

### DIFF
--- a/.github/workflows/bandit_sonar.yaml
+++ b/.github/workflows/bandit_sonar.yaml
@@ -61,7 +61,7 @@ jobs:
     if: |
       (github.event_name == 'push' && github.repository == 'mxcube/mxcubecore') ||
       (github.event_name == 'pull_request' &&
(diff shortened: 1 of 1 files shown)


.github/workflows/bandit_sonar.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
     if: |
       (github.event_name == 'push' && github.repository == 'mxcube/mxcubecore') ||
       (github.event_name == 'pull_request' &&
-       github.event.pull_request.base.repo.full_name == 'mxcube/mxcubecore')
+       github.event.pull_request.head.repo.full_name == 'mxcube/mxcubecore')
```
marcus-oscarsson, line 64: So I changed this to `head` instead of `base`

PR #1353, opened 2025-08-21
Add missing links to dev docs pages

fabcor-maxiv:
Some pages in the dev docs were missing in the index, so they were not visible in the table of contents.


fabcor-maxiv:
Also, one should know that in URLs it is better to use dashes than underscores to separate words. As far as I understand, it is mostly SEO reasons (Search Engine Optimization), in other words it is to make it easier to find things in Google and co. But also some other reasons as well, it is all explained on the Internet. For example here: <https://stackoverflow.com/questions/119312/urls-dash-vs-underscore>.

With Sphinx, if we have a document `dev/foo_bar.md` we end up having `dev/foo_bar.html`. So we should aim for `dev/foo-bar.html` instead.

In general we should be careful when renaming pages, because it potentially creates broken links. But these two pages were basically unlinked and invisible to search engines, and very likely no one ever bookmarked those pages, so it is very likely safe to rename them now.

On the other hand, let's be pragmatic, if we rename our other pages, I doubt we will break anything, but still it might make sense to do it separately.

walesch-yan (approved):

PR #1357, opened 2025-08-25
Handle image mode from camera's get_last_image

fabcor-maxiv:
The `take_snapshot` method of the `SampleView` hardware object calls the `get_last_image` method of the camera hardware object and assumes the result is always RGB.

There are some cases where the camera returns grayscale data (`L` mode). So this change allows `get_last_image` to optionally return image mode.

### DIFF
--- a/mxcubecore/HardwareObjects/SampleView.py
+++ b/mxcubecore/HardwareObjects/SampleView.py
@@ -138,20 +138,27 @@ class SampleView(AbstractSampleView):
 
         self._last_oav_image = path
 
-    def take_snapshot(self, overlay_data=None, bw=False):
+    def take_snapshot(self, overlay_data=None, bw=False) -> Image:
         """
         Get snapshot with overlaid data.
 
         Args:
-            overlay_data (str): base64 encoded image to lay over camera image
-            bw (bool): return grayscale image
+            overlay_data: base64 encoded image to lay over camera image
+            bw: return black and white image
 
         Returns:
-            (Image) rgb or grayscale image
+            RGB or black and white image
         """
-        data, width, height = self.camera.get_last_image()
 
-        img = Image.frombytes("RGB", (width, height), data)
+        last_image = self.camera.get_last_image()
+
+        if len(last_image) == 3:
+            data, width, height = last_image
+            mode = "RGB"
+        elif len(last_image) == 4:
+            data, width, height, mode = last_image
+
+        img = Image.frombytes(mode, (width, height), data)
 
         if overlay_data:
             overlay_data = base64.b64decode(overlay_data)


fabcor-maxiv:
As far as I understand, at MicroMAX, the on-axis camera delivers image data in gray scale for the higher zoom levels (instead of RGB). So we want the `SampleView` to be able to receive an explicit image mode instead of always assuming the data is RGB. This change is meant to be backwards compatible, so if no image mode is delivered RGB is assumed by default.

fabcor-maxiv:
Also I updated the docstring to correct the fact that the `bw` option will deliver a "black and white" image (and not "gray scale" as it is currently documented).

fabcor-maxiv:
> Also I updated the docstring to correct the fact that the `bw` option will deliver a "black and white" image (and not "gray scale" as it is currently documented).

Well, I now tested this option, and it seems to me like this `bw` option does nothing. It seems to me like I with or without this option (i.e. with or without going through `img.convert("1")` I still get a colourful image.

elmjag:
I feel this API for `get_last_image()` method should be documented somewhere, as it's a bit tricky and not obvious. At least we need to write down that `mode` is supposed to be **PIL** `mode`.

However, I'm not sure where. Is there some kind of abstract class for sample view cameras?

fabcor-maxiv:
> Is there some kind of abstract class for sample view cameras?

No, there is no abstract class for the sample view camera, and indeed I feel like that is what is missing. I might be able to write one. Should I do it for this pull request or separately?

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -138,20 +138,27 @@ def save_snapshot(self, path, overlay=None, bw=False):
...
+            bw: return black and white image
 
         Returns:
-            (Image) rgb or grayscale image
+            RGB or black and white image
         """
-        data, width, height = self.camera.get_last_image()
 
-        img = Image.frombytes("RGB", (width, height), data)
+        last_image = self.camera.get_last_image()
```
marcus-oscarsson, line 153: I'm wondering if it would not be better to handle the proposed logic in the `camera` object (LimaTangoVideo): https://github.com/mxcube/mxcubecore/blob/c6e9797d514f3df6278bccd6e559071dd307ac78/mxcubecore/HardwareObjects/TangoLimaVideo.py#L29

We could handle it like we do in the video streamer: https://github.com/mxcube/video-streamer/blob/874577793070609ad37e6c8f22d864b5f2bc1520/video_streamer/core/camera.py#L260

@walesch-yan and I think also @mockoocy implemented this function:
https://github.com/mxcube/video-streamer/blob/874577793070609ad37e6c8f22d864b5f2bc1520/video_streamer/core/camera.py#L234

marcus-oscarsson: The ideal would perhaps be if we could use the video streamer to get the image data, like this we would not need to duplicate this code and the video/image bits would be better encapsulated.

fabcor-maxiv: OK, so @marcus-oscarsson you are suggesting it would be better if the sample view camera hardware object's `get_last_image` method always returns RGB images, do I understand this correctly?

That was my original suggestion as well (before making this pull request). One of the problem with this solution is that the sample view expects raw image data (not a PIL `Image`). So in case the raw image data is not RGB, we get this sequence of events:

1. [sample view camera converts the non-RGB raw image data to an RGB PIL `Image` object](https://github.com/mxcube/mxcubecore/blob/c6e9797d514f3df6278bccd6e559071dd307ac78/mxcubecore/HardwareObjects/TangoLimaVideo.py#L40)
1. [sample view camera converts the RGB PIL `Image` back to raw image data (this time it is RGB)](https://github.com/mxcube/mxcubecore/blob/c6e9797d514f3df6278bccd6e559071dd307ac78/mxcubecore/HardwareObjects/TangoLimaVideo.py#L44)
1. [sample view converts the raw image data to a PIL `Image`](https://github.com/mxcube/mxcubecore/blob/c6e9797d514f3df6278bccd6e559071dd307ac78/mxcubecore/HardwareObjects/SampleView.py#L154)

So that seems like unnecessary back and forth.

Maybe we should decide to only pass PIL `Image` objects around, if that is easier to work with (which seems to be the case).

fabcor-maxiv: The goal for this pull request is to propose a change that is backwards compatible while still improving things quite a lot (no unnecessary conversion to PIL `Image`). If we want to do something else, then I need more time to think about it.

fabcor-maxiv: We do not seem to be using `TangoLimaVideo` at MAX IV currently.

fabcor-maxiv: _[Sorry, I am in "brain dump" mode, writing things as I go, trying to figure things out and untangle this.]_

marcus-oscarsson: I see, it is indeed a bit complicated all of this. I think that we should, longterm, try to use the **VideoStreamer** project, that would be my recommendation. Like this we get  a robust, resource efficient and scalable streaming solution. For this reason; I think that we should perhaps not add a `AbstractCamera/Video` but rather delegate that to the **VideoStreamer**. We can create a HardwareObject that starts, stops and interfaces the streaming process(es) (perhaps via @walesch-yan Argus project) 

For this PR, I would recommend that `get_last_image()` would return a PIL `Image` object. Like this the logic above can be moved up.

If you are not using `TangoLimaVideo` what are you using, I guess its a Tango camera for the BZOOM ?

walesch-yan: I agree with @marcus-oscarsson that using the `video-streamer` here might be beneficial, as it already implements some of the conversions and supports different types of input (MJPEG, Lima, Redis,...), so we would have to do less duplication, and we can always expect the type of frame we get from the stream. This solution might require a little bit of change and enforces the usage of the `video-streamer`. But this might help with cleaning up lots of the different camera hwo files and implementations around them.

On the other hand, for a more backward-compatible solution, I think we can do as you said, by reducing the conversions and fixing the return type to be different from raw image bytes. One other issue might be that, as I know of, there are multiple camera hwos defined in mxcube, but only the `TangoLimaVideo` (and the mockup camera) implements the `get_last_image` function. So we would need an AbstractClass here

fabcor-maxiv: > For this PR, I would recommend that `get_last_image()` would return a PIL `Image` object. Like this the logic above can be moved up.

That would be a backward incompatible change, wouldn't it? I was not planning to deal with this kind of backward incompatible change right now. I guess we would close this pull request and do the conversion to RGB within our current sample view camera hardware object for now. And maybe we would look at this kind of refactoring later.

> I think that we should, longterm, try to use the VideoStreamer project, that would be my recommendation.

Feels like you want to get rid of the hardware object for the sample view camera, is that right? If that is the case, I would not know how to retrieve a frame from the video stream for the snapshots during data collection. For the snapshots via the UI button, I guess we could read the content of the canvas, maybe. Otherwise maybe we can extract an image from the websocket, I guess it must be possible but I would not know how. Or am I misunderstanding something?

> We can create a HardwareObject that starts, stops and interfaces the streaming process(es)

Currently, at MAX IV, the video-streamer process for the on-axis camera feed is always running, independently from MXCuBE (managed by `supervisord`). I guess it feels more stable like this and less of a hassle to deal with.

marcus-oscarsson: > That would be a backward incompatible change, wouldn't it? I was not planning to deal with this kind of backward incompatible change right now. I guess we would close this pull request and do the conversion to RGB within our current sample view camera hardware object for now. And maybe we would look at this kind of refactoring later.

Yes it would in theory be backward incompatible change but it also depends on where its used. I don't think that `get_last_image()` is used in more in one or two places so making that update would be easy. You can perhaps otherwise create a new function `get_last_image_rgb` that does what you need ?

> Feels like you want to get rid of the hardware object for the sample view camera, is that right? If that is the case, I would not know how to retrieve a frame from the video stream for the snapshots during data collection. For the snapshots via the UI button, I guess we could read the content of the canvas, maybe. Otherwise maybe we can extract an image from the websocket, I guess it must be possible but I would not know how. Or am I misunderstanding something?

We would like to move of all the video streaming/handling code and delegate it to the **VideoStreamer** that handles this better. The **VideoStreamer** application can run in a different process or computer and provides as I mentioned above a more robust and resource efficient way to handle video streams. The HardwareObject that interfaces the **VideoStreamer** can know (or needs to) about your **VideoStreamer** process like this the image can be retrieved from it.

fabcor-maxiv: > like this the image can be retrieved from it.

How? Does video-streamer have an API to obtain one frame? I am not following what you mean... :D

marcus-oscarsson: Sort of, there are a few options. 

 - There is the Argus `HardwareObject` still in a separate branch: https://github.com/mxcube/mxcubecore/compare/develop...yw-features-argus#diff-9115696d3e29be8c899e862a3cb8f50960b47362faffb2498bbc6ce38ccc9f8fR22
 - You can use the Redis streamer to get the last image
 - Perhaps less clean, but you can also capture video stream and use OpenCV to get the image

walesch-yan: > * There is the Argus `HardwareObject` still in a separate branch: [develop...yw-features-argus#diff-9115696d3e29be8c899e862a3cb8f50960b47362faffb2498bbc6ce38ccc9f8fR22](https://github.com/mxcube/mxcubecore/compare/develop...yw-features-argus#diff-9115696d3e29be8c899e862a3cb8f50960b47362faffb2498bbc6ce38ccc9f8fR22)

Currently, this `HardwareObject` does not have the possibility to extract single frame, as it is meant to coordinate different streams and send queries to a GRPC server, the extraction happens on a different server for manipulation, but can be added as a single function here as well (this is all still in developpment)

> * You can use the Redis streamer to get the last image

This might be a good solution in my opinion, so there is the possibility to have a second output source for the video-streamer in the form of a Redis stream (we might have to verify the format of the image in that case), the advantage is that additional metadata, like the timestamp, for example, is sent through this as well.

fabcor-maxiv: > > You can use the Redis streamer to get the last image
> 
> This might be a good solution in my opinion, so there is the possibility to have a second output source for the video-streamer in the form of a Redis stream (we might have to verify the format of the image in that case), the advantage is that additional metadata, like the timestamp, for example, is sent through this as well.

I'll take your word for it that it is possible, I have not looked into Redis camera at all, for sure we will not do this change now though. But yes, this seems like it could be a good solution in the future (even if it feels a bit overkill to me for just snapshots).

marcus-oscarsson: Sure its not for now, it was to highlight a possible future direction.  Its also a bit overkill to have all the streaming code and hardware objects for video in MXCuBE if its only used for snapshots ;), one can see it the the other way ;)

For now I would say, that we either keep the PR as it is or if you see an easy way to move/re-use some of the existing logic in `TangoLimaCamera`. What `HardwareObject` are you using in the end ?

fabcor-maxiv: I close this pull request. We solved this in our hardware object. We have a custom hardware object, not sure why, maybe we should check again if we can use one of the available hardware objects.

PR #1362, opened 2025-09-04
units: add function to convert KeV to eV

elmjag:
Expands unit conversion module with KeV to eV function.

(We use this in our yet-to-be upstreamed code)

### DIFF
--- a/mxcubecore/utils/units.py
+++ b/mxcubecore/utils/units.py
@@ -47,6 +47,13 @@ def ev_to_kev(ev: float) -> float:
     return ev / 1000.0
 
 
+def kev_to_ev(ev: float) -> float:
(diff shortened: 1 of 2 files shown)


marcus-oscarsson:
:), I just want to clarify/remind everyone of a previous discussion we had that basically amounted to that; The use of these functions that "convert/handles" orders of magnitude are optional. Its up to each site to decide whether to use them or not. I'm not opposing the change but I don't see the direct use of it.

What I do believe is (more) important is to document the expected input or output units, or comments in inline code.

rhfogh:
@marcus-oscarsson Seconded.

marcus-oscarsson (approved):

PR #1377, opened 2025-09-12
don't use the deprecated __iter__() method in get_object_by_role()

elmjag:
Don't use the deprecated `HardwareObjectNode.__iter__()` method when looking up object by role.

This fixes 1212 warnings when running the tests :)

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -23,6 +23,7 @@ from __future__ import absolute_import
 import ast
 import copy
 import enum
+import itertools
 import logging
 import typing
 import warnings
@@ -520,14 +521,14 @@ class HardwareObjectNode:
         #
         # Look-up object by role the old way.
         #
-        objects = [self]
 
-        for curr in objects:
-            result = curr._objects_by_role.get(role)
-            if result is None:
-                objects.extend(obj for obj in curr if obj)
-
-            else:
+        #
+        # Look for object recursively, starting with our self
+        # and then checking the child objects.
+        #
+        for object in itertools.chain([self], *self.__objects):
+            result = object._objects_by_role.get(role)
+            if result is not None:
                 return result
 
     def _objects_names(self) -> List[Union[str, None]]:


mxcubecore/BaseHardwareObjects.py
```diff
@@ -520,14 +521,14 @@ def get_object_by_role(self, role: str) -> Union["HardwareObject", None]:
...
-            result = curr._objects_by_role.get(role)
-            if result is None:
-                objects.extend(obj for obj in curr if obj)
-
-            else:
+        #
+        # Look for object recursively, starting with our self
+        # and then checking the child objects.
+        #
+        for object in itertools.chain([self], *self.__objects):
```
marcus-oscarsson, line 529: Just being curious, whats the advantage of using `chain` ?

elmjag: `__objects` is a list of lists. That is, it looks something like this:

`[[<objXX>], [objXY], [objXZ]]`

It needs to be flatten, if you want to make a single nested for-loop.

marcus-oscarsson: Oh I see, I thought it was already flat

rhfogh:
It would be nice to get rid of the warning, but I think this is the wrong way to do it. Basically you are inlining the deprecated method; what we ought to do is to stop using it.

The internal attribute __objects is only (?) used in functions that are either deprecated or commented as 'use for XML configured files only'. Rather than getting rid of the warnings, I think we should keep them, and get rid of all the functions using __objects. IMHO the only place where we store contained objects should be in the _objects_by_role dictionary. Who disagrees with this?

Another point: do we really need the functionality of get_object_by_role that loops over all contained objects, recursively, and look for some object anywhere in the tree that happens to have the right role name? Can we not get to a point where we know where the objects are kept and get them where they live, instead of this wide-area searching? What do you say, @beteva ?

rhfogh (changes requested):
I do not think we should do this.

marcus-oscarsson:
I think thats two different discussions, I think this is simply to get rid of the usage of `__iter__` on "self" as is done in `for curr in objects`. Right @elmjag

rhfogh:
Well, let us ask ourselves why we deprecated \_\_iter\_\_. We did this because we thought it was a bad idea to treat hardware objects as lists of contained objects. And we put in the deprecation so we could find the places where this happened and eventually get rid of them. What elmjag is doing is to find a way of treating hardware objects as lists of contained objects while avoiding the warning - to my mind the exact equivalent of simply removing the deprecation warning.

I see only two sensible courses of action here: If we want to get rid of storing contained contained objects in duplicate data structures like \_\_objects, then this is a bad idea, because it hides the places where this is going on and reduces the motivation for fixing it. If we do *not* want to get rid of \_\_objects then let us take that decision up front and keep doing it. In which case we might as well keep \_\_iter\_\_ AFAIAC. Why not?

As I see it the point of deprecation and linting warnings is that they highlight problematic parts of the code and prod you to fix them - and hiding the warnings without fixing the problem defeats the purpose.  If your patient has a fever, you either cure him or accept it. Hiding the thermometer only makes it less likely that you will ever find a solution.

marcus-oscarsson:
The reason why I thought we deprecated, the overloading/use of `__iter__` in this way is because it leads to somewhat odd and unclear code. I don't think its evident that one would iterate over a list of contained HardwareObjects if one does `for _o in myobj`. I would say that that is the main reason to remove the overloading of  `__iter__` or overloading of any python special function, for that matter, unless its immediately clear what it does.

How hardware objects is accessed is another question. As far as I remember we said that we would keep `get_object_by_role` and make it "protected" by renaming it `_get_object_by_role`.

marcus-oscarsson:
As I see it this PR is completely fine and it removes the use of overloaded standard python functions which is very nice

elmjag:
The motivation for this change is to be able to use `get_object_by_role()` for optional objects, without triggering the deprecated warning.

For example this call: https://github.com/mxcube/mxcubecore/blob/f7c8bf2b5283e92d9b29a4f0b231c0938cc7e97a/mxcubecore/HardwareObjects/abstract/ISPyBAbstractLims.py#L35

We, at MAX IV, don't use `icat_client`, thus when we initialize our LIMS we end up  running this code:
https://github.com/mxcube/mxcubecore/blob/f7c8bf2b5283e92d9b29a4f0b231c0938cc7e97a/mxcubecore/BaseHardwareObjects.py#L528

That triggers the deprecation warning.

Currently there is no way to use `get_object_by_role()` for optional objects without deprecation warnings. As a bonus, this change brings us one step closer to be able to remove the `__iter__()` function.

I think the question of deprecating `__objects` is a separate issue. I agree it's a good idea, but we should deal with that separately.

rhfogh:
If you can wait a bit on this one, I am trying to come up with a solution that does not depend on using \_\_objects so that at least we are not further entrenching that.

elmjag:
> In which case we might as well keep `__iter__ `AFAIAC. Why not?
> 

We want to move away from using python magic methods. It make it harder to read our code when somewhat unexpectedly iteration is overloaded for hardware objects.

rhfogh:
I have made an alternative PR, #1378, that should do the same job but also move towards using the yaml-type attributes instead of sticking with the xml-only ones. What do you think?

marcus-oscarsson:
Its possible that I dissect this problem too much but I really think its two separate issues:

  - Overloading of standard python methods often leads to unexpected results and its hard to understand whats going on. In this case its `__iter__`.
  
  - The internal structure(s) for child objects, child objects are currently stored in both `__objects`  and `_objects_by_role`. I agree that its a good idea to sort this one out.

Part of the problem is perhaps that we have not yet done what we said that we would do, namely; rename `get_object_by_role` to `_get_object_by_role`. Like this could add a deprecation warning on `get_object_by_role`, for instance: **"get_object_by_role" should only be used for internal or optional HardwareObject lookup, use "_get_object_by_role" instead.**

Id be happy to merge this PR and look at the second point in another PR.

elmjag:
@rhfogh How about we add a warning on this line 

https://github.com/mxcube/mxcubecore/pull/1377/files#diff-3822e4536634bf372f5d0cc7f0fd470b4a5fc2a2741127801c417464b9480779R532

Something like 'usage of __objects attribute is deprecated'

This way this PR will not hide fact that code is using `__objects` attribute. All 'modern' code, calling `get_object_by_role()` will not show any warning. But the code that does end up looking-up the object the 'old-style' way, will.

elmjag:
> Its possible that I dissect this problem too much but I really think its two separate issues:
> 
> * Overloading of standard python methods often leads to unexpected results and its hard to understand whats going on. In this case its `__iter__`.
> * The internal structure(s) for child objects, child objects are currently stored in both `__objects`  and `_objects_by_role`. I agree that its a good idea to sort this one out.
> 
> Part of the problem is perhaps that we have not yet done what we said that we would do, namely; rename `get_object_by_role` to `_get_object_by_role`. Like this could add a deprecation warning on `get_object_by_role`, for instance: **"get_object_by_role" should only be used for internal or optional HardwareObject lookup, use "_get_object_by_role" instead.**
> 
> Id be happy to merge this PR and look at the second point in another PR.

I agree, that this is two separate issues, best handled in separate PRs.

elmjag:
Ehh, sorry about closing this PR. I think I clicked the wrong button.

rhfogh (approved):
Ok, go ahead. See comments in #1378 for further discussion

marcus-oscarsson:
Thanks !

PR #1378, opened 2025-09-12
Fix get_object_by_role avoiding __objects attribute

rhfogh:
This should have the same effect as PR #1377, but avoids using the __objects attribute, removes some obsolete code and comments, and moves the code base a step towards yaml configuration instead of sticking with xml-based attributes.

This would be a good time to simplify the structure of BaseHardwareObject inheritance - all HardwareObjects are subtypes of ConfiguredObject, HardwareObjectNode, and HardwareObjectMixin so they ought to be merged (and this would get rid of a warning from using ConfiguredObject._hwobj_by_role in a function defined in HardwareObjectNode). But for now let us just do this fix.

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -202,6 +202,40 @@ class ConfiguredObject:
         """
         return list(self._hwobj_by_role.keys())
 
+    def get_object_by_role(self, role: str) -> Union["HardwareObject", None]:
+        """Get hardware object by role.
+
+        Args:
+            role (str): Role.
+
+        Returns:
+            Union[HardwareObject, None]: Hardware object.
+        """
+        role = str(role).lower()
+        result = self.objects_by_role.get(role)
+
+        if result is None:
+            # Does breadth-first search of all contained HWO to look for any HWO
+            # that has the right role name
+            # This entire block is deprecated.
+            #
+            # NB if this block is removed, the entire function can be removed
+            # and replaced by self.objects_by_role.get(role)
+            objects = list(self.objects_by_role.values())
+            for obj in objects:
+                objs_by_role = obj.objects_by_role
+                result = objs_by_role.get(role)
+                if result is None:
+                    objects.extend(objs_by_role.values())
+                else:
+                    warnings.warn(
+                        "%s.%s: get_object_by_role for nested objects is Deprecated."
+                        "Avoid"
+                        % (self.__class__.__name__, role)
+                    )
+                    break
+        return result
+
     def print_log(
         self,
         log_type: str = "HWR",
@@ -489,32 +523,6 @@ class HardwareObjectNode:
         else:
             for obj in self.__objects[index]:
                 yield obj
-
-    def get_object_by_role(self, role: str) -> Union["HardwareObject", None]:
-        """Get hardware object by role.
-
-        Args:
-            role (str): Role.
-
-        Returns:
-            Union[HardwareObject, None]: Hardware object.
-        """
-        role = str(role).lower()
-        result = self._hwobj_by_role.get(role)
-
-        if result is None:
-            # Can we not get rid of this at some point? Please?
-            #
-            # Does breadth-first search of all contained HWO to look for any HWO
-            # that has the right role name
-            objects = list(self._hwobj_by_role.values())
-            for obj in objects:
-                result = obj.get_object_by_role(role)
-                if result is None:
-                    objects.extend(obj._hwobj_by_role.values())
-                else:
-                    break
-        return result
 
     def _objects_names(self) -> List[Union[str, None]]:
         """Return hardware object names.
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -230,8 +230,7 @@ class ConfiguredObject:
                 else:
                     warnings.warn(
                         "%s.%s: get_object_by_role for nested objects is Deprecated."
-                        "Avoid"
-                        % (self.__class__.__name__, role)
+                        "Avoid" % (self.__class__.__name__, role)
                     )
                     break
         return result
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -500,35 +500,18 @@ class HardwareObjectNode:
             Union[HardwareObject, None]: Hardware object.
         """
         role = str(role).lower()
-        #
-        # A hack to emulate get_object_by_role() for objects loaded from YAML config
-        # files.
-        #
-        # When HWOBJ is loaded from YAML, we don't populate it's '_objects_by_role'
-        # dictionary, thus that normal code path to look-up and object by role does
-        # not work.
-        #
-        # However, objects are attached to the parents _hwobj_by_role object via
-        # assignment. Try accessing using that attribute.
-        #
-        if hasattr(self, "_hwobj_by_role"):
-            obj = self._hwobj_by_role.get(role, None)
+        result = self._hwobj_by_role.get(role)
 
-            if obj is not None:
-                return obj
-
-        #
-        # Look-up object by role the old way.
-        #
-        objects = [self]
-
-        for curr in objects:
-            result = curr._objects_by_role.get(role)
-            if result is None:
-                objects.extend(obj for obj in curr if obj)
-
-            else:
-                return result
+        if result is None:
+            # Can we not get rid of this at some point? Please?
+            #
+            # Does depth-first search of all contained HWO to look for any HWO
+            # that has the right role name
+            for obj in self._hwobj_by_role.values():
+                result = obj.get_object_by_role(role)
+                if result is not None:
+                    return result
+        return result
 
     def _objects_names(self) -> List[Union[str, None]]:
         """Return hardware object names.
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -505,12 +505,15 @@ class HardwareObjectNode:
         if result is None:
             # Can we not get rid of this at some point? Please?
             #
-            # Does depth-first search of all contained HWO to look for any HWO
+            # Does breadth-first search of all contained HWO to look for any HWO
             # that has the right role name
-            for obj in self._hwobj_by_role.values():
+            objects = list(self._hwobj_by_role.values())
+            for obj in objects:
                 result = obj.get_object_by_role(role)
-                if result is not None:
-                    return result
+                if result is None:
+                    objects.extend(obj._hwobj_by_role.values())
+                else:
+                    break
         return result
 
     def _objects_names(self) -> List[Union[str, None]]:


rhfogh:
I notice this failed quite a few tests. Probably one should do some abstract class refactoring and look at the tests as well as this. I can give it a go, but better first see what the reactions are to this one.

marcus-oscarsson:
I'm not sure that I understand entirely what you are after here. Is it to fix how the `HardwareObjects` are stored internally or is it how the lookup is made ?

I think this is very welcome if the aim is to remove duplication between `__objects`, `__object_names` and `_objects_by_role`.

Probably somewhat adjacent to this the usage of `get_object_by_role`. My recollection is that we decided to keep it and rename it to `_get_object_by_role` and that it has a place for "optional" hardware objects, meaning lookup from `self`, do we all agree on that ?

rhfogh:
This has a two-fold purpose: First it gets rid of the deprecation warning - which is the same that @elmjag wanted to do. Second it is a step towards cleaning up the implementation, which is something that was missing from elmjag's proposal. As it stands I simply think this is a better way of achieving the same object, and I hope that we can evaluate it on that basis, without prejudice to any deeper disagreements.


I think the heavily linter-dominated approach to cleanup that we have currently is not only missing the main problems, but actually tends to make them harder to solve. The basic reason we are getting these deprecation warnings is that we are in the middle of an unresolved change-over between XML-based and yaml-based implementation, with different functions, internal attributes etc. Meanwhile our code is, in this area - a mess. Which becomes obvious as soon as you look for a few minutes about how the code is organised. This is quite legitimate - we are in the middle of a complex transition - but the main problem here is *not* that we are using a dodgy syntax, but that our code organisation is way off. And by concentrating on silencing the warnings while keeping the code structure as-is, we are making it look like everything is fine and there is less need for getting on to the bits that most need fixing. We are sweeping things under the carpet as it were. Of course it is much easier to get rid of the warnings by minor syntax changes than by refactoring our hardware object handling, but surely it is the refactoring that we really need, not just getting rid of warnings for its own sake? 


Then there is a separate question (which you probably picked up from my in-code comments). On the usage of get_object_by_role, we agreed that it was needed to deal with "optional" hardware objects. Only for that you do not need having a depth-first search of an object tree looking for somewhere where there might be an object with that role. I am not aware of the use case  where you need to get hold of an object but are unable to decide or figure out where in the object tree it ought to live. I am sorry if it is me being obtuse, but did we ever agree on that part? Let me point out that as this code was originally written, the depth-first search was only carried out for xml-configured objects; it was never implemented for yaml-configured objects. I am willing to bow to a consensus on this point, but on general principles I would argue that it is better to give your objects a determined location where you can find them, than to put them in an indeterminate location that you have to search for.

marcus-oscarsson:
> This has a two-fold purpose: First it gets rid of the deprecation warning - which is the same that @elmjag wanted to do. Second it is a step towards cleaning up the implementation, which is something that was missing from elmjag's proposal. As it stands I simply think this is a better way of achieving the same object, and I hope that we can evaluate it on that basis, without prejudice to any deeper disagreements.

Now I understand. I think @elmjag PR solves one of the issues in an acceptable way. I agree that there is a second much deeper problem that also needs to be solved. However the later will need much more testing and discussing.

> 
> I think the heavily linter-dominated approach to cleanup that we have currently is not only missing the main problems, but actually tends to make them harder to solve. The basic reason we are getting these deprecation warnings is that we are in the middle of an unresolved change-over between XML-based and yaml-based implementation, with different functions, internal attributes etc. Meanwhile our code is, in this area - a mess. Which becomes obvious as soon as you look for a few minutes about how the code is organised. This is quite legitimate - we are in the middle of a complex transition - but the main problem here is _not_ that we are using a dodgy syntax, but that our code organisation is way off. And by concentrating on silencing the warnings while keeping the code structure as-is, we are making it look like everything is fine and there is less need for getting on to the bits that most need fixing. We are sweeping things under the carpet as it were. Of course it is much easier to get rid of the warnings by minor syntax changes than by refactoring our hardware object handling, but surely it is the refactoring that we really need, not just getting rid of warnings for its own sake?
> 

I'm not certain that we can say that we have a linter-dominated approach. I agree that there sometimes seems to be a tendency, not only in this project but in others as well, to satisfy linters instead of focusing on the more important structural issues. Both aspects is of-course important to look at, we want both code that is easy to understand, follows accepted good practices and is well structured. I think there are plenty of people amongst us that are looking at the structural issues as well. You are right that moving to YAML configuration is a good opportunity to cleanup some very old code. 

We should in this case then remove `__objects` and `__object_names` entirely and only rely on `_objects_by_role`.

> Then there is a separate question (which you probably picked up from my in-code comments). On the usage of get_object_by_role, we agreed that it was needed to deal with "optional" hardware objects. Only for that you do not need having a depth-first search of an object tree looking for somewhere where there might be an object with that role. I am not aware of the use case where you need to get hold of an object but are unable to decide or figure out where in the object tree it ought to live. I am sorry if it is me being obtuse, but did we ever agree on that part? Let me point out that as this code was originally written, the depth-first search was only carried out for xml-configured objects; it was never implemented for yaml-configured objects. I am willing to bow to a consensus on this point, but on general principles I would argue that it is better to give your objects a determined location where you can find them, than to put them in an indeterminate location that you have to search for.

Is not the lookup method a non question if the agreed mechanism of accessing a hardware object is via the "root" object in this case `beamline.[some-object]`. There is no depth in a list of objects. The method `_get_object_by_role` can then be used from a given hardware object to access a direct child

As I see it, this PR that addresses three issues:

1.  - Removing the call to `__itter__` on self (the implementation on `HardwareObject` still exists though so we should then remove that)

2.  - Removing the duplication of references to contained objects, `__objects`, `__object_names`
  
4.  - Removing the possibility to use `get_object_by_role` and replace it with `_get_object_by_role` that only performs direct lookup on an object.

The two last points, especially the very last would need some care full planing. We can probably implement `get_object_by_role` that gives a deprecation warning if the depth is greater than 1. I would perhaps recommend starting with that and item 1 and 2 in the list above.

beteva:
> I am not aware of the use case where you need to get hold of an object but are unable to decide or figure out where in the object tree it ought to live.

Such an object for us is bliss. It is only used at ESRF, so we do not want to pollute the beamline object with a place holder for it. On the other hand we need it in several hardware objects, so get_object_by_role is used in this case. @rhfogh is this what you were asking? Do we want to have something like beamline.my_synchrotron.whatever_we_need ?

elmjag:
> This should have the same effect as PR #1377, but avoids using the __objects attribute, removes some obsolete code and comments, and moves the code base a step towards yaml configuration instead of sticking with xml-based attributes.
> 

This PR does not have the same effect as PR #1377. The #1377 is 'backward compatible' change. The code that relies on using `__objects` attribute will still work. It only removes one call to the deprecated `__iter__` method.

This PR is a breaking change. Any code that relies on `__objects` will need to be updated. Of course we can work on that, but I think that is a bigger project.

I would personally suggest that we first deprecate,  `__objects` and `__objects_names` attribute. Give facilities some time to migrate their hardware objects, and then remove it.

rhfogh:
Admittedly I remember we have discussed this before, but I did not remember a stable outcome.

You are right that one would prefer to avoid having beamline,my_synchrotron, or ESRFBeamline. But could it not be possible for you to decide on a location and put it there? It does not have to be directly under beamline. Could you not decide to put it under, say, ESRFBeamInfo or  ESRFMultiCollect? It would not have to be a location wehere Bliss was necessarily *used* all the time, as long as it was known to live there. You only need the looping search if it is in an *unknown* location under beamline.

My latest PR version puts in the deprecation warning, as suggested by @marcus-oscarsson But we can remove that again if we agree on it.

rhfogh:
@elmjag As I read the code, I do not think that my PR is actually a breaking change (and I did check first).

If you look in  HardwareObjectNode._add_object and HardwareObjectNode.resolve_references and generally do a global search for _objects_by_role and \_\_objects you will find that everything that is in _objects_by_role is also in __objects and vice versa, once object loading is finished.
And if you search for _hwobj_by_role you will find that it is added to only in HardwareRepository.load_from_yaml and HardwareRepository._attach_xml_objects, and that in the latter all objects in _objects_by_role (and hence all objects in \_\_objects) are copied into _hwobj_by_role. 

So, summa summarum, I would say that _hwobj_by_role will contain all hardware objects, no matter how configured, once the loading process is finished.  Which should mean that it will give the same result as the previous code, and as your code. OK, with one exception: The new code will also make nested role search in yaml-configured code. I I am really quite sure that there is no case of yaml-configured code looking for an optional hardware object using get_object_by_role and breaking because there is an object with the name role name lower down.

mxcubecore/BaseHardwareObjects.py
```diff
@@ -202,6 +202,40 @@ def get_roles(self) -> List[str]:
...
+                if result is None:
+                    objects.extend(objs_by_role.values())
+                else:
+                    warnings.warn(
+                        "%s.%s: get_object_by_role for nested objects is Deprecated."
+                        "Avoid"
+                        % (self.__class__.__name__, role)
+                    )
+                    break
+        return result
```
elmjag: This is not really needed.

There is no code that relies on having nested look-up in the `_hwobj_by_role` attribute, as that feature have never existed.

The only code that, hypothetically speaking, is using nested look-up, is the code the looks for objects in `__objects` attribute.

Note, object references are _not_ duplicated between `_hwobj_by_role` and  `__objects` attributes. The objects references are stored _either_ in `_hwobj_by_role` _or_  `__objects`.

rhfogh: As per my last comment, I do not think this is the case any longer (though it probably was in an earlier version).

rhfogh:
My PR still fails, near as I can see because while all actual HardwareObjects will have get_object_by_role, moving it to a different superclass gives problems with the tests. But there is no point in trying to clear that up unless we first agree that this is actually the way to go.

rhfogh:
@marcus-oscarsson My latest PR version aims to solve your 1) and to do a safe part of 2) (doing all of it will indeed require more care). As for 3) I think there is an alternative (specifically using hwobj.objects_by_role.get()), but first we need to agree with Antonia whether it is an OK thing to do.

elmjag:
> @elmjag As I read the code, I do not think that my PR is actually a breaking change (and I did check first).
> 

If you trace calls to `get_object_by_role()`, you can see that _some_ hardware objects have only `_hwobj_by_role` attribute populated,  others have only `__objects` populated.

I have applied this [patch](https://gist.github.com/elmjag/f5af5d9f5bc56795a372d2ab11f00f72) to mxcubecore and started the demo beamline. Here is the resulting [output](https://gist.github.com/elmjag/0623ce118dee56e4f896bcee312bf996). You can also see that results from `__objects`  are returned in [some](https://gist.github.com/elmjag/0623ce118dee56e4f896bcee312bf996#file-gistfile1-txt-L16) [cases](https://gist.github.com/elmjag/0623ce118dee56e4f896bcee312bf996#file-gistfile1-txt-L23).

This is with mxcubecore v1.372.0 and mxcubeweb v4.520.0. I used this command to start the demo beamline:

`mxcubeweb-server -s /<...>/mxcube/web/ui/build`

rhfogh:
@elmjag OK, You were right and I as wrong - I shall withdraw this PR. Much as I would have liked there to be, there does not seem to be any way of cleaning up this particular function without doing a lot of work, so feel free to go ahead with #1377. Thanks for doing the extra work to clarify this. 

I could try to sneak out on a technicality - it is strictly speaking true that all objects are present in _hwobj_by_role *once the objects are loaded*, but I had not considered that get_object_by_role was being called from the init() functions *before* the objects had finished loading. I still firmly believe that the right thing to do is to finish the move to yaml configuration and get rid of this mess (including getting rid of get_object_by_role wherever it can be done). Let us hope that some time in the future someone may find the time to look at it.

marcus-oscarsson:
@elmjag, very nice initiative well done !

marcus-oscarsson:
One way to do this would be to deprecate and remove the XML configuration once that we have migrated to YAML and then clean up or rationalize the object loading.

rhfogh:
@marcus-oscarsson I'd be in favour of that. Do you think you can muster a consensus that making this happen is worth the work it would take?

marcus-oscarsson:
Yes, I believe/hope so. I think that we need to have some patience but its good that we are already exploring and looking in to what needs to be done. I think it will be much easier once we have a migrated to YAML and have a mechanism for validating that the configuration files are what we expect. The conversion tool that you wrote helps alot and I think the transition to YAML can be quite fast. 

We probably need to get the `AbstractDiffractometer` in place and discuss object access again to make sure that we get it right.

PR #1382, opened 2025-09-16
tests: fix deprecated warnings in beamline tests

elmjag:
Use `hwobj.id` attribute and `get_by_id()` method in the tests.

This fixes these warnings:

```
test/test_beamline_ho_id.py::TestBeamlineHoId::test_beamline_id
  /<...>/mxcube/core/test/test_beamline_ho_id.py:46:
    UserWarning: Beamline.get_hardware_object is Deprecated.
    Use get_by_id instead
    ho = test_object.get_hardware_object("diffractometer.sampx")

test/test_beamline_ho_id.py::TestBeamlineHoId::test_beamline_id
  /<...>/mxcube/core/test/test_beamline_ho_id.py:47:
    UserWarning: Beamline.get_id is Deprecated.
    Use hwobj.id instead
    ho_id = test_object.get_id(ho)
```

### DIFF
--- a/test/test_beamline_ho_id.py
+++ b/test/test_beamline_ho_id.py
@@ -38,11 +38,9 @@ class TestBeamlineHoId:
     def test_beamline_id(self, test_object):
         # Test if we can retrieve a object located directly on
         # the beamline object
-        ho = test_object.get_hardware_object("diffractometer")
-        ho_id = test_object.get_id(ho)
-        assert "diffractometer" == ho_id
+        ho = test_object.get_by_id("diffractometer")
+        assert "diffractometer" == ho.id
 
         # Test if we can get an object further down the structure
-        ho = test_object.get_hardware_object("diffractometer.sampx")
-        ho_id = test_object.get_id(ho)
-        assert "diffractometer.sampx" == ho_id
+        ho = test_object.get_by_id("diffractometer.sampx")
+        assert "diffractometer.sampx" == ho.id


marcus-oscarsson:
Just a thought is it really a good idea to deprecate `get_id` and use `.id` instead. I think the mixup/confusion with the python id() lies closer to hand if we use `.id`. I know that the naming is partly my initiative, but is it only me or is there a risk for confusion ?  We could perhaps call things function something else, like `hardware_object_id`, `hwobj_id`, `full_address` or `full_name` ? 

I'm otherwise fine with this and its a very welcome contribution.

fabcor-maxiv:
Yes, it also seems to me like `id` is not a good name: too generic, too often used, too much at risk of introducing confusing.

elmjag:
> Just a thought is it really a good idea to deprecate `get_id` and use `.id` instead. I think the mixup/confusion with the python id() lies closer to hand if we use `.id`. I know that the naming is partly my initiative, but is it only me or is there a risk for confusion ? We could perhaps call things function something else, like `hardware_object_id`, `hwobj_id`, `full_address` or `full_name` ?
> 
> I'm otherwise fine with this and its a very welcome contribution.

Personally I think maybe it's not optimal, but fine.

Python does not have a `foo.id()` method. It have `id(foo)` function. I don't there is a huge risk of confusing `foo.id` with `foo.id()`. `foo.id()` expression give you an error.

Also, the `id()` is quite obscure function. I don't think I have ever seen it being used in 'normal' python code.

marcus-oscarsson:
Ok, Ill put it in the list of things to consider :), its quite low priority after all an I will not let it hold this PR back. I just wanted to know what your thoughts where.

marcus-oscarsson (approved):

marcus-oscarsson:
thanks :+1:  you did well to merge, sorry Ive been a bit busy

PR #1383, opened 2025-09-16
add 'Logging' section to developer documentation

elmjag:
Document the convention of how logging shall be performed in hardware objects code.

Here is the rendered [result](https://mxcubecore--1383.org.readthedocs.build/en/1383/dev/contributing.html#logging).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -285,6 +285,28 @@ One can also run the Ruff linter and Ruff formatter locally with commands like
 To get Ruff to automatically fix some of the issues it detects,
 one can run a command like this: `poetry run ruff check --fix`.
 
+#### Logging
+
+When logging hardware object messages, use the `self.log` or `self.user_log` attributes.
+
+These attributes refer to logger objects from the standard `logging` module.
+`self.log` sends messages to `HWR` log, `user_log` to `user_level_log` log.
+The `self.log` will prefix log entries with the hardware object's class name.
+
+Below is an example of logging in a hardware object.
+
+```python
+class MyHardwareObject(HardwareObject):
+    def my_method(self):
+        # log messages to 'HWR' log
+        self.log.info("an information level message")
+        self.log.warning("a warning message")
+
+        # log message to 'user_level_log' log
+        self.user_log.info("an information level message")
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
Do you feel like adding something like this?

```python
try:
    output = transform(input)
except Exception:
    # The `exception` method of the logger adds the traceback to the log entry
    # On the other hand, variables must be added explicitly to the log entry
    # These two elements are essential for proper troubleshooting
    self.log.exception("Could not transform: %s", input)
```

elmjag:
> Do you feel like adding something like this?
> 
> ```python
> try:
>     output = transform(input)
> except Exception:
>     # The `exception` method of the logger adds the traceback to the log entry
>     # On the other hand, variables must be added explicitly to the log entry
>     # These two elements are essential for proper troubleshooting
>     self.log.exception("Could not transform: %s", input)
> ```

I feel this should be it's own dedicated paragraph and example. I'll make another PR with that addition.

fabcor-maxiv (approved):

CONTRIBUTING.md
```diff
@@ -285,6 +285,28 @@ One can also run the Ruff linter and Ruff formatter locally with commands like
...
+```python
+class MyHardwareObject(HardwareObject):
+    def my_method(self):
+        # log messages to 'HWR' log
+        self.log.info("an information level message")
+        self.log.warning("a warning message")
+
+        # log message to 'user_level_log' log
+        self.user_log.info("an information level message")
+        self.user_log.error("an error have been encountered")
```
marcus-oscarsson, line 307: Should we mention that `debug` and `exception` should not be used on the `user_log`.

elmjag: Sure. I think we need to add a rationale why `debug` and `exception` should be avoided. 

Does this text make sense:

> Avoid using `debug()` and `exception()` level logging to the `user_log`. The user log is aimed to convey information to the users. The debug and exception logging should be used for logging information suitable for developers, to  help with trouble-shooting.

marcus-oscarsson: Perfect, I guess we should add some logic in the `user_log` that relays debug and exception levels to `HWR` ?

PR #1375, opened 2025-09-11
#1374 Added lims name

antolinos:
This closes #1374 

Still to be tested in real environment

PR #1384, opened 2025-09-17
Add energy scan in ICATLIMS. 

beteva:
Cleanup. Remove unnecessary HO file.


ruff.toml
```diff
@@ -1797,26 +1797,13 @@ convention = "google"
...
-    "PTH113",
-    "PTH118",
-    "PTH123",
-    "RET502",
+    "S108",
     "S603",
-    "SIM115",
-    "T201",
-    "TRY300",
+    "T201"
```
fabcor-maxiv: Please, do not add any new items in these "per-file ignore" lists. If we do want to ignore some rules in the code, then we have to add the "ignores" directly in the code, on the line of the rule violation, for example: `  # noqa: T201 we use print because...`.

beteva: @fabcor-maxiv, I think there is a problem somewhere in ruff. I fixed  ESRFEnergyScan.py with isort, but ruff still complains about the imports. Any help will be appreciated.

fabcor-maxiv: @beteva, there is no issue in the GitHub Actions workflows, so I do not know what problem you are referring to. I downloaded the file and running Ruff locally did not bring up any import-related issue either.

beteva: OK, I'll remove the exception, see if it passes in github. It is I001 for the ESRFEnergyScan.py

fabcor-maxiv: Note that while Ruff aims to be (at least partially) compatible with Isort, they are still two different tools that will provide different outputs under some circumstances. I recommend to use Ruff only. We could discuss and decide to disable the "import sorting" features of Ruff and use the real Isort tool instead.

beteva: Well, I've only used isort because ruff complained about the sorting and I could not see what the problem was.

fabcor-maxiv: Ah ok, I understand. I thought Ruff would be able to reorder imports on its own, I was not expecting that one would need to use isort for this.

beteva: Apparently the problem is gone. One rule exception less!

fabcor-maxiv: > Apparently the problem is gone. One rule exception less!

Great! :)

mxcubecore/HardwareObjects/ESRF/ESRFEnergyScan.py
```diff
@@ -1,16 +1,52 @@
...
+
+"""
+Example xml_ configuration:
+
+.. code-block:: xml
+
+ <object class="ESRF.ESRFXRFSpectrum">
+   <object href="/bliss" role="controller"/>
+   <cfgfile>/users/blissadm/local/beamline_configuration/misc/15keV.cfg</cfgfile
+>
```
fabcor-maxiv:
```suggestion
   <cfgfile>/users/blissadm/local/beamline_configuration/misc/15keV.cfg</cfgfile>
```

mxcubecore/HardwareObjects/ESRF/ESRFEnergyScan.py
```diff
@@ -393,5 +422,5 @@ def StoreEnergyScanThread(db_conn, scan_info):
             else:
                 asso = {"blSampleId": blsample_id, "energyScanId": escan_id}
                 db_conn.associate_bl_sample_and_energy_scan(asso)
-    except Exception as e:
-        print(e)
+    except Exception as err:
+        print(err)
```
fabcor-maxiv:
```suggestion
        logging.getLogger("HWR").exception("Could not store energy")
```

beteva:
Thanks @fabcor-maxiv

mxcubecore/HardwareObjects/ESRF/ESRFEnergyScan.py
```diff
@@ -226,159 +240,173 @@ def do_chooch(self, elt, edge, directory, archive_directory, prefix):
...
+            eroi_min /= 1000.0
+            eroi_max /= 1000.0
+        self.ctrl.mca.set_roi(
+            eroi_min,
+            eroi_max,
+            channel=1,
+            element=self.energy_scan_parameters["element"],
+            atomic_nb=self.energy_scan_parameters["atomic_nb"],
+        )
+        print(self.ctrl.mca.get_roi())
```
fabcor-maxiv:
```suggestion
        logging.getLogger("HWR").debug("MCA ROI is now: %s", self.ctrl.mca.get_roi())
```

and then it should be possible to remove `T201` from `ruff.toml`.

ruff.toml
```diff
@@ -1795,28 +1795,12 @@ convention = "google"
 ]
 "mxcubecore/HardwareObjects/ESRF/ESRFEnergyScan.py" = [
     "ARG002",
-    "BLE001",
     "C901",
-    "ERA001",
-    "FLY002",
-    "G002",
-    "ICN001",
+    "DTZ005",
```
fabcor-maxiv: To be moved to the file itself.
```suggestion
```

antolinos:
This PR also replaces #1375 and #1391 too

marcus-oscarsson (approved):

Issue #1367, opened 2025-09-05
Ruff configuration - Linting rule exceptions


marcus-oscarsson:
This is issue is meant to contain the discussions on the various linting rules, or categories of rules that we for some reason would like to exclude or configure differently.

fabcor-maxiv:
Disable [Ruff rule N999 `invalid-module-name`](https://docs.astral.sh/ruff/rules/invalid-module-name/) because we have many modules named in `PascalCase` instead of the Python conventional `snake_case`:

* https://github.com/mxcube/mxcubecore/pull/1403

marcus-oscarsson:
Also discussed in https://github.com/mxcube/mxcubecore/issues/1562

PR #1386, opened 2025-09-17
[WIP] AbstractDffractometer, AbstractSampleView and inheriting classes

beteva:
- Create AbstractDiffractometer class:
  -   inherits from HardwareObject and uses wait_ready, abort methods and the state handling.
  -  defines some methods, properties and the DiffractometerHead, DiffractometerPhase enums.
  -  defines some scan types
  -  in the init method defines standard access to different equipment (motors and nstate equipment).
- Change DiffarctometerMockup accordingly
- Implement MicroDiffractometer for the Arinax MD2S
- Refactor AbstractSampleViwe and SampleView to include the methods, previously attached to the diffractometer.

Dependant on [mxcubeweb/pull/1881](https://github.com/mxcube/mxcubeweb/pull/1881)


mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,378 @@
...
+from enum import Enum, unique
+
+from mxcubecore.BaseHardwareObjects import HardwareObject
+
+__copyright__ = """ Copyright © 2010-2022 by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+@unique
+class DiffractometerHead(Enum):
```
rhfogh: Does anyone have a Smargon? If they do which head would that count as?

rhfogh (changes requested):
There seems to be something weird here: Both MicroDiffractometer and AbstractDiffractometer containclasses called AbstractDiffractometer, DifractometerHead, DiffractometerHead, DiffractometerConstraint and the MicroDiffractometer does not inherit from the AbstractDiffractometer.

Is this a temporary measure, or what is going on?

beteva:
Please, do not waste your time to review the code - it is not ready jet.

rhfogh:
@beteva Ah well, I guess it does say WIP in the title ;-)

beteva:
Replaces #695

beteva:
Dear @rhfogh, @all,
 the AbstractDiffractometer/SampleView is ready to be reviewed now.

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+    <object role="fluo_detector" href="/udiff_fluodet"/>
+    <object role="cryostream" href="/udiff_cryostream"/>
+    <object role="front_light" href="/udiff_frontlight_inout"/>
+    <object role="back_light" href="/udiff_backlight_inout"/>
+    <object role="beamstop" href="/udiff_beamstop"/>
+    <object role="capillary" href="/udiff_capillary"/>
+    <object role="aperture" href="/udiff_aperturemot"/>
+    <object role="zoom" href="/udiff_zoom"/>
+  </nstate_equipment>
+</object>
```
fabcor-maxiv: I think this belongs in the class' docstring.

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+    DiffractometerHead,
+    DiffractometerPhase,
+)
+
+__copyright__ = """ Copyright © by the MXCuBE collaboration """
+__license__ = "LGPLv3+"
+
+
+class MicroDiffractometer(AbstractDiffractometer):
+    """Microdiff with Exporter implementation of AbstartDiffractometer"""
```
fabcor-maxiv:
```suggestion
    """Microdiff with Exporter implementation of AbstractDiffractometer"""
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+
+    def abort(self):
+        """Immediately terminate action."""
+        self._exporter.execute("abort")
+
+    @property
+    def _get_hwstate(self) -> str:
+        """Get the hardware state, reported by the MD2 application.
+        Returns:
+            (str): The state.
```
fabcor-maxiv: There has to be an empty line.

There is no need to have types in the docstring that are already documented as type hint annotations.

```suggestion
        """Get the hardware state, reported by the MD2 application.
        
        Returns:
            The state.
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+            return self._exporter.read_property("HardwareState")
+        except AttributeError:
+            return "Ready"
+
+    @property
+    def _get_swstate(self) -> str:
+        """Get the software state, reported by the MD2 application.
+        Returns:
+            (str): The state.
+        """
```
fabcor-maxiv, line 105:
```suggestion
        """Get the software state, reported by the MD2 application.
        
        Returns:
            The state.
        """
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+        return self._exporter.read_property("State")
+
+    def get_state(self):
+        """Get the diffractometer general state.
+        Returns:
+            (enum 'HardwareObjectState'): state
+        """
```
fabcor-maxiv, line 119:
```suggestion
    def get_state(self) -> HardwareObjectState:
        """Get the diffractometer general state."""
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+        except ValueError:
+            self._state = self.STATES.UNKNOWN
+        return self._state
+
+    @property
+    def _ready(self) -> bool:
+        """Get the "Ready" state - software and hardware.
+        Returns:
+            (bool): True if both "Ready", False otherwise.
+        """
```
fabcor-maxiv, line 131:
```suggestion
        """Get the "Ready" state - software and hardware.

        Returns:
            ``True`` if both "Ready", ``False`` otherwise.
        """
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+        return self._get_swstate == "Ready" and self._get_hwstate == "Ready"
+
+    def _wait_ready(self, timeout: [None | float] = None):
+        """Wait timeout seconds until status is ready.
+        Args:
+            timeout(float): Timeout [s]. None means infinite timeout.
+        """
```
fabcor-maxiv, line 139:
```suggestion
        """Wait timeout seconds until status is ready.

        Args:
            timeout: Timeout in seconds. ``None`` means infinite timeout.
        """
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+
+    @property
+    def _ready(self) -> bool:
+        """Get the "Ready" state - software and hardware.
+        Returns:
+            (bool): True if both "Ready", False otherwise.
+        """
+        return self._get_swstate == "Ready" and self._get_hwstate == "Ready"
+
+    def _wait_ready(self, timeout: [None | float] = None):
```
fabcor-maxiv:
```suggestion
    def _wait_ready(self, timeout: float | None = None) -> None:
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+        with Timeout(timeout, RuntimeError("Timeout waiting for status ready")):
+            while not self._ready:
+                sleep(0.5)
+
+    def get_motors(self):
+        """Get the dictionary of all configured motors or the ones to use.
+        Returns:
+            (dict): Dictionary key=role: value=hardware_object
+        """
+        return self.get_movable_motors()
```
fabcor-maxiv:
```suggestion
    def get_motors(self) -> dict[str, HardwareObject]:
        """Get the dictionary of all configured motors or the ones to use.

        Returns:
            Dictionary where keys are roles and values are hardware objects.
        """
        return self.get_movable_motors()
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+        Args:
+            motors_positions_dict (dict): Dictionary {motor_role: target_value}.
+            simultaneous (bool): Move the motors simultaneously (True - default) or not.
+            timeout (float): optional - timeout [s],
+                             if timeout = 0: return at once and do not wait,
+                             if timeout is None: wait forever (default).
+        Raises:
+            TimeoutError: Timeout
+            KeyError: The name does not correspond to an existing motor
+        """
```
fabcor-maxiv, line 162:
```suggestion
    def set_value_motors(
        self,
        motors_positions_dict: dict,
        simultaneous: bool = True,
        timeout: float | None = None,
    ) -> None:
        """Move specified motors to the requested positions.

        Args:
            motors_positions_dict: Dictionary ``{motor_role: target_value}``.
            simultaneous: Move the motors simultaneously (``True``, default) or not.
            timeout: optional - timeout [s],
                             if timeout = 0: return at once and do not wait,
                             if timeout is None: wait forever (default).
        Raises:
            TimeoutError: Timeout
            KeyError: The name does not correspond to an existing motor
        """
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+                self.wait_ready(timeout)
+
+    def get_value_motors(self, motors_list: [list | None] = None) -> dict:
+        """Get the positions of diffractometer motors. If the motors_list
+           is empty, return the positions of all the availble motors.
+        Args:
+            motors_list (list): List of motor roles.
+        Returns:
+            (dict): role: position dictionary
+        """
```
fabcor-maxiv:
```suggestion
    def get_value_motors(self, motors_list: list[str] | None = None) -> dict:
        """Get the positions of diffractometer motors.
        
        If ``motors_list`` is empty,
        return the positions of all the available motors.

        Args:
            motors_list: List of motor roles.

        Returns:
            Position dictionary
        """
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+        motors_positions_dict = super().get_value_motors(motors_list)
+        if not self.in_kappa_mode:
+            motors_positions_dict.update({"kappa": None, "kappa_phi": None})
+        return motors_positions_dict
+
+    def get_movable_motors(self):
+        """Get the dictionary of all the motors which can be used.
+        Returns:
+            (dict): Dictionary key=role: value=hardware_object
+        """
```
fabcor-maxiv:
```suggestion
    def get_movable_motors(self) -> dict[str, HardwareObject]:
        """Get the dictionary of all the motors which can be used.

        Returns:
            Dictionary where keys are roles and values are hardware objects.
        """
```

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+                motors_dict.update(elem)
+
+        return motors_dict
+
+    @property
+    def get_head_type(self) -> DiffractometerHead:
+        """Get the head type
+        Returns:
+            head_type(enum): Head type
+        """
```
fabcor-maxiv:
```suggestion
        """Get the head type.

        Returns:
            Head type.
        """
```

ruff.toml
fabcor-maxiv, line 1: Please, let's not add any exception in this file. From now on, exceptions should be added directly in the code, on the same line where the violation appears.

ruff.toml
```diff
@@ -24,6 +24,8 @@ extend-ignore-names = [
     "HWR",  # `HWR` is often used as a shorthand for `mxcubecore.HardwareRepository`.
 ]
 
+[lint.pylint]
+max-args = 10 # some methods require more than 5 arguments - PLR0913
```
fabcor-maxiv, line 30: Surely this can also be solved by adding an exception in the code where the violation appears. I see no reason to change the global value, for just a handful of exceptions.

mxcubecore/queue_entry/base_queue_entry.py
```diff
@@ -923,7 +923,7 @@ def center_before_collect(view, dm, queue, sample_view):
 
     queue.pause(True)
     pos, shape = None, None
-
+    dm = dm or None
```
fabcor-maxiv, line 926: That is strange. What is the intention here? This deserves an explanation.

mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
```diff
@@ -16,409 +16,102 @@
...
+    <device role="sampx" hwrid="/diff-sampx-mockup"/>
+    <device role="sampy" hwrid="/diff-sampy-mockup"/>
+    <device role="kappa" hwrid="/diff-kappa-mockup"/>
+    <device role="kappa_phi" hwrid="/diff-kappaphi-mockup"/>
+  </motors>
+  <nstate_equipment>
+    <object role="fshutter" href="/fast-shutter-mockup"/>
+    <object role="beamstop" href="/beamstop-mockup"/>
+  </nstate_equipment>
+ </object>
```
fabcor-maxiv, line 40: This belongs in the class docstring.

mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
```diff
@@ -16,409 +16,102 @@
...
+        if self.in_kappa_mode:
+            self.update_state(HardwareObjectState.BUSY)
+            time.sleep(random.uniform(0, 2.2))
+        self.update_state(HardwareObjectState.READY)
+
+    def get_pixels_per_mm(self):
+        """Get the pixel/mm values.
+        Returns:
+            (tuple): x,y [pixel/mm]
+        """
```
fabcor-maxiv, line 84:
```suggestion
    def get_pixels_per_mm(self) -> tuple[float, float]:
        """Get the pixel/mm values.

        Returns:
            ``x`` and ``y`` values in pixels per millimetre.
        """
```

mxcubecore/HardwareObjects/abstract/AbstractSampleView.py
```diff
@@ -46,15 +46,18 @@ def __init__(self, name):
...
 
     @property
     def camera(self):
         return self.get_object_by_role("camera")
 
     @abc.abstractmethod
     def get_snapshot(
-        self, overlay: Union[bool, str] = True, bw=False, return_as_array=False
+        self,
+        overlay: Union[bool, str] = True,
```
fabcor-maxiv:
```suggestion
        overlay: bool | str = True,
```

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+        self.username = name
+        self.current_phase = None
+        self.head_type = None
+        self.current_constraint = None
+        self.timeout = 3  # default timeout 3 s
+
+    def init(self):
+        """Initialise username property.
+        Initialise the equipment, defined in the configuration file
+        """
```
fabcor-maxiv, line 219:
```suggestion
    def init(self) -> None:
        """Initialise username property.

        Initialise the equipment, defined in the configuration file.
        """
```

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+                _hobj = self.get_object_by_role(role)
+                self.nstate_equipment_hwobj_dict[role] = _hobj
+                setattr(self, role, _hobj)
+                self.connect(_hobj, "valueChanged", _hobj.update_value)
+            except KeyError:
+                logging.getLogger("HWR").warning(
+                    "No nstate (discrete positions) equipment configured"
+                )
+
+    def get_motors(self) -> dict:
```
fabcor-maxiv, line 248:
```suggestion
    def get_motors(self) -> dict[str, HardwareObject]:
```

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+        Returns:
+            Dictionary key=role: value=hardware_object
+        """
+        return self.motors_hwobj_dict.copy()
+
+    def get_nstate_equipment(self) -> dict:
+        """Get the dictionary of all the nstate (discrete positions) equipment.
+        Returns:
+            Dictionary key=role: value=hardware_object
+        """
```
fabcor-maxiv, line 261:
```suggestion
    def get_nstate_equipment(self) -> dict[str, HardwareObject]:
        """Get the dictionary of all the nstate (discrete positions) equipment.

        Returns:
            Dictionary key=role: value=hardware_object
        """
```

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -61,53 +67,280 @@ def combine_images(img1, img2):
...
+            y = dy + (phiz * p_y) + beam_position[1]
+
+        except AttributeError as err:
+            raise NotImplementedError from err
+        return x, y
+
+    def start_manual_centring(self, nb_click=3):
+        """Do the manual centring procedure.
+        Args:
+           nb_click (int): Number of clicks.
```
fabcor-maxiv:
```suggestion
    def start_manual_centring(self, nb_click: int = 3):
        """Do the manual centring procedure.

        Args:
           nb_click: Number of clicks.
```

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -124,21 +357,24 @@ def get_snapshot(self, overlay=None, bw=False, return_as_array=False):
 
         return buffered
 
-    def save_snapshot(self, path, overlay=None, bw=False):
+    def save_snapshot(
+        self, filename: str, overlay: [str | None] = None, bw: bool = False
+    ):
```
fabcor-maxiv, line 421:
```suggestion
    def save_snapshot(
        self, filename: str, overlay: str | None = None, bw: bool = False
    ) -> None:
```

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -124,21 +357,24 @@ def get_snapshot(self, overlay=None, bw=False, return_as_array=False):
...
         Save a snapshot to file.
 
         Args:
-            path (str): The filename.
-            overlay(str): Image data with shapes and other items to display on the snapshot
-            bw(bool): return grayscale image
+            filename(str): The filename.
+            overlay(str): Image data with shapes and other items to display
+                          on the snapshot
+            bw(bool): return grayscale image. Default False
```
fabcor-maxiv:
```suggestion
            filename: The filename.
            overlay: Image data with shapes and other items to display
                          on the snapshot
            bw: return grayscale image. Default False
```

fabcor-maxiv (commented):
Random comments (I did not get through everything).

fabcor-maxiv (commented):
Random comments (I did not get through everything).

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Actually it is not HardwareObject, but can be any type of NState

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Actually it is not HardwareObject but any type of Motor (microdiff, blis, sardana...)

mxcubecore/HardwareObjects/MicroDiffractometer.py
beteva: We do not import HardwareObjectState, but use self.STATES instead. I think it is too much to import it only for the sake of the documentation.

mxcubecore/HardwareObjects/MicroDiffractometer.py
beteva: We need to make a proper documentation what are the reserved roles for a diffractometer. This is here just for convenience and will be replaced in near future by an yaml format file, and will be removed from this file.

mxcubecore/HardwareObjects/MicroDiffractometer.py
beteva: > There has to be an empty line.

I was relying on the pre-commit to sort this kind of formatting.

fabcor-maxiv: Fair enough :)

Sadly nearly all docstring-related rules are disabled globally currently:

<https://github.com/mxcube/mxcubecore/blob/d295958e141b51088967eb38c1af7c4cd1d4eb55/ruff.toml#L11>

I did this to make the transition easier. At the time I thought I would get around to improving this situation quickly after creating this situation, but that did not happen. I should come up with a plan how to get out of this inconvenient situation.

mxcubecore/HardwareObjects/MicroDiffractometer.py
fabcor-maxiv: * I think you can use quotes to avoid the import: `-> "HardwareObjectState"`.

* Also one can use something like this:

  ```python
  import typing

  if typing.TYPE_CHECKING:
      import something
  ```

  -- https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

* And finally, this kind of things get imported anyway, don't they? So I doubt it has a negative impact if we import it just for the type hint annotation (but I might be wrong).

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Would a lower class do, like AbstractMotor? Alternatively, dict[str, HardwareObject] at least excludes a lot.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Again, a lower abstract class??

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+                    "No nstate (discrete positions) equipment configured"
+                )
+
+    def get_motors(self) -> dict:
+        """Get the dictionary of all configured motors or the ones to use.
+
+        Returns:
+            Dictionary key=role: value=hardware_object
+        """
+        return self.motors_hwobj_dict.copy()
```
rhfogh, line 254: Try using types.MappingPropcyType (avoids the copy and is still read-only)

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def get_nstate_equipment(self) -> dict:
+        """Get the dictionary of all the nstate (discrete positions) equipment.
+        Returns:
+            Dictionary key=role: value=hardware_object
+        """
+        return self.nstate_equipment_hwobj_dict.copy()
```
rhfogh, line 262: Try using types.MappingPropcyType (avoids the copy and is still read-only)

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+    States:
+        HardwareObjectStates: READY, BUSY, FAULT
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.motors_hwobj_dict = {}
```
rhfogh, line 207: motors_hwobj_dict and nstate_equipment_hwobj_dict should start with underscore and not be accessed from outside the object.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def init(self):
+        """Initialise username property.
+        Initialise the equipment, defined in the configuration file
+        """
+        self.username = self.get_property("username") or self.username
+        # motors
+        for role in self.config.motors:
+            try:
+                _hobj = self.get_object_by_role(role)
```
rhfogh, line 225: This looks rather clunky. Is this temporary for xml config, or is this the way we want configuration to look, or what? Would it work with yaml config?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+        # use only the available motors
+        mot_hwobj_dict = self.get_motors()
+
+        tout = timeout
+        if simultaneous:
+            tout = 0
+
+        self.update_state(HardwareObjectState.BUSY)
+        for key, val in motors_positions_dict.items():
```
rhfogh, line 294: Normally I would do
```
hwobj = mot_hwobj_dict.get(key)
if hwobj is None:
  ...
else:
  ...
```
avoiding the try: ... except: ...

Is this better?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+    def get_nstate_equipment(self) -> dict:
+        """Get the dictionary of all the nstate (discrete positions) equipment.
+        Returns:
+            Dictionary key=role: value=hardware_object
+        """
+        return self.nstate_equipment_hwobj_dict.copy()
+
+    # -------- Motor Groups --------
+
+    def set_value_motors(
```
rhfogh, line 266: Should this not be called 'set_motor_values'? You are setting values, not motor hwobj.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+                msg = f"Invalid motor name {key}"
+                raise RuntimeError(msg) from err
+
+        # wait for the end of move of all the motors, if needed
+        if simultaneous:
+            for key in motors_positions_dict:
+                mot_hwobj_dict[key].wait_ready(timeout)
+        self.update_state(self.get_state())
+
+    def get_value_motors(self, motors_list: list | None = None) -> dict:
```
rhfogh, line 307: Should this not be called 'get_motor_values'? You are setting values, not motor hwobj.


Also, should the parameter not be called motor_role_list or similar? motors_list should IMHO be a list of motor objects.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+                try:
+                    mot_pos_dict[role] = float(motor.get_value())
+                except TypeError:
+                    msg = f"No value for {role}"
+                    logging.getLogger("HWR").warning(msg)
+            return mot_pos_dict
+
+        for motor in motors_list:
+            try:
+                mot_pos_dict[str(motor)] = float(mot_hwobj_dict[motor].get_value())
```
rhfogh, line 333: Why 'str(motor)'? What is the type of the list elements anyway, if not a string?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+        """
+        mot_pos_dict = {}
+
+        # use only the available motors
+        mot_hwobj_dict = self.get_motors()
+
+        if not motors_list:
+            for role, motor in mot_hwobj_dict.items():
+                try:
+                    mot_pos_dict[role] = float(motor.get_value())
```
rhfogh, line 325: Why float(...)? Are these not floats?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+                mot_pos_dict[str(motor)] = float(mot_hwobj_dict[motor].get_value())
+            except KeyError:
+                msg = f"Invalid motor name {motor}"
+                logging.getLogger("HWR").exceptionb(msg)
+            except TypeError:
+                msg = f"No value for {motor}"
+                logging.getLogger("HWR").warning(msg)
+        return mot_pos_dict
+
+    def get_state_motors(self, motors_list: list | None = None) -> dict:
```
rhfogh, line 342: Again, surely this shold be 'get_motor_states()'? We are returnign states, not motor hwobj.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+    def get_head_enum(self) -> DiffractometerHead:
+        """Get the diffractometer head Enum. Used when no import possible.
+        Returns:
+            DiffractometerHead Enum.
+        """
+        return DiffractometerHead
+
+    # -------- Phases --------
+
+    def set_phase(self, value: DiffractometerPhase | str, timeout: float | None = None):
```
rhfogh, line 466: Is it not possible to avoid the 'object-or-string' parameter, decide on one of them, and let the caller get the right one? This kind of multi-type parameter si a permanent source of confusion.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def get_phase(self) -> DiffractometerPhase:
+        """Get the current phase.
+
+        Returns:
+            DiffractometerPhase member.
+        """
+        return self.current_phase
+
+    def get_phase_list(self) -> list:
```
rhfogh, line 495: A phase list shoudl be a list of phase objects. Can we rename this (and others) so we can see when we are talking abotu a string and when we are talking about an object or an enum value?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+    def set_constraint(self, value, timeout=None):
+        """Sets diffractometer to selected constraint.
+
+        Args:
+            value (Enum): DiffractometerConstraint member.
+            timeout (float): optional - timeout [s],
+                             if timeout = 0: return at once and do not wait,
+                             if timeout is None: wait forever (default).
+        """
+        if isinstance(value, DiffractometerConstraint):
```
rhfogh, line 536: Again, can we not have a clear type, rather than this string-or-object?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+        else:
+            constraint = self.value_to_enum(value, DiffractometerConstraint)
+
+        self._set_constraint(constraint)
+        self._update_value(constraint, value_cmp=self.get_constraint())
+        if timeout == 0:
+            return
+        self.wait_ready(timeout)
+
+    def _set_constraint(self, value):
```
rhfogh: Should this not be an abstract function, so that subclasses have to implement it? Or, if the default behaviour is a no-op, should that not be shown clearly, e.g. by having the body be 'pass'? How is this supposed to behave if the subclass does not implement the function?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+        if not isinstance(value, DiffractometerPhase):
+            value = self.value_to_enum(value, DiffractometerPhase)
+        if value != DiffractometerPhase.UNKNOWN:
+            self._set_phase(value)
+            self.update_phase(value)
+            if timeout == 0:
+                return
+            self.wait_ready(timeout)
+
+    def _set_phase(self, value: DiffractometerPhase):
```
rhfogh, line 484: Should this not be an abstract function, so that subclasses have to implement it? Or, if the default behaviour is a no-op, should that not be shown clearly, e.g. by having the body be 'pass'? How is this supposed to behave if the subclass does not implement the function?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+        """Return a list of all the defined in DiffractometerPhase phases."""
+        phase_list = []
+        for member in DiffractometerPhase:
+            _nam = member.name
+            if _nam not in ["IN", "OUT", "UNKNOWN"]:
+                phase_list.append(_nam)
+        return phase_list
+
+    @property
+    def get_phase_enum(self) -> DiffractometerPhase:
```
rhfogh, line 505: As previous comment, is this necessary? Could it not be replaced by a class attribute?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+    def do_still_scan(self, *args, **kwargs):
+        """Do a zero oscillation acquisition."""
+        raise NotImplementedError
+
+    def do_characterisation_scan(self, *args, **kwargs):
+        """Do characterisation."""
+        raise NotImplementedError
+
+    def _update_value(self, value=None, value_cmp=None):
```
rhfogh, line 585: Surely this should be removed (or alternatively renamed).  A DIffractometer does not have a value, If it had, what would it be? When and why is this called?

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+
+    @property
+    def _ready(self) -> bool:
+        """Get the "Ready" state - software and hardware.
+        Returns:
+            (bool): True if both "Ready", False otherwise.
+        """
+        return self._get_swstate == "Ready" and self._get_hwstate == "Ready"
+
+    def _wait_ready(self, timeout: [None | float] = None):
```
rhfogh: How does this relate to the standard wait_ready function of HardwareObjkects? If it more or less does the same we should use (or maybe override) the superclass functino. If it does something different it should be documented clearly what the difference is, to avoid confusion.

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+            (enum 'HardwareObjectState'): state
+        """
+        try:
+            self._state = ExporterStates(self._get_swstate)
+        except ValueError:
+            self._state = self.STATES.UNKNOWN
+        return self._state
+
+    @property
+    def _ready(self) -> bool:
```
rhfogh, line 126: See comment to _wait_ready

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -0,0 +1,475 @@
...
+            motors_list (list): List of motor roles.
+        Returns:
+            (dict): role: position dictionary
+        """
+        motors_positions_dict = super().get_value_motors(motors_list)
+        if not self.in_kappa_mode:
+            motors_positions_dict.update({"kappa": None, "kappa_phi": None})
+        return motors_positions_dict
+
+    def get_movable_motors(self):
```
rhfogh: Why do you need this? Why not just override get_motors?

mxcubecore/queue_entry/base_queue_entry.py
beteva: dm is not supposed to be used any more, as we now use sample view for the centring. I do not know, though, if it does not break some logic, so for time being I've left it, but set it to None. Should be removed at near future.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: I would disagree. They are used in the SampleView HO to check the centring motors..

mxcubecore/queue_entry/base_queue_entry.py
rhfogh: It looks like it is not use in that function, so that line could maybe be removed without loss

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Well, you could simply call get_motors() or get_nstate_equipment() to get hold of a copy of the same dictionaries - you have those already. Or, if you go with types.MappingProxyType (which I only discovered recently) you can export that without making it modifiable from the outside

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: > This looks rather clunky. Is this temporary for xml config, or is this the way we want configuration to look, or what? 

My intention was, that you get all the equipment, accessible via beamline.diffractometer to be organised in two groups - motors and nstate equipment. Also for time being we do have xml files configuration, so we need to use the get_object_by_role, especially as we really rely on the roles to make a "standard" diffarctometer interface.
What would you suggest instead?

> Would it work with yaml config?

Yes, it should, although not tried yet. I've tested with the xml/yaml branch.

rhfogh: The main suggestion is that when we finish sorting out the xml/yaml configuration, this should be replaced by something simpler and cleaner - I know we are not there yet, so I was mostly asking if you intended this to be permanent.

As it looks to me, this code is 1) keeping two duplicate copies of the motor roles (why not stick to one, and have everybody look in the one authorised location?). 2) setting attributes on the AbstractDiffractometer that are not documented or findable by a linter, and that could add arbitrary attributes depending on what is in the configuration files.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: I have to think about it.
In the latest implementation of the Microdiff, you have the notion of constraints and even if the motor HO is valid, not all of them can be moved under certain constraints. The get_motors will give you the names of the currently movable motors, but nothing prevents you to put as a parameter a motor which is not accessible at that moment.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: I wanted to keep the get_value/set_value model, even if semantically is not right, but am ready to change the name.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: You are right, as we manipulate a list of roles , the parameter should be called motor_role_list

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This is belt and braces, as the input comes from the UI. We can remove the casting.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Same as above - belt and braces 😄

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Same as for the values - wanted to keep get_value logic.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: It is an Enum for the web and string for the Qt UI. Glad to change if possible.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Well, motor.get6_value() can surely be relied on to return a float (uless it is None, in which case the float call would not work either).

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This is used by the UI to display the labels of the phases in the pull down menu. I thought it will be better if the logic of how you get this list is in the HO and not on the frontend, but do not have a strong opinion.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This is a brain new concept. I do not have a clear idea of how we are going to use it in the UI, so left the possibility to have a string as an input. Could be removed if we agree on something.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: My taste would be for a function like get_value_motors to have a defined input type (in this case List[str] or list[str|None], and to leave it to the UI-handling code to do the type conversion. That keeps it closer to the place where the UI input is happening.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Good point. Looks like maybe the Qt people should change it to an enum.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: I did not set it as abstract, as nobody is using it right now. Could be changed in the future.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: get_phase_labels ?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Again fair enough. It ought to be cleared up once people start using it, but probably not yet, then.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Having pass in the body is a default behaviour, there is no need to set it explicitly. I was waiting for a discussion to come to an agreement if this is to be an abstract method or not.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This was ment mostly for convenience, so you do not need to import the respective classes form the AbstractDiffractometer. I am using it in the tests. 

> Could it not be replaced by a class attribute?

Maybe it could.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This is ment to be a generic method to check if the value to be set is different from the current one one and to emit valueChanged signal if needed. It is true that a diffractometer has no value, but this is an attempt to reduce the number of emitted signals. Maybe I am on a wrong track?
It is used, e.g. in the set_constraint method.

mxcubecore/HardwareObjects/MicroDiffractometer.py
beteva: This is specific for the Microdiff way of knowing if the java application is ready to take commands. The MD2 is very particular, as it only can handle one command at a time and if the application does not report Ready, we cannot send another command.  It is not based on events.

mxcubecore/HardwareObjects/MicroDiffractometer.py
beteva: Well spotted!  This should be get_motors.

mxcubecore/HardwareObjects/MicroDiffractometer.py
rhfogh: That makes sense, but it would be good if it said in the documentation to the function how it differed and when you should use what. If for some reason I had to check if a diffractometer was ready, I would br very confused about what to do.

beteva: In principle you should only use the standard wait_ready, if you need to use it from another HO or UI. The _wait_ready and  _read methods are MD2 specific and should be used only in the MicroDiffractometer HO.

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -454,31 +690,30 @@ def inc_used_for_collection(self, cpos):
...
 
     SHAPE_COUNT = 0
 
-    def __init__(self, mpos_list=[], screen_coord=(-1, -1)):
-        object.__init__(self)
+    def __init__(self, mpos_list=None, screen_coord=(-1, -1)):
         Shape.SHAPE_COUNT += 1
         self.t = "S"
         self.id = ""
         self.cp_list = []
```
rhfogh, line 774: Three points here:

-1- It is bad to have a directly accessible list and also to have a function like get_centred_positions() that returns it. Have one way to access that list. 

-2- It is a long-term risk to return internal modifiable lists - you get weird side effects if people start modifying them ('encapsulation'). A bit of Stack Overflow gives two solutions: return a tuple copy (better IMHO), or have only a function or property access and type the return value as Sequence - that way linters will protest if you start modifying the list (even if Python itself will do it). EIther way the internal list shoudl be _cp_list

-3- here I am not quite sure, but would it not be better to have the centred_positions as a property than as a function call? I'd say that they are a property of the object, not something that you generate when you call the function. centred_position would give rise to the same discussion

rhfogh:
There are a lot of changes here - this is a very welcome and long-awaited PR. There are also a lot of comments. Not sure of the best way forward.

There seems to be three main general discussion points:

-1- The format of doc strings.

-2- Naming of certain functions

-3- The question of exposing internal lists, of using getter functions v. direct access v. properties, and being consistent about it all.

On -1- there are just a lot of instances. On -2- I clearly prefer my own versions, but I am just one person, Should we ask for other opinions? On -3- we could in principle postpone the problem, but we will never have a better chance to harmonise things than we do now.

Do we need some kind of discussion here? Do you think you can already make enough modifications that it should sort things out @beteva? Or do you want to merge and then look at some of the questions in successive PRs?

mxcubecore/HardwareObjects/SampleView.py
beteva: In this particular case I've left what was before. You should see this PR as only the first step to clean the mess of the SampleView and sample centring. For time being I've concentrated on removing from the diffractometer object everything which concerns the centring and moved the functionality to the SampleView. I did not clean the SampleView itself, although this is needed.
@rhfogh Your suggestions are very valuable. May I suggest that we  validate the diffractometer part first and than concentrate on better centring and sample view code.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: As some of the motors and the actuators are optional (e.g. the fluo detector actuator or the sample vertical/horizontal pseudo motors), this was a way to allow them to be configured. Maybe not the best, I will try to see how to change this.

beteva:
@rhfogh, thanks a lot for the constructive summary. 
> Do we need some kind of discussion here?


My idea was that it is easier to discuss over an implementation which is working with a mock setup and can be also tested on a real beamline. Of course all the suggestions are welcome and if people feel like it, we could add the discussion for the next developer’s meeting. Otherwise, we could simply exchange mails in order not to flood the others.


> Do you think you can already make enough modifications that it should sort things out @beteva? Or do you want to merge and then look at some of the questions in successive PRs?

For what concerns the diffractometer part I will apply most of the @fabcor-maxiv suggestions for the doc strings and some of your suggestions, as well as whatever we decide after the discussion  Once this done, it will be probably the best to accept the PR. Only after this we could start to clean and improve the sample view / centring.

rhfogh (approved):
As discussed.

rhfogh:
@beteva OK sounds fine. I have approved the PR.

mxcubecore/HardwareObjects/MicroDiffractometer.py
fabcor-maxiv: OK, this still seems strange to me to add this in the wrong place when we already know what the correct place is, but I won't argue further.

fabcor-maxiv:
We should not let the format of the docstrings block other improvements to be made (formatting can be fixed later with the help of a tool). What matters is that we document things correctly now while we are working on it and we have everything in mind. I mean that the content of the docstrings should be correct, and type hints as well and so on.

ruff.toml
marcus-oscarsson: What we discussed and decided during our previous meetings was that we may want to make certain deviations from whats proposed by default by RUFF. There may be several reasons for these deviations ranging from historical decisions (that we can't easily handle on a case to base basis) to that we simply don't agree with a certain rule.  There is an issue for these discussions if needed https://github.com/mxcube/mxcubecore/issues/1367.

I think disabling the `N999` rule below is alright. I agree that it would be good to rename the module so that it conforms with snake case. Its something we are all aware of  and we are used to see in this code base so there is little risk for confusion. However in this case the linter will be "in the way" every time we use this import and we have to add an explicit exception. 

I believe  that the later is more annoying when we work with the code on an everyday basis. Once we have refactored/renamed the module names we can simply remove the N999 rule globally instead of in xyz places in the code.

For the `FBT001` (and `FBT002`) rules, I simply don't agree. Who does not understand `def round_number(number, up=True):` or `def move_motor(value, wait=True):` How is the other proposed solution in the [documentation](https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/) (using the enum) more clear ?

To me its instead counter intuitive to look for a method called `round_number_up` or `move_motor_and_wait` , `move_motor_async`. Its perhaps useful in more uncertain situations but than I'm more inclined to say that its the method name in combination with the boolean argument that is a problem. `FBT002` would make more sense in certain situations.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+@unique
+class DiffractometerPhase(Enum):
+    """Enumeration diffractometer phases"""
+
+    UNKNOWN = "Unknown"
+    CENTRE = "Centring"
+    COLLECT = "DataCollection"
+    SEE_BEAM = "BeamLocation"
```
marcus-oscarsson, line 97: I'm hesitating about the word `SEE`, perhaps simply `BEAM` or `SAMPLE` or `VIEW` instead of `SEE` ? Not sure ...

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -0,0 +1,461 @@
...
+    TRANSFER = "Transfer"
+    SEE_SAMPLE = "LightSample"
+
+
+@unique
+class DiffractometerConstraint(Enum):
+    """Enumeration diffractometer constraint types"""
+
+    UNKNOWN = "Unknown"
+    RELEASE = "Normal"
```
marcus-oscarsson, line 106: `ROTATION` or `FREE` instead of `RELEASE ?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: I agree with @rhfogh that it if we have methods to access these objects, we should use them and not the structure. I would also add a `_`

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: I don't think the structure `motors_hwobj_dict` is actually needed, the object already exists on  `self.config.motors:`. 

I would also suggest, if possible (i think it is), that we fix a reference frame for the AbstractDiffractometer and that we predefine the motor names that move along those axes. Basically what we are all defacto using today. Like this we can perhaps even avoid the `setattr`

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: Can't we simply return a tuple of motors, they all have `.name` right ?

marcus-oscarsson: That would go for all the `get_` functions

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: :), ahh hard one. Id also go for the more palatable `set_motor_values`

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: Its perhaps safer to check that all keys passed exist before moving any motor ?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: Yes thats enough

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: Id be happy if we did not using `str`

ruff.toml
rhfogh: For the specific case of  FBT002 using keyword-only arguments would be OK (the other solutions are horrendous). But is that the most important problem we have? I could think of several other refactorings (unencapsulated internal collections, the residue of XML configuration, still with us after two years, ...) that I think are rather more important, but that I cannot force people to do simply by setting a line in ruff.toml.

If we were creating a project from scratch we could enforce all the rules at once without too much hassle - if we wanted. For the large, existing code-base that we have, setting these rules are basically a way of forcing people to address these stylistic quirks *at once*. That is not a good idea. The sensible practice would be to 1) Identify the problem, 2) decide that this has top priority now, 3) refactor to remove rule-breaks, 4) only THEN enforce the rules. Just adding the rules up front distorts attention from  problems that are more important or urgent, and incentivises people to find quick workarounds that satisfy the linter but do not make the code better or solve underlying problems. 

For something like N999: If you think that having CamelCase ModuleNames is the most urgent problem right now, you should refactor the code base yourself to remove them and make a PR, or convince some other people to do it with you. If you do not want to do that, because you have more urgent things to do, you should not use Ruff as a tool to force other people to do it for you.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: @marcus-oscarsson Yes!

mxcubecore/HardwareObjects/abstract/AbstractSampleView.py
```diff
@@ -112,53 +117,36 @@ def backlight(self):
...
-    # def create_auto_line(self):
-    #     return
+    def move_to_beam(self, x: float, y: float):
+        """Move the sample to the x,y coordinates"""
 
-    # @abc.abstractmethod
-    # def create_grid(self, spacing):
-    #     return
+    @abc.abstractmethod
+    def get_positions(self) -> dict:
```
marcus-oscarsson, line 131: How is this method used, I'm wondering if it can't simply by on `diffractometer`

marcus-oscarsson:
Big thanks @beteva it starts to look really nice. I have one thing that I think would be quite useful and important and its to fix a reference frame and standard motor names. It does not mean that we can't have additional motors but it would be great if we could fix a set of concepts, basically the ones the we are already using. Ill try to outline something here knowing that I will leave out some bits, but please feel free to complete:

  - Beam direction (x̂)
  - Goniometer orientation, vector
  - Motor names with description that tells which direction they are moving
      - Alignment X 
      - Alignment Y
      - Alignment Z
      - Phi (Or kappa phi)
      - Kappa
      - Omega
      - Sample X
      - Sample Y
      - Sample Vertical
      - Sample Horizontal

The rest is more or less details.

When it comes to the concrete implementation of the Microdiff we discussed some idea of  being able to read available motors directly from the MD software and add the necessary motors dynamically (to avoid lots of files). It seems like the possibility to do this is a bit limited on the MD side at the moment, but it would be a very nice feature.

marcus-oscarsson:
That should probably also contain the direction in which the OAV is looking

rhfogh:
@marcus-oscarsson A reference frame and motor names and directions would be really great. It is needed also for recentring, and for the GPhL workflow, and it would be nice to be able to get that from MXCuBE rather than configuring it for GPhL separately, if possible. 

There are various niggles one might want to discuss

- Names without spaces?

- Are the centring motors in the goniostat coordinate system and the goniostat in some lab coordinate system? Is that how you position the goniostat relative to the detector (up, down, sideways)?

- Rotation direction of rotation motors v. orientation of the apparatus.

- Zero point of kappa and omega (you can describe the goniostat in internal cordinates using the angles between the three rotation axes and the two zero points. This is more intelligible, but possibly less practical).

Or do you think I am pushing this too far? It might all need some discussion after the PR is accepted.

marcus-oscarsson:
I guess we should not push it to far to start with but it would be good with some basic concepts. 

> Names without spaces?

Sure I guess in code or in a schema that would be lowercase with underscore so that we have the same names in code ?

> Are the centring motors in the goniostat coordinate system and the goniostat in some lab coordinate system? Is that how you position the goniostat relative to the detector (up, down, sideways)?

I was thinking in more in the lab reference frame I guess, i.e with the beam in the X^ direction and the motors defined similar to today y and z perpendicular to the beam (x) and then omega as a rotation in the direction of the beam or function of the kappa motors.

I guess we can assume a certain orientation to make things easier to understand but it would be good if we could define a orientation vector, (up, left down, right) (of the gonio)

> Zero point of kappa and omega (you can describe the goniostat in internal cordinates using the angles between the three rotation axes and the two zero points. This is more intelligible, but possibly less practical).

Can't really answer this one, not sure its so easy to find those zero points ?

You are right that including all of this in this PR is a bit much, but perhaps defining the motor names and the "standard" gonio orientation relative to the beam would be a good start ?

ruff.toml
fabcor-maxiv: If we decide to add new exceptions, it is perfectly fine, but they should NOT be added to the `[lint.per-file-ignores]` table, except for things like [this](https://github.com/mxcube/mxcubecore/blob/b4b1a39ae72ff8bfbf6218749f54834a4ee2ff2f/ruff.toml#L32-L38).

And also I recommend that adding a global exception should be done in its own pull request so that we can handle the whole code base at once. I can do one such pull request for `N999`, for example.

For `FBT001` the issue is when calling the function: `foo = round_number(number, True)` is valid code but more confusing to read than `foo = round_number(number, up=True)`. Keyword-only argument would probably be my preferred solution as well (but I have not checked if it is applicable here). If you decide to disable this rule globally, please do it in a dedicated pull request. Or please disable it at the file level or at the line level, but not in the `[lint.per-file-ignores]` table.

For `max-args` as well, it should probably be its own pull request if we decide to change its value globally, so that we can handle the whole code base at once and remove all occurrences of `PLR0913` in `ruff.toml`.

rhfogh: > And also I recommend that adding a global exception should be done in its own pull request so that we can handle the whole code base at once

Yes, that sounds right. But, by the same token, adding a new _rule_  should _also_ be done in its own pull request, so we can agree about doing it and handle the whole code base at once. It may be rather late now, but it was not good practice that the first that you saw about those 800 Ruff rules was when your PRs started failing.

fabcor-maxiv: https://github.com/mxcube/mxcubecore/pull/1403

beteva:
> I have one thing that I think would be quite useful and important and its to fix a reference frame and standard motor names. It does not mean that we can't have additional motors but it would be great if we could fix a set of concepts, basically the ones the we are already using.

Funny, I've done this at the same time you were writing the comments, but did not have time to push. Now it is done and it will be good if the people concerned look at the conversions and the fixed naming to approve or disagree.

beteva:
> Or do you think I am pushing this too far? It might all need some discussion after the PR is accepted.

I think it is important that we agree on most of the this. @rhfogh I think most of your questions were addresses in the diffractometer.md and in the AbstractDiffractometer description. Now I see that I've forgotten some, which will be done now.

mxcubecore/HardwareObjects/abstract/AbstractSampleView.py
beteva: This is only concerning the centring motors, which are kn own by the sample view. Other possibility would be to have get_centring_motor_values in the diffractometer. Not sure which is more consistent.

rhfogh: CentringMotors are a bit of a strange beast. As I understand it, they are basically wrappers around the actual motors, with reference position, unit conversion, and direction added. Ideally, I would say, these should be united with the normal motors, as a subtype, so that we did not have to have the complex wrapping, or two different objects representing the same motor. This may not be possible, of course, certainly not in the short term. But until we decide how we want to handle this long-term, it may not be a  bad idea to keep the centring motors limited to the place where they are now.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: This is the the phase which allows you to see the beam on a scintillator. BEAM would be misleading. VIEW_BEAM is OK

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Maybe FREE than. This is when you are released of all the constraints.

PR #1389, opened 2025-09-18
docs: expand 'Logging' section

elmjag:
Add information on usage of `exception()` and `debug()` logging methods to the 'Coding convention and style guidelines' section.

Adds a bit more text, as discussed in https://github.com/mxcube/mxcubecore/pull/1383.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -307,6 +307,26 @@ class MyHardwareObject(HardwareObject):
         self.user_log.error("an error have been encountered")
 ```
 
+When logging a caught exception, consider using `self.log.exception()`.
+The `exception()` method adds raised exception's type, message and traceback to the log entry.
+Include any relevant variables to the log message explicitly.
+These details are essential for proper troubleshooting.
+
+Below is an example of logging an exception.
+
+```python
+try:
+    output = transform(input)
+except Exception:
+    # log the raised exception
+    self.log.exception("Could not transform: %s", input)
+```
+
+Avoid using `debug()` and `exception()` level when logging to `user_log`.
+The user log is aimed to convey information to the users.
+The debug and exception logging should be used for logging information suitable for developers,
+to help with troubleshooting.
+
 ### Continuous integration (CI)
 
 GitHub Action are used for continues integration


CONTRIBUTING.md
```diff
@@ -307,6 +307,26 @@ class MyHardwareObject(HardwareObject):
         self.user_log.error("an error have been encountered")
 ```
 
+When logging raised exceptions, use the `self.log.exception()`.
+The `exception()` method adds raised exception's type, message and traceback to the log entry.
```
marcus-oscarsson, line 311: Not sure about this paragraph, it feels like we are documenting a standard python behavior.

elmjag: In sense yes, but the motivation is that we run into this type of code in MXCuBE:

```
try:
  some_thing()
catch Exception as e:
  self.log.error(f"cryptic error message: {e}")
```
 
Which makes it hard to figure out _where_ this exception comes from. I feel it would be helpful to explicitly mention `exception()` exist, and that it should be used.

I think many developers are not aware it exist. For example I did not know this, until @fabcor-maxiv showed me.

CONTRIBUTING.md
```diff
@@ -307,6 +307,26 @@ class MyHardwareObject(HardwareObject):
...
+
+```python
+try:
+    output = transform(input)
+except Exception:
+    # log the raised exception
+    self.log.exception("Could not transform: %s", input)
+```
+
+Avoid using `debug()` and `exception()` level logging to the `user_log`.
```
fabcor-maxiv:
```suggestion
Avoid using `debug()` and `exception()` level when logging to `user_log`.
```

fabcor-maxiv (approved):

CONTRIBUTING.md
marcus-oscarsson: Hm, :thinking:, I thought it was more or less well known but anyhow. This section perhaps belongs more in a section about Exception handling. 

We should probably then add that we should in most cases NOT log with `exception` if the exceptions is re-raised, and we should probably also in addition to the `exception` log message also make a `user.warning/info` log  message ?

fabcor-maxiv: > I thought it was more or less well known

Sadly not. :/

This is a problem that is plaguing us. I know we have an awful lot of those cases in the MAX IV internal code base we inherited of. I could also track some recent cases we spotted in the shared MXCuBE code base:

* <https://github.com/mxcube/mxcubecore/pull/1384#discussion_r2355362907>
* <https://github.com/mxcube/mxcubecore/pull/1375#discussion_r2343379736>
* <https://github.com/mxcube/mxcubeweb/pull/1717#discussion_r2174629847>
* <https://github.com/mxcube/mxcubeweb/pull/1672>
* <https://github.com/mxcube/mxcubeweb/pull/1694#discussion_r2152475961>

To me this is a very serious issue. Makes troubleshooting extremely difficult when there is no traceback to start from.

marcus-oscarsson: Well its one thing to know about a feature and not using it for whatever reason that may be. I think many people already know about the feature but the exception handling in general is not done very well. It would in my opinion help more if we had a section about exception handling in general where this could also be mentioned. 

Simply saying that we should always use `exception` when logging exceptions is a bit of an over simplification as I see it.

fabcor-maxiv: I want both. It would be great to have a section about exceptions, for sure. But I think it is good to also have a reminder in the "logging" section.

We can probably tweak the phrasing:

```diff
-When logging raised exceptions, use the `self.log.exception()`.
+When logging a caught exception, consider using `self.log.exception()`.
```

marcus-oscarsson: Ok, I guess thats fine for now

marcus-oscarsson:
Please feel free to update/expand the section about exception logging. Ill merge this in the mean time.

elmjag:
> Please feel free to update/expand the section about exception logging. Ill merge this in the mean time.

Let's merge this as-is, and then we'll work on a new PR, which adds an 'Exceptions' section?

marcus-oscarsson:
Sure, no problem, thanks !

PR #1392, opened 2025-09-22
tango: make `polling` optional argument

elmjag:
Make it possible to create tango attribute channels without polling when using the `add_attribute_channel()` utility function.

### DIFF
--- a/mxcubecore/utils/tango.py
+++ b/mxcubecore/utils/tango.py
@@ -14,7 +14,7 @@ def add_attribute_channel(
     hwo: HardwareObject,
     tango_device: str,
     attribute_name: str,
-    polling: int,
+    polling: int | None = None,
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Ok, do you think we can manage to integrate this utility into the command handling

marcus-oscarsson (approved):

marcus-oscarsson:
Thanks, nothing to say really. I think it would be great if this utility could get integrated into the Command/Channel object somehow. So that we in the long run have one standard way of doing things.

elmjag:
> Thanks, nothing to say really. I think it would be great if this utility could get integrated into the Command/Channel object somehow. So that we in the long run have one standard way of doing things.

Yes, I agree. We'll need to look into that, at some point.

PR #1396, opened 2025-09-25
Bug fix in centring and take_snapshot when plate mode.

beteva:
1. Do not move to 90 degrees if plate mode when taking snapshots..
2. Do not move to centring phase if not needed.

### DIFF
--- a/mxcubecore/HardwareObjects/MiniDiff.py
+++ b/mxcubecore/HardwareObjects/MiniDiff.py
@@ -1026,7 +1026,7 @@ class MiniDiff(HardwareObject):
             )
             HWR.beamline.sample_view.save_snapshot(path=image_path)
             # do not move 90 degrees if not needed
-            if not self.in_plate_mode() or snapshot_index > 0:
+            if not self.in_plate_mode() and snapshot_index < len(image_path_list) - 1:
                 self.phiMotor.set_value_relative(90, timeout=5)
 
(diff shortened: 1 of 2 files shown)


mxcubecore/HardwareObjects/MiniDiff.py
```diff
@@ -1024,7 +1025,9 @@ def take_snapshot(self, image_path_list: list) -> None:
                 f"Taking {snapshot_index + 1} sample snapshot(s)"
             )
             HWR.beamline.sample_view.save_snapshot(path=image_path)
-            self.phiMotor.set_value_relative(90, timeout=5)
+            # do not move 90 degrees if not needed
+            if not self.in_plate_mode() or snapshot_index > 0:
+                self.phiMotor.set_value_relative(90, timeout=5)
```
walesch-yan: I do not see the relevance of `or snapshot_index > 0`, so it only comes into effect if we are in plate_mode and want to take more than 1 snapshot, which should not be possible anyways.

I would suggest to change to `if not self.in_plate_mode() and snapshot_index < len(image_path_list) - 1` so that we do not turn when in `plate_mode` and also not after the last snapshot was taken?

beteva: Thanks @walesch-yan, I've applied your suggestion

marcus-oscarsson (approved):

PR #1399, opened 2025-10-02
Remove lnls_sol from configuration

RafaelLyra8:


rhfogh (approved):
This is fine. If you want it in at all, it should be in mxcube_configuration anyway.

fabcor-maxiv:
Added to the list here:

* <https://github.com/mxcube/mxcubecore/issues/1226>

PR #1400, opened 2025-10-03
Support beamline config file with 'yaml' extension.

elmjag:
Make it possible to use ~~beamline_config.y**a**ml~~ beamline.y**a**ml, as config file, in addition to beamline_config.yml.

As all other YAML config files use `yaml`, it feels a bit odd that beamline_config file _must_ have a `yml` extension.

### DIFF
--- a/mxcubecore/HardwareRepository.py
+++ b/mxcubecore/HardwareRepository.py
@@ -79,7 +79,11 @@ _instance = None
 TIMERS = []
 
 beamline = None
-BEAMLINE_CONFIG_FILE = "beamline_config.yml"
+#
+# Supported beamline configuration filenames.
+# The first name in the list takes precedence.
+#
+BEAMLINE_CONFIG_FILES = ["beamline.yaml", "beamline_config.yml"]
 
 
 def load_from_yaml(
@@ -394,8 +398,16 @@ def init_hardware_repository(
     logging.getLogger("HWR").info("Hardware repository: %s", configuration_path)
     _instance = __HardwareRepositoryClient(configuration_path)
     _instance.connect()
+
+    beamline_config_file = _instance.find_beamline_config_file()
+    if beamline_config_file.endswith(".yml"):
+        warn(  # noqa: B028 we don't care about stacklevel for this warning
+            f"Config file '{beamline_config_file}' have deprecated extension 'yml', "
+            "change to 'yaml'."
+        )
+
     beamline = load_from_yaml(
-        BEAMLINE_CONFIG_FILE,
+        beamline_config_file,
         role="beamline",
         yaml_export_directory=yaml_export_directory,
     )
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareRepository.py
```diff
@@ -455,6 +459,21 @@ def connect(self):
...
 
+    def find_beamline_config_file(self) -> str | None:
+        """Find beamline configuration file name.
+
+        Look in the repository paths for beamline config file.
+
+        Returns:
+            The found config file name, or None if not found.
+        """
+        for file_name in BEAMLINE_CONFIG_FILES:
```
rhfogh, line 478: How about raising a Deprecation Warning if the name is *.yml'? Long term we should have only one name here, which should be .yaml, no?

elmjag: Oh yeah, good idea. Added a warning [here](https://github.com/mxcube/mxcubecore/pull/1400/files#diff-e3ba2a0b991b4725ee3e2e98deb77bf4777747fb7cdc071954d85a44297968f5R404).

marcus-oscarsson: yes

elmjag:
A question, should we support `beamline.yaml` as well?

Normally other hardware object config files does not contain `_config` suffixes in their filenames. I think it would make it more uniform if `Beamline` hardware object could use `beamline.yaml` filename.

marcus-oscarsson:
I think it should be one of the two, its true that `beamline.yaml` is what one would probably look for knowing how the rest works. Its a bit odd to reach for `beamline-config`

marcus-oscarsson:
I'd vote for `beamline.yaml` ...

mxcubecore/HardwareRepository.py
rhfogh: Can I propose putting B028 on the global exclusion list? It clutters up the code. Clearly whoever is responsible for Ruff thinks that Python should have stacklevel=2 as the default, but why should we care? Can anyone come with an argument for enforcing this rule, beyond 'Ruff say so, and of course we have to obey'?

If you really think it is worth the trouble, the solution would be to write a wrapper function (as proposed in the Python documentation), so that you call 'mxube_warn' with whatever default parameters you think we should use.

elmjag:
Ok, let's support `beamline.yaml` and `beamline_config.yml` then, right?

Where `beamline_config.yml`  is deprecated.

marcus-oscarsson:
excellent :)

mxcubecore/HardwareRepository.py
marcus-oscarsson: I would not have anything against ignoring that rule. I see some sense in using `stacklevel=2`, it would make it easier to find from where its triggered. It would be quite easy to fix with `sed/grep` I guess ;). But I would not mind ignoring this one.

mxcubecore/HardwareRepository.py
```diff
@@ -394,8 +398,16 @@ def init_hardware_repository(
     logging.getLogger("HWR").info("Hardware repository: %s", configuration_path)
     _instance = __HardwareRepositoryClient(configuration_path)
     _instance.connect()
+
+    beamline_config_file = _instance.find_beamline_config_file()
+    if beamline_config_file.endswith(".yml"):
+        warn(  # noqa: B028 we don't care about stacklevel for this warning
+            f"Config file '{beamline_config_file}' have deprecated extension 'yml', "
```
fabcor-maxiv, line 405:
```suggestion
            f"Configuration file '{beamline_config_file}' has deprecated extension 'yml', "
```

PR #1401, opened 2025-10-03
Rename / bug fix for get_full_path function

rhfogh:
Clean-up / rename, changing get_full_path to get_full_paths. The function returns two paths, so the name ought to be plural.
Also includes apparent bug fix in ICATLims.__to_sample
Coupled with parallel PR in mxcubeweb

The autotest fails, but it does not look like the failure has anything to do with the changes in the PR.

### DIFF
--- a/mxcubecore/HardwareObjects/ICATLIMS.py
+++ b/mxcubecore/HardwareObjects/ICATLIMS.py
@@ -325,31 +325,30 @@ class ICATLIMS(AbstractLims):
                     self.__get_sample_information_by(sample_sheet_id)
                 )
                 if sample_information is not None:
-                    if len(HWR.beamline.session.get_full_path("", "")) > 0:
-                        destination_folder = (
-                            HWR.beamline.session.get_base_process_directory()
-                        )
-                        msg = "Download restource: "
-                        msg += f"sample_sheet_id={sample_sheet_id} "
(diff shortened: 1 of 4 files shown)


mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -325,31 +325,30 @@ def __to_sample(
                     self.__get_sample_information_by(sample_sheet_id)
                 )
                 if sample_information is not None:
-                    if len(HWR.beamline.session.get_full_path("", "")) > 0:
```
rhfogh, line 328: get_full_paths always returns a tuple with two elements, so the if statement does not work as you would think.
Also, BTW, the parameters default to "", so setting them explicitly to "" is unnecessary (as well as very hard to understand the meaning of).

marcus-oscarsson:
thanks @rhfogh, just out of curiosity how come that stumbled on to that, are you also trying ICAT ?

rhfogh:
@marcus-oscarsson No, no time for ICATting at the moment, but I came across get_full_path, wondered what it was, realised that I wanted it renamed, and then did a global search for uses before I started making changes.

marcus-oscarsson:
Ok, thanks

marcus-oscarsson (approved):

PR #1403, opened 2025-10-06
Ignore Ruff rule `N999:invalid-module-name`

fabcor-maxiv:
We have many modules that do not follow the Python naming convention. We have many modules named in `PascalCase` instead of `snake_case`.

This naming decision was made a long time ago. Rectifying this is not a priority for now but should be done eventually.


rhfogh (approved):

marcus-oscarsson:
Thanks, for taking the initiative to do this

PR #1406, opened 2025-10-07
Attempt to handle `except` cases in a somewhat better way

marcus-oscarsson:
This PR tries to do something with all the `except` cases that may potentially hide an issue.

Replaced all pass statements with a `exception` logger statement and inserted a a `exception` logger statement before return statements.

### DIFF
--- a/mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
+++ b/mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
@@ -55,6 +55,8 @@ class DigitalZoomMotor(AbstractMotor, HardwareObject):
             self.camera = self.get_object_by_role("camera")
         except KeyError:
             logging.getLogger("HWR").warning("DigitalZoomMotor: camera not defined")
+
+            logging.getLogger("HWR").exception("")
             return
         try:
             self.read_only = not (self.camera.zoom_exists())
--- a/mxcubecore/HardwareObjects/DESY/P11Collect.py
+++ b/mxcubecore/HardwareObjects/DESY/P11Collect.py
@@ -1137,6 +1137,8 @@ class P11Collect(AbstractCollect):
                         os.mkdir(path, mode=0o777)
                     except RuntimeError:
                         self.log.debug("mkdir failed:", str(sys.exc_info()))
+
+                        logging.getLogger("HWR").exception("")
                         return False
                 else:
                     self.log.debug("dir not found:", str(sys.exc_info()))
--- a/mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
+++ b/mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
@@ -1473,6 +1473,8 @@ class P11NanoDiff(GenericDiffractometer):
                 self.log.rror(
                     "motor_positions is not a dict and cannot be converted using as_dict()"
                 )
+
+                logging.getLogger("HWR").exception("")
                 return
 
         # Move translation motors
--- a/mxcubecore/HardwareObjects/EMBL/EMBLBeamlineTest.py
+++ b/mxcubecore/HardwareObjects/EMBL/EMBLBeamlineTest.py
@@ -116,6 +116,8 @@ class EMBLBeamlineTest(HardwareObject):
                 logging.getLogger("HWR").warning(
                     "BeamlineTest: Unable to create test directories"
                 )
+
+                logging.getLogger("HWR").exception("")
                 return
 
         self.results_list = []
--- a/mxcubecore/HardwareObjects/ESRF/ID29HutchTrigger.py
+++ b/mxcubecore/HardwareObjects/ESRF/ID29HutchTrigger.py
@@ -42,6 +42,8 @@ class ID29HutchTrigger(BaseHardwareObjects.HardwareObject):
             self.card, self.channel = map(int, PSSinfo.split("/"))
         except Exception:
             logging.getLogger().error("%s: cannot find PSS number", self.name())
+
+            logging.getLogger("HWR").exception("")
             return
 
         if self.device is not None:
--- a/mxcubecore/HardwareObjects/abstract/AbstractVideoDevice.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractVideoDevice.py
@@ -42,7 +42,7 @@ import numpy as np
 try:
     import cv2
 except ImportError:
-    pass
+    logging.getLogger("HWR").exception("")
 
 from mxcubecore.BaseHardwareObjects import HardwareObject
 
@@ -120,7 +120,7 @@ class AbstractVideoDevice(HardwareObject):
         try:
             self.cam_gain = float(self.get_property("gain"))
         except TypeError:
-            pass
+            logging.getLogger("HWR").exception("")
 
         try:
             self.cam_exposure = float(self.get_property("exposure"))
--- a/mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
+++ b/mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
@@ -16,6 +16,8 @@ def get_image(url: str) -> Optional[bytes]:
         logging.getLogger("user_level_log").warning(
             "Failed to fetch image from %s: %s", url, e
         )
+
+        logging.getLogger("HWR").exception("")
         return None
 
 
@@ -28,6 +30,8 @@ def get_image_size(url: str) -> Tuple[int, int]:
         logging.getLogger("user_level_log").warning(
             "Failed to get image size from %s: %s", url, e
         )
+
+        logging.getLogger("HWR").exception("")
         return (0, 0)
 
 
@@ -147,6 +151,8 @@ def get_processing_plan(
         logging.getLogger("user_level_log").warning(
             "Error getting processing plan from  %s : %s", url, e
         )
+
+        logging.getLogger("HWR").exception("")
         return None
 
 
@@ -174,4 +180,6 @@ def send_data_collection_info_to_crims(
         return True
     except requests.RequestException as e:
         logging.getLogger("user_level_log").warning("Request to %s failed: %s", url, e)
+
+        logging.getLogger("HWR").exception("")
         return False
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -486,7 +486,7 @@ class HardwareObjectNode:
         try:
             index = self.__objects_names.index(object_name)
         except ValueError:
-            pass
+            logging.getLogger("HWR").exception("")
         else:
             for obj in self.__objects[index]:
                 yield obj
--- a/mxcubecore/Command/Epics.py
+++ b/mxcubecore/Command/Epics.py
@@ -141,7 +141,7 @@ class EpicsCommand(CommandObject):
         try:
             callback = self.__value_changed_callback_ref()
         except Exception:
-            pass
+            logging.getLogger("HWR").exception("")
         else:
             if callback is not None:
                 callback(value)
@@ -155,7 +155,7 @@ class EpicsCommand(CommandObject):
             try:
                 poller.restart(1000)
             except Exception:
-                pass
+                logging.getLogger("HWR").exception("")
 
     def get_pv_value(self):
         """wrapper function to pv.get() in order to supply additional named parameter"""
--- a/mxcubecore/Command/Pool.py
+++ b/mxcubecore/Command/Pool.py
@@ -141,9 +141,9 @@ def process_tango_events():
                     try:
                         receiver_cb(event.attr_value.value)
                     except AttributeError:
-                        pass
+                        logging.getLogger("HWR").exception("")
             except KeyError:
-                pass
+                logging.getLogger("HWR").exception("")
 
 
 class PoolChannel(ChannelObject):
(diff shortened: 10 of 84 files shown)


mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -16,6 +16,8 @@ def get_image(url: str) -> Optional[bytes]:
         logging.getLogger("user_level_log").warning(
             "Failed to fetch image from %s: %s", url, e
         )
+
+        logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 20: I write it only here but I guess this applies to all hardware objects, right?

```suggestion
        self.log.exception("")
```

fabcor-maxiv (commented):
Great! Much needed! Should hopefully make troubleshooting much easier.

fabcor-maxiv:
I think this might make it possible to remove a bunch of linter exceptions (in the "per file" lists).

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
marcus-oscarsson: That will be done in a second pass as we are not sure that all the statements belong to a HardwareObject ...

marcus-oscarsson:
This will probably trigger a bunch of other linting errors that I'm not sure how we should handle at the moment ...

marcus-oscarsson:
Please have a look, this PR simply changes the except case in the following way:
 - Replaces pass statements with a `exception` logger statement
 - inserts a `exception` logger statement before a return
 - import logging if not already done

We can do a `second` pass to go from `logging.getLogger` to `self.log` in a followup PR

marcus-oscarsson:
I would also like to replace the use of the `traceback` module ...

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -147,6 +151,8 @@ def get_processing_plan(
         logging.getLogger("user_level_log").warning(
             "Error getting processing plan from  %s : %s", url, e
         )
+
+        logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 155:
```suggestion
```

Maybe this one is not necessary. Exception handling seems under control, but of course I might not have all the details.

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -28,6 +30,8 @@ def get_image_size(url: str) -> Tuple[int, int]:
         logging.getLogger("user_level_log").warning(
             "Failed to get image size from %s: %s", url, e
         )
+
+        logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 34:
```suggestion
```

Maybe this one is not necessary. Exception handling seems under control, but of course I might not have all the details.

mxcubecore/HardwareObjects/abstract/AbstractVideoDevice.py
```diff
@@ -42,7 +42,7 @@
 try:
     import cv2
 except ImportError:
-    pass
+    logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 45:
```suggestion
```

Maybe this one is not necessary. Exception handling seems under control, but of course I might not have all the details.

mxcubecore/HardwareObjects/DESY/DigitalZoomMotor.py
```diff
@@ -55,6 +55,8 @@ def init(self):
             self.camera = self.get_object_by_role("camera")
         except KeyError:
             logging.getLogger("HWR").warning("DigitalZoomMotor: camera not defined")
+
+            logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 59:
```suggestion
            logging.getLogger("HWR").exception("DigitalZoomMotor: camera not defined")
```

mxcubecore/HardwareObjects/DESY/P11Collect.py
```diff
@@ -1137,6 +1137,8 @@ def check_path(self, path=None, force=False):
                         os.mkdir(path, mode=0o777)
                     except RuntimeError:
                         self.log.debug("mkdir failed:", str(sys.exc_info()))
+
+                        logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 1141:
```suggestion
                        self.log.exception("mkdir failed: %s", str(sys.exc_info()))
```

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -98,6 +98,8 @@ def get_value(self):
 
         except Exception as e:
             logging.error(f"Error getting machine values: {e}")
+
+            logging.getLogger("HWR").exception("")
             return {}  # Return empty dictionary if there's an error
```
fabcor-maxiv, line 101:
```suggestion
        except Exception:
            logging.exception("Error getting machine values")
            return {}  # Return empty dictionary if there's an error
```

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -108,6 +110,8 @@ def get_current(self):
             return round(current_value, 2)
         except AttributeError:
             logging.error("Error reading 'current': Attribute 'current' not found.")
+
+            logging.getLogger("HWR").exception("")
```
fabcor-maxiv:
```suggestion
            logging.getLogger("HWR").exception("Error reading 'current': Attribute 'current' not found.")
```

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -116,6 +120,8 @@ def get_lifetime(self):
             return round(lifetime_value, 2)
         except AttributeError:
             logging.error("Error reading 'lifetime': Attribute 'lifetime' not found.")
+
+            logging.getLogger("HWR").exception("")
             return 0
```
fabcor-maxiv, line 119:
```suggestion
except AttributeError:
            logging.getLogger("HWR").exception("Error reading 'lifetime': Attribute 'lifetime' not found.")
            return 0
```

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -124,6 +130,8 @@ def get_maschine_energy(self):
             return round(energy_value, 2)
         except AttributeError:
             logging.error("Error reading 'energy': Attribute 'energy' not found.")
+
+            logging.getLogger("HWR").exception("")
             return 0
```
fabcor-maxiv, line 127:
```suggestion
except AttributeError:
            logging.getLogger("HWR").exception("Error reading 'energy': Attribute 'energy' not found.")
            return 0
```

mxcubecore/HardwareObjects/DESY/P11MachineInfo.py
```diff
@@ -132,6 +140,8 @@ def get_message(self):
             return message_value
         except AttributeError:
             logging.error("Error reading 'message': Attribute 'message' not found.")
+
+            logging.getLogger("HWR").exception("")
             return "Message unavailable"
```
fabcor-maxiv, line 135:
```suggestion
        except AttributeError:
            logging.getLogger("HWR").exception("Error reading 'message': Attribute 'message' not found.")
            return "Message unavailable"
```

mxcubecore/HardwareObjects/DESY/P11NanoDiff.py
```diff
@@ -1473,6 +1473,8 @@ def move_motors(self, motor_positions, timeout=10):
                 self.log.rror(
                     "motor_positions is not a dict and cannot be converted using as_dict()"
                 )
+
+                logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 1477:
```suggestion
                self.log.exception(
                    "motor_positions is not a dict and cannot be converted using as_dict()"
                )
```

mxcubecore/HardwareObjects/EMBL/EMBLBeamlineTest.py
```diff
@@ -116,6 +116,8 @@ def start_test_queue(self, test_list, create_report=True):
                 logging.getLogger("HWR").warning(
                     "BeamlineTest: Unable to create test directories"
                 )
+
+                logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 120:
```suggestion
                logging.getLogger("HWR").exception(
                    "BeamlineTest: Unable to create test directories"
                )
```

mxcubecore/HardwareObjects/ESRF/ID29HutchTrigger.py
```diff
@@ -42,6 +42,8 @@ def init(self):
             self.card, self.channel = map(int, PSSinfo.split("/"))
         except Exception:
             logging.getLogger().error("%s: cannot find PSS number", self.name())
+
+            logging.getLogger("HWR").exception("")
```
fabcor-maxiv, line 46:
```suggestion
            logging.getLogger("HWR").exception("%s: cannot find PSS number", self.name())
```

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -1927,6 +1927,8 @@ def calculate_recentring(self, okp, ref_okp, ref_xyz):
                 err.returncode,
                 err.output,
             )
+
+            logging.getLogger("HWR").exception("")
```
fabcor-maxiv:
```suggestion
            logging.getLogger("HWR").exception(
                "Recen failed with returncode %s. Output was:\n%s",
                err.returncode,
                err.output,
            )
```

fabcor-maxiv (commented):
OK, I started trying to write fix suggestions, but then realised that this is probably all automatically generated and you might not be interested in manual edits at this stage... I do not know exactly what the plan is. Anyway, while I fully agree with the initiative, I am not 100% convinced that the approach is ideal. In any case, it seems better than doing nothing, but could you tell what the next steps planned are?

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -1927,6 +1927,8 @@ def calculate_recentring(self, okp, ref_okp, ref_xyz):
                 err.returncode,
                 err.output,
             )
+
```
rhfogh: No, there was already a logger message before the return

rhfogh:
I agree with @fabcor-maxiv - this is a very good initiative, but I am not sure it should be merged as it stands. This is clearly done by script, and I notice in at least one case the script has added a log message before a return - even though there was a log message there already. I think someone ought to go through these by hand, one by one, correct as appropriate, and maybe change to get_log (or whatever) It is easier to do it now, when the script output show you exactly where you should look, than it would be to do it after this has been merged.. Simple scripts are very useful, but  - pace Kahnemann ;-)  - they are not a full substitute for human interaction.

marcus-oscarsson:
> OK, I started trying to write fix suggestions, but then realised that this is probably all automatically generated and you might not be interested in manual edits at this stage... I do not know exactly what the plan is. Anyway, while I fully agree with the initiative, I am not 100% convinced that the approach is ideal. In any case, it seems better than doing nothing, but could you tell what the next steps planned are?

So yes you are right I did not manually edit all the files, I made a script to do it for me. And yes I partly sold my sole to the devil and asked an AI tool to generate parts of that script. 

The plan is to: 

    1. Replace all the simple cases of pass and return with an exception logger (this PR)
    3. Get rid of `traceback` in `except`
    5. Make use of self.log where applicable 

I still don't know what to do with `error`, `warning`, `info` messages, but at the moment I thought it was better to also ad a separate `exception` log. That's how I read the intent of those logger statements; that the author wanted a separate "user-friendly" log entry. I think that we do on most cases also want a `exception` log.

Id propose that if there are files (or a individual change) that you'd like to remove from the PR that we do so and then we can continue with the next steps.

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
marcus-oscarsson: Yes but it does not give us the traceback. I think this exception is actually quite well handled so we can remove the "fix" added by this PR. Its not the case however in most other cases.

marcus-oscarsson:
> I agree with @fabcor-maxiv - this is a very good initiative, but I am not sure it should be merged as it stands. This is clearly done by script, and I notice in at least one case the script has added a log message before a return - even though there was a log message there already. I think someone ought to go through these by hand, one by one, correct as appropriate, and maybe change to get_log (or whatever) It is easier to do it now, when the script output show you exactly where you should look, than it would be to do it after this has been merged.. Simple scripts are very useful, but - pace Kahnemann ;-) - they are not a full substitute for human interaction.

You are right we need to go through this and remove the cases we don't want "fixed". This goes through all `except` clauses not only the ones for `HardwareObjects` so we can't easily apply the `self.log` construct without making the script quite complex.

What I propose that we do, and that will also take care of the other cases that are still not updated to `self.log` is to make another script that only changes HardwareObjects and make the necessary update.

marcus-oscarsson:
I wanted to do these changes in a separate PR but if you prefer I can make the change to `self.log`. It will however be a big change because Ill then change all logger statements.

marcus-oscarsson:
Alright I did like this;
 - I made a separate PR for the switch to `self.log` its made to this branch, I think it will make the review easier
 - I removed the cases that guard against import of libraries that does no exist
 - I removed cases where the exception was handled somewhat alright
 - I kept the cases where I thought the exception class is to general for simply logging with `error` or `warning` and I added an additional `exception` log

rhfogh (approved):
I am happy with this this as it stands now.

And, yes, it is very good that you got this done :-)

elmjag:
image

:smile_cat:

PR #1407, opened 2025-10-07
Replaced logging.getLogger("HWR") with self.log

marcus-oscarsson:
Replaced logging.getLogger("HWR") with self.log

### DIFF
--- a/mxcubecore/HardwareObjects/CentringMath.py
+++ b/mxcubecore/HardwareObjects/CentringMath.py
@@ -1,4 +1,3 @@
-import logging
 import math
 
 import numpy
@@ -38,8 +37,10 @@ class CentringMath(Procedure):
                     "type": axis["type"],
                     "direction": eval(axis["direction"]),
                     "motor_name": axis["motorname"],
-                    "motor_HO": HWR.get_hardware_repository().get_hardware_object(
-                        axis["motorHO"]
+                    "motor_HO": (
+                        HWR.get_hardware_repository().get_hardware_object(
+                            axis["motorHO"]
+                        )
                     ),
                 }
 
@@ -80,7 +81,7 @@ class CentringMath(Procedure):
     def listOfCentringsToScreen(self, list_of_centring_dicts):
         self.factorize()
         lst = []
-        logging.getLogger("HWR").debug(
+        self.log.debug(
             " in listOfCentringToScreen - %s points in list "
             % len(list_of_centring_dicts)
         )
--- a/mxcubecore/HardwareObjects/EMBLFlexHarvester.py
+++ b/mxcubecore/HardwareObjects/EMBLFlexHarvester.py
@@ -208,7 +208,7 @@ class EMBLFlexHarvester(EMBLFlexHCD):
         try:
             _tt = time.time()
             self._wait_busy(300)
-            logging.getLogger("HWR").info(f"Waited SC activity {time.time() - _tt}")
+            self.log.info(f"Waited SC activity {time.time() - _tt}")
         except Exception:
             logging.getLogger("user_level_log").error(
                 "ERROR While Waited SC activity to start"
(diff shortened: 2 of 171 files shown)


mxcubecore/HardwareObjects/CentringMath.py
```diff
@@ -38,8 +37,10 @@ def get_gonio_axes():
                     "type": axis["type"],
                     "direction": eval(axis["direction"]),
                     "motor_name": axis["motorname"],
-                    "motor_HO": HWR.get_hardware_repository().get_hardware_object(
-                        axis["motorHO"]
+                    "motor_HO": (
+                        HWR.get_hardware_repository().get_hardware_object(
+                            axis["motorHO"]
+                        )
```
fabcor-maxiv, line 43: Unrelated change, but I guess it is fine, although surprising that it happened now, since we did not update the Ruff version, unless maybe you used a Ruff version different than the one pinned in the project.

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
```diff
@@ -237,7 +237,7 @@ def _do_load(self, sample=None):
                 logging.getLogger("user_level_log").error(
                     "ERROR While SC activity After Loaded Sample "
                 )
-                logging.getLogger("HWR").error(msg)
+                self.log.error(msg)
                 logging.getLogger("user_level_log").error(msg)
```
fabcor-maxiv, line 241: Right, we would need to do those as well, the user log, but it is much less important right now.

mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -750,8 +750,12 @@ def add_beamline_configuration_metadata(self, metadata, beamline_config):
...
                 "detector_px": "InstrumentDetector01_beam_center_x",
                 "detector_py": "InstrumentDetector01_beam_center_y",
-                "beam_divergence_vertical": "InstrumentBeam_vertical_incident_beam_divergence",
-                "beam_divergence_horizontal": "InstrumentBeam_horizontal_incident_beam_divergence",
+                "beam_divergence_vertical": (
+                    "InstrumentBeam_vertical_incident_beam_divergence"
+                ),
+                "beam_divergence_horizontal": (
+                    "InstrumentBeam_horizontal_incident_beam_divergence"
+                ),
```
fabcor-maxiv, line 758: Ah, another formatter hiccup I guess :)

fabcor-maxiv (approved):
Difficult to be confident when there are so many lines changed, but is seems fine :)

I am glad to see this happening!

I spotted some unrelated changes, I only marked two of them. My guess is that maybe you used a different Ruff formatter version, could that be?

mxcubecore/HardwareObjects/ICATLIMS.py
marcus-oscarsson: Yes, not sure what happened there ....

mxcubecore/HardwareObjects/EMBLFlexHarvester.py
marcus-oscarsson: Yepp

marcus-oscarsson: I thought Ill do those separately

marcus-oscarsson:
> Difficult to be confident when there are so many lines changed, but is seems fine :)

Yes I know, but I thought its better to get done with it :)

> 
> I am glad to see this happening!
> 
> I spotted some unrelated changes, I only marked two of them. My guess is that maybe you used a different Ruff formatter version, could that be?

I have no idea why it does that, I have `ruff 0.9.10`.

fabcor-maxiv:
> > I spotted some unrelated changes, I only marked two of them. My guess is that maybe you used a different Ruff formatter version, could that be?
> 
> I have no idea why it does that, I have `ruff 0.9.10`.

Strange... Anyway, does not matter, the CI check is green.

PR #1408, opened 2025-10-08
Dosc: Add ``Exceptions`` section

dominikatrojanowska:
This is a continuation of https://github.com/mxcube/mxcubecore/pull/1389 discussion

This is my proposition (based on discussion, best practices and frequently occurring imperfections in our code base) of what could be included in this section.

Feel free to add comments here or within related PR https://github.com/mxcube/mxcubecore/pull/1409

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -327,6 +327,126 @@ The user log is aimed to convey information to the users.
 The debug and exception logging should be used for logging information suitable for developers,
 to help with troubleshooting.
 
+#### Exceptions
+
+Using `try/except` clause to catch exceptions apply the following rules. Unless you have strong arguments to do it another way.
+
+- Do not catch *blind* and *bare* exceptions, unless there is no better option and there is a real need to catch all exceptions. In this case make sure to log the whole traceback by using **exception** method of the logger object *Blind* exception catches `BaseException` which is a parent of all exception classes (`KeyError`, `Exception`, `SystemExit` etc.). This may lead to unpredicted behaviour for instance with interrupting the program (ctrl+C)
+  `Incorrect code`
+
+  ```python
+  try:
+      # some code
+  except BaseException:
+      # handle exception
+  ```
+
+  *Bare* except clause comes down to the same as *blind* exception. Avoid it from the same reason.
+  `Incorrect code`
+
+  ```python
+  try:
+      # some code
+  except:
+      # handle exception
+  ```
+
+  Instead use a `try/except` with explicitly catched exception (list), or if there is a reason, with `Exception` base class that handles most of the common cases except system errors (like `KeyboardInterrupt` and `SystemExit`).
+  `Correct code`
+
+  ```python
+  try:
+      # some code
+  except KeyError:
+      # handle exception
+  except Exception:
+      # handle other exceptions
+  ```
+
+- Log full traceback when logging during handling exceptions. This may be done by running **exception** method of the logging object, which is preffered way. If the logging level really have to be at another level set `exc_info` flag to `True`.
+  `Correct code`
+
+  ```python
+  try:
+      # do something
+  except Exception:
+      logging.getLogger("").debug(msg, exc_info=True)
+  ```
+
+  However remember that this is the worse version:
+  `Incorrect code`
+
+  ```python
+  try:
+      # do something
+  except Exception:
+      logging.getLogger("").error(msg, exc_info=True)
+  ```
+
+  of below example, that will be always the preffered one:
+  `Correct code`
+
+  ```python
+  try:
+      # do something
+  except Exception:
+      logging.getLogger("").exception(msg)
+
+  ```
+
+- Re-raising exception is fine, however aviod redundant code:
+  `Incorrect code`
+
+  ```python
+  try:
+      # do something
+  except Exception:
+      raise
+  ```
+
+  Instead simple skip `try/except` at all:
+  `Correct code`
+
+  ```python
+  # do something
+  ```
+
+  Another example:
+  `Incorrect code`
+
+  ```python
+  try:
+    # do something
+  except Exception as exc:
+    # do something else
+    raise exc
+  ```
+
+  Instead:
+  `Correct code`
+
+  ```python
+  try:
+    # do something
+  except Exception:
+    # do something else
+    raise
+  ```
+
+- Importing the modules that may be not present or need alternatives does not requier neither re-raising exceptions nor logging:
+  `Correct code`
+
+  ```python
+  try:
+      import Queue as queue
+  except ImportError:
+      import queue
+  ```
+
+- ...
+
+<!-- WIP -->
+
 ### Continuous integration (CI)
 
 GitHub Action are used for continues integration


CONTRIBUTING.md
```diff
@@ -327,6 +327,57 @@ The user log is aimed to convey information to the users.
...
+
+  ```python
+  try:
+      # some code
+  except Exception:
+      self.log.exception("Some good explenation what happend")
+      # handle exception
+  ```
+
+  or do both.
```
rhfogh: There are cases where exceptions are used for non-exceptional situations, and here one should neither re-raise the error or log an error message. Both examples are from current mxcubecore code:

- When importing modules that may not be present, or where you need one of several alternatives. e.g.:
```
try:
     import Queue as queue
except ImportError:
    import queue
```

- When it is (felt to be) quicker or neater to assume something works than to test for it first, e.g.:
```
for xml_path in something:
    try:
        attribute_set_2 = s.attributes[xml_path]
    except KeyError:
        return False
    else:
        if not attribute_set.issubset(attribute_set_2):
            return False
return True
```

These are AFAIAC perfectly legitimate approavhes, so I think the documentation should allow explicitly for this kind of usage instead of just saying "unless you have strong arguments for doing it the other way".

dominikatrojanowska: I can agree with importing, but I think it would be a good manner to add comment in the second case... but I will look more carefully into our code base and think about it :)
 
I can modify the description a bit and add exclusions covering this topic (will include the first one for now)

fabcor-maxiv: Both these examples showed by @rhfogh seem like perfectly normal Python code indeed. Many words have been written on this topic, this article for example: <https://realpython.com/python-lbyl-vs-eafp/#the-pythonic-way-to-go-lbyl-or-eafp>.

I think this pull request should focus and what was the essence of the previous discussion which was actually about logging. I think that in short, what we should say here is something along the lines of:

> Do not catch [blind exceptions](https://docs.astral.sh/ruff/rules/blind-except/) (which is obviously very poor coding in general, not just in Python, not just in MXCuBE) but if you really do not know how to make it better and if you really do have to catch all exceptions then make sure to log the whole traceback by using the `exception` method of the logger object.

That's pretty much what we agreed on in the previous discussion, and would be a great start that we could expand later on.

dominikatrojanowska: I changed whole section a bit, trying to address you suggestions :)

CONTRIBUTING.md
```diff
@@ -327,6 +327,126 @@ The user log is aimed to convey information to the users.
...
+  ```python
+  try:
+      # some code
+  except KeyError:
+      # handle exception
+  except Exception:
+      # handle other exceptions
+  ```
+
+- Log full traceback when logging during handling exceptions. This may be done by running **exception** method of the logging object, which is preffered way. If the logging level really have to be at another level set `exc_info` flag to `True`.
```
dominikatrojanowska, line 366: I wrote this, however this case appears quite often: https://github.com/mxcube/mxcubecore/blob/95c26995f51a1f8974a58c1bf91bdebad7bd712f/mxcubecore/HardwareObjects/Energy.py#L39 as the example. If we agree it's ok, I will change the description

rhfogh:
This still leaves the impression that not raising or logging exceoption is always naughty, except in the specific case of imports. I think that is unfortunate, especially in the current climate of 'obey the linter or else!'. Python programmers are expected to behave as sensible adults, and should not be discouraged from writing sensible code in order to comply with a formal rule.
 
Could we not write something like :

> In Python exceptions are sometimes used to deal with situations that are not exceptional, but as an alternative to `if` statements and tests where an exception is the most appropriate way of solving a problem. In these cases exceptions do not need to be either re-raised or logged (nothing exceptional is going on), but unless the situation is obvious it is recommended to add a comment explaining it. Note that when executable tasks or hardware object loading fails it *is* important to log the error so that people can see why the failure happened. One example of a 'constructive' exception would be to check for the presence of an importable module:  EXAMPLE HERE

marcus-oscarsson:
I think that the two, perhaps, most important parts is:

   - **Errors should never be hidden,** always make sure that an exception is at-least logged unless it can't be handled in evident and robust way. Make sure that context is kept by using `raise from`  when raising a new exception in a except clause
 
   - **The code should be easy to read/understand,** Limit the statements in the try clause so that its clear which statement it is that throws the actual exceptions (the other statements do not need to be an the try. If there are multiple add a comment.

dominikatrojanowska:
I feel like this
> This still leaves the impression that not raising or logging exception is always naughty

is quite opposite to this
> * **Errors should never be hidden,** always make sure that an exception is at-least logged unless it can't be handled in evident and robust way

Do we agree that this a robust and evident way where the logging could be skipped?
```
try:
     value = dictionary[key]
except KeyError:  # or wahtever
    return False
return value 
```

dominikatrojanowska:
> * Make sure that context is kept by using `raise from`  when raising a new exception in a except clause

I think this is a redundant part of code, according to https://docs.astral.sh/ruff/rules/verbose-raise/
I checked that in terminal:

```python
try:
    1/0
except Exception as exc:
    raise KeyError from exc
```
otput:
```
  File "<stdin>", line 2, in <module>
ZeroDivisionError: division by zero

The above exception was the direct cause of the following exception:

  File "<stdin>", line 4, in <module>
KeyError
```

Without `raise ... from ...`:
```python
try:
    1/0
except Exception:
    raise KeyError
```
output:
``` 
  File "<stdin>", line 2, in <module>
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

  File "<stdin>", line 4, in <module>
KeyError
```
The only difference visible is this middle sentence.  
The underlying difference is described here: https://docs.python.org/3/library/exceptions.html#exception-context
Since the outcome is quite the same, the question is if we care to which attribute `__context__` or `__cause__` is the handled exception attached. AFAIAC this makes no difference from the debugging perspective, meaning the whole traceback is displayed. 

If the final decision is to stick to the first option, then the general ruff rule will be better instead of what we have now.

rhfogh:
@dominikatrojanowska I would prefer this style (as would you):
```

try:
    1/0
except Exception:
    raise KeyError
```
As being simpler, but one could live with either

fabcor-maxiv:
Again, in a first round I would stick to some simple statements that @elmjag had intended to have in <https://github.com/mxcube/mxcubecore/pull/1389> in the first place. We should not aim at teaching Python here. I think a simple paragraph with two or three sentences that addresses the most pressing issue in our code base is more than enough. The most pressing issue being that some actual errors are either swallowed without a trace in the log or logged without the full stack trace.

It is kind of shocking to me that we have to write this kind of things in MXCuBE documentation at all, since there is nothing specific to MXCuBE here. But, obviously there is an issue in MXCuBE code, and I guess having such a mention in our documentation can make sense, just so that we can point at it when we see such things in code review.

rhfogh:
@dominikatrojanowska I would almost *) agree with the 'Errors should never be hidden' statement - but that is not actually in the text of the PR. I am commenting on the PR as it stands.

*) I would still add a sentence like "Where exceptions are used as an integral part of the code flow rather than for detecting errors, log messages will generally not be needed."

@fabcor-maxiv I agree that it ought to be enough to address the most pressing issues of our code base and leave the rest to short, general advice. The problem is that the use of Ruff is extremely prescriptive and does not by any means limit itself to 'pressing issues in the code base'. And, as we have seen over recent months, people are actually doing considerable work to satisfy Ruff even where there is no objective need, or where the fixes applied do not solve the underlying problem but only sweeps it under the carpet. If we could get back to concentrating on the problems that actually need fixing, we would not need to agonise over the documentation.  But as long as people are compelled to do a lot of work to respect the letter of the guidelines, with little regard to whether this is useful in any deeper sense, I think we are forced to spend a lot of energy up front in making sure that the letter of the guidelines are as precise and sensible as we can make them, and do not encourage making unnecessary changes.

marcus-oscarsson:
Ill repeat what I said above and I really think this section does not need to contain much more than something like:

> * **Errors should never be hidden,** always make sure that an exception is at-least logged unless it can't be handled in evident and robust way. Make sure that context is kept by using `raise from`  when raising a new exception in a except clause
> 
> * **The code should be easy to read/understand,** Limit the statements in the try clause so that its clear which statement it is that throws the actual exceptions (the other statements do not need to be an the try. If there are multiple add a comment.

Exceptions in python are sometimes used as a sort control statement as is mentioned above. It naturally sounds weird to say that an exception is part of an expected code flow, but it is a very common pattern. I would not really consider such cases exception handling but part of normal code flow. 

Exception handling is about handling an unexpected, erroneous or unhandled case. Its in such cases very often important to be able to find the stacktrace, not always but almost always. In some cases we can handle the exception and simply log with `warning` or perhaps `error` but I would advocate that having the stacktrace would not be bad.  I think that its important that its done in a homogeneous way using `logging.exception`, and not any combination of traceback or sys with print, so that the stack trace is logged on file (or wherever its stored)

rhfogh:
@marcus-oscarsson What you say is very sensible. But I have the following situation in mind:

- We leave the documentation as you suggest.
- Some months from now, someone, linter or person, will flag some code that contains an exception that is not logged, fail it, point to the documentation and say '_You bad. Fix!_'. 
- Whoever wrote that, probably me, will then have to argue about the interpretation of a text that we all have passed and waste everybody's time, when it will be so much easier to just abide by the letter of the documentation, never mind  that it might be '_a very common pattern_' and should not really '_be considered as exception handling'_.  @fabcor-maxiv wants some text that he can _'point at it when we see such things in code review.'_ What is that for, except to say _'It says so there. Comply!_'

I would love it if you could convince me that such a thing would never happen, but currently my experience suggests otherwise. Meanwhile I would rather have the discussion now, when we are writing out the rules we have to follow, than later, when it will all be considered long settled.

marcus-oscarsson:
I see your concern. I hope that not anyone of us would critic code that is correct by using documentation to confound the actual intention of the code or documentation in question. 

To me "...it can't be handled in evident and robust way." adds enough of room to add any cases you would not need to log but we can expand it to be broader somehow. It can be rephrased. But I don't think we need to list all "valid" and "invalid" cases/examples of writing `try-except`.

rhfogh:
@marcus-oscarsson I do not at all share your optimism. But I have had my say. I will not suggest further changes to this PR.

dominikatrojanowska:
closing this PR, the new one will be provided including this discussion :)

marcus-oscarsson:
When it comes to documenting this I believe that its perhaps easier to do so in more general terms without listing to much code examples. We can point to the official python documentation for functionality that we think is important to consider.

The most important parts would be to mention that;

  - The code should be easy to understand
  - The developers get the right information to be able to find the issue
  - The users get the right information to be able to understand "roughly" whats going on

We should configure the linter so that it tells us when we handle exceptions in way that we consider incorrect but still ignore the rules that we think are not useful in our project. That may not be possible in all cases and then we have to refer to the paragraph above and that the code is peer-reviewed.

rhfogh:
@marcus-oscarsson Your latest comment is strongly seconded.

PR #1409, opened 2025-10-08
Reworking on exception handling imperfections

dominikatrojanowska:
According to Exceptions section in the documentation and this PR: https://github.com/mxcube/mxcubecore/pull/1408

### DIFF
--- a/mxcubecore/BaseHardwareObjects.py
+++ b/mxcubecore/BaseHardwareObjects.py
@@ -690,7 +690,7 @@ class HardwareObjectMixin(CommandContainer):
             try:
                 _attr = getattr(self, attr_name)
             except AttributeError:
-                logging.getLogger("HWR").error(
+                logging.getLogger("HWR").exception(
                     f"{attr_name} configured as exported for {self.name} but not implemented"
                 )
                 continue
--- a/mxcubecore/Command/Epics.py
+++ b/mxcubecore/Command/Epics.py
@@ -111,8 +111,8 @@ class EpicsCommand(CommandObject):
                     if ret is not None:
                         self.emit("commandReplyArrived", (ret, str(self.name())))
                         return ret
-                except Exception as e:
-                    logging.getLogger("HWR").error(
+                except Exception:
+                    logging.getLogger("HWR").exception(
                         "%s: an error occured when getting value with Epics command %s",
                         str(self.name()),
                         self.pv_name,
@@ -126,8 +126,8 @@ class EpicsCommand(CommandObject):
                     value = args[0]
                     wait = kwargs.get("wait", False)
                     self.pv.put(value, wait=wait)
-                except:
-                    logging.getLogger("HWR").error(
+                except Exception:
+                    logging.getLogger("HWR").exception(
                         "%s: an error occured when putting a value with Epics command %s",
                         str(self.name()),
                         self.pv_name,
--- a/mxcubecore/Command/Pool.py
+++ b/mxcubecore/Command/Pool.py
@@ -79,7 +79,7 @@ class PoolCommand(CommandObject):
             self.device = PyTango.DeviceProxy(self.device_name)
         except PyTango.DevFailed as traceback:
             last_error = traceback[-1]
-            logging.getLogger("HWR").error(
+            logging.getLogger("HWR").exception(
                 "%s: %s", str(self.name()), last_error["desc"]
             )
             self.device = None
@@ -103,10 +103,8 @@ class PoolCommand(CommandObject):
                 ret = tango_cmd_object(
                     args
                 )  # eval('self.device.%s(*%s)' % (self.command, args))
-            except PyTango.DevFailed as error_dict:
-                logging.getLogger("HWR").error(
-                    "%s: Tango, %s", str(self.name()), error_dict
-                )
+            except PyTango.DevFailed:
+                logging.getLogger("HWR").exception("%s: Tango", str(self.name()))
             except Exception:
                 logging.getLogger("HWR").exception(
                     "%s: an error occured when calling Tango command %s",
@@ -191,7 +189,7 @@ class PoolChannel(ChannelObject):
             self.device = PyTango.DeviceProxy(self.device_name)
         except PyTango.DevFailed as traceback:
             last_error = traceback[-1]
-            logging.getLogger("HWR").error(
+            logging.getLogger("HWR").exception(
                 "%s: %s", str(self.name()), last_error["desc"]
             )
         else:
--- a/mxcubecore/Command/Tine.py
+++ b/mxcubecore/Command/Tine.py
@@ -66,10 +66,6 @@ class TineCommand(CommandObject):
                 self.tineName, self.commandName, commandArgument, self.timeout
             )
             self.emit("commandReplyArrived", (ret, str(self.name())))
-        except IOError as ex:
-            logging.getLogger("user_level_log").exception("TINE")
-            self.emit("commandFailed", str(ex))
-            raise
         except Exception as ex:
             logging.getLogger("user_level_log").exception("TINE")
             self.emit("commandFailed", (str(ex)))
@@ -79,7 +75,7 @@ class TineCommand(CommandObject):
         result = None
         try:
             result = tine.get(self.tineName, self.commandName, self.timeout)
-        except (IOError, Exception):
+        except Exception:
             logging.getLogger("HWR").exception("")
         return result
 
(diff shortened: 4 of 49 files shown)


mxcubecore/BaseHardwareObjects.py
```diff
@@ -1016,7 +1017,8 @@ def disconnect(self, signal, slot, sender=None): ...
                 signal = sender
                 sender = self
             else:
-                raise ValueError("invalid slot (None)")
+                msg = "invalid slot (None)"
+                raise ValueError(msg)
```
dominikatrojanowska: The ruff complains about this rule https://docs.astral.sh/ruff/rules/raise-vanilla-args/ but my precommit hints indicates: `TRY003 Avoid specifying long messages outside the exception class`  

This proposed solution seems somehow strange to me. Another option is to disable that rule in inline comment https://github.com/mxcube/mxcubecore/pull/1409/files#diff-53c8f01f41b001eb75e17aa77ab3dbcaeb9a6a5ac7421ef391c9dc9484dae136R143  or to exclude it generally in `ruff.toml` :thinking:

marcus-oscarsson: The rule is actually good and not so strange, it prevents the actual message from being printed as part of the stacktrace (sort of an issue with python itself I guess but still). Otherwise "invalid slot (None)" will be part of the stack trace which sometimes can be confusing for longer messages.

rhfogh: Ah, there is some confusion here. Rule TRY003 (referred to by @dominikatrojanowska and matching her pointed-to example) seems to be saying that it is bad to pass strings to raised exceptions - apparently the message should be defined in a custom exception class. I'd say that this is an extremely bad idea for our purposes, and that TRY003 should be permanently and globally disabled. 

The example @dominikatrojanowska points to says : 
`                raise RuntimeError("Processing stalled")`
I would say that RuntimError with a passed-in string is *exactly* what you should do with an error condition that does not match any of the generic cases that we know of and that is not particularly easy to generalise. 
If we are to follow Ruff, we should make a custom class called 'ProcessingStalledError' or something similar. Now if we were making a general framework of moderate size we might decide to create a family of custom errors to cover the generic things that might go wrong.  For the large and very varied code base we have this seems like a hopeless endeavour and a waste of time to try, but if we do want that, we should start by making the classes of generic exceptions, in a thought-out way and then refactor the entire code base before we put in Ruff rules to follow. And then we would *still* have lots of exceptions that did not merit a separate class and should not be cluttered with individual noqa lines, but at least we would have achieved something.

The Ruff statement Marcus has in mind and that is shown in line 1019 is EM101 (https://docs.astral.sh/ruff/rules/raw-string-in-exception/). IMHO this is *not* a good rule, just  pointless. By following that rule you make the code longer and more complex to read, and you make the error message no more informative or intelligible than it was before, just slightly neater. Have a look. In fact I strongly suspect that whoever invented that rule suffered from OCD,. Since it is only a minor irritant we can live with it, but I would love to put that rule under 'permanently disabled' as well.

marcus-oscarsson: Sorry, it seems that I mixed up the rules. I kind of like the rule `EM101` but can definitely live without it. I do think that the stack traces could be better formatted. 

I don't think its a good idea to implement specific exception sub-classes unless you intend to later use those to handle some specific cases. I think `raise RuntimeError("Processing stalled")` is just fine unless you need to somewhere in the code make the difference between the various errors later in the code.

mxcubecore/HardwareObjects/MD2Motor.py
```diff
@@ -119,20 +119,22 @@ def get_dynamic_limits(self):
...
-            if low_lim == float(1e999) or hi_lim == float(1e999):
-                raise ValueError
-            return low_lim, hi_lim
         except Exception:
-            return (-1e4, 1e4)
+            low_lim, hi_lim = -1e4, 1e4
+        else:
+            if low_lim == float(1e999) or hi_lim == float(1e999):
+                low_lim, hi_lim = -1e4, 1e4
+        return low_lim, hi_lim
```
fabcor-maxiv, line 127: Are we sure about this?

mxcubecore/HardwareObjects/MD2Motor.py
```diff
@@ -119,20 +119,22 @@ def get_dynamic_limits(self):
...
```
fabcor-maxiv, line 137: And this?

ruff.toml
fabcor-maxiv, line 1: Thanks for cleaning this up as well :)

fabcor-maxiv (commented):
Seems generally fine.

Was it automatically generated by a tool or done manually?

Some of those changes are obviously straightforward, but some seem a bit more tricky, so I am a bit worried that they might be backwards incompatible changes.

mxcubecore/BaseHardwareObjects.py
dominikatrojanowska: ok, I will at this rule as general exclusion

mxcubecore/HardwareObjects/MD2Motor.py
dominikatrojanowska: yep, this one (and the next one) was manual, and it's not so tricky :) should I explain? 
I limited to one return statement and removed the `ValueError` raising that would be caught anyway by following `except`

beteva:
Hi, it is very good initiative to do this cleaning.
I have a question  - should we move the legacy hardware objects (like MD2Motor, replaced by ExporterMotor) to some legacy code directory or purely remove them, so the cleanup is not done for obsolete files?

marcus-oscarsson:
I think we can start to remove files

dominikatrojanowska:
>  should we move the legacy hardware objects (like MD2Motor, replaced by ExporterMotor) to some legacy code directory or purely remove them, so the cleanup is not done for obsolete files?

> I think we can start to remove files

rather in another PR :)

mxcubecore/Command/Pool.py
```diff
@@ -191,7 +189,7 @@ def __init__(
             self.device = PyTango.DeviceProxy(self.device_name)
         except PyTango.DevFailed as traceback:
             last_error = traceback[-1]
-            logging.getLogger("HWR").error(
+            logging.getLogger("HWR").exception(
                 "%s: %s", str(self.name()), last_error["desc"]
             )
```
dominikatrojanowska, line 194: Hm, I'm not sure what to do with such code... 
I guess, we are consciously limiting the PyTango traceback since it may be long?  
If not my suggestion is:
```python
        except PyTango.DevFailed:
            logging.getLogger("HWR").exception(self.name())
```

fabcor-maxiv:
I have not checked, and hopefully I am worried for nothing, but doesn't this conflict with <https://github.com/mxcube/mxcubecore/pull/1410>, in the sense that it attempts to make similar changes in the similar places?

dominikatrojanowska:
> I have not checked, and hopefully I am worried for nothing, but doesn't this conflict with #1410, in the sense that it attempts to make similar changes in the similar places?

Rebased, so should be fine :) those PRs are simlar but not exactly the same. Here, in contrast to https://github.com/mxcube/mxcubecore/pull/1410/ site-specific HWO are skipped, left for site developers.

mxcubecore/Command/Tine.py
```diff
@@ -66,21 +66,15 @@ def __call__(self, *args, **kwargs):
                 self.tineName, self.commandName, commandArgument, self.timeout
             )
             self.emit("commandReplyArrived", (ret, str(self.name())))
-        except IOError as strerror:
-            logging.getLogger("user_level_log").exception("TINE: %s" % strerror)
-            self.emit("commandFailed", (strerror))
-            raise strerror
         except Exception as ex:
```
dominikatrojanowska, line 69: IOError inherits from Exception

mxcubecore/HardwareObjects/MD2Motor.py
fabcor-maxiv: The original code is so strange to me this breaks my head. But the change seems correct to me.

ruff.toml
```diff
@@ -14,6 +14,7 @@ ignore = [
     "PLR2004",  # magic-value-comparison: ignored, too many hits (249)
     "UP",  # pyupgrade: ignored for now, can be fixed automatically by ruff itself
     "G004", # Allow uses of f-strings to format logging messages
+    "TRY003", # Allow to add message when instantiating built-in exception
```
fabcor-maxiv: I'd prefer such a change to be done in its own pull request.

As far as I understand there is only one case of `TRY003` that is impacted by this change, isn't it?

dominikatrojanowska: https://github.com/mxcube/mxcubecore/pull/1414

mxcubecore/Command/Epics.py
```diff
@@ -126,8 +126,8 @@ def __call__(self, *args, **kwargs):
                     value = args[0]
                     wait = kwargs.get("wait", False)
                     self.pv.put(value, wait=wait)
-                except:
-                    logging.getLogger("HWR").error(
+                except Exception:
+                    logging.getLogger("HWR").exception(
```
marcus-oscarsson, line 130: Not certain Id do like this, id probably keep the error but possibly also add the `exception` log

marcus-oscarsson:
I would recommend to split this PR into several small to avoid or at-least control unexpected regression with some of the logical rewrites. It looks alright but there are a few and one can easily make a mistake. Would it be possible to split this PR so that the we separate between the simpler additions of logging.exception and the re-ordering of try-except-else statements ?

marcus-oscarsson:
Ill close this PR please feel free to open several smaller PRs that are easier to review and test

PR #1418, opened 2025-10-21
changed beamline.yaml to beamline_config.yaml

beteva:
@elmjag  in your previous PR you've changed beamline_config to beamline. Was this for purpose? I find it better with _config, as indeed it describes the configuration of the beamline, not the beamline itself.

### DIFF
--- a/mxcubecore/HardwareRepository.py
+++ b/mxcubecore/HardwareRepository.py
@@ -83,7 +83,7 @@ beamline = None
 # Supported beamline configuration filenames.
 # The first name in the list takes precedence.
 #
-BEAMLINE_CONFIG_FILES = ["beamline.yaml", "beamline_config.yml"]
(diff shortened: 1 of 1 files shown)


fabcor-maxiv:
Why not both? But I rather agree with @elmjag's and @marcus-oscarsson's [point of view](https://github.com/mxcube/mxcubecore/pull/1400#issuecomment-3365743893) that no other file seems to have the `_config` suffix so that seems odd. Why the `_config` suffix only for `beamline`? Isn't the beamline an hardware object like any other?

elmjag:
The consensus in the PR was `beamline.yaml` feels most logical, as this is the naming schema we use for all other hardware objects.

Here is the discussion:
https://github.com/mxcube/mxcubecore/pull/1400#issuecomment-3365736003

beteva:
OK, the conversation has slipped out of my memory. I withdraw the PR

PR #1421, opened 2025-10-21
Increase sleep time in order for epics PV to connect

RafaelLyra8:
Without this increase sleep time, so EPICS PVs may not connect.


marcus-oscarsson:
Thanks, could you re-base and then merge please :)

marcus-oscarsson (approved):

PR #1423, opened 2025-10-21
Update LNLS Classes 

RafaelLyra8:
- Remove all of LNLS Classes (This was done because they are not being currently used and we are updating our classes in the transition to MXCuBE 4)
- Add authentication with ICAT


marcus-oscarsson:
Thanks very nice to see, so this should I guess contain mostly changes to your site specific files. I think there is a rebase missing ? This PR somehow contain previous commits done other people.

PR #1424, opened 2025-10-21
Create LNLSLIMS for LNLS Authentication

RafaelLyra8:
- Implement ICATLIMS fix in order to authenticate with both username and password. This was done because the pyicat plus used in the LNLS needs both username and password for the login.
- Implement LNLSLIMS


marcus-oscarsson:
Thanks !

PR #1425, opened 2025-10-21
Cleanup Hardware Objects LNLS

RafaelLyra8:
Remove all of LNLS Classes (This was done because they are not being currently used and we are updating our classes in the transition to MXCuBE 4)


fabcor-maxiv (approved):
Godspeed!

PR #1426, opened 2025-10-22
Fix file mode bits

fabcor-maxiv:
Some files had mode bits different from `644` for no reason.


fabcor-maxiv:
There are a lot more of such files, but they are in non-shared directories (site-specific directories for example).

rhfogh (approved):
Fine with me

PR #1428, opened 2025-10-22
Clean up `/docs/conf.py`

fabcor-maxiv:
`copyright` is a built-in Python variable. From Sphinx's point of view, `project_copyright` is an alias of `copyright`.

The parent directory `/docs` is not meant to be an importable package.


marcus-oscarsson:
Thanks, could you please rebase and merge ?

marcus-oscarsson (approved):

PR #1429, opened 2025-10-22
Let Ruff format Python code embedded in docstrings

fabcor-maxiv:


marcus-oscarsson:
Ok, if this works well but is it really necessary ?

marcus-oscarsson:
I'm hesitant but I'm fine with it works well. Please feel free to rebease and merge

PR #1430, opened 2025-10-23
Remove value_changed call at init

RafaelLyra8:
This was done because this function will always break, as the self.__value_changed_callback_ref will only be set in the poll function. This was discovered due to the log added by Marcus some weeks ago.


marcus-oscarsson (approved):

marcus-oscarsson:
Thanks, well done. Feel free to rebase and merge

fabcor-maxiv:
This probably could use some squashing of some sort. I do not think this should be two commits.

PR #1434, opened 2025-10-27
abstract collect HWO: simplify loading of `undulators` property

elmjag:
No need to iterate over undulators property value, as it's already a list.

Use default value of empty list when fetching undulators property with `get_property()` method. This way we get rid of the scary looking traceback, for configurations that don't specify undulators property.


marcus-oscarsson:
Thanks !

fabcor-maxiv (approved):

PR #1435, opened 2025-10-28
Improve EPICSMotors and EPICSActuator classes

RafaelLyra8:
- Create a simplified EPICSMotor class with the standart EPICSMotors fields already configured (This simplifies the YAML configuration file of every motor, reducing from 9 parameters to be be configured to just 1)
- Fix _wait_actuator function that wasn't working on both EPICSMotors and EPICSActuator

### DIFF
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -16,7 +16,6 @@ class EPICSActuator(AbstractActuator):
         super().__init__(name)
         self.setpoint = None
         self._nominal_limits = (-1e4, 1e4)
-        self.default_timeout = 180
         if not self.unit:
             self.unit = 10**-3
 
@@ -53,7 +52,7 @@ class EPICSActuator(AbstractActuator):
 
     def set_value(self, value, timeout: float = 0):
         if not timeout:
-            timeout = self.default_timeout
+            timeout = None
         try:
             super().set_value(value, timeout)
         except ValueError:
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -1,11 +1,11 @@
 import threading
-import logging
 import time
 
 import gevent
 import numpy as np
 
 from mxcubecore.HardwareObjects.abstract.AbstractActuator import AbstractActuator
+
 
 class EPICSActuator(AbstractActuator):
     ACTUATOR_VAL = "val"  # setpoint
@@ -16,7 +16,6 @@ class EPICSActuator(AbstractActuator):
         self.setpoint = None
         self._nominal_limits = (-1e4, 1e4)
         self.default_timeout = 180
-        self.log = logging.getLogger("HWR")
         if not self.unit:
             self.unit = 10**-3
 
@@ -28,19 +27,19 @@ class EPICSActuator(AbstractActuator):
         readback = self.get_value()
         if not readback:
             return False
-        return not np.isclose(
-            readback, setpoint, rtol=self.unit, atol=self.unit
-        )
+        return not np.isclose(readback, setpoint, rtol=self.unit, atol=self.unit)
 
     def wait_ready(self, timeout):
         self._wait_task = threading.Event()
         try:
             with gevent.Timeout(timeout, exception=TimeoutError):
-                while self.hasnt_arrived(self.setpoint) and not self._wait_task.is_set():
+                while (
+                    self.hasnt_arrived(self.setpoint) and not self._wait_task.is_set()
+                ):
                     time.sleep(0.15)
         except TimeoutError:
             pvname = self.get_channel_object("rbv").command.pv_name
-            self.log.error(f"{pvname} motion has timed out.")
+            self.print_log(level="error", msg=f"{pvname} motion has timed out.")
         self.update_state(self.STATES.READY)
 
     def get_value(self):
@@ -54,7 +53,16 @@ class EPICSActuator(AbstractActuator):
     def set_value(self, value, timeout: float = 0):
         if not timeout:
             timeout = self.default_timeout
-        super().set_value(value, timeout)
+        try:
+            super().set_value(value, timeout)
+        except ValueError:
+            pvname = self.get_channel_object("rbv").command.pv_name
+            msg = f"{pvname} value {value} is outside motor limits"
+            msg += f": {self._nominal_limits}"
+            self.print_log(
+                level="error",
+                msg=msg,
+            )
 
     def abort(self):
         if self._wait_task is not None:
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -1,0 +1,80 @@
+import numpy as np
+import gevent
+from mxcubecore.HardwareObjects.abstract.AbstractActuator import AbstractActuator
+
+
+class EPICSActuator(AbstractActuator):
+    """EPICS actuator class"""
+
+    ACTUATOR_VAL = "val"  # setpoint
+    ACTUATOR_RBV = "rbv"  # readback
+
+    def __init__(self, name):
+        super(EPICSActuator, self).__init__(name)
+        self.__wait_actuator_task = None
+        self._nominal_limits = (-1e4, 1e4)
+
+    def init(self):
+        """Initialization method"""
+        super(EPICSActuator, self).init()
+        gevent.spawn(self._watch)
+        self.update_state(self.STATES.READY)
+        self.old_value = None
+        if not self.unit:
+            self.unit = 0
+
+    def _watch(self):
+        """Watch Actuator current value and update it on the UI."""
+        while True:
+            gevent.sleep(0.25)
+            if self._nominal_value != self.old_value:
+                self.old_value = self._nominal_value
+            self.update_value()
+
+    def hasnt_arrived(self, setpoint):
+        return not np.isclose(self.get_value(), setpoint, rtol=self.unit, atol=self.unit)
+
+    def _wait_actuator(self, setpoint):
+        """Wait actuator to be ready."""
+        while self.hasnt_arrived(setpoint):
+            gevent.sleep(0.25)
+        self.update_state(self.STATES.READY)
+
+    def get_value(self):
+        """Override AbstractActuator method."""
+        return self.get_channel_value(self.ACTUATOR_RBV)
+
+    def _set_value(self, value):
+        """Override AbstractActuator method."""
+        self.set_channel_value(self.ACTUATOR_VAL, value)
+
+    def set_value(self, value, timeout=0):
+        """Override AbstractActuator method."""
+        """ Set actuator to value.
+        Args:
+            value: target value
+            timeout (float): optional - timeout [s],
+                            If timeout == 0: return at once and do not wait
+                            (default);
+                            if timeout is None: wait forever.
+        Raises:
+            ValueError: Invalid value or attemp to set read only actuator.
+            RuntimeError: Timeout waiting for status ready  # From wait_ready
+        """
+        if self.read_only:
+            raise ValueError("Attempt to set value for read-only Actuator")
+        if self.validate_value(value):
+            self.update_state(self.STATES.BUSY)
+            self._set_value(value)
+            self.__wait_actuator_task = gevent.spawn(lambda: self._wait_actuator(value))
+        else:
+            raise ValueError(
+                "Invalid value %s; limits are %s" % (value, self.get_limits())
+            )
+
+    def abort(self):
+        """Imediately halt movement. By default self.stop = self.abort"""
+        if self.__wait_actuator_task is not None:
+            self.__wait_actuator_task.kill()
+        self.update_state(self.STATES.READY)
+
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -37,7 +37,7 @@ class EPICSActuator(AbstractActuator):
     def _wait_actuator(self, setpoint):
         """Wait actuator to be ready."""
         while self.hasnt_arrived(setpoint):
-            gevent.sleep(0.25)
+            gevent.sleep(0.15)
         self.update_state(self.STATES.READY)
 
     def get_value(self):
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -21,7 +21,7 @@ class EPICSActuator(AbstractActuator):
         self.update_state(self.STATES.READY)
         self.old_value = None
         if not self.unit:
-            self.unit = 0
+            self.unit = 10**-3
 
     def _watch(self):
         """Watch Actuator current value and update it on the UI."""
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -20,16 +20,9 @@ class EPICSActuator(AbstractActuator):
     def init(self):
         """Initialization method"""
         super(EPICSActuator, self).init()
-        gevent.spawn(self._watch)
         self.update_state(self.STATES.READY)
         if not self.unit:
             self.unit = 10**-3
-
-    def _watch(self):
-        """Watch Actuator current value and update it on the UI."""
-        while True:
-            gevent.sleep(0.25)
-            self.update_value()
 
     def hasnt_arrived(self, setpoint):
         return not np.isclose(
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -1,4 +1,5 @@
 import threading
+import logging
 import time
 
 import gevent
@@ -6,45 +7,41 @@ import numpy as np
 
 from mxcubecore.HardwareObjects.abstract.AbstractActuator import AbstractActuator
 
-
 class EPICSActuator(AbstractActuator):
     ACTUATOR_VAL = "val"  # setpoint
     ACTUATOR_RBV = "rbv"  # readback
 
     def __init__(self, name):
         super(EPICSActuator, self).__init__(name)
-        self._wait_task = None
         self.setpoint = None
         self._nominal_limits = (-1e4, 1e4)
         self.default_timeout = 180
+        self.log = logging.getLogger("HWR")
+        if not self.unit:
+            self.unit = 10**-3
 
     def init(self):
         self.update_state(self.STATES.READY)
         self.connect(self.get_channel_object("rbv"), "update", self.update_value)
-        if not self.unit:
-            self.unit = 10**-3
 
     def hasnt_arrived(self, setpoint):
+        readback = self.get_value()
+        if not readback:
+            return False
         return not np.isclose(
-            self.get_value(), setpoint, rtol=self.unit, atol=self.unit
+            readback, setpoint, rtol=self.unit, atol=self.unit
         )
-
-    def _wait_thread(self, setpoint, timeout):
-        try:
-            with gevent.Timeout(timeout, exception=TimeoutError):
-                while self.hasnt_arrived(setpoint) and not self._wait_task.is_set():
-                    time.sleep(0.15)
-        except TimeoutError:
-            pvname = self.get_channel_object("rbv").command.pv_name
-            self.log(f"{pvname} motion has timed out.")
-        self.update_state(self.STATES.READY)
 
     def wait_ready(self, timeout):
         self._wait_task = threading.Event()
-        thread = threading.Thread(
-            target=self._wait_thread, args=(self.setpoint, timeout)
-        )
-        thread.start()
+        try:
+            with gevent.Timeout(timeout, exception=TimeoutError):
+                while self.hasnt_arrived(self.setpoint) and not self._wait_task.is_set():
+                    time.sleep(0.15)
+        except TimeoutError:
+            pvname = self.get_channel_object("rbv").command.pv_name
+            self.log.error(f"{pvname} motion has timed out.")
+        self.update_state(self.STATES.READY)
 
     def get_value(self):
         return self.get_channel_value(self.ACTUATOR_RBV)
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -12,7 +12,7 @@ class EPICSActuator(AbstractActuator):
     ACTUATOR_RBV = "rbv"  # readback
 
     def __init__(self, name):
-        super(EPICSActuator, self).__init__(name)
+        super().__init__(name)
         self.setpoint = None
         self._nominal_limits = (-1e4, 1e4)
         self.default_timeout = 180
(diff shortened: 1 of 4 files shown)


mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def __init__(self, name):
+        super(EPICSActuator, self).__init__(name)
+        self.__wait_actuator_task = None
+        self._nominal_limits = (-1e4, 1e4)
+
+    def init(self):
+        """Initialization method"""
+        super(EPICSActuator, self).init()
+        gevent.spawn(self._watch)
```
marcus-oscarsson: Do you really need to perform this polling of the value like this cant it be done via the EPICS command/channels ?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,90 @@
...
+        self.update_state(self.STATES.READY)
+        self.old_value = None
+        if not self.unit:
+            self.unit = 10**-3
+
+    def _watch(self):
+        """Watch Actuator current value and update it on the UI."""
+        while True:
+            gevent.sleep(0.25)
```
marcus-oscarsson: You are "observing" `self._nominal_value` in the "background" but at the same time your the one setting `self._nominal_value`. I'm perhaps missing something but I don't think this would be needed ?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def get_value(self):
+        """Override AbstractActuator method."""
+        return self.get_channel_value(self.ACTUATOR_RBV)
+
+    def _set_value(self, value):
+        """Override AbstractActuator method."""
+        self.set_channel_value(self.ACTUATOR_VAL, value)
+
+    def set_value(self, value, timeout=0):
```
marcus-oscarsson: Are you sure that you need to overload `set_value`, its perhaps sufficient to overload, `wait_ready`.
If you look at:

   https://github.com/mxcube/mxcubecore/blob/bc5381a8ce804193c5cc925eee6b9cc51d284474/mxcubecore/HardwareObjects/abstract/AbstractActuator.py#L179C17-L179C41

You can see that it does basically the same as your `set_value`.

You could add a `self.update_state(self.STATES.BUSY)` in your `_set_value` and implement the `__wait_actuator_task` in `wait_ready`, or similair.

I would use a event mechanism instead of the spawn (if its available)

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
```diff
@@ -0,0 +1,345 @@
...
+        # Atualiza posição atual do feixe
+        self.beam_position[0], self.beam_position[1] = (
+            HWR.beamline.beam.get_beam_position_on_screen()
+        )
+        print(f"Posição atual do feixe na tela: {self.beam_position}")
+
+        # Armazena clique do usuário
+        self.last_centred_position[0] = x
+        self.last_centred_position[1] = y
+        print(f"Última posição clicada pelo usuário: x={x}, y={y}")
```
marcus-oscarsson: Be careful with using prints, these are not logged/saved anywhere. They are also in Spanish. I don't see it as an issue as its in your site specific code, but you may want to change it to English.

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
```diff
@@ -0,0 +1,345 @@
...
+        self.frontlight = None
+        self.frontlightswitch = None
+        self.phi = None
+        self.phiy = None
+        self.phiz = None
+        self.sampx = None
+        self.sampy = None
+
+    def init(self):
+        """
```
marcus-oscarsson, line 55: I think you can remove these old documentation placeholders :)

marcus-oscarsson:
Very nice, I'm a bit puzzled by how the the `EPICSActuator` handles updates, maybe there is a good reason for it that I don't see.

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
RafaelLyra8: This will be removed in a future version, I only focused on the EpicsMotor and EpicsActuator classes and left the old code for the LNLSDiffractometer that is being used. I'll update it in another PR.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
RafaelLyra8: I think using the EPICS command would really be better, I'll study how to implement it.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
RafaelLyra8: It's really not necessary, I saw that the update_value function in the AbstractActuator already checks if the PV value has changed or not. I'll remove it.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,90 @@
...
+    ACTUATOR_RBV = "rbv"  # readback
+
+    def __init__(self, name):
+        super(EPICSActuator, self).__init__(name)
+        self.__wait_actuator_task = None
+        self._nominal_limits = (-1e4, 1e4)
+
+    def init(self):
+        """Initialization method"""
+        super(EPICSActuator, self).init()
```
beteva: As we dropped the python2 support, you may want to use python3 only syntax

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
beteva: Not only unnecessary, but may induce errors, as update_value will not see the change of the value and not emit the signal.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,90 @@
...
+        )
+
+    def _wait_actuator(self, setpoint, timeout):
+        """Wait actuator to be ready."""
+        start = time.time()
+        while self.hasnt_arrived(setpoint):
+            gevent.sleep(0.15)
+            cur = time.time()
+            if (cur - start) > timeout:
+                raise TimeoutError
```
beteva: May I suggest again the use of pyton3 syntax:
with gevent.Timeout( timeout):
            while hasnt_arrived(setpoint):
                time.sleep(0.15)

RafaelLyra8:
The update are currently being handled with the watcher method (self._watch), but as you mentioned before, this would be better if it was handled by the Command.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
beteva: Agree with @marcus-oscarsson.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
```diff
@@ -0,0 +1,79 @@
...
+        self._instantiate_attributes()
+        self.get_limits()
+        self.get_velocity()
+        self.get_precision()
+        super().init()
+
+    def _wait_actuator(self, value):
+        """Override EPICSActuator method."""
+        while not self.done_movement() or self.hasnt_arrived(value):
+            time.sleep(0.25)
```
beteva: You've removed the timeout mechanism from the wait command, Does that mean that you always want to wait forever?

beteva (commented):
Nice job.
It is a bit unfortunate that you implement your diffractometer now, with the GenericDiffractometer. If you can wait for another few weeks, you may want to base it on the new AbstractDiffarctometer abstract class.

RafaelLyra8:
> Nice job. It is a bit unfortunate that you implement your diffractometer now, with the GenericDiffractometer. If you can wait for another few weeks, you may want to base it on the new AbstractDiffarctometer abstract class.

This is only an initial version, based on the current class that is currently being used, I will refactor its code in the next few weeks, and it shouldn't be a problem to wait for the new Abstract Diffractometer class as the new version of mxcube will take some months before going for production.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
RafaelLyra8: No, I will implement this. But, as of right now, the timeout isn't being set as I didn't really understand how can I configure the devices timeout, is it in the YAMLs?

beteva: There is no rule of how to do it. In general we assume that timeout of None is waiting forever, 0 means return immediately without waiting the execution to finish and a float value is in s to wait a reasonable amount of time. In general timeout is a parameter of the method. If you want, you can add default value in the configuration file of each device (the motors in this case), What I do in general is to set default value to None in the call of the method and when calling the method, set it to some value if needed. You can have a look in the ExporterMotor.py to see how the timeout is implemented for this kind of motors.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
rhfogh: set_value was written to be generic, calling _set_value for the actual setting. If you need something additional done I would even say to do the additional stuff and then call the AbstractActuator function. Better not duplicate that if not absolutely necessary.

marcus-oscarsson (approved):

marcus-oscarsson:
Great, nicely done. Could you please rebase and then you can merge ?

**edit**: For some reason the changes you made, also made it look like most comments where addressed but there still seems to be quite a few remaining right ?, Did you plan to look a those as well ?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,70 @@
...
+        self.update_state(self.STATES.READY)
+        self.connect(self.get_channel_object("rbv"), "update", self.update_value)
+
+    def hasnt_arrived(self, setpoint):
+        readback = self.get_value()
+        if not readback:
+            return False
+        return not np.isclose(readback, setpoint, rtol=self.unit, atol=self.unit)
+
+    def wait_ready(self, timeout):
```
rhfogh: The superclass function has the timeout parameter optional. You cannot make it mandatory in a subclass

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
```diff
@@ -0,0 +1,97 @@
...
+        super().init()
+
+    def wait_ready(self, timeout):
+        self._wait_task = threading.Event()
+        timeout = abs(self.get_value() - self.setpoint) / self.get_velocity()
+        timeout += 2 * self.get_acceleration()
+        # Timeout tolerance
+        timeout += 10
+        try:
+            with gevent.Timeout(timeout, exception=TimeoutError):
```
rhfogh, line 52: What would be the reason for not using the existing machinery, with ready_event?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
```diff
@@ -0,0 +1,97 @@
...
+                msg=f"{pvname} motion has timed out.",
+            )
+        self.update_state(self.STATES.READY)
+
+    def abort(self):
+        self.set_channel_value(self.MOTOR_STOP, 1)
+        super().abort()
+
+    def get_limits(self):
+        try:
```
rhfogh, line 71: Do the limits vary over time? If not, why not use the existing machinery?

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
```diff
@@ -0,0 +1,345 @@
...
+        Descript. :
+        """
+        # self.image_width = 100
+        # self.image_height = 100
+
+        GenericDiffractometer.init(self)
+        # Bzoom: 1.86 um/pixel (or 0.00186 mm/pixel) at minimum zoom
+        self.x_calib = 0.00186
```
rhfogh: This should be in a configuration file rather than hardcoded, on general principles. Or if you really, really, have to hardcode it, at least put it in a named constant at the top of the file.

rhfogh (changes requested):
See comments to code

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,70 @@
...
+
+from mxcubecore.HardwareObjects.abstract.AbstractActuator import AbstractActuator
+
+
+class EPICSActuator(AbstractActuator):
+    ACTUATOR_VAL = "val"  # setpoint
+    ACTUATOR_RBV = "rbv"  # readback
+
+    def __init__(self, name):
+        super(EPICSActuator, self).__init__(name)
```
beteva:
```suggestion
        super().__init__(name)
```

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,70 @@
...
+        self.setpoint = None
+        self._nominal_limits = (-1e4, 1e4)
+        self.default_timeout = 180
+        if not self.unit:
+            self.unit = 10**-3
+
+    def init(self):
```
beteva, line 22:
```suggestion
    def init(self):
        super().init()
```
Is it for purpose you are initialising the parent class?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
beteva:
```suggestion
    def wait_ready(self, timeout=None):
```

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -0,0 +1,70 @@
...
+        return self.get_channel_value(self.ACTUATOR_RBV)
+
+    def _set_value(self, value):
+        self.setpoint = value
+        self.update_state(self.STATES.BUSY)
+        self.set_channel_value(self.ACTUATOR_VAL, value)
+
+    def set_value(self, value, timeout: float = 0):
+        if not timeout:
+            timeout = self.default_timeout
```
beteva, line 60: You are excluding the default value (timeout=0) - is it deliberate? This also means that you cannot set timeout to None, which is the default behaviour for the rest of the systems.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def init(self):
+        self._motor_channels = {}
+        self._instantiate_attributes()
+        self.get_limits()
+        self.get_velocity()
+        self.get_precision()
+        super().init()
+
+    def wait_ready(self, timeout):
```
beteva:
```suggestion
    def wait_ready(self, timeout=None):
```

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
beteva: I believe this is the concrete implementation for getting the limits.

beteva (commented):
I presume the diffractometer is a temporary solution, so would not comment on it.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
beteva: I do not think it is resolved.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
beteva: The code did not change, as it has been requested by @rhfogh.

beteva (changes requested):
Could you, please, apply the requested changes and not mark them as resolved without being changed. Thank you

pedrobenetton:
> Could you, please, apply the requested changes and not mark them as resolved without being changed. Thank you

Hi, I didn't mark them as resolved. Currently I updated this branch with the most recent commits from mxcube:develop and now I'm looking at everything that has been commented.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
pedrobenetton: We did this because of the `AbstractActuator` logic at the `set_value` function:
If there is no timeout, `wait_ready` is never called, and then the `update_state` never happens, and the motor never returns to the `STATES.READY` state. After a few discussions it was clear that 180 seconds was more than enough for anything in the beamline for a `default_timeout`.

beteva: > If there is no timeout, `wait_ready` is never called

This is only true if timeout=0. If timeout=None, it waits forever. In your code you exclude both 0 and None. That means:
if no 0 - you do not have the possibility to run two actions together.
if no None - you have to be sure that your default value is big enough for every set_value call or you specify appropriate timeout every time you do set_value.
This is up to you.

beteva:
Welcome to the wonderful world of rebasing 😄
I believe you've made a slight mistake, as there are MAXIV commits in your branch.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
pedrobenetton: I tried to implement ready_event with our `self.done_movement()` and `self.hasnt_arrived()` functions in both EPICSActuator and EPICSMotor. I think it is being properly used now.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
pedrobenetton: Since the function call was not receiving anything from the `return`, I changed it from `get_limits()` to `set_limits()` and I think it's being used correctly now.

beteva: Well, set and get do not have the same usage. set_limits is needed if you want to change the limits and not use the initially configured ones. You do not want to set limits if you only want to read them. I think your initial implementation of get_limits was correct. You were probably missing the concrete reading from the epics configuration if you did not see anything returned.
You still need get_limits method with reading of the epics configuration. Otherwise you will get (None, None), as this is the default value of self._nominal_limits if you do get_limits before set_limits.
You may want to check how get_limits is implemented in BlissMotor or ExporterMotor

pedrobenetton: I did read BlissMotor and ExporterMotor and it seems that I should just revert this last commit. What I meant about the `return` was that `get_limits()` gets called at `init()` and doesn't do anything with what is returned, the returned values are correct.

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
pedrobenetton: I have moved the values to the configuration file.

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
```diff
@@ -0,0 +1,345 @@
...
+        # Atualiza posição atual do feixe
+        self.beam_position[0], self.beam_position[1] = (
+            HWR.beamline.beam.get_beam_position_on_screen()
+        )
+        print(f"Posição atual do feixe na tela: {self.beam_position}")
+
+        # Armazena clique do usuário
+        self.last_centred_position[0] = x
+        self.last_centred_position[1] = y
+        print(f"Última posição clicada pelo usuário: x={x}, y={y}")
```
beteva: This is not resolved. Please, see the @marcus-oscarsson comment.

beteva (approved):

mxcubecore/HardwareObjects/LNLS/LNLSDiffractometer.py
RafaelLyra8: This was an old version of the LNLSDiffractometer that will be refactored completely in another PR, I didn't include it in this one because it would become really big.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
rhfogh: IMHO this is bad - you are breaking the behaviour of the superclass. It says in the relevant documentation that if timeout=0 do not wait, and if timeout==None wait forever. Anyone calling that function is entitled to rely on that behaviour. If it is the case that the state  never returns to READY if the timeout is 0 then that is a bug ino the rest of the code, and that should be fixed, not worked around. Anyway, I wonder. Surely the actuator will return to ready when the motor has finished moving and sends the relevant signal??

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
rhfogh: I do not really understand. The existing wait_ready function has only two lines:
```
        with Timeout(timeout, RuntimeError("Timeout waiting for status ready")):
            self._ready_event.wait(timeout=timeout)
```
The only thing you need to  do is to ensure that self._ready_event is set and cleared appropriately by the code elsewhere, you do not need wait_ready to do anything. If you have chosen to bypass that and re-implement the event waiting functionality afresh, what is the reason for that? What do you need that cannot be done in the existing framework, or why is it not possible for the rest of the code to set and clear that event appropriately?

beteva: Hi @rhfogh, I do not know the requirements of the epics motors, but it might be that you should alway give some (timeout=0 might not be allowed). @RafaelLyra8, could you, please, confirm this.

RafaelLyra8: I decided to set the timeout automatically in order to simplify the configuration file and because the EPICS Motor Record, which is the standard EPICS abstraction that I want this class to encompass, contains the setpoint, readback and velocity parameters with which I can discover the motion duration and I added some tolerance time of motion.
I also reimplemented the wait_ready class because the EPICS Motor Record provide the information if the motion has finished, that's why I implemented the done_movement method, and in order to encompass cases that this logic might be faulty I added another validation which verify if the readback is the same as the setpoint, considering some tolerance.
We could leave the class using the standard implementation of timeout but because of the motor record standards, I wanted to create this class in order to simplify the configuration. This simplification made motor XML that used to have 15 lines without considering timeouts and limits to YAMLs with 7 with those considerations, and these considerations being standardized between all motors in order to simplify maintenance.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
RafaelLyra8: I did this because I always wanted to call the wait_ready method of the EPICSMotor, because the timeout is dynamically set there, and the only way I could do that, without changing the Abstract classes, was setting the timeout. Another way that could be done would be adding a new parameter to the set_value in the AbstractActuator with a default value that allows sub classes to set the timeout dynamically.

rhfogh: OK, it sounds like this solution does what is needed, and we should certainly not make everything more complicated just to be standardising. I'll stop being obstructive ;-)

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSMotor.py
rhfogh: Good. Again, it is better with a simpler solution than standardising and making it all more difficult. Thanks for your patient explanations.

rhfogh (approved):
Lots of good work here. Thanks for your long answers to my comments.

PR #1439, opened 2025-10-30
Session HWO: drop unused `email_extension` attribute

elmjag:
The Session `email_extension` attribute is not used anywhere, remove it.


mockoocy (approved):

marcus-oscarsson:
Not sure whats going on with the doc build, it takes forever.

PR #1442, opened 2025-10-31
1437 Remove hard-coded endpoints

bodinm:
- use `pyicat-plus` library instead of hard-coded endpoints
- upgrade pyicat-plus version
- remove `ICATSession`, it is defined as a dict in pyicat-plus

=> this changes require an upgrade of `pyicat-plus` library to 0.11.0

### DIFF
--- a/mxcubecore/HardwareObjects/ICATLIMS.py
+++ b/mxcubecore/HardwareObjects/ICATLIMS.py
@@ -1070,10 +1070,8 @@ class ICATLIMS(AbstractLims):
             Optional[SampleInformation]: Returns a SampleInformation object or None.
         """
         try:
-            result = self.icatClient.get_sample_files_information_by(
-                sample_id=sample_id
-            )
-            SampleInformation.parse_obj(result)
+            result = self.icatClient.get_sample_files_information_by(sample_id)
+            return SampleInformation.parse_obj(result)
         except requests.exceptions.HTTPError as e:
             if e.response.status_code == 404:
                 logger.info("Sample %s not found (404)", sample_id)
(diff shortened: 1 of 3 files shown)


bodinm:
tested on the simulator

mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -60,7 +58,7 @@ def get_lims_name(self) -> List[Lims]:
         ]
 
     def _create_icat_session(self, user_name: str, password: str):
-        self.icat_session: ICATSession = self.icatClient.do_log_in(password)
```
marcus-oscarsson, line 63: @bodinm, @antolinos are you sure about this, its perhaps "safer" with the ICATSession model ?

marcus-oscarsson:
Looks more or less ok, just one question about the `ICATSession` model ?

mxcubecore/HardwareObjects/ICATLIMS.py
bodinm: https://gitlab.esrf.fr/icat/pyicat-plus/-/blob/main/src/pyicat_plus/client/interface.py?ref_type=heads#L346

`do_log_in` returns a `dict`

- `Session` has been deprecated (and is a `dict`)
https://gitlab.esrf.fr/icat/pyicat-plus/-/blob/main/src/pyicat_plus/client/models/session.py?ref_type=heads

So I think it's safer to use `dict`, as `Session` will be removed from pyicat-plus in a future release

antolinos: The reason the endpoints were hardcoded is that I believe we should use only pyicat-plus for data ingestion, e.g., `store_dataset,` and rely on a pure ICAT+ client written in Python with models for LIMS communication.

The main rationale behind this is that while some facilities can use ICAT, they most likely cannot use pyicat-plus because it's highly tailored to our data ingestion process.

This isn't urgent, but it would be great to discuss it in the coming months. We could even prepare for a major clean-up.

marcus-oscarsson (approved):

PR #1444, opened 2025-11-04
beamline actions improvement

epanepucci:
beamline actions improvement

- readability
  - shorten logging lines
  - refactor long values into variables

- feature: disabled directive
  - at times, we may want to temporarily disable and action so a `disabled` directive has been added
  - the directive does not need to be present
  - the `disabled` directive is *false* by default so it does not modify existing behavior

- feature: read a pure `yaml` configuration file
  - allow for the configuration file to be 100% yaml

Example of a pure YAML config file:
```yaml
configuration:
  commands:
    - type: controller
      name: Open Safety Shutter
command:
HardwareObjects.MAXIV.BioMAX.beamline_actions.OpenSafetyShutter
      disabled: False # new feature

    - type: controller
      name: Close Safety Shutter
command:
HardwareObjects.MAXIV.BioMAX.beamline_actions.CloseSafetyShutter
      disabled: False

    - type: controller
      name: Prepare Open Hutch
command:
HardwareObjects.MAXIV.BioMAX.beamline_actions.PrepareOpenHutch
      disabled: False
```

### DIFF
--- a/mxcubecore/HardwareObjects/BeamlineActions.py
+++ b/mxcubecore/HardwareObjects/BeamlineActions.py
@@ -207,8 +207,9 @@ class BeamlineActions(HardwareObject):
                     _cls = self._get_command_object_class(command["command"])
                 except:
                     self.log.exception(f"failed to load annotated action {action}")
-                fname = camel_to_snake(_cls.__name__)
-                _cls_inst = _cls(self, fname.replace("_", " ").title(), fname)
+                else:
+                    fname = camel_to_snake(_cls.__name__)
+                    _cls_inst = _cls(self, fname.replace("_", " ").title(), fname)
 
                 self._annotated_command_dict[fname] = _cls_inst
                 setattr(self, attrname, getattr(_cls_inst, fname))
--- a/mxcubecore/HardwareObjects/BeamlineActions.py
+++ b/mxcubecore/HardwareObjects/BeamlineActions.py
@@ -205,7 +205,7 @@ class BeamlineActions(HardwareObject):
             if command["type"] == "annotated":
                 try:
                     _cls = self._get_command_object_class(command["command"])
-                except:
+                except Exception:
                     self.log.exception(f"failed to load annotated action {action}")
                 else:
                     fname = camel_to_snake(_cls.__name__)
--- a/mxcubecore/HardwareObjects/BeamlineActions.py
+++ b/mxcubecore/HardwareObjects/BeamlineActions.py
@@ -19,6 +19,7 @@ from mxcubecore.utils.conversion import camel_to_snake
 class ControllerCommand(CommandObject):
     def __init__(self, name, cmd=None, username=None, klass=None):
         CommandObject.__init__(self, name, username)
+        self.log = logging.getLogger("HWR")
 
         if not cmd:
             self._cmd = klass()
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -185,23 +186,27 @@ def _get_command_object_class(self, path_str):
...
                 continue
 
             if command["type"] == "annotated":
-                _cls = self._get_command_object_class(command["command"])
+                try:
+                    _cls = self._get_command_object_class(command["command"])
+                except:
+                    self.log.exception(f"failed to load annotated action {action}")
                 fname = camel_to_snake(_cls.__name__)
                 _cls_inst = _cls(self, fname.replace("_", " ").title(), fname)
```
dominikatrojanowska: if `_cls = self._get_command_object_class(command["command"])` would fail, this line will fail again as `_cls` won't be defined 

```suggestion
                else:
                    fname = camel_to_snake(_cls.__name__)
                    _cls_inst = _cls(self, fname.replace("_", " ").title(), fname)
```

epanepucci: nice catch @dominikatrojanowska I'll merge this tomorrow as soon as I test it.

mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -19,6 +19,7 @@
 class ControllerCommand(CommandObject):
     def __init__(self, name, cmd=None, username=None, klass=None):
         CommandObject.__init__(self, name, username)
+        self.log = logging.getLogger("HWR")
```
beteva, line 22: This is not needed -see [Logging]( https://github.com/mxcube/mxcubecore/blob/develop/CONTRIBUTING.md#logging)

beteva:
Hi @epanepucci, nice idea to have the disable feature. Thanks for the cleanup as well.

mxcubecore/HardwareObjects/BeamlineActions.py
fabcor-maxiv: Oh? I thought this is needed here because it is not a Hardware Object. We need to inherit from `HardwareObjectMixin`, right? Is it the case here?

https://github.com/mxcube/mxcubecore/blob/8a41fca7cb197999d46030019ed2dd5a62dd36f0/mxcubecore/BaseHardwareObjects.py#L645-L646

beteva:
I would like to know how the disabled option look on the GUI - is the disabled action greyed out or anything else? The GUI shows what is in the command_list. Being present in the command_list, can it be executed by inadvertence?

mxcubecore/HardwareObjects/BeamlineActions.py
beteva: You are right @fabcor-maxiv, CommandObject does not inherit from HardwareObjects.

epanepucci: like @fabcor-maxiv mentioned, I had to add it for the current code-base.

epanepucci:
> I would like to know how the disabled option look on the GUI - is the disabled action greyed out or anything else? The GUI shows what is in the command_list. Being present in the command_list, can it be executed by inadvertence?

the **disabled** action is completely ignored so it never shows up in the GUI.

mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -185,25 +186,30 @@ def _get_command_object_class(self, path_str):
...
                 self.log.warning(msg)
                 continue
 
             if command["type"] == "annotated":
-                _cls = self._get_command_object_class(command["command"])
-                fname = camel_to_snake(_cls.__name__)
-                _cls_inst = _cls(self, fname.replace("_", " ").title(), fname)
+                try:
+                    _cls = self._get_command_object_class(command["command"])
+                except:
```
epanepucci:
```suggestion
                except Exception:
```

epanepucci (commented):
complying with linter

dominikatrojanowska (dismissed):

mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -185,25 +186,30 @@ def _get_command_object_class(self, path_str):
         return _cls
 
     def init(self):
```
epanepucci:
```suggestion
    def init(self): # noqa: C901
```

mockoocy:
```suggestion
    def init(self): # noqa: C901
```
Ruff complains that this function is too complicated ( I guess it counts number of control statements or something, I am not sure). 

You can either silence it (as in suggestion) or split code in different methods, whatever you think would be better.
I imagine silencing this rule is good enough, since it follows the same kind of flow (if -> try/except/else, elif->try/except/else and so on) and is not really confusing.

mxcubecore/HardwareObjects/BeamlineActions.py
```diff
@@ -184,26 +185,31 @@ def _get_command_object_class(self, path_str):
 
         return _cls
 
-    def init(self):
-        command_list = ast.literal_eval(
-            self.get_property("commands").strip().replace("\n", "")
-        )
+    def init(self): # noqa: C901
```
epanepucci:
```suggestion
    def init(self):  # noqa: C901
```

marcus-oscarsson:
Thanks @epanepucci !

marcus-oscarsson (approved):

PR #1445, opened 2025-11-04
ESS: Add NICOS classes (for connection, actuator and motor)

laispc:
Hi all,

Sharing the site-specific classes used at ESS for controlling devices through NICOS framework (see: gitlab.com/nicos-controls). NICOS is the default control layer for the experiments at ESS.

The covered devices (for now) are actuators and motors (with the classes NICOSActuator and NICOSMotor). A common connection class (NICOSConnection) is used by them to communicate with the instrument/beamline NICOS server.

Tested with: mxcubeweb==v4.231.0, mxcubecore==v1.188.0

Feel free to send comments/feedback. Thank you!

### DIFF
--- a/mxcubecore/HardwareObjects/ESS/NICOSActuator.py
+++ b/mxcubecore/HardwareObjects/ESS/NICOSActuator.py
@@ -15,38 +15,38 @@ Example of config file:
 </object>
 """
 
-from gevent import monkey
-monkey.patch_all()
-
 import time
+
 import gevent
+from gevent import monkey
 
 from mxcubecore.HardwareObjects.abstract import AbstractActuator
 
 from .nicos_connection import connect_to_nicos
 
+monkey.patch_all()
+
 
 class NICOSActuator(AbstractActuator.AbstractActuator):
     """NICOS Actuator class
-    
+
     Controls a NICOS EpicsAnalogMoveable device."""
 
     def __init__(self, name):
         super().__init__(name)
         self.__wait_actuator_task = None
-        self._nominal_limits = (-1E4, 1E4)
+        self._nominal_limits = (-1e4, 1e4)
         self.last_target_value = None
         self.ERROR_READBACK = 0
 
     def init(self):
-        """ Initialization method """
+        """Initialization method"""
         super(NICOSActuator, self).init()
 
-        host = self.get_property("host") # Create NICOS connection using the config
+        host = self.get_property("host")  # Create NICOS connection using the config
         port = self.get_property("port")
         user = self.get_property("user")
-        pw = self.get_property("password") # TODO: Improve this to be safer.
-        self.nicos_cli = connect_to_nicos(host, port, user, pw)
+        pw = self.get_property("password")  # Improve this to be safer.
         self.device_name = self.get_property("device_name")
         self.nicos_cli = connect_to_nicos(host, port, user, pw)
 
@@ -64,7 +64,7 @@ class NICOSActuator(AbstractActuator.AbstractActuator):
         self.update_state(self.STATES.READY)
 
     def _watch(self):
-        """ Watch actuator current value and update it on the UI."""
+        """Watch actuator current value and update it on the UI."""
         while True:
             time.sleep(0.3)
             self.update_value()
@@ -80,22 +80,22 @@ class NICOSActuator(AbstractActuator.AbstractActuator):
     def _wait_actuator(self):
         """Wait actuator to be at target."""
         self.MOVING = 1
-        while (not self.done_movement()):
+        while not self.done_movement():
             time.sleep(0.3)
         self.update_specific_state(None)
         self.MOVING = 0
 
     def get_value(self):
-        """ Override AbstractActuator method."""
+        """Override AbstractActuator method."""
         readback_val = self.nicos_cli.get_dev_param_value(self.device_name)
         if readback_val is None:
             self.ERROR_READBACK = 1
             return 0
         self.ERROR_READBACK = 0
         return readback_val
-    
+
     def _set_value(self, value):
-        """ Override AbstractActuator method."""
+        """Override AbstractActuator method."""
         self.last_target_value = value
         self.update_state(self.STATES.BUSY)
 
@@ -103,28 +103,28 @@ class NICOSActuator(AbstractActuator.AbstractActuator):
         self.nicos_cli.process_command(line)
 
         self.__wait_actuator_task = gevent.spawn(self._wait_actuator)
-        
+
     def abort(self):
         """Override HardwareObject method."""
         super().abort()
         line = "stop('{}')".format(self.device_name)
-        ret = self.nicos_cli.process_command(line)
+        self.nicos_cli.process_command(line)
         self.MOVING = 0
         self.reset()  # Clean state on NICOS
 
         if self.__wait_actuator_task is not None:
             self.__wait_actuator_task.kill()
         self.update_state(self.STATES.READY)
-    
+
     def done_movement(self):
-        """ Return whether actuator is at target position or not."""
+        """Return whether actuator is at target position or not."""
         if self.get_value() == self.last_target_value:
             self.reset()
             return True
         return False
 
     def reset(self):
-        """Reset NICOS device. This can be useful to be sure the device is in 
+        """Reset NICOS device. This can be useful to be sure the device is in
         a health state on the NICOS side."""
         line = "reset('{}')".format(self.device_name)
         self.nicos_cli.process_command(line)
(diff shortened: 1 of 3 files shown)


mxcubecore/HardwareObjects/ESS/NICOSActuator.py
```diff
@@ -0,0 +1,100 @@
...
+        user = self.get_property("user")
+        pw = self.get_property("password") # TODO: Improve this to be safer.
+        self.nicos_cli = connect_to_nicos(host, port, user, pw)
+        self.device_name = self.get_property("device_name")
+        self.update_state(self.STATES.READY)
+
+    def _wait_actuator(self):
+        """ Wait actuator to be ready."""
+        time.sleep(0.3)
+        self.update_state(self.STATES.READY)
```
beteva: You actually do not check the state, only set it to READY. Is this for purpose?

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
```diff
@@ -0,0 +1,100 @@
...
+                    self._set_value(value)
+                    new_value = self._wait_actuator(value)
+            else:
+                self._set_value(value)
+                self.__wait_actuator_task = gevent.spawn(self._wait_actuator)
+        else:
+            raise ValueError("Invalid value %s; limits are %s"
+                             % (value, self.get_limits())
+                             )
+
```
beteva: Maybe just overloading the _set_value will save you from duplicating the code. If you need, you can add the update_state call in your _set_value.
 I also see that your messages contain Motor.
Also in AbstractActuator set_value we have a lock mechanism, to be sure about the execution order.

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
```diff
@@ -0,0 +1,100 @@
...
+from .nicos_connection import connect_to_nicos
+
+
+class NICOSActuator(AbstractActuator.AbstractActuator):
+    """NICOS actuator class
+    
+    This class is based on LNLS.EPICSActuator."""
+
+    def __init__(self, name):
+        super(NICOSActuator, self).__init__(name)
```
beteva: Maybe some python3 syntax would be easier to read

`super().__init__(name)`

beteva:
Hi @laispc, nice to have a new type of protocol added to MXCuBE.

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
```diff
@@ -0,0 +1,100 @@
...
+        if self.read_only:
+            raise ValueError("Attempt to set value for read-only Actuator")
+        if self.validate_value(value):
+            self.update_state(self.STATES.BUSY)
+            if timeout or timeout is None:
+                with gevent.Timeout(
+                    timeout, RuntimeError("Motor %s timed out" % self.username)
+                ):
+                    self._set_value(value)
+                    new_value = self._wait_actuator(value)
```
beteva: Maybe I am missing something , but I do not see how this timeout works - the _wait actuator only sleeps for 0.3s and there is no other test for whatever value or state.

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
```diff
@@ -0,0 +1,100 @@
...
+Example of config file:
+
+<object class="ESS.NICOSActuator">
+    <host>my-nicos-server-hostname</host>
+    <port>1234</port>
+    <user>myuser</user>
+    <password>mypassword</password>
+    <device_name>my_nicos_device</device_name>
+    <username>Kappa</username>
+    <motor_name>Kappa</motor_name>
```
beteva: This is probably a legacy. I do not see you using the actuator_name, though

mxcubecore/HardwareObjects/ESS/NICOSMotor.py
```diff
@@ -0,0 +1,84 @@
...
+        while (not self.done_movement()):
+            time.sleep(0.3)
+        self.update_specific_state(None)
+        self.moving = 0
+    
+    def reset(self):
+        """Reset NICOS device. This can be useful to be sure the device is in 
+        a health state."""
+        line = "reset('{}')".format(self.device_name)
+        ret = self.nicos_cli.process_command(line)
```
beteva: Maybe you do not nead the ret value

beteva (commented):
Nice start. Waiting for the rest to come!

beteva:
To prevent the linting errors, did you try:
poetry run pre-commit run --all-files

marcus-oscarsson:
Well done !

@beteva already pointed at some interesting bits as well :)

I think the logic in [nicos_connection.py](https://github.com/mxcube/mxcubecore/pull/1445/files#diff-6e3f0e3cacfbb3ad3d69f4124cc3980dfa9f5f527ebfa11841911f6d264908bf) could be added as commands and channels, see https://github.com/mxcube/mxcubecore/tree/develop/mxcubecore/Command.

I think this is fine if it works for you, and your probably alone to use NICOS ? I guess it would be a cleaner solution long term but I understand that its more complicated.

laispc:
Hi all, thanks a lot for the review, @beteva and @marcus-oscarsson !
I could only come back to this now. I will send the suggested changes soon, so keep the PR open for a bit more, please. Thanks for the patience. :)

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
laispc: Fixed. This part was a bit undefined indeed because I was using/testing more with NICOSMotor. I moved everything to NICOSActuator, so now this method looks better. :)

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
laispc: All fixed. Thanks for the advice! Overloading _set_value() is much better and enough. I could remove the duplicate set_value() after doing this.

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
laispc: Good catch. Fixed!

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
laispc: Fixed. This part was working but I agree it was very confusing. It could be removed. :)

mxcubecore/HardwareObjects/ESS/NICOSActuator.py
laispc: Nice! Thanks for highlighting this. I could remove these from the config files indeed, and everything still works. Fixed!

mxcubecore/HardwareObjects/ESS/NICOSMotor.py
laispc: I agree. Removed. Thanks! :D

laispc:
@beteva thanks a lot for all the suggestions! I agreed and implemented then all. I've just sent the commits, and the pipeline finally looks happy (thanks againfor the patience). :sparkles: 

Also, by following the suggestions, I decided to merge NICOSMotors methods into NICOSActuator for simplification, as the motors we are using/testing ATM behave basically like actuators. I had to preserve the NICOSMotor class just for mxcube web UI to work (it ignores the motors if I configure then as NICOSActuators).
Another alternative would be to have just NICOSMotors, but the above approach looked more compatible with what we have configured in NICOS so far. Let me know what you think.

@marcus-oscarsson yes, for now we are the only ones using NICOS for MXCuBE.
Sounds like a good idea having a Command class for NICOS in a future version/MR, I will try this.
But first I want to try more simple ways to talk to NICOS server (or at least simplify a bit more the current one). I am still testing some alternatives. After this, I will try the Command class. :D

laispc:
Probably it's better to squash the commits into a single one after the review. I didn't find the option here, so please let me know if I should re-send it all squashed.

marcus-oscarsson (approved):

marcus-oscarsson:
Thanks, sorry for taking so long to approve !

marcus-oscarsson:
You can re-base and merge :), my apologies again for the delay !

laispc:
Thanks, @marcus-oscarsson . Don't worry, there was absolutely no delay, actually it was really fast. \O/
Rebase done.
I am not authorized to merge.

beteva:
Dear @laispc, could you, please rebase your branch and I will merge it afterwards.

marcus-oscarsson:
@laispc: You can rebase and merge :)

laispc:
@beteva rebase done. Feel free to merge.
BTW, this is what I see at the moment:
Screenshot from 2026-04-14 11-27-19

marcus-oscarsson:
Oh, I see you should be able to merge

marcus-oscarsson:
Ok, I think it should be alright for next time :)

PR #1447, opened 2025-11-05
Sync processing plan: check cache

bodinm:


mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -353,14 +353,18 @@ def __get_or_download_plan_resources(
...
         cache_key = (sample_sheet_id, protein_acronym)
-        if cache_key in self._downloads_cache:
-            logger.debug(f"Reusing cached downloads for {cache_key}")
-            return self._downloads_cache[cache_key]
-
         sample_information = self.__get_sample_information_by(sample_sheet_id)
         if not sample_information:
             return []
 
+        cached = self._downloads_cache.get(cache_key)
```
marcus-oscarsson, line 360: Just a question, is it really necessary to cash these, are they very big ?

marcus-oscarsson (approved):

mxcubecore/HardwareObjects/ICATLIMS.py
bodinm: PDB files are relatively small (a few KB), but HKL files can be several MB. As the download happens  when we synchronize the samples from DRAC, re-downloading these files each time slows down the process.

marcus-oscarsson: Ok, I see. Cash handling usually comes with some caveats so sometimes its easier to just avoid it ;)

bodinm: Sure, if caching gives us trouble later, we can just remove it.

PR #1455, opened 2025-11-27
Fields modeled as str converting to str before passing to `Session` and `Porposal` models

marcus-oscarsson:
Fields modeled as str converting to str before passing to `Session` and `Porposal` models


marcus-oscarsson:
@antolinos, perhaps due to the switch to Pydantic v2 models

antolinos:
> @antolinos, perhaps due to the switch to Pydantic v2 models

Indeed, it is because of v2

[Validator behavior changes](https://docs.pydantic.dev/latest/migration/#typeerror-is-no-longer-converted-to-validationerror-in-validators)

PR #1456, opened 2025-11-27
yaml purification

epanepucci:
- allows reading yaml for attributes formerly being only eval or ast.literal_eval
- use default_value in get_property() calls
- replace some eval by ast.literal_eval
- this commit should not modify the existing behaviour

### DIFF
--- a/mxcubecore/HardwareObjects/GenericDiffractometer.py
+++ b/mxcubecore/HardwareObjects/GenericDiffractometer.py
@@ -21,6 +21,7 @@
 GenericDiffractometer
 """
 
+import ast
 import copy
 import enum
 import json
@@ -355,12 +356,9 @@ class GenericDiffractometer(HardwareObject):
         self.back_light_switch = self.get_object_by_role("backlightswitch")
 
         # Channels -----------------------------------------------------------
-        ss0 = self.get_property("used_channels")
-        if ss0:
-            try:
-                self.used_channels_list = eval(ss0)
-            except Exception:
-                self.log.exception("")
+        self.used_channels_list = self.get_property("used_channels", [])
+        if isinstance(self.used_channels_list, str):
+            self.used_channels_list = ast.literal_eval(self.used_channels_list)
 
         for channel_name in self.used_channels_list:
             self.channel_dict[channel_name] = self.get_channel_object(channel_name)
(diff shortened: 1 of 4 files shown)


mxcubecore/HardwareObjects/GenericDiffractometer.py
```diff
@@ -482,33 +481,32 @@ def init(self):
             self.log.exception("")
 
         # Other parameters ---------------------------------------------------
-        try:
-            self.zoom_centre = eval(self.get_property("zoom_centre"))
-        except Exception:
-            self.zoom_centre = {"x": 0, "y": 0}
-            self.log.warning("Diffractometer: " + "zoom centre not configured")
+        self.zoom_centre = self.get_property("zoom_centre", {"x": 0, "y": 0})
```
rhfogh: Here and in the following I'd get the property, set the default (and warn) if it is None, and set the value otherwise. In principle you could be configuring this to  {"x": 0, "y": 0} (in a mock, for instance) and you do not want to warn if it is actually configured.

mxcubecore/HardwareObjects/GenericDiffractometer.py
```diff
@@ -482,33 +481,32 @@ def init(self):
...
 
         self.reversing_rotation = self.get_property("reversing_rotation")
-        try:
-            # grid_direction describes how a grid is collected
-            # 'fast' is collection direction and 'slow' describes
-            # move to the next collection line
-            self.grid_direction = eval(self.get_property("grid_direction"))
-        except Exception:
-            self.grid_direction = {"fast": (0, 1), "slow": (1, 0), "omega_ref": 0}
+        # grid_direction describes how a grid is collected
```
rhfogh, line 503: See comment for zoom_centre

rhfogh (changes requested):
Very good, just taht one point.

mxcubecore/HardwareObjects/GenericDiffractometer.py
epanepucci: thanks! did not think of that use case.

mxcubecore/HardwareObjects/GenericDiffractometer.py
```diff
@@ -482,33 +481,35 @@ def init(self):
             self.log.exception("")
 
         # Other parameters ---------------------------------------------------
-        try:
-            self.zoom_centre = eval(self.get_property("zoom_centre"))
-        except Exception:
+        self.zoom_centre = self.get_property("zoom_centre")
```
rhfogh, line 484: Good change. Still, do you need to do 'eval' if the property is a string, or is there no need?

mxcubecore/HardwareObjects/GenericDiffractometer.py
rhfogh: Do you need to do 'eval' if the property is a string, or is there no need?

epanepucci: thanks! I added more comprehensive checks now. but I think that's eno9ugh until pydantic comes around. let me know otherwise.

rhfogh (approved):
That is *definitely* enough tests ;-) . Please feel free to merge.

marcus-oscarsson:
:), thanks. Just as a side comment, much of this will be done by Pydantic once we get that into place as well.

marcus-oscarsson:
You can merge if you are done of course :)

epanepucci:
> You can merge if you are done of course :)

I'm done with these updates but I don't have privileges to merge.
When do you think the pydantic validation will be implemented?

rhfogh:
I'll merge it, then

marcus-oscarsson:
> When do you think the pydantic validation will be implemented?

I hope soon, not sure when but I can try to draft something during our winter shutdown (if I get the time)

PR #1458, opened 2025-12-02
Tango Poller: handle numpy array shape changes

elmjag:
Add support for polling tango attribute values, which are numpy arrays that change shapes. 

Currently if, for example, a tango attribute changes value from `[1, 2]` to `[]`, poller will fail. It will fail, because numpy does not support `==` operation on arrays of different 'shapes'.

Adds code to explicitly check if shapes are different between old and new values.

### DIFF
--- a/mxcubecore/Poller.py
+++ b/mxcubecore/Poller.py
@@ -67,6 +67,34 @@ def poll(
     return poller
 
 
+def _compare_numpy_array(new, old) -> bool:
+    """Check if a numpy array is equal to `old` value.
+
+    Deal with all the special cases when comparing either:
+
+      - numpy array `new` and non-numpy value `old`
+      - two numpy arrays `new` and `old`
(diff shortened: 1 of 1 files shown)


epanepucci:
I like this. It's more robust.

marcus-oscarsson:
Well done @elmjag. Thanks !

marcus-oscarsson (approved):

marcus-oscarsson:
Did you think more about integrating this as well: https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/utils/tango.py

elmjag:
> Did you think more about integrating this as well: https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/utils/tango.py

No, I did not think too much about it so far. I'll try to look at this in the coming weeks.

Issue #1459, opened 2025-12-02
Bug for MX_xBeam and MX_yBeam in ICATLIMS.py


olofsvensson:
The values for MX_xBeam and MX_yBeam are wrong in ICATLIMS.py.

I have a patch ready which fixes this problem.

olofsvensson:
Fixed in PR https://github.com/mxcube/mxcubecore/pull/1460

PR #1460, opened 2025-12-02
Issue #1459 : Fix for wrong MX_xBeam and MX_yBeam

olofsvensson:


mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -910,8 +913,8 @@ def store_common_data(self, datacollection_dict: dict) -> dict:
             "MX_beamShape": shape.value,
             "MX_beamSizeAtSampleX": bsx,
             "MX_beamSizeAtSampleY": bsy,
-            "MX_xBeam": bsx,
-            "MX_yBeam": bsy,
+            "MX_xBeam": xbeam,
+            "MX_yBeam": ybeam,
```
marcus-oscarsson:
```suggestion
            "MX_yBeam": datacollection_dict.get("yBeam"),
```

marcus-oscarsson: And the same for xBeam then :)

marcus-oscarsson:
Thanks, very nice :) :+1: . I guess there is no need for those intermediate variables if they are only used once ?

beteva:
Attention:  bsx and bsy are directly read from the HO and not passed as a dictionary key, as maybe not all the collection methods are having them (namely XRF, Escan).
Also there is a cleanup foreseen as there are two redundant values - the MX_xBeam and MX_beamSizeAtSampleX (the same goes for y), but we were waiting for Alex to confirm with the scientists which is the one needed in DRAC.

olofsvensson:
@beteva , MX_xBeam and MX_beamSizeAtSampleX  are **not** redundant: MX_xBeam is the beam coordinate on the detector and MX_beamSizeAtSampleX  is the beam size at the sample.

beteva:
Thank @olofsvensson fro explaining this. Now I understand better.  Your changes make sense and it is good that you use get, as they are indeed not present for XRF and energy scan.

mxcubecore/HardwareObjects/ICATLIMS.py
beteva: According to Alex it is better with a default value, rather than None
```
"MX_xBeam": datacollection_dict.get("xBeam", "N/A"),
"MX_yBeam": datacollection_dict.get("yBeam", "N/A"),
```

olofsvensson: Thanks @marcus-oscarsson and @beteva for suggestions! I have implemented them in the PR.

beteva (approved):
Thanks @olofsvensson

olofsvensson:
Then is it ok for me to merge? I ask as it's the first time I do this for MXCuBE...

beteva:
In principle it should be the reviewer, once the test have passed correctly, unless he gives you green light to do it yourself. You can merge if you want. otherwise i can do it.

olofsvensson:
Ok, please do it you Antonia.

In the DAU it's the assignee who does the merge after having the green light from the reviewer but every project can have different rules so I'm glad that I asked before merging! :-)

PR #1463, opened 2025-12-04
Fixed ISPyBDataAdapter `utf_decode`

### LINKED ISSUE #1457: decode issue in ISPyBDataAdapter
@marcus-oscarsson, I believe this could be a side effect of migrating to pydantic v2.0
I am getting the following error when synchronizing. Have you?
```
2025-12-01 10:57:33,130 |HWR    |ERROR  | 
  File "/home/demariaa/Software/workspaces/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py", line 60, in utf_decode
    res_d[key] = value.decode("utf8", "ignore")
                 ^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'decode'
2025-12-01 10:57:33,130 |HWR    |ERROR  | 
  File "/home/demariaa/Software/workspaces/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py", line 60, in utf_decode
    res_d[key] = value.decode("utf8", "ignore")
                 ^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'decode'
2025-12-01 10:57:33,130 |HWR    |ERROR  | 
  File "/home/demariaa/Software/workspaces/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py", line 60, in utf_decode
    res_d[key] = value.decode("utf8", "ignore")
                 ^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'decode'
2025-12-01 10:57:33,131 |HWR    |ERROR  | 
  File "/home/demariaa/Software/workspaces/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py", line 60, in utf_decode
    res_d[key] = value.decode("utf8", "ignore")
                 ^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'decode'
2025-12-01 10:57:33,131 |HWR    |ERROR  | 
  File "/home/demariaa/Software/workspaces/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py", line 60, in utf_decode
    res_d[key] = value.decode("utf8", "ignore")
                 ^^^^^^^^^^^^
AttributeError: 'datetime.datetime' object has no attribute 'decode'
2025-12-01 10:57:33,131 |HWR    |ERROR  | 
  File "/home/demariaa/Software/workspaces/mxcube/mxcubecore/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py", line 60, in utf_decode
    res_d[key] = value.decode("utf8", "ignore")
                 ^^^^^^^^^^^^
AttributeError: 'datetime.datetime' object has no attribute 'decode'

```

marcus-oscarsson:
Replacing `logging.getLogger("HWR").exception("")` with `pass` as its causing the recursion to fail. Closes #1457

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py
+++ b/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py
@@ -56,10 +56,10 @@ def utf_decode(res_d):
     for key, value in res_d.items():
         if isinstance(value, dict):
             utf_decode(value)
-        try:
+        try:  # noqa: SIM105
             res_d[key] = value.decode("utf8", "ignore")
-        except Exception:
-            logging.getLogger("HWR").exception("")
+        except Exception:  # noqa: S110
+            pass
 
     return res_d
 


mockoocy:
I think that a slight improvement to the function could be done since it's being touched :)
I believe that an `isinstance(value, bytes)` check could be introduced instead of `except Exception`, or just catch `AttributeError`.

Even more alternatively, the whole function could be rewritten to get rid of in-place mutations to something like 
```python
def utf_decode(obj):
    if isinstance(obj, d): 
        return {key: utf_decode(val) for key,val in obj.items()}
    if isinstance(obj, bytes): 
        return obj.decode("utf8")
    return obj 
```
But, that's a bit extra...  Do as you deem fit :)

marcus-oscarsson:
Actually I prefer not to touch it too much as its been like this for a very long time. Any other change needs more extensive testing. We are phasing out ISPyB and wont use this code for "much" longer. However I'm happy if you guys follow up with some other improvement.

epanepucci:
looks like this PR reverts back to the original **pass** so it's already been tested.
from my side it can be merged already.

as for **phasing out ispyb**, I don't think MAXIV is quite there yet :-D

marcus-oscarsson:
> looks like this PR reverts back to the original **pass** so it's already been tested.
> from my side it can be merged already.

Ok, yes that is what it does. I just found the fix a bit odd. Not certain why the `log` statement does not work while the `pass` does.

marcus-oscarsson:
> as for **phasing out ispyb**, I don't think MAXIV is quite there yet :-D

Sure no problem, thats why I though that any other improvement better come from you guys because we can't really test it reliably in the future.

PR #1466, opened 2025-12-08
Energy scan bug fixes

beteva:
- Wait for the energy to change.
- Change the directory for DRAC display.
- Ignore the get_scan_data call if not implemented - data only needed for the Qt GUI.


marcus-oscarsson (approved):

marcus-oscarsson:
Thanks @beteva

PR #1473, opened 2025-12-10
Fixed undefined variable fname for `set_rotation_axis_position`

marcus-oscarsson:
The variable `fname` is used in the except and else clauses


walesch-yan (approved):
Looks good to me!

PR #1475, opened 2025-12-11
tango command: use `attribute_query()` to check if attribute exists

elmjag:
When setting up a TangoChannel use `DeviceProxy.attribute_query()` method to check if an attribute exist. This way, we don't need look in the list of all device attributes. This list can potentially be quite long.

Move the code for checking if an attribute exists to it's own dedicated function, to improve code readability.

### DIFF
--- a/mxcubecore/Command/Tango.py
+++ b/mxcubecore/Command/Tango.py
@@ -143,6 +143,23 @@ class E:
         self.event = event
 
 
+def _device_has_attribute(device: DeviceProxy, attribute_name: str) -> bool:
+    """Check if a tango device has an attribute."""
+
+    try:
+        device.attribute_query(attribute_name)
+    except PyTango.DevFailed as ex:
+        if ex.args[0].reason == "API_AttrNotFound":
+            # query failed with 'attribute not found' error
+            return False
+
+        # unexpected exception, re-raise
+        raise
+
+    # attribute query was successful, thus we know the attribute exits
+    return True
+
+
 class TangoChannel(ChannelObject):
     _tangoEventsQueue = queue.Queue()
     _eventReceivers = {}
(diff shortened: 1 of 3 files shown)


mxcubecore/Command/Tango.py
```diff
@@ -143,6 +143,23 @@ def __init__(self, event):
         self.event = event
 
 
+def _device_have_attribute(device: DeviceProxy, attribute_name: str) -> bool:
+    """Check if a tango device have an attribute."""
+
+    try:
+        device.attribute_query(attribute_name)
```
beteva, line 150: Simple curiosity - is attribute_query available with all the Tango versions?

mxcubecore/Command/Tango.py
```diff
@@ -143,6 +143,23 @@ def __init__(self, event):
         self.event = event
 
 
+def _device_have_attribute(device: DeviceProxy, attribute_name: str) -> bool:
```
marcus-oscarsson: Nice ! Should it be `_device_has_attribute` ?

marcus-oscarsson:
Looks good to me in general, just a question regarding have/has. Antonia also has  a point about the availability of the feature ?

marcus-oscarsson (approved):

mxcubecore/Command/Tango.py
elmjag: > Simple curiosity - is attribute_query available with all the Tango versions?

Ah, good question!

I checked, it exists in PyTango 9.3.6. This is the oldest version MXCuBE supports, so it should be fine.

marcus-oscarsson: Perfect ok then

mxcubecore/Command/Tango.py
elmjag: Yes, it should be `_device_has_attribute` , changed.

PR #1488, opened 2026-01-26
Implement EPICSNState class (LNLS)

pedrobenetton:
This PR makes minor adjustments in our EPICSActuator class and implements EPICSNState class.
The EPICSNState class is already working with the backlight, frontlight and zoom components.
The EPICSActuator class is already working with: flux, energy, wavelength, cryojet temperature, detector distance, resolution, transmission (some are read_only, some are read/write).


beteva (commented):
Very nice to add  the documentation.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -22,6 +22,16 @@ def __init__(self, name):
...
         self.update_state(self.STATES.READY)
         self.connect(self.get_channel_object("rbv"), "update", self.update_value)
+        low_limit = self.get_property("low_limit")
+        high_limit = self.get_property("high_limit")
+        tolerance = self.get_property("tolerance")
+        if (low_limit is not None) and (high_limit is not None):
+            limits = (low_limit, high_limit)
+            self.set_limits(limits)
+        if tolerance is not None:
+            self.unit = tolerance
```
marcus-oscarsson: How is `self.unit` used, is it used as a sort of step size ?. It seems odd to assign `tolerance` to something called `unit` ?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -0,0 +1,126 @@
...
+      values: {'LEVEL0': 1, 'LEVEL1': 0}
+    """
+
+    def init(self):
+        super().init()
+        limits = self._nominal_limits
+        self.set_limits(limits)
+        self._initialise_values()
+        current_value = self.get_value()
+        self.update_value(current_value)
```
marcus-oscarsson, line 41: Can or should this update_value be handled via `self._set_value` ?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+    def update_value(self, value=None) -> None:
+        if value is None:
+            return
+        if not isinstance(value, Enum):
+            value = self.value_to_enum(value)
+        super().update_value(value)
+
+    def hasnt_arrived(self, setpoint):
```
marcus-oscarsson, line 68: I understand what this is doing, I'm wondering though. Can this be integrated into some sort of state maybe re-using `BUSY` are `MOVING` states ?. Then you could use something like `is_moving` or `is_ready` ?

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -0,0 +1,126 @@
...
+        if not isinstance(setpoint, Enum):
+            setpoint = self.value_to_enum(setpoint)
+        readback = self.get_value()
+        if not readback:
+            return False
+        return setpoint != readback
+
+
+class EPICSNStateInterval(EPICSNState):
```
marcus-oscarsson, line 77: Nice :)

marcus-oscarsson (approved):

marcus-oscarsson:
Looks good, thanks alot for the nice PR. I made a few remarks that may be interesting to look at. Its at the level where you know best what to do, if you are happy with the code like this just merge it, otherwise have a look a the comments and do the changes you think are necessary.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -69,6 +79,10 @@ def set_value(self, value, timeout: float = 0):
                 msg=msg,
             )
 
+    def set_limits(self, limits=(None, None)):
+        self._nominal_limits = limits
+        self.emit("limitsChanged", (self._nominal_limits,))
+
```
beteva: This is also handled the same way in AbstractActuator, so no need to repeat the code

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -22,6 +22,16 @@ def __init__(self, name):
     def init(self):
         self.update_state(self.STATES.READY)
         self.connect(self.get_channel_object("rbv"), "update", self.update_value)
+        low_limit = self.get_property("low_limit")
+        high_limit = self.get_property("high_limit")
+        tolerance = self.get_property("tolerance")
+        if (low_limit is not None) and (high_limit is not None):
+            limits = (low_limit, high_limit)
+            self.set_limits(limits)
```
beteva: The property default_limits, was meant exactly for setting the low and high limits at init. Do you really need to create new properties? This whole chunk (except for the tolerance) is doing exactly what is in the init of the AbstractActuator:
```
limits = self.get_property("default_limits")
        if limits:
            try:
                self._nominal_limits = tuple(literal_eval(limits)) ...
```
 You simply have to define the default_limits property as a list [low, high]

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -22,6 +22,16 @@ def __init__(self, name):
...
+        if tolerance is not None:
+            self.unit = tolerance
+        if self.get_property("read_only"):
+            self.read_only = True
```
beteva: This is already inherited from AbstractActuator, so you do not need to redo it.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -0,0 +1,126 @@
...
+    def _initialise_values(self):
+        values = self.get_property("values")
+        self.VALUES = Enum(
+            "ValueEnum",
+            dict(values, **{item.name: item.value for item in BaseValueEnum}),
+        )
+
+    def update_value(self, value=None) -> None:
+        if value is None:
+            return
```
beteva, line 63: There are many places where update_value is used with no argument. In this case it gets the value first and than there is a mechanism to decide whether a signal will be sent or not. In your case you change the logic and do nothing - could you explain us the reason. Thank you

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -0,0 +1,126 @@
...
+        val:
+          suffix: ''
+    configuration:
+      low_limit: 0
+      high_limit: 1
+      values: {'LEVEL0': 15000, 'LEVEL1': 0}
+    """
+
+    def init(self):
+        super().init()
```
beteva: This is not needed, you can remove it.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def update_value(self, value=None) -> None:
+        if value is None:
+            return
```
beteva, line 112: The same as above - could you, please, explain why you are changing the logic of update_value.

beteva:
Nice work @pedrobenetton . My comments are mostly to brush it up and to understand some of the features.

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
pedrobenetton: `self.unit` wasn't being used, so I simply removed it. It was an idea we had some time ago but we forgot to remove it form the `init`

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
pedrobenetton: I will take a look on this, and also on the suggestion for `hasnt_arrived`, thanks :)

mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
pedrobenetton: Hi Antonia! All of your other suggestions were correctly pointed out and I fixed them. Regarding this one: I did this because during the movement of any NState device, the device would stay in the value UNKNOWN for a brief amount. Here is a recording that shows it:

https://github.com/user-attachments/assets/3284125b-60ea-4cd8-853f-8611cdac6118

I realised that this was only happening when `update_value` was used with no argument, just how you pointed it. I added the `return` for these cases, and it worked: frontend stopped displaying this behaviour. Can you tell me where this mechanism is so I can take a better look?

marcus-oscarsson:
Thanks for the nice work @pedrobenetton

PR #1489, opened 2026-01-26
EpicsMotorDetachable

RafaelLyra8:
Class that implements an interface of motor record IOC and accepts a presence PV, that indicated if the motor is or is not present on the facility. If it is present, it actuates as an EpicsMotor, if it is not, then the widget receives the state OFF and remains disabled.


marcus-oscarsson (approved):

marcus-oscarsson:
Please go head and re-base and merge :)

PR #1490, opened 2026-01-26
Bluesky Http Server Integration

RafaelLyra8:
Create a Command to interact with the Bluesky Http Server API.


marcus-oscarsson:
Very nice !, thanks !

marcus-oscarsson (approved):

marcus-oscarsson:
I think all the EPICS work you are doing could be presented at the next meeting or at least during a developers meeting. If you are up for it ?

RafaelLyra8:
Yes, we could present our work during the next meeting.

PR #1500, opened 2026-02-06
Make HO compliant with bioth xml and yaml configuration

beteva:
The incompatibility is with properties of type tuple, list, dictionary.
**get_property** returns a string which we have to convert with xml, an tuple, list or dictionary with the yaml.
The fix is that we only convert with literal_eval if the return value is a string.

### DIFF
--- a/mxcubecore/HardwareObjects/mockup/MotorMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/MotorMockup.py
@@ -32,13 +32,10 @@ Example of xml config file
 </object>
 """
 
-import ast
 import time
+from ast import literal_eval
 
-from mxcubecore.HardwareObjects.abstract.AbstractMotor import (
-    AbstractMotor,
-    MotorStates,
-)
+from mxcubecore.HardwareObjects.abstract.AbstractMotor import AbstractMotor, MotorStates
 from mxcubecore.HardwareObjects.mockup.ActuatorMockup import ActuatorMockup
 
 __copyright__ = """ Copyright © 2010-2020 by the MXCuBE collaboration """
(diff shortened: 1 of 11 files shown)


marcus-oscarsson (approved):

mxcubecore/HardwareObjects/mockup/MotorMockup.py
```diff
@@ -69,11 +66,12 @@ def init(self):
...
             self.update_limits(DEFAULT_LIMITS)
-        try:
-            _wr = self.get_property("wrap_range")
-            self._wrap_range = DEFAULT_WRAP_RANGE if not _wr else ast.literal_eval(_wr)
-        except (ValueError, SyntaxError):
-            self._wrap_range = DEFAULT_WRAP_RANGE
+
+        self._wrap_range = DEFAULT_WRAP_RANGE
+        _wr = self.get_property("wrap_range")
+        if isinstance(_wr, str):
```
rhfogh: What happens if _wr is not a sring, but a lis t or tuple of floats - as it arguably should be? Should this not be changed?

rhfogh (changes requested):
Nice work. Just the one comment to address.

mxcubecore/HardwareObjects/mockup/MotorMockup.py
beteva: Thanks @rhfogh for outlining it. I've just kept the previous logic, which was never taking anything else than a string. I'll add the rest, just curious to see that nobody have complained up to now.
Actually, looking at the code, we should not even use literal_eval, as it should only be float.

PR #1502, opened 2026-02-06
logging and ruff complaint

epanepucci:
- use self.log instead of root logger
- use contextlib.suppress('Exception') instead of try/except/pass; this obviates that this exception is really not an exception
- prefix unused variable to please ruff


mockoocy (approved):

marcus-oscarsson:
As you probably have understood `GenericDiffractometer` is not replaced by `AbstractDiffractometer`

Issue #1504, opened 2026-02-13
Recent Diffractometer changes have broken Mockup


rhfogh:
Trying to run a mockup with the most recent mxcubecore found it non-functional, with numerous errors. Fixes for some of the simpler ones are in a recent pull request, but not all. This (re)raises some questions:

- Did anyone test the mock version before pushing and merging those changes?
- Where can I go to find a mock configuration that works with the new code? Without that, how do I fix problems?  (yes, I know I have promised to do some clean-up here, but I have postponed that till 1) I have a known-to-be-working version to work from, and 2) we have finished the move to yaml configuration)
- What is the correct way of avoiding this kind of problem?

Finally, three concrete problems that need fixing, even after the parallel PR:

1) (This is on Qt, but comes back to a missing Diffractometer.omega)
```
Traceback (most recent call last):
  File "/home/rhfogh/pycharm/mxcubeqt/mxcubeqt/configuration.py", line 82, in load_brick
    new_instance = class_obj(None, brick_name)
  File "/home/rhfogh/pycharm/mxcubeqt/mxcubeqt/bricks/task_toolbox_brick.py", line 67, in __init__
    self.task_tool_box_widget = TaskToolBoxWidget(self)
  File "/home/rhfogh/pycharm/mxcubeqt/mxcubeqt/widgets/task_toolbox_widget.py", line 69, in __init__
    self.discrete_page = CreateDiscreteWidget(self.tool_box, "Discrete")
  File "/home/rhfogh/pycharm/mxcubeqt/mxcubeqt/widgets/create_discrete_widget.py", line 44, in __init__
    CreateTaskBase.__init__(
  File "/home/rhfogh/pycharm/mxcubeqt/mxcubeqt/widgets/create_task_base.py", line 96, in __init__
    HWR.beamline.diffractometer.omega.connect(
  File "/home/rhfogh/pycharm/mxcubecore/mxcubecore/CommandContainer.py", line 311, in __getattr__
    return super().__getattribute__(attr)
AttributeError: 'DiffractometerMockup' object has no attribute 'omega'
```

2) 

  File "/home/rhfogh/pycharm/mxcubeqt/mxcubeqt/bricks/tree_brick.py", line 445, in logged_in
    and HWR.beamline.diffractometer.use_sample_changer()

  File "/home/rhfogh/pycharm/mxcubecore/mxcubecore/CommandContainer.py", line 311, in __getattr__
    return super().__getattribute__(attr)

[...]

PR #1505, opened 2026-02-13
Simple bug fixes following AbstractDiffractometer changes

rhfogh:
This removes some of the errors that prevent the mockup from being run, and should hopefully be obvious enough to  pass. It does not remove all of them, though, so this could not be tested. See Issue #1504 for more info on this.

There are consequent problems in both mxcubeqt and mxcubeweb - I have made parallel PRs there too

### DIFF
--- a/mxcubecore/HardwareObjects/ALBA/ALBAMiniDiff.py
+++ b/mxcubecore/HardwareObjects/ALBA/ALBAMiniDiff.py
@@ -280,8 +280,8 @@ class ALBAMiniDiff(GenericDiffractometer):
         """
         return True
 
-    # TODO:Implement dynamically
-    def in_plate_mode(self):
+    @property
+    def in_plate_mode(self) -> bool:
         """
         Overrides GenericDiffracometer method.
         """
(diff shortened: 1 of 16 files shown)


beteva:
@rhfogh,
thanks for spotting some inconsistency in the hardware objects.
The AbstarctCollect has to be changed indeed.
The MiniDiff, GenericDiffractometer and their inheritance should not be affected by the AbstractDiffractometer, so the in_plate_mode being a method for them is correct.
The rest is already corrected in mxcubecore v2.4.

rhfogh:
@beteva Not sure I agree with you there. It may work if you are always within a single beamline, but there is code, like our workflows or the mock classes, that have to work with several. So if you write something like
`if HWR.beamline.diffractometer.in_plate_mode():` - as it still says in AbstractCollect after fetching the latest develop, BTW - is that supposed to work or not? Surely there should be _one, consistent_ way of checking if the diffractometer is in plate mode across all active diffractometer classes? I mean we are not supposed to do `if isinstance(HWR.diffractometer, AbstractDiffractometer)` in AbstractCollect, are we? Besides the fix is very simple - just change in_plate_mode to a property everywhere it appears, as I did in my PR.

I guess we had better agree on this before we do anything with my outstanding PRs.

beteva:
> I guess we had better agree on this before we do anything with my outstanding PRs.

Definitely agree with you.

> Besides the fix is very simple - just change in_plate_mode to a property everywhere it appears, as I did in my PR.

This is a bit more tricky. Your changes are absolutely valid, but only can be used with matching mxcubeweb and mxcubeqt, where in_palte_mode is also invoked.

rhfogh:
Well, yes. Which is why I made matching PRs for all three repositories. If we want this to be harmonised, we can do it.

rhfogh:
Closed as superseded by changes from elsewhere and by a modified PR.

PR #1509, opened 2026-02-27
AbstractDiffractometer adaption and GPhL updates

rhfogh:
Two parts to this PR:
- Fixed bugs and cleaned up code related to recent AbstractDiffractometer merge. Code now works on mock web version (for rhfogh)
- Adapted configuration, strategies and log output for GPhL workflow

NB mxcubeqt still does NOT work with the  AbstractDiffractometer change, as the Qt SampleVIew class (QtGraphicsManager) needs a complete refactoring.


marcus-oscarsson:
@rhfogh, well done, thanks for doing these fixes !

marcus-oscarsson (approved):

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -378,44 +378,36 @@ def get_state_motors(self, motors_list: list | None = None) -> dict:
...
     # -------- Head Type and Modes --------
 
-    @property
-    def get_head_type(self) -> DiffractometerHead:
-        """Get the head type
-        Returns:
-            DiffractometerHead member.
-        """
-        return self.head_type
-
```
beteva, line 388: This is meant as a place holder for reading of the head type. The reading is an action, different for different type of diffractometers. I do not think it should be removed. Also it is normal that there is no set function for it - it comes from the hardware.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -378,44 +378,36 @@ def get_state_motors(self, motors_list: list | None = None) -> dict:
...
         """Check if the head is a plate."""
 
-        return self.get_head_type == DiffractometerHead.PLATE
+        return self.head_type == DiffractometerHead.PLATE
 
     @property
     def in_kappa_mode(self) -> bool:
         """Check if the head is MiniKappa."""
 
-        return self.get_head_type == DiffractometerHead.MINI_KAPPA
```
beteva, line 399: See my above comment.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -378,44 +378,36 @@ def get_state_motors(self, motors_list: list | None = None) -> dict:
...
     def in_injector_mode(self) -> bool:
         """Check if there is injector on the head."""
         return (
-            self.get_head_type == DiffractometerHead.SSX
+            self.head_type == DiffractometerHead.SSX
             and self.current_constraint == DiffractometerConstraint.INJECTOR
         )
 
     @property
-    def get_head_enum(self):
```
beteva, line 418: This is also debatable - why do you want to remove get_?

test/test_diffractometer.py
```diff
@@ -65,10 +65,10 @@ def test_get_phase_list(self, test_object):
         # subtract one for the UNKNOWN
         assert len(phase_enum) - 1 == len(test_object.get_phase_list())
 
-    def test_get_head_type(self, test_object):
-        head_enum = test_object.get_head_enum
+    def test_head_type(self, test_object):
+        head_enum = test_object.head_enum
         # in the mockup we set the head to be minikappa
-        assert test_object.get_head_type == head_enum.MINI_KAPPA
+        assert test_object.head_type == head_enum.MINI_KAPPA
```
beteva, line 71: As I''ve commented above - I do not quite agree with removing of the get_head_type method.

beteva (changes requested):
The changes you've made are very welcome. There is only one problem - the get_head_type -the implementation you've done is fine with a mock class, but requires that self.head_type is properly assigned. As the assignment depends on the hardware and is an action, I've thought that having a method (even if it is a property)  is more "generic" than just having the self.head_type. otherwise it will depend on specific methods for each diffractometer implementation and might induce error, if there is hardware which changes the head easily,  I think we should keep the method.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: As I see it, something that starts with 'get_' should be a function, and be called as a function, i.e. `obj.get_head_type()`.
Properties should be equivalent of attributes, so a property called 'head_type' returns the head_type, and a property called 'get_head_type' should return a _function_. 
A property called 'get_head_type' that returns the head_type value is strongly misleading, IMHO.
So if it is better to have a function for this purpose, there is nothing wrong with that. But I would much prefer to avoid have a property with a function_type name.

rhfogh:
We could have head_type as a property and change it depending on the implementation, just like we could have get_head_type() as a function and change it according to the implementation. But functions are fine. My problem is with the name matching the nature.

marcus-oscarsson:
@beteva is on vacation at the moment so we are still waiting for her input on this, please stand by :) (unfortunately GitHub does not play any nice waiting music :rofl:)

rhfogh:
@marcus-oscarsson Thanks for letting me know. Clearly this one cannot be merged until there is agreement with Antonia.

Issue #1520, opened 2026-03-10
Modify data portal 'MX_kappa_settings_id'


olofsvensson:
Up till now the 'MX_kappa_settings_id' uploaded to the data portal was an uuid4 identifier which was displayed by the data portal:

<img width="730" height="757" alt="Image" src="https://github.com/user-attachments/assets/18c70793-a522-4324-b0f4-094ec59d9397" />

The patch I'm going to submit as a PR will change the id to something like this:

<img width="830" height="694" alt="Image" src="https://github.com/user-attachments/assets/3d2a3d10-eec3-40e8-a91e-e811a32ef344" />

Note that the double "Kappa Kappa" will be solved by Alejandro by removing the "hardcoded" "Kappa" in the data portal.

Note that this cannot be done in the ESRF BES workflows because for the GPhL workflow it is doing the opening of the kappa and the data collections.

rhfogh:
Very good initiative. 

Sorry for my ignorance - but what is implied by the 'ESRF BES workflows'? Is this improvement something that will work for GPhL workflows, or is it the contrary, that it will _not_ work for them?

olofsvensson:
The BES stands for the "Beamline Expert System", a name dating back to the days when we had Passerelle EDM. The ESRF BES workflows are the workflow implementation on the MX beamlines. All the MXPress workflows are BES workflows but there are also more non-MXPress workflows in BES.

The proposed change will improve the display in the data portal both for GPhL workflows and non-GPhL (BES) workflows.

olofsvensson:
Issue closed.

PR #1521, opened 2026-03-10
Issue #1520 : Changed MX_kappa_settings_id

olofsvensson:
See issue #1520 

Let me know if there is a better way of implementing this functionality, i.e. getting the "kappa" and "kappa_phi" motor positions.

### DIFF
--- a/mxcubecore/HardwareObjects/ICATLIMS.py
+++ b/mxcubecore/HardwareObjects/ICATLIMS.py
@@ -1245,9 +1245,9 @@ class ICATLIMS(AbstractLims):
             if "kappa" in motor_positions:
                 kappa_pos = round(float(motor_positions["kappa"]), 1)
                 kappa_phi_pos = round(float(motor_positions["kappa_phi"]), 1)
-                MX_kappa_settings_id = f"Kappa: {kappa_pos}, Phi: {kappa_phi_pos}"
+                mx_kappa_settings_id = f"Kappa: {kappa_pos}, Phi: {kappa_phi_pos}"
             else:
-                MX_kappa_settings_id = None
+                mx_kappa_settings_id = None
 
             metadata.update(
                 {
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
@olofsvensson There seems to be a conflict with the previous PR ?

mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -1223,6 +1223,15 @@ def finalize_data_collection(self, datacollection_dict):
             proposal = f"{HWR.beamline.session.proposal_code}"
             proposal += f"{HWR.beamline.session.proposal_number}"
 
+            diffractometer_hwobj = HWR.beamline.diffractometer
+            motor_positions = diffractometer_hwobj.get_positions()
+            if "kappa" in motor_positions:
```
marcus-oscarsson, line 1245: @beteva: Is this compatible with the new `AbstractDiffractomter`, if not can the code be made compatible with both `MicroDiff` and `MicroDiffractometer` objects ?

olofsvensson:
@marcus-oscarsson Which previous PR?

olofsvensson:
Ok I see it:

image

I'll fix the branch.

mxcubecore/HardwareObjects/ICATLIMS.py
marcus-oscarsson: So this would get the current motor positions. In this case where these two motor are at the end of the data collection. Is that what you want ?

An approach that would be compatible with both `Microdiff` and `MicroDiffractomter` objects would be to simply access the `kappa` and `kappa_phi` members directly on the diffractometer object. Something like:

```
kappa_pos = diffractometer_hwobj.kappa.get_value() if hasattr(diffractometer_hwobj, "kappa") else ""

kappa_phi_pos = diffractometer_hwobj.kappa_phi.get_value() if hasattr(diffractometer_hwobj, "kappa_phi") else ""

mx_kappa_settings_id = f"Kappa: {kappa_pos}, Phi: {kappa_phi_pos}" if kappa_pos else ""
```

marcus-oscarsson:
Could you please rebease this branch please ?

olofsvensson:
I usually don't rebase branches in other projects - can you please send me the commands I'm supposed to use or point me to documentation on how to rebase for MXCuBE? I'm a bit afraid of messing up the repository...

marcus-oscarsson:
You will only do it on your branch, so you would locally do:

```
git fetch all
git checkout olofsvensson:issue_1520
git rebase origin/develop
```

Then you push back to ` olofsvensson:issue_1520` i.e

```
git push -f origin issue_1520
```

marcus-oscarsson:
Thanks @olofsvensson

marcus-oscarsson (approved):

PR #1523, opened 2026-03-12
Bluesky output

pedrobenetton:
This PR creates a poller that logs all outputs from bluesky server to the mxcube logs for LNLS

### DIFF
--- a/mxcubecore/Command/BlueskyHttpServer.py
+++ b/mxcubecore/Command/BlueskyHttpServer.py
@@ -110,7 +110,9 @@ class BlueskyHttpServerCommand(CommandObject):
         )
         return self.format_response(response)["console_output_uid"]
 
-    def show_console_output(self, value=[]):
+    def show_console_output(self, value=None):
+        if value is None:
+            value = []
         for console_msg in value:
             new_console_line = console_msg["msg"]
             new_console_line = new_console_line.strip()
--- a/mxcubecore/Command/BlueskyHttpServer.py
+++ b/mxcubecore/Command/BlueskyHttpServer.py
@@ -24,8 +24,10 @@ import time
 
 import gevent
 import requests
+import logging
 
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Ok, did you think about or foresee some security layer like HTTPS ~and token authentication~ ?

Noticed that you already have token authentication in place

mxcubecore/Command/BlueskyHttpServer.py
```diff
@@ -37,13 +39,30 @@ class BlueskyHttpServerCommand(CommandObject):
...
     _execute_path = "api/queue/item/execute"
+    _console_uid_path = "api/console_output/uid"
+    _console_path = "api/console_output_update"
 
     def __init__(self, name, url, timeout=5, **kwargs):
         CommandObject.__init__(self, name, **kwargs)
         self.username = ""
         self._default_timeout = timeout
         self._url = f"{url}/" if url[-1] != "/" else url
         self._headers = {"Authorization": f"ApiKey {os.environ['AUTH_KEY']}"}
```
marcus-oscarsson, line 50: Oh, ok so thats done :)

marcus-oscarsson (approved):

mxcubecore/Command/BlueskyHttpServer.py
marcus-oscarsson: I assume that you are also using HTTPS then  ?

marcus-oscarsson:
You can rebase and merge :)

beteva (approved):

mxcubecore/Command/BlueskyHttpServer.py
pedrobenetton: Hi Marcus! Currently, it is only HTTP, but yes we use an auth key for it

PR #1524, opened 2026-03-13
Second try at rewording axis and motor definitions

rhfogh:
Modified motor definitions following discussion in https://github.com/mxcube/mxcubecore/discussions/1470

This PR  has corrected the mistakes of the previous attempt.

-    Definitions follow the original system of @beteva, but are tightened to avoid any possible ambiguity. x, y and z form a right-handed orthogonal axis system with z in the direction of gravity, y orthogonal to gravity and beam, and x approximately along the beam. Note that the definitions use the nominal beam direction, as the motor vectors are not supposed to be update when the actual beam drifts.

-    sample_horizontal and sample_vertical definitions have been changed, as the original set seemed inconsistent. Please check that the current definitions match actual use.

There are some questions (apart from the obvious whether these are all correct):

- What are the directions of sample_horizontal and sample_vertical (+Y or -Y, +Z or -Z)? Are they exactly on the definition axes, or are they variable and linked to hardware, e,g, to the camera? Are they identical to the CentringMath CameraAxes? 

- Does the Z axis point up or down - comments to the previous PR were a bit ambiguous on that point.


mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
 actuator object. This allows to use them in a standard way by the MXCuBE
 application (web or Qt).
 There is also a convention of the direction of the motors:
-  - x axis is parallel to the beam. Positive direction is from right to left.
-  - y axis is perpendicular to the beam, parallel to the ground. Positive
-      direction is from left to right when facing the beam.
-  - z axis is perpendicular to the floor. Positive direction is top down.
+- z axis is in the direction of gravity. Positive direction is downwards.
```
marcus-oscarsson: That's fine, lets decide that the direction is "downwards"

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
...
+- y axis is perpendicular to the z axis and to the nominal beam direction.
```
marcus-oscarsson, line 29: I would not make it complicated by using the term nominal beam direction. I think its enough to say that x is in the direction of the beam

rhfogh: I can see your point, but does it then mean that if the beam direction changes by a few fractions of a degree the Y axis direction also changes so you would have to recalculate the motor directions? This is a definition after all. How do you avoid that one?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
...
+  Positive direction is so that x,y,z would form a right-handed coordinate system.
```
marcus-oscarsson, line 30: Not sure that's true the positive direction is currently out from the ring, so that would not create a right-handed coordinate system, right ?

We can perhaps agree on that it should be a right handed system, I don't think it will change too much.

For the Micro-diffractomter, the directions will depend on the placement of the diffractometer, positive away from the MD.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
...
+- x axis is orthogonal to the y and z axes (and approximately parallel to the beam).
+  Positive direction is from the sample towards the detector.
+
 
 Here follows the list of the fixed roles and the description of the
 corresponding objects, accessible via beamline.diffractometer hardware object.
 
 1. Motor objects (roles)  and their functionality:
   omega - the rotation axis, independent of the orientation (up, down or side).
```
marcus-oscarsson, line 39: Do we add `chi` to describe the orientation of the rotation axis ?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
...
-          Pisitive direction is clockwise.
+          Positive direction is a right-handed rotation.
+  kappa - the rotation axis between the omega and kappa_phi axes.
+          Positive direction is a right-handed rotation. Not present on all goniostats.
+  kappa_phi - the rotation axis directly attached to the sample.
+          Positive direction is a right-handed rotation. Not present on all goniostats.
   sampx - centring table x axis
   sampy - centring table y axis
   focus - alignment table x axis
```
marcus-oscarsson, line 47: should we call these alignment-x-y-z ?

marcus-oscarsson:
> * What are the directions of sample_horizontal and sample_vertical (+Y or -Y, +Z or -Z)? Are they exactly on the definition axes, or are they variable and linked to hardware, e,g, to the camera? Are they identical to the CentringMath CameraAxes?
> 

As I mentioned above its currently depending on how the MD is positioned but the idea with this is that it should not be. Here positive is away from the ring, all our MD's are positioned in such a way.

Perhaps using the OAV, as not everybody has a ring to refer to, is a good idea. Meaning that towards the right or up on the OAV are positive and left or down are negative ? 

> * Does the Z axis point up or down - comments to the previous PR were a bit ambiguous on that point.

I think any direction would be fine, its a bit odd to have a positive direction going "downwards" but otherwise it would be fine.

rhfogh:
@marcus-oscarsson I am happy with Z either up or down (in fact up does make slightly more sense), I just took that from what @beteva wrote. SO I oculd easily change teh direction of teh Y and Z axes if htaat is what we prefer.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: I don't think it really matters once the direction is established the rest is set. I think its well understood that X is parallel to the "calibrated" beam direction and that its not being updated real time. 

Its pretty cool if someone decides/manages to update the direction of X in realtime but it would not change anything once your in that frame of reference.

beteva:
@rhfogh

> I am happy with Z either up or down (in fact up does make slightly more sense), I just took that from what @beteva wrote

What I wrote was simply how the calculation in the sample_centring were made. I do agree that Z positive upwards is more natural, so this should be changed. Taking the OAV as a reference is a good idea. Probably for many synchrotrons it would coincide with the beam. We already assume that all diffractometers have OAV at the same place, as we have the notion of front and back light.

rhfogh:
As of 20260317 I have changed the direction of the Z axis from 'downwards' to 'upwards'. Some outstanding questions:

- It still says that " y axis is perpendicular to the z axis and to the nominal beam direction."
Marcus quite rightly thinks this sounds complicated. The problem is that if we define the axes relative to the actual beam direction - and the beam can change direction, if only slightly - that would in principle require that we recalculated all the motor directions every time the beam moved. How do we want to resolve this?

- I have not put directions for the OAV pseudomotors, and have added 'approximately' to the directions. Is this OK?
I have an underlying plan: The next PR will propose to define motors the way they are defined in CentringMath, where each motor has an explicit direction vector (in the official coordinate system) and the order of motors from goniostat to sample is given. This frees you from having to make assumptions about exact motor directions, and allows you to write/use centring code (like CentringMath) that is independent of motor names and motor directions. 
But before we get to that we had better agree on the definitions.

marcus-oscarsson:
> As of 20260317 I have changed the direction of the Z axis from 'downwards' to 'upwards'. Some outstanding questions:
> 
>     * It still says that " y axis is perpendicular to the z axis and to the nominal beam direction."
>       Marcus quite rightly thinks this sounds complicated. The problem is that if we define the axes relative to the actual beam direction - and the beam can change direction, if only slightly - that would in principle require that we recalculated all the motor directions every time the beam moved. How do we want to resolve this?

I would call it parallel to known/calibrated beam position in that case


> 
>     * I have not put directions for the OAV pseudomotors, and have added 'approximately' to the directions. Is this OK?

For the moment that is fine Id say, I think using OAV, left, right, up and down for reference would be the easiest. 

>       I have an underlying plan: The next PR will propose to define motors the way they are defined in CentringMath, where each motor has an explicit direction vector (in the official coordinate system) and the order of motors from goniostat to sample is given. This frees you from having to make assumptions about exact motor directions, and allows you to write/use centring code (like CentringMath) that is independent of motor names and motor directions.
>       But before we get to that we had better agree on the definitions.

What does the ordering add when we have the names and the directions ?

marcus-oscarsson:
Thanks for the updates @rhfogh

marcus-oscarsson (commented):
Great, well done ! :) I think that we should present this on the next dev meeting as well to make sure that it reaches everyone !

marcus-oscarsson (approved):

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Good point of adding chi. It is not a motor, only an angle, so it should be added to the description of omega, explaining what chi = 0, 90 and -90 means.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Also the way we use it (no matter what the direction of the goniostat is, it is just the focus, not taken into account in any of the centring routines.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: agree with @marcus-oscarsson. The whole point of having a convention is to help people understand how they should configure their axes and how they are used in the sample centring routines. I believe the more words we use to explain (e.g. positive direction away from the MD), the better.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: A do not mind, although I like phix being called focus, as it is really what it is and is not used in the alignmet/centring calculations. Phiy, phiz being called alignmenty, alignmetz may perturm the scientists (the names come from them).

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
...
   phiy - alignment table y axis
   phiz - alignment table z axis
-  sample_horizontal - x axis, combination of sampx and sampy. Equivalent to
-                      sampx at omega=0 and sampy for omega=90.
-  samle_vertical - y axis, combination of sampx and sampy. Equivalent to
-                   sampy at omega=0 and sampx for omega=90.
+  sample_horizontal - Pseudomotor approximately along the y axis.
+         Generated by phiy or a combination of sampx, sampy depending on goniostat
+  sample_vertical - Pseudomotor approximately along the z axis.
+         Generated or phiz or a combination of sampx, sampy depending on goniostat
```
beteva, line 53: This is a bit vague - there should be examples, otherwise people would not know what it represents exactly, if they do not have a Microdiff, that calculates for them. Not quite sure phiy or phiz have any role to play here, but I may be wrong.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -25,26 +25,32 @@
...
 application (web or Qt).
 There is also a convention of the direction of the motors:
-  - x axis is parallel to the beam. Positive direction is from right to left.
-  - y axis is perpendicular to the beam, parallel to the ground. Positive
-      direction is from left to right when facing the beam.
-  - z axis is perpendicular to the floor. Positive direction is top down.
+- z axis is in the direction of gravity. Positive direction is upwards.
+- y axis is perpendicular to the z axis and to the nominal beam direction.
+  Positive direction is so that x,y,z would form a right-handed coordinate system.
+- x axis is orthogonal to the y and z axes (and approximately parallel to the beam).
```
beteva, line 31: As @marcus-oscarsson said, we assume that the beam does not change direction, so we might as well drop the "aprroximately", as it brings confusion.

diffractometer.md
```diff
@@ -4,22 +4,30 @@ This allows them to be used in a standard way by the MXCuBE application (web or
 
 We also introduce a convention about the direction of the alignment and centring motors:
 
-- x axis is parallel to the beam. Positive direction is from right to left.
-- y axis is perpendicular to the beam, parallel to the ground. Positive direction is from left to right, facing the beam.
-- z axis is perpendicular to the floor. Positive direction is top down.
+- z axis is in the direction of gravity. Positive direction is upwards.
+- y axis is perpendicular to the z axis and to the nominal beam direction.
+  Positive direction is so that x,y,z would form a right-handed coordinate system.
+- x axis is orthogonal to the y and z axes (and approximately parallel to the beam).
```
beteva, line 10: See the comments above

diffractometer.md
```diff
@@ -4,22 +4,30 @@ This allows them to be used in a standard way by the MXCuBE application (web or
...
 - **sampy** - centring table y axis
 - **focus** - alignment table x axis
 - **phiy** - alignment table y axis
 - **phiz** - alignment table z axis
-- **sample_horizontal** - x axis, combination of sampx and sampy. Equivalent to sampx at omega=0 and sampy for omega=90.
-- **sample_vertical** - y axis, combination of sampx and sampy. Equivalent to sampy at omega=0 and sampx for omega=90.
+- **sample_horizontal** - Pseudomotor approximately along the y axis.
+  Generated by phiy or a combination of sampx, sampy depending on goniostat
+- **sample_vertical** - Pseudomotor approximately along the z axis.
+  Generated or phiz or a combination of sampx, sampy depending on goniostat
```
beteva, line 30: See the comments above

beteva (commented):
@rhfogh, thank you for the efforts you do to make the convention clearer and more universal. I've made some comments, we need to agree on - they are made in the AbstractDiffractometer part, but are relevant for the convention.md, as they should contain the same information.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Not sure I understand this. As I understand it, chi is the angle between the phi and the omega axes, which can never be +/- 90 on a minikappa. What do you mean by chi?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: Well, if sample_horizontal and sample_vertical are orthogonal to the beam (as Marcus told me they were) then one of them is an omega-dependent linear combination of sampx and sampy, and the other one is either phiy or phiz. If, on the other hand, both  sample_horizontal and sample_vertical are just linear combinations of sampx and sampy, then one of them is along the beam, and the names are wrong because they would both be horizontal for a vertical goniostat. 

What instrument or piece of hardware are sample_horizontal and sample_vertical tied to? Are they not the two axes of the OAV?

rhfogh:
> > As of 20260317 I have changed the direction of the Z axis from 'downwards' to 'upwards'. Some outstanding questions:
> > ```
> > * It still says that " y axis is perpendicular to the z axis and to the nominal beam direction."
> >   Marcus quite rightly thinks this sounds complicated. The problem is that if we define the axes relative to the actual beam direction - and the beam can change direction, if only slightly - that would in principle require that we recalculated all the motor directions every time the beam moved. How do we want to resolve this?
> > ```
> 
> I would call it parallel to known/calibrated beam position in that case
> 
> > ```
> > * I have not put directions for the OAV pseudomotors, and have added 'approximately' to the directions. Is this OK?
> > ```
> 
> For the moment that is fine Id say, I think using OAV, left, right, up and down for reference would be the easiest.
> 
> > ```
> >   I have an underlying plan: The next PR will propose to define motors the way they are defined in CentringMath, where each motor has an explicit direction vector (in the official coordinate system) and the order of motors from goniostat to sample is given. This frees you from having to make assumptions about exact motor directions, and allows you to write/use centring code (like CentringMath) that is independent of motor names and motor directions.
> >   But before we get to that we had better agree on the definitions.
> > ```
> 
> What does the ordering add when we have the names and the directions ?

@marcus-oscarsson @beteva 
I am basing this on the way Gleb set it up for CentringMath. The thing is that when you have to deal with multiple synchrotrons and multiple ways of mounting different goniostats it gets hard to have a set of motor names that fits well everywhere. Just a couple of problems:
- the omega angle where sampx or sampy are along the beam could be any number, e.g. 23.47deg,  and vary over time, 
- Which motor to use for centring, phix or phiy, depends on how the goniostat is mounted.
- So does anything you define as 'horizontal' or 'vertical'. 
- I am not sure whether you can rely on the positive movement or rotation direction always being the same relative to the diffractometer - certainly not across different models and makes.

I think the best way to do it is to separate things out:
- There is a single clear definition that describes a coordinate system, nothing more, linked to the beam and gravity so it is always clearly defined.
- Motor names are defined by how they are named in the hardware, which is always clear which motor has which name. People can interact with the motors by name when they now how they are set up locally, but the code does not depend on it. 
- You then have a mapping between the hardware motors and the goniostat set-up that tells you the direction of all motors at omega=0, and the order they are working in. That allows you to do all the centring calculations in the same way with the exact same code regardless of which motors are named what or how the goniostat is mounted (see CentringMath). It also means that you do not rely on assumptions about the exact relative angles of the motor directions, let alone about where sampx points at omega=0.
- You can still have it so that you know which motors always point in which direction at the ESRF and work with that. It is just that instead of putting it in the basic definitions where things get complicated, it is a matter of putting the appropriate motors with their direction vectors in the ESRF-specific config file and telling your people about it.
- As a side issue you get rid of the need for specifying 'direction' - the directions in the config file are the ones that define positive rotation.  The only thing you need in addition (and that only for shadowing calculations) is a single boolean telling you whether the omega rotation axis (positive rotation)  points towards the sample or towards the diffractometer.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Well, again I am only translating the code as it is now (not written by me at first place 😄). Chi is at 0 when yoy have horizontal omega spindle. +90 with the up and -90 with the down oriented omega spindle. It is only used for sample centring and I do not know at all how it is used with minikappa. If you find too disturbing to use chi, we can find some other way to define the orientation of the omega axis of the goniometer.

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
marcus-oscarsson: I think thats alright, I mean if we say that the motor that moves along Y and Z are phiy and phiz respectively (although I prefer the names alignment-y and -z, used to "align" the rotation axis in the beam path). Then sample movement along one of those axes will be parallel to either phiy and phiz. I think thats clear.

I guess it could be useful to, for better understanding, point out that horizontal and vertical refers to the OAV (looking at the sample towards the detector). But that is in principle already given as horizontal and vertical can only mean one thing as we have defined it to be movements in the y-z plane.

rhfogh: @marcus-oscarsson SOunds fine. In that case we can get rid of the 'approximate' and simply say they are horizontal and vertical pseudomotors. I would still prefer to leave the direction of motion out of it and give the axis directions as direction vectors in the configuration. That avoids messy questions about exactly how people set up their OAVs, or which direction we are looking at when we define them,

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
rhfogh: OK, now I understand what you mean. I would say that it is simpler and more general to give the direction vector for omega (defined as positive rotation), possibly with an additional Boolean to say whether the goniostat is above or below (left or right) of the sample. Vectors require fewer definitions.

marcus-oscarsson:
> I am basing this on the way Gleb set it up for CentringMath. The thing is that when you have to deal with multiple synchrotrons and multiple ways of mounting different goniostats it gets hard to have a set of motor names that fits well everywhere. Just a couple of problems:
> 
>     * the omega angle where sampx or sampy are along the beam could be any number, e.g. 23.47deg,  and vary over time,
> 
>     * Which motor to use for centring, phix or phiy, depends on how the goniostat is mounted.
> 
>     * So does anything you define as 'horizontal' or 'vertical'.
> 
>     * I am not sure whether you can rely on the positive movement or rotation direction always being the same relative to the diffractometer - certainly not across different models and makes.

This is exactly why we are agreeing on a common coordinate system definition so that the terms such as 'horizontal' and 'vertical' are well defined. I think we have more or less narrowed down on that definition now. With this definition 
motors such as `phi/alignment-x` or `phi/alignment-y` can not mean anything else. The same for the centering motors sampx and sampy.

This definition needs to be mapped to both the specific diffractometer implementation such as MicroDiffractometer.py or whatever implementation you have. The actual motors performing those movements may be called something else internally or can even by performed by several motors. However the motors we define in our convention need to be present and to respect the `AbstractDiffractometer` concept. 

These concepts will as you mention be used by the centering routine via the `SampleView` hardware object, but also when we move the sample around in the OAV. Where horizontal and vertical movements of the sample become necessary.

The centering routine  can use some other convention internally, but then needs to map to this. As you mention some sort of order could work but the easiest for most people would be to map the convention we have defined to whatever the sample centering routine uses. The absolutely easiest, in my opinion, is that the centering routine simply uses the convention we have defined. As long as its not some sort of external routine that we can't change.

> 
> 
> I think the best way to do it is to separate things out:
> 
>     * There is a single clear definition that describes a coordinate system, nothing more, linked to the beam and gravity so it is always clearly defined.

I think we are almost there with this one :)

> 
>     * Motor names are defined by how they are named in the hardware, which is always clear which motor has which name. People can interact with the motors by name when they now how they are set up locally, but the code does not depend on it.

Well, we have a convention definition that needs to be respected, `alignment-x,y,z` should always mean the same. The internals, as mentioned above, could then be hidden in a specific implementation.

> 
>     * You then have a mapping between the hardware motors and the goniostat set-up that tells you the direction of all motors at omega=0, and the order they are working in. That allows you to do all the centring calculations in the same way with the exact same code regardless of which motors are named what or how the goniostat is mounted (see CentringMath). It also means that you do not rely on assumptions about the exact relative angles of the motor directions, let alone about where sampx points at omega=0.

I think adding the order here make things a unnecessarily complicated, what would that order be. A mapping between the conventions would be easier to use (if they are not the same ?)

> 
>     * You can still have it so that you know which motors always point in which direction at the ESRF and work with that. It is just that instead of putting it in the basic definitions where things get complicated, it is a matter of putting the appropriate motors with their direction vectors in the ESRF-specific config file and telling your people about it.

Not sure I understand what you mean here.

> 
>     * As a side issue you get rid of the need for specifying 'direction' - the directions in the config file are the ones that define positive rotation.  The only thing you need in addition (and that only for shadowing calculations) is a single boolean telling you whether the omega rotation axis (positive rotation)  points towards the sample or towards the diffractometer.

I think we should define the directions, otherwise moving the sample across the beam with common control logic would be complicated. At some-point we would need to know what left and right is on the OAV, for instance.

beteva:
Dear @rhfogh, I think we are close to a consensus 🎆

> * Motor names are defined by how they are named in the hardware, which is always clear which motor has which name. People can interact with the motors by name when they now how they are set up locally, but the code does not depend on it.

Motor roles are important,  because they are used by their roles to pass the positions for the centring routines. In This is what we are trying to determine in a very pragmatic way. They can be named differently in the hardware, but configured by their roles in the way, corresponding to the convention.

> * As a side issue you get rid of the need for specifying 'direction' - the directions in the config file are the ones that define positive rotation.  The only thing you need in addition (and that only for shadowing calculations) is a single boolean telling you whether the omega rotation axis (positive rotation)  points towards the sample or towards the diffractometer.

This I do not quite understand. Unfortunately the motors can be mounted differently even for a similar equipment, so we need to define directions.

beteva (approved):

rhfogh:
@marcus-oscarsson @beteva
OK I see you have both approved this PR, so I shall merge it. We can then see about discussing further changes later as and when I may make another PR.

PR #1527, opened 2026-03-16
improve subdirectory sanitazation

walesch-yan:
The current `Session` object has a `get_full_paths` function, which merges the subdirectory entered by the user with the corresponding base image paths and processed data paths; the subdirectory is not properly sanitized before this merge happens.

One issue resulting from this: when a user starts the subdir folder with a slash, the returned paths completely ignore the base paths and return the subdirectory as a full path. The same happens if a user enters a double slash (//) inside their subdirectory path (ex., test//test-test, then Session returns /test-test as full_path. (Luckily, MXCuBE fails this in a later stage of the process, because it cannot recreate the subdir from the full path.))

This PR adds a subdirectory sanitization function to prevent the above scenario.

### DIFF
--- a/mxcubecore/HardwareObjects/ESRF/ESRFSession.py
+++ b/mxcubecore/HardwareObjects/ESRF/ESRFSession.py
@@ -1,4 +1,5 @@
 import glob
+import logging
 import os
 import time
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


mockoocy (approved):
Very nice! 🥳

woutdenolf:
@walesch-yan Using `Path.resolve()` will break anything downstream that uses this path from another machine at the ESRF due to the /data symbolic link.

Edit: solution in https://github.com/mxcube/mxcubecore/pull/1541.

PR #1528, opened 2026-03-17
Implement EPICSActuatorBluesky

pedrobenetton:
This PR implements the `EPICSActuatorBluesky` class, which will be used for devices that are moved with a bluesky plan, but have an RBV that can be read by MXCuBE as if it was a regular `EPICSActuator`. Because of that, only the `_set_value` function needs to be overwritten. Our energy will be moved via bluesky plans.

### DIFF
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -5,6 +5,7 @@ import gevent
 import numpy as np
 
 from mxcubecore.HardwareObjects.abstract.AbstractActuator import AbstractActuator
+from mxcubecore import HardwareRepository as HWR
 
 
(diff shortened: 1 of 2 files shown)


mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
```diff
@@ -74,3 +75,52 @@ def abort(self):
...
+    configuration:
+    tolerance: 0.01
+    plan_name: "move_energy_and_phase"
+    plan_parameter: "energy"
+    default_limits: (5, 20)
+    """
+
+    def __init__(self, name):
+        super().__init__(name)
+        self._bluesky_api = HWR.beamline.get_object_by_role("bluesky")
```
beteva, line 111: It is probably better to have this in the `init()`, rather than` __init__()`

beteva (approved):
Well done.

PR #1529, opened 2026-03-17
Implement LNLSResolution

pedrobenetton:
We are finishing the implementation of the UI components, and we understood that the resolution will need a specific class for itself instead of using `AbstractResolution`. The reasons are:

- Our energy component is an `EPICSActuatorBluesky` component and does not have a `get_wavelength` attribute
- We don't have a `_hwr_detector.distance` component: we use `EPICSMotor` and a configuration file.
- Instead of having `_hwr_detector`, we just need to add pixel_size_mm,  n_pixels_x and n_pixels_y to the yaml file of `ResolutionVirtualMotor`, and that is sufficient for calculating the resolution. Our beam center has its own EPICS PVs, that are also passed in the yaml file. MXCuBE doesn't need to know extra information about the detector because our data collection will be handled by bluesky plans.

### DIFF
--- a/mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
+++ b/mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
@@ -59,7 +59,9 @@ class ResolutionVirtualMotor(EPICSMotor):
         self.dx = self.n_pixels_x * self.pixel_size_mm
         self.dy = self.n_pixels_y * self.pixel_size_mm
         super().init()
-        self.connect(self.get_channel_object("wavelength_rbv"), "update", self.update_value)
+        self.connect(
+            self.get_channel_object("wavelength_rbv"), "update", self.update_value
+        )
 
     def get_limits(self):
         llm, hlm = super().get_limits()
--- a/mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
+++ b/mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
@@ -56,8 +56,7 @@ class ResolutionVirtualMotor(EPICSMotor):
         super().init()
 
     def get_limits(self):
-        llm = self.get_channel_value(self.MOTOR_LLM)
-        hlm = self.get_channel_value(self.MOTOR_HLM)
+        llm, hlm = super().get_limits()
         low_limit = self.distance_to_resolution(llm)
         high_limit = self.distance_to_resolution(hlm)
         return (low_limit, high_limit)
--- a/mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
+++ b/mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
@@ -25,19 +25,24 @@ class ResolutionVirtualMotor(EPICSMotor):
         "":
     "MX2:cam1:BeamX":
         channels:
-        beam_x:
-            suffix: ""
-            polling_period: 200
+            beam_x:
+                suffix: ""
+                polling_period: 200
     "MX2:cam1:BeamY":
         channels:
-        beam_y:
-            suffix: ""
-            polling_period: 200
+            beam_y:
+                suffix: ""
+                polling_period: 200
+    "MX2:cam1:Wavelength":
+        channels:
+            wavelength_rbv:
+                suffix: ""
+                polling_period: 200
     configuration:
-    tolerance: 0.01
-    pixel_size_mm: 0.172
-    n_pixels_x: 1475
-    n_pixels_y: 1679
+        tolerance: 0.01
+        pixel_size_mm: 0.172
+        n_pixels_x: 1475
+        n_pixels_y: 1679
     """
 
     BEAM_X_RBV = "beam_x"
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
```diff
@@ -0,0 +1,112 @@
...
+        self.pixel_size_mm = self.get_property("pixel_size_mm")
+        self.n_pixels_x = self.get_property("n_pixels_x")
+        self.n_pixels_y = self.get_property("n_pixels_y")
+        self.dx = self.n_pixels_x * self.pixel_size_mm
+        self.dy = self.n_pixels_y * self.pixel_size_mm
+        super().init()
+
+    def get_limits(self):
+        llm = self.get_channel_value(self.MOTOR_LLM)
+        hlm = self.get_channel_value(self.MOTOR_HLM)
```
beteva: Is this different from super().get_limits()?

beteva (commented):
I understand that you do not have the distance from the detector, but you do need the energy to calculate the wavelength and vice/versa. Did you decide to not display the energy and do you have only fixed energy beamlines?
What I do not see in your implementation is how you update the resolution, if you change the energy (if you do) and the distance motor position (this you certainly do). You can have a look at the init of the AbstractResolution to see what I mean. Or maybe you do not need it, because you do not display the distance, but only use it in an acquisition plan?

pedrobenetton:
> I understand that you do not have the distance from the detector, but you do need the energy to calculate the wavelength and vice/versa. Did you decide to not display the energy and do you have only fixed energy beamlines? What I do not see in your implementation is how you update the resolution, if you change the energy (if you do) and the distance motor position (this you certainly do). You can have a look at the init of the AbstractResolution to see what I mean. Or maybe you do not need it, because you do not display the distance, but only use it in an acquisition plan?

We do change our energy. We have an EPICS PV that does the energy/wavelength calculation and gives the wavelength directly. Because of that, I created an EPICSActuator component for wavelength and made it read-only. This is why I use:
`wavelength = self.wavelength.get_value()`

The way we change energy is with bluesky, in the other PR I created:
https://github.com/mxcube/mxcubecore/pull/1528

rhfogh:
@pedrobenetton  If I understand your comments right, that would mean that it is not possible to query or set resolution or detector position at LNLS the way it is done at all other beamlines. If that is true, it would mean that code that is written to work at multiple beamlines, including the GPhL workflow, could not work at LNLS. Did I get it right?

pedrobenetton:
> @pedrobenetton If I understand your comments right, that would mean that it is not possible to query or set resolution or detector position at LNLS the way it is done at all other beamlines. If that is true, it would mean that code that is written to work at multiple beamlines, including the GPhL workflow, could not work at LNLS. Did I get it right?

@rhfogh I took a quck look at `GphlWorkflow.py` and it looks like it would not work, because currently LNLS does not have `HWR.beamline.detector`. What we have is `HWR.beamline.detector_distance` and `HWR.beamline.resolution`.

rhfogh:
@pedrobenetton Well, if you ever want to try out the GPhL workflows, we would have to look at that.

pedrobenetton:
> @pedrobenetton Well, if you ever want to try out the GPhL workflows, we would have to look at that.

@rhfogh I understand. I think the main concern is that we intend to do all data collection (or any other device plan/orchestration) using bluesky, and the data processing using prefect.
I don't think it will be hard to implement an automation of GPHL Autoproc as a processing option directly in our prefect workflows, for example. It is something I have in mind to do in the future.

mxcubecore/HardwareObjects/LNLS/LNLSResolution.py
pedrobenetton: It isn't. I adjusted it, because using super().get_limits() is better

beteva:
> We do change our energy. We have an EPICS PV that does the energy/wavelength calculation and gives the wavelength directly.

This I understand and is perfectly fine. I am just afraid that you will be missing the connection between changing the resolution and changing the wavelength, so in the GUI you would not see it.

rhfogh:
@pedrobenetton I agree that the autoPROC processing can surely be implemented. Anyway that does not interact with MXCUBE, so all it needs is the images to process - my comments do not apply to that at all. What I was thinking of was the GPhL Experiment workflow that derives and runs acquisition strategies, and that _does_ interact with the hardware objects at runtime. To be sure you can live without it - while you absolutely cannot live without a functioning hardware  control system - but if you ever want to try out our workflow we would, as I said, have to look at all this

pedrobenetton:
> > We do change our energy. We have an EPICS PV that does the energy/wavelength calculation and gives the wavelength directly.
> 
> This I understand and is perfectly fine. I am just afraid that you will be missing the connection between changing the resolution and changing the wavelength, so in the GUI you would not see it.

That was correct. I used the connect function to also update resolution when wavelength changes.

beteva (approved):
Well done.

PR #1530, opened 2026-03-17
Trigger sample_info_changed when the state of a sample changes

marcus-oscarsson:
Trigger sample_info_changed when the state of a sample changes


beteva (approved):
Nice

PR #1534, opened 2026-03-18
EDNA characterisation error handling improvemnt

marcus-oscarsson:
Making sure that the exception stack trace is logged. I also took the opportunity to add some code that parses the EDNA results and can be used with some test input to generate a diffraction plan.


beteva (approved):
Thanks

PR #1537, opened 2026-03-23
Use ``msg`` to improve logging of unsuccesfull login to ISPYB

dominikatrojanowska:
`msg` was defined and never used.   
In case of `self.login_ok = False` for "ldap" type, the debug message is a bit wired, though: `ISPYB login not ok: <login_id>`

### DIFF
--- a/mxcubecore/HardwareObjects/ProposalTypeISPyBLims.py
+++ b/mxcubecore/HardwareObjects/ProposalTypeISPyBLims.py
@@ -42,19 +42,23 @@ class ProposalTypeISPyBLims(ISPyBAbstractLIMS):
         """
         Authentication step based on the authServerType
         """
-
+        msg = ""
         if self.authServerType == "ldap":
             self.log.debug("Starting LDAP authentication %s" % user_name)
             ok = self.ldap_login(user_name, psd)
         elif self.authServerType == "ispyb":
             self.log.debug("ISPyB login")
-            ok, _ = self.ispyb_login(user_name, psd)
+            ok, msg = self.ispyb_login(user_name, psd)
         else:
             raise Exception("Authentication server type is not defined")
 
         if not ok:
             # refuse Login
-            self.log.error("Authentication with %s failed" % self.authServerType)
+            self.log.error(
+                "Authentication with %s failed. %s",
+                self.authServerType,
+                msg.capitalize(),
+            )
             raise Exception("Authentication failed")
 
         self.log.debug("User %s logged in %s" % (user_name, self.authServerType))
--- a/mxcubecore/HardwareObjects/UserTypeISPyBLims.py
+++ b/mxcubecore/HardwareObjects/UserTypeISPyBLims.py
@@ -87,10 +87,10 @@ class UserTypeISPyBLims(ISPyBAbstractLIMS):
             login_name = self._translate(proposal_code, "ldap") + str(proposal_number)
 
         # Authentication
+        msg = ""
         if self.authServerType == "ldap":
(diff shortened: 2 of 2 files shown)


mockoocy (approved):

mxcubecore/HardwareObjects/UserTypeISPyBLims.py
```diff
@@ -99,9 +99,8 @@ def login(self, loginID, psd, ldap_connection=None) -> LimsSessionManager:
             raise Exception("Authentication server type is not defined")
 
         if not self.login_ok:
-            msg = "%s." % msg.capitalize()
-            # refuse Login
-            self.log.error("ISPyB login not ok")
+            # refuse login
+            self.log.error("ISPyB login not ok. %s", msg.capitalize())
```
marcus-oscarsson:
```suggestion
            self.log.error(f"ISPyB authentication for user {msg} failed.")
```

marcus-oscarsson: I also find the current message a bit short and odd, maybe the above is a bit nicer ?

dominikatrojanowska: I would say it's wired to set `msg=loginId` in ldap method and then adding `ISPYB` word into the error message regardless of ispyb/ldap method...  
I will propose something more generic then

marcus-oscarsson: We can perhaps add the method  as well then ```ISPyB authentication for user {msg}  with method: {method}failed```. The fact is that the method can at the moment either be LDAP or ISPyB.

mxcubecore/HardwareObjects/ProposalTypeISPyBLims.py
```diff
@@ -42,19 +42,19 @@ def _authenticate(self, user_name, psd):
...
             self.log.debug("ISPyB login")
-            ok, _ = self.ispyb_login(user_name, psd)
+            ok, msg = self.ispyb_login(user_name, psd)
         else:
             raise Exception("Authentication server type is not defined")
 
         if not ok:
             # refuse Login
-            self.log.error("Authentication with %s failed" % self.authServerType)
+            self.log.error("Authentication with %s failed. %s" % (self.authServerType, msg))
```
marcus-oscarsson: What will this print in the end ?

mxcubecore/HardwareObjects/UserTypeISPyBLims.py
mockoocy: I think this suggestion goes against Ruff :D 
https://docs.astral.sh/ruff/rules/logging-f-string/

marcus-oscarsson: We are ignoring that rule, or should be, in the global ignore list.

mxcubecore/HardwareObjects/ProposalTypeISPyBLims.py
dominikatrojanowska: `Authentication with ispyb failed. Invalid credentials.` where Invalid credentials is just one example, or: ` Authentication with ldap failed. ` where the message is always empty string as `ldap_login` seems to return single value

marcus-oscarsson: I would leave out the message (`msg`) and simply return; `Authentication with {method} failed for {user}.` If you really think its necessary to add the contents of `msg` I would add, `... reason: {msg}` at the end of the message. But then we should be sure that we know what `msg` could contain so that no sensitive information is logged.

PR #1539, opened 2026-03-25
MultiCollect for use with BLISS and Pilatus 4

marcus-oscarsson:
This PR adds a BLISS compatible `MutliCollect` HardwareObject that uses the Pilatus4 detector.

### DIFF
--- a/mxcubecore/HardwareObjects/LimaPilatus4Detector.py
+++ b/mxcubecore/HardwareObjects/LimaPilatus4Detector.py
@@ -39,7 +39,10 @@ class LimaPilatus4Detector(LimaEigerDetector):
         threshold = self.get_channel_object("threshold_energy").get_value()
         # set the other thresholds to the same value as 1
         for chn in range(2, 5, 1):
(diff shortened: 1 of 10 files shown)


mxcubecore/HardwareObjects/LimaPilatus4Detector.py
```diff
@@ -0,0 +1,46 @@
...
+        working_energy_chan = self.get_channel_object("photon_energy")
+        working_energy = working_energy_chan.get_value() / 1000.0
+        if math.fabs(working_energy - energy) > 0.1:
+            egy = int(energy * 1000.0)
+            working_energy_chan.set_value(egy)
+
+        threshold = self.get_channel_object("threshold_energy").get_value()
+        # set the other thresholds to the same value as 1
+        for chn in range(2, 5, 1):
+            url = f"http://lid30a1pilatus4dcu/detector/api/1.8.0/config/threshold/{chn}/energy"
```
beteva: We should see how to change this to not contain the url. I can take care of.

beteva (changes requested):

beteva (approved):

PR #1541, opened 2026-03-26
Do not resolve symlinks in path handling

marcus-oscarsson:
Replaced `path.resolve` with `normpath` and `path.absolute` so that paths are stripped from `./` `../` and absolute but symlinks are NOT resovled

### DIFF
--- a/mxcubecore/HardwareObjects/ESRF/ESRFSession.py
+++ b/mxcubecore/HardwareObjects/ESRF/ESRFSession.py
@@ -2,6 +2,7 @@ import glob
 import logging
 import os
 import time
+from pathlib import Path
 from typing import Tuple
 
 from mxcubecore import HardwareRepository as HWR
@@ -43,7 +44,7 @@ class ESRFSession(Session.Session):
 
         :returns: Tuple with the full path to image and processed data
         """
-        subdir = self._sanitize_subdir(subdir).resolve()
+        subdir = Path(self._sanitize_subdir(subdir))
 
         if not subdir.is_relative_to(self.get_base_image_directory()):
             error_message = "Invalid subdirectory"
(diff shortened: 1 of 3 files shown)


rhfogh (changes requested):
Good to make the change.
But if I read the Python documentation correctly,,  os.path.abspath will make absolute and remove '..' without touching symlinks - in a single operation. Quotes from Python docs:

[[3](https://docs.python.org/3/library/pathlib.html#id5)]
[os.path.abspath()](https://docs.python.org/3/library/os.path.html#os.path.abspath) removes “..” components without resolving symlinks, which may change the meaning of the path, whereas [Path.absolute()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.absolute) leaves any “..” components in the path.

 os.path.abspath(path)

    Return a normalized absolutized version of the pathname path. On most platforms, this is equivalent to calling normpath(join(os.getcwd(), path)).

marcus-oscarsson:
Yes, I guess using `os.path.abspath` might be faster. I replaced the instance where a `Path` object is not used down stream. For the rest of the cases `normpath` is used.

marcus-oscarsson:
ahh, RUFF my dear friend:  `PTH100 os.path.abspath() should be replaced by Path.resolve()` ... I would if they did the same thing. Soon, someone has to write a checker that checks the linting rules ...

marcus-oscarsson:
@matyson, @epanepucci : Maybe someone at MAXIV knows why but it seems like some MAIV tango related tests takes a extremely long time to complete ?

epanepucci:
> @matyson, @epanepucci : Maybe someone at MAXIV knows why but it seems like some MAIV tango related tests takes a extremely long time to complete ?

I haven't really paid attention, usually I go for a coffee after pushing something :-D
I'll ask the experts.

marcus-oscarsson:
It might have been a GitHub issue, seems more or less back to normal now :)

PR #1536, opened 2026-03-20
Sync TaskPrerequisites with ContextMenu

mockoocy:
Currently ContextMenu component on MXCuBE-Web side respects following prerequisites:
- point
- no_shape
- no_shape_2d
- line
- grid

The "grid" constant was missing from TaskPrerequisites. Also, the already existing "mesh" and "chip" prerequisites are nowhere to be found in MXCuBE-Web.

---

I will also note that the original list of values was derived from hardware objects in this repo, `mesh` and `chip` were present [there](https://github.com/mxcube/mxcubecore/blob/4bf73a083ed38793e7a04fb417bc2674dec2c148/mxcubecore/HardwareObjects/ESRF/queue_entry/fast_char_collection.py#L45), maybe they were used in the past?

PR #1542, opened 2026-03-26
Remove TaskPrerequisite.MESH from test collection.

mockoocy:
As of https://github.com/mxcube/mxcubecore/pull/1536 the MESH Enum item has been removed, as it is not used by the frontend application.

An alternative - reverting it's removal is not considered, since the MESH requirement would not differ substantially from already present GRID.

---

note: I've tested it by selecting `Test Collection` task from the context menu. Previously that option was not showing up in the UI.

### DIFF
--- a/mxcubecore/queue_entry/test_collection.py
+++ b/mxcubecore/queue_entry/test_collection.py
@@ -78,7 +78,6 @@ class TestCollectionQueueEntry(BaseQueueEntry):
         TaskPrerequisite.POINT,
         TaskPrerequisite.LINE,
         TaskPrerequisite.CHIP,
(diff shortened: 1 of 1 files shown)


mockoocy:
Also, I doubt that any site would be using `MESH` in their tasks. If it were to break someone's workflow, then it could be perhaps reverted?

The motivation behind the the #1536 PR was the confusion arousing in our dev team, once we've discovered that `MESH` is not doing anything.

marcus-oscarsson:
Thanks for the quick fix !

marcus-oscarsson (approved):

PR #1546, opened 2026-03-27
Minor fixes after move to AbstractDiffractometer

marcus-oscarsson:
This PR contains some changes necessary after the move to AbstractDiffractometer mostly concerning the split of diffractometer - sample view and method name changes:

   - run_script -> run_custom_script
   - move_motors -> set_value_motors

### DIFF
--- a/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
+++ b/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
@@ -233,6 +233,8 @@ class IdentifiedElement(MessageData):
             self._id = value
         else:
             self._id = uuid.uuid1()
+
+
 #            raise TypeError("UUID input must be of type uuid.UUID")
 
 
--- a/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
+++ b/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
@@ -232,7 +232,8 @@ class IdentifiedElement(MessageData):
         elif isinstance(value, uuid.UUID):
             self._id = value
         else:
-            raise TypeError("UUID input must be of type uuid.UUID")
+            self._id = uuid.uuid1()
+#            raise TypeError("UUID input must be of type uuid.UUID")
 
 
 # Sync with Java 4/5/2017
--- a/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
+++ b/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
@@ -232,10 +232,7 @@ class IdentifiedElement(MessageData):
         elif isinstance(value, uuid.UUID):
             self._id = value
         else:
-            self._id = uuid.uuid1()
-
-
-#            raise TypeError("UUID input must be of type uuid.UUID")
+            raise TypeError("UUID input must be of type uuid.UUID")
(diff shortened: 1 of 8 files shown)


mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
```diff
@@ -1135,7 +1135,7 @@ def query_collection_strategy(self, geometric_strategy):
             dose_label = "Characterisation dose (MGy)"
             if not self.config.settings.get("recentre_before_start"):
                 # replace planned orientation with current orientation
-                current_pos_dict = HWR.beamline.diffractometer.get_positions()
+                current_pos_dict = HWR.beamline.sample_view.get_positions()
```
marcus-oscarsson, line 1138: @rhfogh: we made these changes. You have likely already done these but this is what we had on ID30A1 and these changes does not seem to already be present. Let me know if you would like to keep them in this PR, or do something else

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
```diff
@@ -232,7 +232,8 @@ def __set_id(self, value):
         elif isinstance(value, uuid.UUID):
             self._id = value
         else:
-            raise TypeError("UUID input must be of type uuid.UUID")
+            self._id = uuid.uuid1()
```
marcus-oscarsson: @rhfogh, @olofsvensson: This change is not really related to this PR, but this is the only other GPHL related change I have and I thus decided to include it as we had a change to the GPHL workflows related to the AbstractDiffractometer as well.

So we found that we needed sort of hack this to get things running, maybe this is not needed any longer. Let me me know if it can be removed ?

mxcubecore/HardwareObjects/Gphl/GphlWorkflow.py
rhfogh: I cannot find those changes in my code, but they are definitely needed. Thanks. for taking care of it.

marcus-oscarsson: Ill keep them then :)

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
rhfogh: This is really not a change I like, as it happens. The way the code should work, if you pass in a uuid it should be set, if you pass in None it should generate a new uuid, and if you pass in an invalid value it should throw an error. With the change invalid values are silently ignored and replace with a new uuid, potentially messing up what people think are the object identifiers. What _should_ happen is that you throw a TypeError, and the calling code makes sure that it passes in correct values. So I think this change should be removed, and if that means you get errors they should be fixed in another way.

rhfogh (changes requested):
OK apart from the change with uuid handling in GphlMessages

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
marcus-oscarsson: Ok, lets remove it

rhfogh (approved):
Excellent!

PR #1548, opened 2026-03-27
Added the possiblity to use soft interlcok for n-state devices such as the back light

marcus-oscarsson:
The MD does not interlock the back-light in data collection phase which allows the user to move the back light in while collecting.

This PR adds a simple why of disabling a n-state device such as a backlight so that it can be enabled/disabled during data collection.

### DIFF
--- a/mxcubecore/HardwareObjects/ESRF/MD2MultiCollect.py
+++ b/mxcubecore/HardwareObjects/ESRF/MD2MultiCollect.py
@@ -70,6 +70,9 @@ class MD2MultiCollect(ESRFMultiCollect):
         diffr.frontlightswitch.set_value(diffr.frontlightswitch.VALUES.IN)
         # diffr.frontlight.set_value(2)
 
+        back_light_switch = diffr.get_object_by_role("BackLightSwitch")
+        back_light_switch.disable()
+
         # move to DataCollection phase
         logging.getLogger("user_level_log").info("Moving MD2 to DataCollection")
         # AB next line to speed up the data collection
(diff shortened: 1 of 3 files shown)


mxcubecore/HardwareObjects/ESRF/MD2MultiCollect.py
```diff
@@ -70,6 +70,9 @@ def do_prepare_oscillation(self, *args, **kwargs):
         diffr.frontlightswitch.set_value(diffr.frontlightswitch.VALUES.IN)
         # diffr.frontlight.set_value(2)
 
+        back_light_switch = diffr.get_object_by_role("BackLightSwitch")
+        back_light_switch.disable()
+
```
beteva:
```suggestion
        diffr.backlightswitch.disable()
```
No need to get the object, it is already part of the NState objects of the diffractometer.

mxcubecore/HardwareObjects/ESRF/MD2MultiCollect.py
```diff
@@ -79,6 +82,7 @@ def do_prepare_oscillation(self, *args, **kwargs):
     def data_collection_cleanup(self):
         HWR.beamline.diffractometer.wait_status_ready(10)
         self.close_fast_shutter()
+        HWR.beamline.diffractometer.get_object_by_role("BackLightSwitch").enable()
```
beteva:
```suggestion
        HWR.beamline.diffractometer.backlightswitch.enable()
```
As above

beteva (approved):

PR #1552, opened 2026-03-30
Make SampleView compatible with UP/DOWN diffractometer.

beteva:
Some changes to make the centring and the move to beam compatible with both horizontally and vertically oriented goniometer.
Add super class initialisation in the init method of Abstract and Micro Diffractometer.


marcus-oscarsson (approved):

marcus-oscarsson:
Awesome, thanks @beteva

PR #1555, opened 2026-04-01
Changing all occurrences of `centering` to `centring`

marcus-oscarsson:
Changing all occurrences of `centering` to `centring`, including file names


marcus-oscarsson:
Dear all, we have once said that we should use British English and not US. We found quite a lot of occurrences of centering instead of centring. I made this PR to replace all those.  

Let me know what you think, there where also some files specifically `xray_centering.py`, @rhfogh, @olofsvensson

rhfogh (approved):
Worth doing, and thanks for taking the time. 

Looking at the page I'd say we now need to muzzle SonarCloud just as much as we need muzzling ruff. See #1562 for discussion - this is not about the changes in the PR.

olofsvensson:
I agree with @rhfogh , it's a good idea. We would need to do the same on the BES (workflow) side.

marcus-oscarsson:
@olofsvensson: Do you want that we prepare a branch with the fixes on the WF side and test this on a beamline before merging ?

olofsvensson:
@marcus-oscarsson I had a look through the changes in this MR and they are ok with the current status of the BES workflows. The name of the workflows in the BES already have the correct spelling:

https://gitlab.esrf.fr/sb/bes/-/blob/main/src/bes/flows/SmallXrayCentring.json?ref_type=heads

and

https://gitlab.esrf.fr/sb/bes/-/blob/main/src/bes/flows/XrayCentring.json?ref_type=heads

The spelling inside the BES is inconsistent and I'll fix it, but it's not affected by this MR.

Therefore you can go ahead and merge it any time.

marcus-oscarsson:
Excellent, @olofsvensson. Thanks for the update

marcus-oscarsson:
Ok pay attention, I'm merging this one !

PR #1557, opened 2026-04-01
[ESRFSession] - Removed check that will always end up with an exception

marcus-oscarsson:
Removed fix that will always end up with an exception. The statement `subdir.is_relative_to(self.get_base_image_directory()):` will always raise as `subdir` is simply a string without the root/base `directory`


beteva (approved):
Thanks @marcus-oscarsson

PR #1560, opened 2026-04-07
Implement EpicsToggle

pedrobenetton:
This PR implements EPICSToggle, a class used for EPICSNState devices that always receive the same input when switching between its states. It will be used for our safety shutter:

image

### DIFF
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
@@ -119,3 +119,38 @@ class EPICSNStateInterval(EPICSNState):
             return
         enum_value = values_list[0] if value > values_list[1].value else values_list[1]
         super().update_value(enum_value)
+
+class EPICSToggle(EPICSNState):
+    """
+    This class is a workaround for devices that exist in a discrete number of states
+    but take always the same input when switching between these states.
+    Example: a safety shutter with RBVs 0 for open and 1 for closed, but input value
+    must always be 1 for either opening/closing.
+
+    YAML Example
+    ------------
+
+    %YAML 1.2
+    ---
+    class: LNLS.EPICS.EPICSNState.EPICSToggle
+    epics:
+    "MNC:A:PPS01:":
+        channels:
+            rbv:
+                suffix: 'PG_STATUS'
+                polling_period: 200
+            val:
+                suffix: 'OEAOPENCLOSE'
+    configuration:
+        default_limits: (0, 1)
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/LNLS/EPICS/EPICSNState.py
```diff
@@ -119,3 +119,39 @@ def update_value(self, value=None) -> None:
...
+        input_value: 1
+    """
+
+    def init(self):
+        super().init()
+        self.input_value = self.get_property("input_value")
+
+    def _set_value(self, value):
+        value = self.input_value
+        super()._set_value(value)
```
beteva, line 157: Do you pass via value just to make ruff happy 😄 
Can you, instead of the input_value use the self.VALUES.Closed.value?

pedrobenetton: Yes... Is there a workaround for ruff? The whole point is that I am not going to use value.

I would rather keep input_value defined in configuration file, because this will make the class more generic and reusable for other cases (e.g. another component where input_value must be 1 but self.VALUES.Closed.value is 0 and self.VALUES.Open.value is 1).

beteva, line 157:
```suggestion
    def _set_value(self, value):  # noqa: ARG002
        super()._set_value(self.input_value)
```
There are two ways of making a work around which makes ruff ignore the rule (any rule):
- either you use the # noqa: rule_number in your code - this is good if you are offending one or two rules only
-  or in the ruff.toml file you can add your HO and the rules you want to exclude.

beteva (approved):
Once all the test have passed, you may merge. 
Out of curiosity, how this is used if you always set the same value? Do you never open the shutter from the GUI?

pedrobenetton:
> Once all the test have passed, you may merge. Out of curiosity, how this is used if you always set the same value? Do you never open the shutter from the GUI?

It works because the setpoint PV works as a signal to the hardware. It is always at 0, and when it is set to 1, it changes the state of the shutter based on the value of the readback PV (if it was open it closes, and vice-versa). After that, the setpoint PV immediatly goes back to 0 again.
The readback PV works as an Nstate device: 1 for closed and 0 for open.

beteva:
@pedrobenetton, have you seen my reply for the ruff rules and the suggestion?

pedrobenetton:
> @pedrobenetton, have you seen my reply for the ruff rules and the suggestion?

Sorry, I missed that, i will open a new PR

PR #1566, opened 2026-04-13
Removing obsolete dict for resolution - upper and lower values

### LINKED ISSUE #1565: MX_resolution value
`MX_resolution` is expected to be a float, but in some cases it is a json like

```
{
    "upper": 2
},
```

From @olofsvensson  @demariaa
```
The 'MX_resolution' is coming from here:
https://github.com/mxcube/mxcubecore/blob/44c49a7ac730d80e2d2771e5b52096a572248e28/mxcubecore/HardwareObjects/ICATLIMS.py#L1263

Marcus, Antonia and Yan, can you see where the "upper" value in "resolution" is coming from?
```

marcus-oscarsson:
The resolution values have for historical reasons been passed in a dictionary with the following format

```
{
"upper": x,
"lower": y,
}
```

However, in practice only `data_collect_parameters["resolution"]["upper"]` is used. Replacing all occurrences of `["upper"]` or similair with  `data_collect_parameters["resolution"]`

closes #1565

### DIFF
--- a/mxcubecore/HardwareObjects/Gphl/CollectEmulator.py
+++ b/mxcubecore/HardwareObjects/Gphl/CollectEmulator.py
@@ -155,12 +155,9 @@ class CollectEmulator(CollectMockup):
         # get resolution limit and detector distance
         detector_distance = data_collect_parameters.get("detector_distance", 0.0)
         if not detector_distance:
-            dd = data_collect_parameters.get("resolution")
-            # resolution may not be set - if so you should take the current value
-            if dd:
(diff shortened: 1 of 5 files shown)


rhfogh:
Good initiative - but I think you missed one.
AbstractCollect, line 239:
                resolution = self.current_dc_parameters.get("resolution").get("upper")

rhfogh (changes requested):
Approved once you fix the missing one,

marcus-oscarsson:
Ah, forgot to add that file :), done

Issue #1562, opened 2026-04-09
Review linting strategy


beenje:
After some internal discussion at MAX IV I'm creating this issue to start the discussion about the linting strategy with `ruff` in mxcube.

I wasn't here when the previous decisions were made, so I might be missing something, but I think I understand the reasoning:
- select `ALL` rules by default
- add ignore per file as it's impossible to fix everything in one go
- when people work on a file, try to fix some issues and remove some ignored rules
- bit by bit the whole project will improve

After a few months, the feedback from MAX IV is that in practice that doesn't really work.
People tend to add more ignore than removing them.

I think it's because this is adding too much overhead and extra work:

- there are too many rules, some quite opinionated like [SIM108](if-else-block-instead-of-if-exp), which don't bring much benefit 
- the current `ruff.toml` is huge and not easy to maintain - again too many rules and ignore
- removing some ignore and fixing them is quite some work and the benefit not always obvious

I'd propose to move to a new strategy:
1. agree on a very limited set of rules and apply them everywhere (meaning no ignore by default - there can of course be exception).
1. later, if we see another rule that is useful, add it and apply it on the full codebase

That will make the code more consistent as right now there are many different rules per file.

The default rules used by ruff would be a good start: `["E4", "E7", "E9", "F"]`. We should be able to easily add isort `I` (and maybe some that were already applied?).
This is what we did in the [sardana project](https://gitlab.com/sardana-org/sardana/-/blob/develop/ruff.toml).
At MAX IV, in most projects we also enable flake8-bugbear (`B`): `select = ["E4", "E7", "E9", "F", "B", "I"]`, but we could start without.

This is just a proposal.
What is the community opinion?
Are you happy with the current rules? Or do you have another approach in mind?

rhfogh:
I strongly support this proposal.

One drawback of the 'select ALL rules by default' approach is that some of the rules actually make things worse, and quite a few are arbitrary and do not give much improvement.

Another is that the approach has promoted a working style where some people spend energy silencing ruff errors _without considering the actual code quality_! There was for instance the rule that said that unused function parameters caused ruff errors, unless they began with underscore. And someone spent a lot of energy adding those underscores 1) without matching function signatures. 2) without looking at the real problem, which is that unused function parameters are often an indicator of bad code that should be fixed, not swept under the carpet. 

[...]

PR #1574, opened 2026-04-16
Updated ruff.toml

marcus-oscarsson:
This PR updates the current `ruff.toml` with a restricted set of rules instead as we have done until now selecting `ALL` and blacklisting the rules on a case by case basis.

This PR selects the rules that we have discussed in https://github.com/mxcube/mxcubecore/issues/1562, these may need further refinement after discussion but I understand it as we are starting to reach a set of baseline rules.

We have to continue to discuss and come up with a plan for

- How to address the files that are currently in the per file ignore list 
- Remove the `noqa` statements

Ive added the script that I used to modify the current `ruff.toml` in `utils`

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -47,8 +47,6 @@ convention = "google"
     "F841",
     "S307",
 ]
-"mxcubecore/Command/Exporter.py" = []
-"mxcubecore/Command/Mockup.py" = []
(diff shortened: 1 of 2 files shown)


rhfogh (approved):
FWIW (I am not the most ruffy guy around) this looks good.

PR #1584, opened 2026-04-20
line and mesh scan for the MD3 (up) version of the diffractometer.

beteva:
The line and the mesh scan take parameters in different order for the startScan4DEx and startRasterScanEx commands.


beteva:
Replaced by #1615

PR #1585, opened 2026-04-21
Fixed issue with resolution beeing removed from data collection parameters

marcus-oscarsson:
While removing resolution["upper"] some other handling also changed which removed resolution from the parameters all together which is obviously wrong.

Fixed issue with resolution being removed from data collection parameters and removed some sort of default parameter handling that also seems incorrect.

### DIFF
--- a/mxcubecore/model/queue_model_objects.py
+++ b/mxcubecore/model/queue_model_objects.py
@@ -2759,17 +2759,6 @@ def to_collect_dict(data_collection, sample, centred_pos=None):
         }
     ]
 
-    # NBNB HACK. These start life as default values, and you do NOT want to keep
-    # resetting the beamline to the current value,
-    # as this causes unnecessary hardware activities
-    # So remove them altogether if the value is (was explicitly set to)  None or 0
-    dd = result[0]
-    for tag in ("detector_distance", "energy", "transmission"):
-        if tag in dd and not dd[tag]:
-            del dd[tag]
-    resolution = dd.get("resolution")
-    if resolution is not None:
-        del dd["resolution"]
     return result
 
 
--- a/mxcubecore/HardwareObjects/abstract/AbstractCollect.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractCollect.py
@@ -225,33 +225,25 @@ class AbstractCollect(HardwareObject, object):
             # ----------------------------------------------------------------
             # Set data collection parameters
 
-            if "transmission" in self.current_dc_parameters:
-                log.info(
-                    "Collection: Setting transmission to %.2f",
-                    self.current_dc_parameters["transmission"],
-                )
-                self.set_transmission(self.current_dc_parameters["transmission"])
-
+            transmission = self.current_dc_parameters.get("transmission")
             wavelength = self.current_dc_parameters.get("wavelength")
             energy = self.current_dc_parameters.get("energy")
             detector_distance = self.current_dc_parameters.get("detector_distance")
-            try:
-                resolution = self.current_dc_parameters.get("resolution")
-            except AttributeError:
-                resolution = None
+            resolution = self.current_dc_parameters.get("resolution")
+
+            if transmission:
+                log.info(
+                    "Collection: Setting transmission to %.2f",
+                    transmission,
+                )
+                self.set_transmission(transmission)
 
             if wavelength:
                 # Wavelength (not having a default) overrides energy
-                log.info(
-                    "Collection: Setting wavelength to %.4f",
-                    self.current_dc_parameters["wavelength"],
-                )
+                log.info("Collection: Setting wavelength to %.4f", wavelength)
                 energy = HWR.beamline.energy.calculate_energy(wavelength)
             elif energy:
-                log.info(
-                    "Collection: Setting energy to %.4f",
-                    self.current_dc_parameters["energy"],
-                )
+                log.info("Collection: Setting energy to %.4f", energy)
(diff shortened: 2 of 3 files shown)


mxcubecore/model/queue_model_objects.py
```diff
@@ -2759,17 +2759,6 @@ def to_collect_dict(data_collection, sample, centred_pos=None):
...
-    # NBNB HACK. These start life as default values, and you do NOT want to keep
-    # resetting the beamline to the current value,
-    # as this causes unnecessary hardware activities
-    # So remove them altogether if the value is (was explicitly set to)  None or 0
-    dd = result[0]
-    for tag in ("detector_distance", "energy", "transmission"):
-        if tag in dd and not dd[tag]:
-            del dd[tag]
-    resolution = dd.get("resolution")
-    if resolution is not None:
```
marcus-oscarsson, line 2771: The line causing the issue is this line

mxcubecore/model/queue_model_objects.py
```diff
@@ -2759,17 +2759,6 @@ def to_collect_dict(data_collection, sample, centred_pos=None):
...
     ]
 
-    # NBNB HACK. These start life as default values, and you do NOT want to keep
-    # resetting the beamline to the current value,
-    # as this causes unnecessary hardware activities
-    # So remove them altogether if the value is (was explicitly set to)  None or 0
-    dd = result[0]
-    for tag in ("detector_distance", "energy", "transmission"):
-        if tag in dd and not dd[tag]:
-            del dd[tag]
```
mockoocy, line 2769: I wonder how safe is it to remove that code.

By quick inspection, these are set to `0.0` in case there was error reading either one of them. I am not aware if it is possible to have such a value otherwise (maybe through dc form, but that is still a bogus value then).

If that's the case I am perfectly fine with removing this hacky code. Especially that it is not the place to handle that logic anyway.

Also, I am not even sure this code prevent any hardware movements? I'd imagine that trying to put `0.0` value for detector distance or energy would surely yield an error.

marcus-oscarsson: I had a think about it since I made the PR and the line that I actually wanted to remove was the one I commented on above, related to `resolution`. That is clearly an issue.

I think you are mostly right and I agree that its not the place to handle an incorrect value, that is also why I decided to remove this piece of code in the first place.

The way that it was working is that if the parameter is not present the code that performs the collection will not use that parameter hence not move any instrument, but rather use the current value.

The code in `AbstractMutiCollect` would need to be updated so that check for a "valid" value is used instead of checking if the parameter is present. I think that @beteva was looking at this ?

We can keep the # NBNB Hack code for the time being, but I would be great if we could get rid of this code. 

Thinking about it, it don't see why transmission, energy or resolution, would ever be passed as None or 0 ?

beteva: > Thinking about it, it don't see why transmission, energy or resolution, would ever be passed as None or 0 ?

By the nature of the beamlines and the experiments, they should neither be None nor 0. Those are clearly invalid values, which makes the code even more important to remove.

mockoocy: I would also like to see this code removed here. Just wanted to put more focus on it, as it seems brittle and scary to remove :grin:

mockoocy (approved):

mxcubecore/model/queue_model_objects.py
marcus-oscarsson: I agree, at least not from a UI perspective but perhaps from a workflow perspective, @rhfogh @olofsvensson. I'm sure there is a reasonable explanation for why this code is present

olofsvensson: First time I see this code... It seems to have been added in this commit:

https://github.com/mxcube/mxcubecore/commit/ad8c79b7c266d8ee7669088a6b224cd2b61094ba

marcus-oscarsson: Thanks @olofsvensson, @rhfogh: Do you think we can remove this code and perhaps move it elsewhere if still needed ?

rhfogh:
Could you wait with merging this while I check it out? From memory this was not a matter of fixing incorrect values, but a deliberate way to avoid resetting the hardware if you were doing several scans after one another, and the values had not changed. So you set resolution to 0 as a way of saying 'just stick to the current value'. But I'll try to figure out if that has been superseded by now.

rhfogh (changes requested):
Looking at the code it seems that AbstractCollect.do_collect will - correctly - ignore wavelength, energy, detector_distance or resolution if they are either missing or set to zero. That being so there should be no need for this code, and it can be removed (I also did not find any case where these were set to zero, but maybe I did not look enough, and anyway this should be a legitimate operation.)
So, yes, by all means remove it, and if that unexpectedly breaks something we shall just have to deal with it. 
To be on the safe side I think we should change the code in AbstractCollect.do_collect so that it will do nothing if the transmission is set to 0, to put it in line with the other parameters.

marcus-oscarsson:
@rhfogh would you mind pushing those changes to this branch or make a PR to this branch, or did I misunderstand you are you fine with merging this ?

marcus-oscarsson:
I made the necessary changes to `AbstractMultiCollect` because thats the collect routine we are using. If you want to make a change to `AbstractCollect`, please go-ahead or let me know if I can merge ?

rhfogh:
I made a PR to resolution-fix that should be handled before this one.

marcus-oscarsson:
Thanks for the input everyone and thanks for the PR @rhfogh

PR #1588, opened 2026-04-22
Make sure wait_status_ready is only called if the timeout is not 0.

beteva:
Add check to avoid executing wait_status_ready if not necessary.


mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -341,7 +341,8 @@ def do_oscillation_scan(
         self._exporter.write_property("ScanNumberOfFrames", number_of_images)
         scan_params = f"1\t{start:0.3f}\t{(end - start):0.3f}\t{exptime:0.3f}\t1"
         self._exporter.execute("startScanEx", (scan_params,))
-        self.wait_status_ready(timeout)
+        if timeout != 0:
+            self.wait_status_ready(timeout)
```
marcus-oscarsson: Could we not just move this down to wait_status_ready so that 0 means `"no wait"`:

Basically: 
```
timeout = None   # Wait forever
timeout = 0           # No timeout
timeout=x             # (where X a float > 0),  timeout is X
```

beteva: > Could we not just move this down to wait_status_ready


Well, I was considering it, but than decided to be  **in line with the BaseHardwareObjects** wait_ready, where there is no check for 0 timeout. Maybe this is for semantic reasons - waiting for some time, function not called when no wait 😄 
Both approaches are good, I will do whatever the majority says.

rhfogh:
It is not crucial, but I would vote for doing it in with wait_status_ready, like @marcus-oscarsson proposed. It is less code around, and it is potentially confusing that timeout=None is a significant value, but timeout=0 skips the function.

Actually in wait_ready you ultimate end up calling gevent.event.wait(timeout-timeout), and if timeout=0 that should return immediately anyway. Is there any special benefit in calling gevent.event.wait(timeout=0)? Unless this is the case I would propose putting the special handling of timeout==0 in both wait_ready and wait_status_ready, just to save on a function call.

beteva:
> if timeout=0 that should return immediately anyway. Is there any special benefit in calling gevent.event.wait(timeout=0)? 

In the methods that specify timeout and use it as a parameter for the wait_status_ready, the timeout = 0 would correspond to an action to be executed, but not waiting for it to finish. Now if we pass timeout=0, but the action has finished (e.g. nothing moves as already in position). than we end with a timeout error. That's why we need to check if the polling in the wait_status_ready has to be executed.
```
with Timeout(0, RuntimeError("Timeout waiting for status ready")):
   while not True:
```
This returns immediately.
```
with Timeout(0, RuntimeError("Timeout waiting for status ready")):
  while not False:
```
This returns RuntimeError: Timeout waiting for status ready. This corresponds to the case where the action has finished,  before the call to wait_status_ready.

Now whether the test for 0 should be implemented in wait_status_ready (and wait_ready) or before calling wait_status_ready (wait_ready) for me makes no difference. As I said, it is up to the community to decide.
I personally prefer the way @marcus-oscarsson suggested - check in the wait method. Simply I would advocate for both wait_ready and wait_status_ready (and whatever other wait method is needed) to be consistent.

rhfogh:
@beteva That makes perfect sense. Thanks. Looks like we all agree on doing it in the wait_... function.

mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -537,7 +539,8 @@ def get_beam_position(self) -> tuple:
     def run_custom_script(self, script_cmd: str, timeout: float | None = None):
         """Run custom script."""
         self.run_script(script_cmd)
-        # Wait for script to start before checking status,
-        # can perhaps be improved to wait for ready-busy-ready ?
-        time.sleep(0.4)
-        self.wait_status_ready(timeout)
+        if timeout != 0:
```
marcus-oscarsson, line 542: Do we need this if ?

beteva: The custom scripts need some time to change the application status. This is just to be sure that we do not have racing conditions.

marcus-oscarsson (approved):

PR #1589, opened 2026-04-22
Resolution fix

rhfogh:
Rebased on develop, made changed as discussed, and cleaned up parameter value handling for log messages


marcus-oscarsson:
Thanks @rhfogh, wonderful !

PR #1590, opened 2026-04-22
Lnls resolution adjustment

pedrobenetton:
This PR deletes the LNLSResolution class. This is done because we realized it would be better to create a Soft-IOC at EPICS level rather than having the resolution/wavelength/energy/detector_distance logic implemented at every application. Because of that, now our resolution is simply an EPICSMotor class.
However, `EnergyAdapter` tries to call `HWR.beamline.resolution.get_limits_for_wavelength`  at `get_resolution_limits_for_energy` when data collection is about to start. Because of that, we implemented the class at our `EPICSMotor.py` file just to cover this case.


marcus-oscarsson:
Nice, do you think that, at some stage, it would be a good idea to move the `EPICS` folder to the root of `HardwareObjects` ? Like this others using EPICS could use the same, or is the EPICS objects very LNLS specific somehow ?

marcus-oscarsson (approved):

marcus-oscarsson:
You can rebase and merge ! :)

beteva (approved):
Well done

pedrobenetton:
> Nice, do you think that, at some stage, it would be a good idea to move the `EPICS` folder to the root of `HardwareObjects` ? Like this others using EPICS could use the same, or is the EPICS objects very LNLS specific somehow ?

For now we are still testing if everything is alright, but yes in the future we certainly want to do that

PR #1593, opened 2026-04-24
Replaced 'overlap' with '-offset'. First pass

rhfogh:
This is a first pass of replacing overlap with o0ffset, essentially a renaming and taking offset == -overlap.
The following files have been skipped as being probably coupled to code outside MXCuBE:
EMBLCollect, EMBLXrayImaging, ESRFMetadataManagerClient, ICATLIMS, XSDataMXCuBEv1_3, XSDataMXCuBEv1_4, XSDataMXv1.
The EMBL files coudl likely be ignored until such a time as P14 again cyncs back to the mainstream, but the rest of these files would have to be addressed by someone who understands them before we move ahead with the PR. 

The rest shoudl not be a problem:

Config files for embl-hh and config files in mxcubecore.configuration have also been skipped.
Changes in configuration files should be simple to carry out, but as the config files in the repositories have little connection to the ones actually in use it seemed simpler to leave those changes to the individual beamlines.

There should maybe be a consequential change in MXLIMS as well, but I can handle that easily enough.

### DIFF
--- a/mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
+++ b/mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
@@ -318,8 +318,7 @@ class PX2Collect(AbstractCollect, HardwareObject):
                 self.store_image_in_lims_by_frame_num(last_frame)
             if (
                 self.current_dc_parameters["experiment_type"] in ("OSC", "Helical")
-                and self.current_dc_parameters["oscillation_sequence"][0]["offset"]
-                == 0
+                and self.current_dc_parameters["oscillation_sequence"][0]["offset"] == 0
                 and last_frame > 19
             ):
                 self.trigger_auto_processing("after", self.current_dc_parameters, 0)
--- a/mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
+++ b/mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
@@ -138,29 +138,20 @@ class PX2Collect(AbstractCollect, HardwareObject):
 
         osc_seq = parameters["oscillation_sequence"][0]
         fileinfo = parameters["fileinfo"]
-        sample_reference = parameters["sample_reference"]
         experiment_type = parameters["experiment_type"]
         energy = parameters["energy"]
         transmission = parameters["transmission"]
         resolution = parameters["resolution"]
 
         exposure_time = osc_seq["exposure_time"]
-        in_queue = parameters["in_queue"] != False
 
-        overlap = osc_seq["overlap"]
         angle_per_frame = osc_seq["range"]
         scan_start_angle = osc_seq["start"]
         number_of_images = osc_seq["number_of_images"]
         image_nr_start = osc_seq["start_image_number"]
 
         directory = fileinfo["directory"]
-        prefix = fileinfo["prefix"]
         template = fileinfo["template"]
-        run_number = fileinfo["run_number"]
-        process_directory = fileinfo["process_directory"]
-
-        # space_group = str(sample_reference['space_group'])
-        # unit_cell = list(eval(sample_reference['cell']))
 
         self.emit("collectStarted", (self.owner, 1))
         self.emit("progressInit", ("Data collection", 100))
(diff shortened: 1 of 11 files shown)


marcus-oscarsson:
Thanks @rhfogh, for the PR. I think we need to test this on the beamlines before mergning. It will take a few days until I can get the time. 

@olofsvensson: Would this change be alright on the ESRF WF side, to replace `overlap` with `offset` ?

olofsvensson:
Hi @marcus-oscarsson , yes, no problem from my side. I just checked and "overlap" is not used extensively in the BES so it should be straight forward to change it to "offset".

One has though be aware that "overlap" is is used in ISPyB and I think we will not change ISPyB since we will soon phase it out. Overlap is also used in the ESRF data portal but it would be straight-forward to change it to "offset".

mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
```diff
@@ -138,29 +138,20 @@ def data_collection_hook(self):
 
         osc_seq = parameters["oscillation_sequence"][0]
         fileinfo = parameters["fileinfo"]
-        sample_reference = parameters["sample_reference"]
```
marcus-oscarsson, line 141: Why is this being removed, I guess its not being used ?

mxcubecore/model/common.py
```diff
@@ -27,7 +27,6 @@ class LegacyParameters(BaseModel):
     take_dark_current: int
     inverse_beam: bool
     num_passes: int
-    overlap: float
```
marcus-oscarsson, line 30: This should then be replaced by offset ?, we don't know if this is used or not at this stage. So better keep rename it to offset

mxcubecore/HardwareObjects/abstract/ISPyBValueFactory.py
```diff
@@ -205,11 +205,11 @@ def from_data_collect_parameters(ws_client, mx_collect_dict):
             data_collection.axisStart = osc_seq["start"]
 
             data_collection.axisEnd = float(osc_seq["start"]) + (
-                float(osc_seq["range"]) - float(osc_seq["overlap"])
+                float(osc_seq["range"]) + float(osc_seq["offset"])
             ) * float(osc_seq["number_of_images"])
 
             data_collection.axisRange = osc_seq["range"]
-            data_collection.overlap = osc_seq["overlap"]
+            data_collection.offset = osc_seq["offset"]
```
marcus-oscarsson:
```suggestion
            data_collection.overlap = osc_seq["offset"]
```

marcus-oscarsson: `data_collection` here is the ISPyB data model that is not being changed so it needs to be ` data_collection.overlap = osc_seq["offset"]`

marcus-oscarsson:
There seems to be some unrelated changes that I presume is "dead" code but its difficult to know when scanning the code for potential issues with `offset` vs `overlap`. Could I suggest to do the "dead" code removal in a separate PR.

mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
```diff
@@ -267,22 +256,6 @@ def data_collection_hook(self):
             )
             experiment.execute()
 
-        # for image in range(number_of_images):
```
marcus-oscarsson, line 270: Ok, this is clearly not used

mxcubecore/model/common.py
rhfogh: I have now replaced it by 'offset', but I do not understand what this does. I had assumed that these were parameters that were not used and should be ignored - and 'offset'  is as of now a parameter that is used and should not be ignored. But maybe it means something different in context? I shall leave it to you to decide.

mxcubecore/HardwareObjects/SOLEIL/PX2/PX2Collect.py
rhfogh: None of these are being used + but I put them back in for clarity

mxcubecore/HardwareObjects/abstract/ISPyBValueFactory.py
rhfogh: That should be '- osc_seq["offset"]'. I have fixed it.

rhfogh:
@marcus-oscarsson I have reinstated the dead code. Some other time.

olofsvensson:
@marcus-oscarsson , please let me know when you plan to deploy these changes because the deploy must be synchronised with a similar deploy on the BES side.

marcus-oscarsson:
Yes ofcourse, we also need to test this specifically. Let me know when you have made the change on the BES side and Ill try to get some time to test next week or so.

mxcubecore/model/common.py
marcus-oscarsson: Your right that `offset` should be moved from `LegacyParameters` as it is indeed used routinely as was `overlap`. It was just tucked into the `LegacyParameters`.

marcus-oscarsson:
This will require that the same is done on the UI sides, I will make PR for the `mxcubeweb` part

marcus-oscarsson (approved):

PR #1596, opened 2026-04-28
Compatibility fix for GPhL workflow, and some linting clean-up

rhfogh:
Fixed a breakage in most recent commit, having to do with a version mismatch for the GPhL workflow. 
Also fixed Ruff violations and cut down on noqAS


marcus-oscarsson (approved):

marcus-oscarsson:
Thanks

PR #1600, opened 2026-04-29
Implement LNLSSampleChanger and LNLSBeamlineActions

pedrobenetton:
This PR implements the classes necessary for operating our sample changer Staubli robot. I started with SampleChangerMockup and BeamlineActionsMockup.

The Home, Soak and Dry actions are meant to be available at this tab with LNLSBeamlineActions :
image

I realized that when the user is running a beamline action, in other words, when the UI is in this moment:
image

it is possible for the user to press F5 to get rid of this modal component, and then try to move a motor or the backlight. If this happens, the robot stops its movement and a staff member needs to manually operate it back to a safe position. `update_diffractometer_states` and the `EPICSRetsrictedMovement` class are security measurements for that :)

All the 5 actions inherit form the same class, LNLSSampleChangerAction, because they are all bluesky plans. For the Home, Soak, and Dry, there is the Abort option, and that is implemented via LNLSBeamlineActions. LNLSSampleChanger class is the one that calls Mount and Unmount actions, and they don't seem to have an abort option.

I also saw that unmount triggered this modal at UI:
image

But mount does trigger anything at UI. This the reason I did `self.frontend_application.server.emit("sc", frontend_msg, namespace="/hwr")` at LNLSSampleChanger, so that I could trigger this modal when a sample is being mounted:

image

In both mount and unmount, the same problem applies: user could simply press F5 (or click cancel) and then try to move a motor. Again, this is why `update_diffractometer_states` and the EPICSRetsrictedMovement class are being used

The remainng functions at LNLSSampleChanger are basically functions that allows us to get the Puck names from our EPICS PVs (they are configured at an external application), and if they are changed, clicking on the button "Get Samples from SC" will always update them in the UI.

### DIFF
--- a/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
+++ b/mxcubecore/HardwareObjects/LNLS/EPICS/EPICSActuator.py
@@ -158,11 +158,13 @@ class LNLSEnergy(AbstractEnergy, EPICSActuatorBluesky):
 
     pass
 
+
 class EPICSRestrictedMovement:
     """
     This class is meant for devices that should not move while
     the LNLSStaubli is moving.
     """
+
     def init(self):
         super().init()
         self.sc = HWR.beamline.get_object_by_role("sample_changer")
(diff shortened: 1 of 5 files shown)


marcus-oscarsson (approved):

marcus-oscarsson:
Looks good, you can rebase and merge. Really nice work guys !

PR #1608, opened 2026-05-19
Added warning when `overlap` is used

marcus-oscarsson:
This PR adds a warning `overlap` is used

### DIFF
--- a/mxcubecore/HardwareObjects/Beamline.py
+++ b/mxcubecore/HardwareObjects/Beamline.py
@@ -336,6 +336,18 @@ class Beamline(HardwareObject):
             else:
                 params.update(dd0)
 
+        if "overlap" in params:
+            self.log.warning(
+                "The 'overlap' parameter is deprecated. Please use 'offset' instead."
+            )
+
+            overlap = params.pop("overlap")
+
+            if overlap:
+                overlap = -overlap
+
+            params["offset"] = overlap
+
         for tag, val in params.items():
             setattr(acq_parameters, tag, val)
 


mxcubecore/HardwareObjects/Beamline.py
```diff
@@ -336,6 +336,12 @@ def get_default_acquisition_parameters(self, acquisition_type="default"):
             else:
                 params.update(dd0)
 
+        if "overlap" in params:
+            self.log.warning(
+                "The 'overlap' parameter is deprecated. Please use 'offset' instead."
+            )
+            params["offset"] = params.pop("overlap")
+
```
marcus-oscarsson: We could be a bit more drastic and `raise` here but thats maybe a bit radical ?

rhfogh: I  cannot get the 'propose change; to work, but I would propose this instead:

```
        if "overlap" in params:
            self.log.warning(
                "The 'overlap' parameter is deprecated. Please use 'offset' instead."
            )
            overlap = params.pop("overlap")
            if overlap:
                overlap = -overlap
            params["offset"] = overlap
```

Offset is -1 times overlap.
Or we can do an error if you think it is cleaner - teach those guys to keep up with the changes!

rhfogh (changes requested):

mxcubecore/HardwareObjects/Beamline.py
marcus-oscarsson:
```suggestion
        if "overlap" in params:
            self.log.warning(
                "The 'overlap' parameter is deprecated. Please use 'offset' instead."
            )
            overlap = params.pop("overlap")
            if overlap:
                overlap = -overlap
            params["offset"] = overlap
```

rhfogh (changes requested):
Sorry to be picky - but woudl it not be better to get rid of those empty lines in the block?

marcus-oscarsson:
Actually I prefer it like that, I added those lines by purpose:

I find that this to be easier to read:
```
        if "overlap" in params:
            self.log.warning(
                "The 'overlap' parameter is deprecated. Please use 'offset' instead."
            )

            overlap = params.pop("overlap")

            if overlap:
                overlap = -overlap

            params["offset"] = overlap

Compared to:

```
        if "overlap" in params:
            self.log.warning(
                "The 'overlap' parameter is deprecated. Please use 'offset' instead."
            )
            overlap = params.pop("overlap")
            if overlap:
                overlap = -overlap
            params["offset"] = overlap
```

rhfogh (approved):
OK, it is just style anyway. By all means commit it as is.

Issue #1609, opened 2026-05-19
Latest changes related to GPhL workflow on MASSIF 1


olofsvensson:
I will submit an MR with the latest changes applied to mxcubecore on MASSIF 1 for running the GPhL workflows.

marcus-oscarsson:
Wonderful, thanks @olofsvensson

olofsvensson:
Patch merged.

PR #1610, opened 2026-05-19
Issue #1609 : Patches for GPhL workflow on MASSIF 1

olofsvensson:
These are the latest patches for mxcubecore for getting the GPhL workflows running on MASSIF 1.

### DIFF
--- a/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
+++ b/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
@@ -227,12 +227,12 @@ class IdentifiedElement(MessageData):
 
     def __set_id(self, value):
         """Setter for uuid - accessible only within this class"""
-        if isinstance(value, uuid.UUID):
-            self._id = value
-        elif value is None:
+        if value is None:
             self._id = uuid.uuid1()
+        elif isinstance(value, uuid.UUID):
+            self._id = value
         else:
-            self._id = uuid.UUID(value)
+            raise TypeError("UUID input must be of type uuid.UUID")
 
 
 # Sync with Java 4/5/2017
--- a/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
+++ b/mxcubecore/HardwareObjects/Gphl/GphlMessages.py
@@ -227,12 +227,12 @@ class IdentifiedElement(MessageData):
 
     def __set_id(self, value):
         """Setter for uuid - accessible only within this class"""
-        if value is None:
-            self._id = uuid.uuid1()
-        elif isinstance(value, uuid.UUID):
+        if isinstance(value, uuid.UUID):
             self._id = value
+        elif value is None:
+            self._id = uuid.uuid1()
         else:
-            raise TypeError("UUID input must be of type uuid.UUID")
+            self._id = uuid.UUID(value)
 
 
 # Sync with Java 4/5/2017
--- a/mxcubecore/HardwareObjects/ICATLIMS.py
+++ b/mxcubecore/HardwareObjects/ICATLIMS.py
@@ -1197,6 +1197,11 @@ class ICATLIMS(AbstractLims):
                 # The "complete" entry in metadata must be set to False in order to
                 # group multi-wedge reference image data collection for characterisation
                 metadata["complete"] = False
+            elif scan_type == "OSC":
+                # In case the experiment_type is "OSC" and doesn't have
+                # "datacollection" in the dataset name, we set it to "datacollection".
+                # This happens for data collected by GPhL workflows.
+                scan_type = "datacollection"
 
             workflow_params = datacollection_dict.get("workflow_parameters", {})
             workflow_type = workflow_params.get("workflow_type")
--- a/mxcubecore/HardwareObjects/XMLRPCServer.py
+++ b/mxcubecore/HardwareObjects/XMLRPCServer.py
@@ -112,11 +112,7 @@ class XMLRPCServer(HardwareObject):
         msg = "XML-RPC server listening on: %s:%s" % (self.host, self.port)
         self.log.info(msg)
 
-        self.connect(
-            HWR.beamline.gphl_workflow,
-            "gphl_workflow_finished",
-            self._async_job_completed,
-        )
+        self._connect_gphl_workflow_finished()
 
         self._server.register_introspection_functions()
         self._server.register_function(self.start_queue)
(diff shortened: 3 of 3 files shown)


mxcubecore/HardwareObjects/Gphl/GphlMessages.py
```diff
@@ -227,12 +227,12 @@ def id_(self):
...
-        if value is None:
-            self._id = uuid.uuid1()
-        elif isinstance(value, uuid.UUID):
+        if isinstance(value, uuid.UUID):
             self._id = value
+        elif value is None:
+            self._id = uuid.uuid1()
         else:
-            raise TypeError("UUID input must be of type uuid.UUID")
+            self._id = uuid.UUID(value)
```
olofsvensson: For @rhfogh : This was necessary as the workflow id coming from the workflow is not an uuid as it has to pass through the XML RPC server.

mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -1197,6 +1197,11 @@ def finalize_data_collection(self, datacollection_dict):
                 # The "complete" entry in metadata must be set to False in order to
                 # group multi-wedge reference image data collection for characterisation
                 metadata["complete"] = False
+            elif scan_type == "OSC":
```
marcus-oscarsson, line 1200: yes, ok for now :)

mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -1197,6 +1197,11 @@ def finalize_data_collection(self, datacollection_dict):
...
+                # In case the experiment_type is "OSC" and doesn't have
+                # "datacollection" in the dataset name, we set it to "datacollection".
+                # This happens for data collected by GPhL workflows.
+                scan_type = "datacollection"
```
olofsvensson, line 1204: For @marcus-oscarsson : This is a temporary fix until we have sorted out the experiment_type / scan_type.

marcus-oscarsson (approved):

marcus-oscarsson:
Thanks, excellent. Looks good to me. Maybe @rhfogh wants to have a quick look ?

mxcubecore/HardwareObjects/XMLRPCServer.py
```diff
@@ -683,8 +679,22 @@ def add_gphl_workflow(self, parent_node_id, task_dict, workflow_id):
...
+            gphl_workflow,
+            "gphl_workflow_finished",
+            self._async_job_completed,
+        )
+        self.connect(
+            gphl_workflow,
+            "gphl_workflow_finished",
+            self._async_job_completed,
+        )
+
```
olofsvensson, line 697: For @marcus-oscarsson and @rhfogh : Not so elegant workaround for making sure the connection is done between the XMLRPCServer and the GphlWorkflow. The fact that the XMLRPCServer HO is defined earlier than the GphlWorkflow HO in the beamline.yaml file didn't help to make sure the connection is done when starting MXCuBE.

rhfogh: Looks fine except for one thing: Is it OK to leave the connection live and dangling? Or would there be a need to disconnect  again - maybe in _async_job_completed?

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
rhfogh: This is a significant change, as it changes the behaviour for all GPhL messages. It would be useful to know the context, and what constraints there are on the type of value that can come from the server. Will it always be a string version of a UUID, or could it be something else, that might cause an error, and that we would therefore need to allow for somehow? Which objects do we get the error for? If it makes things work for now it is of course good, but it would be nice to know more and then make a decision for a more permanent solution - which might be this one, actually, just would want to be a bit more sure.

olofsvensson: The context is that since the kappa id cannot be transferred as an object from the BES workflow to MXCuBE. It must therefore be transferred as a string. Without this patch the GPhL workflow stopped here with a TypeError "UUID input must be of type uuid.UUID".

We can though solve this problem without touching the GPhL code. The workflow kappa id was passed from the BES workflow to the data portal in order to group data sets belonging to the same kappa settings. This is no longer necessary as the ICATLIMS now sets the data portal kappa id to the actual settings, e.g. "Kappa 34.3 Phi 123.4". Therefore, we could simply remove the kappa id from the BES workflow and set it to None. The GPhL workflow would then use its own settings. Let me know if you prefer this approach.

mxcubecore/HardwareObjects/XMLRPCServer.py
olofsvensson: Well, initially (before the new abstract diffractometer code)  the connection was done once in the init method and hence left "live and dangling" during the whole lifetime of the MXCuBE instance and it worked well. I therefore don't think that it should be a problem to leave it connected.

rhfogh: Makes sense.

mxcubecore/HardwareObjects/Gphl/GphlMessages.py
rhfogh: I am not sure what the best solution is. We would really like to have actual UUID in our workflow, particularly since we would not be in control of IDs coming in from the outside, so we could not know up front if they were unique. One solution would be that outside IDs were guaranteed to be UUIDs - they could be string versions of UUIDs as well, but in that case we would have to keep on top of when they were converted to/from string form. Alternatively we could let the internal GPhL ID be a proper UUID handled by the GPhL workflow (which would amount to passing in the ID as None), and then find another way of keeping and passing along IDs for other systems (your workflow, ICAT etc.), possibly with another attribute. Unless, of course the 'other systems' could work with UID as well.

If it is enough for now to pass in the id as None that might be the simplest, quick fix - assuming it works. But a slightly longer term we ought to coordinate what IDs your side needs and how we need the IDs of the two sides to work with each other. Should we talk about this?

olofsvensson: I think the best is to pass the id as None and let the GPhL workflow handle the id. I'll remove this change from the MR.  Since it's now MXCuBE that sets the kappa settings id for the data portal I can remove the kappa settings id all together from the BES workflows. I'll test this with my MXCuBE simulator (we are now in a short shutdown till the beginning of June).

As we don't have any other id to pass between the workflows I don't think we need to go deeper in solving this now.

rhfogh: OK, if this works for you I am fine with it. As you say, we can always look deeper at ID harmonisation later, if we need to.

rhfogh (approved):
This is fine. That was a very productive discussion, BTW,

olofsvensson:
@marcus-oscarsson , patch rebased and ready to be merged.

marcus-oscarsson:
Thanks !

PR #1612, opened 2026-05-21
Handle DispatcherKeyError in XMLRPCServer

olofsvensson:
I got an "DispatcherKeyError" when testing the patch with the simulator - apologies for not having tested before. This patch should solve this problem.

### DIFF
--- a/mxcubecore/HardwareObjects/XMLRPCServer.py
+++ b/mxcubecore/HardwareObjects/XMLRPCServer.py
@@ -19,7 +19,6 @@ from functools import reduce
 
 import gevent
 import jsonpickle
-from pydispatch.errors import DispatcherKeyError
 
 from mxcubecore import HardwareRepository as HWR
 from mxcubecore.BaseHardwareObjects import HardwareObject
@@ -691,7 +690,7 @@ class XMLRPCServer(HardwareObject):
                 "gphl_workflow_finished",
                 self._async_job_completed,
             )
-        except DispatcherKeyError:
+        except KeyError:
             pass
         self.connect(
             gphl_workflow,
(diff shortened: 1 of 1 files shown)


rhfogh:
@marcus-oscarsson I do see a  problem with this. dispatcher is imported from mxcubecore.dsipatcher, where you get a dispatcher either from louie or from pydispatcher. Yet here we are pretty much assuming that it is a pydispatcher we have, since we are importing the relevant error message from there. I suspect that louie is old code and never used, and that it is actually pydispatcher every time. But someone who knows something about those dispatchers ought to have a look at this, and possibly clean out old Louie.

mockoocy:
@rhfogh I think it's very good that you are bringing the topic of dispatchers in the codebase. I struggle to find a reason why would anyone go out of their way and plug in louie instead of `PyDispatcher` that is shipped with mxubecore by default. I believe that this dispatcher-related logic could be simplified a bit. It's not a lot of code, but lack of stated reasoning behind it confuses me :D

marcus-oscarsson:
Its not so strange at all in-fact its almost enough to read the documentation of pydispatcher https://pydispatcher.sourceforge.net/ to understand why there is also support for Louie. At some point the support for Qt was better in Louie, its perhaps still the case. For sometime it was also the software that was better maintained, as far as I understand.

I don't think anyone is using `louie` today but I would perhaps just double check with the Qt users before removing it.

We can:
 
   - Either remove `louie` which seems indeed to be unused
   - Catch, `KeyError` instead of `DispatcherKeyError`, the later inherits `KeyError`

rhfogh:
@marcus-oscarsson Ah, in that case I would say to catch KeyError instead. That solves the immediate problem, and frees us from having to import specifically from dispatcher. The Louie problem can then be dealt with in its own good time

olofsvensson:
Ok, I'll do the change.

rhfogh (approved):
Looks good

marcus-oscarsson:
Thanks alot @olofsvensson nicely done !

PR #1614, opened 2026-05-22
PyISPyB integration: add REST implementation, corresponding to current SOAP ISPyB solution

dominikatrojanowska:
Replace the legacy SOAP-based ISPyB client with a new PyISPyB REST implementation backed by Keycloak token authentication. Introduce ``PyISPyBDataAdapter`` and ``PyISPyBRestClient``as abstract, reusable components, removing the old ``ISPyBAdapter``. Add adjustable timeouts, improved error handling, and legacy method compatibility. Update tests to reflect the new REST implementation and handle Max IV specific  ``duo`` and ``sdm`` dependencies gracefully.

### DIFF
--- a/mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
+++ b/mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
@@ -13,13 +13,19 @@ class NoSessionException(Exception):
 
 
 class ISPyBLims(UserTypeISPyBLims):
+    def __init__(self, name: str):
+        super().__init__(name)
+        self._duo_api_url = ""
+
     def init(self):
+        self._duo_api_url: str = self.get_property("duo_api_url")
         pyispyb_rest_root = self.get_property("pyispyb_rest_root")
         self._rest_client = PyISPyBRestClient(pyispyb_rest_root)
         super().init()
 
     def _create_data_adapter(self) -> MAXIVPyISPyBDataAdapter:
         return MAXIVPyISPyBDataAdapter(
+            self._duo_api_url,
             self._rest_client,
             self.beamline_name,
         )
--- a/mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
+++ b/mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
@@ -1,241 +1,154 @@
-import json
+# ruff: noqa: TD003, FIX002, ERA001
 import logging
-from json.decoder import JSONDecodeError
-from typing import (
-    Dict,
-    List,
-    Optional,
-)
-from urllib.parse import urljoin
 
-import requests
-from suds import WebFault
+try:
+    # duo is a part of sdm package that lives at MaxIV private repository
+    from duo.UO import RestDuo
+    from sdm.config import DUOPASSWORD, DUOUSER
+except ImportError:
+    RestDuo = None
+    DUOPASSWORD = None
+    DUOUSER = None
 
-from mxcubecore.HardwareObjects.abstract.ISPyBDataAdapter import ISPyBDataAdapter
+from mxcubecore.HardwareObjects.abstract.PyISPyBDataAdapter import PyISPyBDataAdapter
+from mxcubecore.HardwareObjects.abstract.PyISPyBRestClient import PyISPyBRestClient
 from mxcubecore.HardwareObjects.UserTypeISPyBLims import UserTypeISPyBLims
-from mxcubecore.model.lims_session import (
-    LimsSessionManager,
-    Session,
-)
+from mxcubecore.model.lims_session import Proposal, Session
+
+DUO_API_URL = "https://duo-api.maxiv.lu.se"
+LAZY_SESSION_PREFIX = "lazy"
 
 log = logging.getLogger("ispyb_client")
 
-LAZY_SESSION_PREFIX = "lazy"
+
+class NoSessionException(Exception):
+    """Exception raised when no expected session found."""
 
 
-def _get_lazy_session_id(proposal: Dict) -> str:
-    prop_id = proposal["proposalId"]
-    return f"{LAZY_SESSION_PREFIX}{prop_id}"
+class CustomISPyBDataAdapter(PyISPyBDataAdapter):
+    """Extend the standard ISPyB data adapter with MAXIV specific logic."""
 
+    def get_proposals(self):
+        """Override method to filter proposals by the type, state and beamline name.
 
-def _is_lazy_session_id(session_id: str) -> bool:
-    return session_id.startswith(LAZY_SESSION_PREFIX)
-
-
-def _check_ispyb_error_message(response):
-    def _expected_ispyb_err_msg(error_msg):
-        import re
-
-        match = re.match("^JBAS011843: Failed instantiate.*ldap.*ispyb", error_msg)
-        return match is not None
-
-    #
-    # check that we got the 'expected' error message on invalid credentials,
-    # otherwise log the error message, so we don't swallow new error messages
-    #
-    if _expected_ispyb_err_msg(response.text):
-        # all is fine
-        return
-
-    log.warning(
-        "unexpected response from ISPyB\n"
-        + f"{response.status_code} {response.reason}\n{response.text}"
-    )
-
-
-class ISPyBRestClient:
-    def __init__(self, rest_root: str):
-        self._rest_root = rest_root
-
-    def authenticate(self, user_name: str, password: str):
+        Include proposals: of type ``MX`` or ``MB`` in ``Open`` state and assigned
+        to the current beamline. The last is checked via DUO API.
         """
-        authenticate with REST services
+        duo = RestDuo(DUO_API_URL)
+        duo.login(DUOUSER, DUOPASSWORD)
+        beamline_proposals_ids = set(duo.get_beamline_proposals(self.beamline_name))
+        return [
+            proposal
+            for proposal in super().get_proposals()
+            if proposal.code in ["MX", "MB"]
+            and proposal.state == "Open"
+            and int(proposal.number) in beamline_proposals_ids
+        ]
+
+    def create_session(self, proposal: Proposal) -> Session:
+        """Create a new Session object for the given proposal and beamline.
+
+        This is a lazy session creation, done automatically on the fly in case
+        no appropriate session is found for the user, proposal and current day.
+        This session is labelled with ``lazy`` prefix and is not posted to
+        Py-ISPYB service until it is selected.
 
         Args:
-            user_name: Username
-            password: Password
+            proposal: Proposal object to create session for
+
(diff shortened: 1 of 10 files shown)


mxcubecore/model/lims_session.py
```diff
@@ -34,6 +34,8 @@ class Proposal(BaseModel):
     code: str = ""
     number: str = ""
     title: str = ""
+    name: str = ""
```
marcus-oscarsson: Is `name` different from `title` ?

mxcubecore/model/lims_session.py
```diff
@@ -34,6 +34,8 @@ class Proposal(BaseModel):
...
+    state: str = ""
```
marcus-oscarsson: What does `state` mean in this context ?

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -0,0 +1,574 @@
...
+        except TypeError:
+            self.logger.exception(
+                "Invalid date format. start: %s, end: %s, now: %s",
+                start_datetime,
+                end_datetime,
+                today,
+            )
+            return False
+
+    # =========================
```
marcus-oscarsson: We usually don't add this kind of section comments. I think its alright if you think they are useful, otherwise I would maybe comment in the doc string of each function what it is and whats its intended use :)

mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
```diff
@@ -1,241 +1,154 @@
...
-        return
+class NoSessionException(Exception):
+    """Exception raised when no expected session found."""
 
-    log.warning(
-        "unexpected response from ISPyB\n"
-        + f"{response.status_code} {response.reason}\n{response.text}"
-    )
 
+class CustomISPyBDataAdapter(PyISPyBDataAdapter):
```
marcus-oscarsson: I would perhaps call this `MAXIVISPyBDataAdapter` and also change the filename to reflect this ?

marcus-oscarsson:
Looks good overall, just a few comments to consider

mxcubecore/model/lims_session.py
dominikatrojanowska: `name` refers to `[proposal_name](https://github.com/mxcube/mxcubecore/blob/develop/mxcubecore/model/lims_session.py#L67C9-L67C22)` of Session object. I found it redundant to add `proposal_name` field to Proposal object. 
That's what is returner from PyISPyB:
```
{
      "proposalCode": "MX",
      "proposalNumber": "20230002",
      "proposal": "MX20230002",
      "title": "Test In-house Commissioning & SDM",
      "state": "Open",
      "_metadata": {
        "persons": 14,
        "sessions": 514,
        "beamLines": [
          "BioMAX",
          "MicroMAX",
          "mxcube3test"
        ],
        "uiGroups": []
      },
      "proposalId": 216
    },
```
`name` it's generally a concatenation of proposal number and code, while title is something more descriptive

mxcubecore/model/lims_session.py
dominikatrojanowska: Proposal can be in closed or open state. Not sure if any other is possible here. At Max IV we are looking only for open proposals: https://github.com/mxcube/mxcubecore/pull/1614/changes#diff-6a47fcd7804afcce43cfdf9b3ce2bb8721e74ea9ede437f049b53f1f376e2d67R44

mxcubecore/model/lims_session.py
marcus-oscarsson: Alright I see, but do we need it at all then ? I think its always a concatenation of proposal number and code ?

mxcubecore/model/lims_session.py
marcus-oscarsson: I see, I would somehow say that `get_beamline_proposals` should not at all return proposals that are closed then ? Like this you would not need `state`

mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
dominikatrojanowska: good point :) I just followed what was implemented before and didn't consider that. Will implement :)

mxcubecore/model/lims_session.py
dominikatrojanowska: riiight, maybe we don;t need this. I will remove that

mxcubecore/model/lims_session.py
dominikatrojanowska: That's another application endpoint (PyISPyB) that delivers proposals :thinking: I guess in other use cases (app's inner requests for example) it may be required to get all proposals regardless of their state. I can implement or ask about new endpoint for our purpose with the default filtering to get only those in Open state. Let me know :)

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -0,0 +1,555 @@
...
+        except TypeError:
+            self.logger.exception(
+                "Invalid date format. start: %s, end: %s, now: %s",
+                start_datetime,
+                end_datetime,
+                today,
+            )
+            return False
+
+    def __to_proposal(self, proposal: Dict[str, str]) -> Proposal:
```
mockoocy:
```suggestion
    def __to_proposal(self, proposal: dict[str, str]) -> Proposal:
```
`typing.Dict` is deprecated since python 3.9, same for `typing.List`.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -0,0 +1,555 @@
...
+            title=title,
+            comments=session.get("comments"),
+            start_datetime=start_datetime,
+            end_datetime=end_datetime,
+            # TODO@dominikatrojanowska: check if needed
+            # start_date=datetime.strftime(start_datetime, "%Y%m%d"),
+            # end_date=datetime.strftime(end_datetime, "%Y%m%d"),
+            # start_time=start_datetime,
+            # end_time=end_datetime,
+            # nbShifts can be None, convert to empty string for consistency
```
mockoocy: did you check? :D

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -0,0 +1,108 @@
...
+        except Exception:
+            log.exception(
+                "Authentication failed. Status code: %s, Response: %s",
+                response.status_code,
+                response.text,
+            )
+        if token is None:
+            err = "Authentication failed. No token received."
+            raise NoTokenException(err)
+        return token
```
mockoocy: This exception handling is a bit odd to me.
It seems like `response` is assumed to be a dict here, in that case we don't need to guard `response.get` call. 
The `except` path would also raise another exception, unless the `response` object is of type `requests.Response` - but it does not seem like the indented behaviour?

I imagine something like this could work
```suggestion
    def _get_auth_token(self, response: dict) -> str:
        token = response.get("token", None)
        if token is None:
            err = "Authentication failed. No token received."
            raise NoTokenException(err)
        return token
```

ruff.toml
```diff
@@ -964,6 +964,10 @@ convention = "google"
     "F841",
     "S110",
 ]
+"mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py" = [
+    "ARG002",
+    "TRY301",
+]
```
mockoocy: I don't think this is necessary, since we're not using these rules since https://github.com/mxcube/mxcubecore/pull/1574.

mockoocy (changes requested):

mxcubecore/model/lims_session.py
marcus-oscarsson: Maybe you can just filter on retrieval so that you don't need `state` once everything is retrieved ? There is no need to pass it around if its not used.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -0,0 +1,548 @@
...
+                Expected to contain "collection_id" key.
+
+        Returns:
+            Tuple containing data collection id and data collection group id.
+        """
+        dc_id, dc_group_id = 0, 0
+        if "collection_id" in mx_collection:
+            mx_collection["group_id"] = self._store_data_collection_group(
+                mx_collection
+            )["dataCollectionGroupId"]
```
dominikatrojanowska: According to [discussion](https://github.com/mxcube/mxcubecore/pull/1621/changes#r3427799851) and the relevant change will be added after merging https://github.com/mxcube/mxcubecore/pull/1621/changes

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
dominikatrojanowska: The implementation changed and I think this comment is outdated already. Take a look: https://github.com/mxcube/mxcubecore/pull/1614/changes#diff-643ac4888f415e9de261c606c7779209b143b53fd803c228c7cf0301e9e0c5e0R134

mxcubecore/model/lims_session.py
dominikatrojanowska: ok, removed this `state` field

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -76,6 +76,40 @@ def __to_session(self, session: dict) -> Session:
             is_scheduled_beamline=True,  # MAX IV does not care about this value
         )
 
+    @staticmethod
+    def __to_sample(sample_data: dict) -> dict:
```
rhfogh, line 80: There is nothing wrong with  it - but why are you using double_underscore? As I understand it, that is specific for functions that cannot be called even from subclasses, whereas the normal syntax for private functions is a single underscore.

dominikatrojanowska: That's quite right about _private and __protected members. Actually there is something called name mangling to access even the protected but generally it should be avoided. 

This implementation bases on ISPyBDataAdapter and originally these helpers are protected members. I've just find no reason to change that.

rhfogh: Fair enough. I have a suspicion that in many cases (possibly including this one) people are using protected members without justification, but it certainly makes more sense to leave things as they are and do the nitpicking and clean-up later (if at all).

marcus-oscarsson:
Looks good as far as I can see, not sure if there are still changes requested @mockoocy. Otherwise I think we can rebase and merge

mockoocy (approved):

PR #1617, opened 2026-05-28
Set Diffractometer to ready once all motors have moved

rhfogh:
In my (mock) execution, the diffractometer remains BUSY when motors have moved. This change makes it go back to being READY

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
@@ -315,7 +315,7 @@ class AbstractDiffractometer(HardwareObject):
         if simultaneous:
             for key in motors_positions_dict:
                 mot_hwobj_dict[key].wait_ready(timeout)
-        self.update_state(HardwareObjectState.READY)
+        self.update_state()
 
     def get_value_motors(self, motors_list: list | None = None) -> dict:
         """Get the positions of diffractometer motors. If the motors_list is
--- a/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
@@ -315,7 +315,7 @@ class AbstractDiffractometer(HardwareObject):
         if simultaneous:
             for key in motors_positions_dict:
                 mot_hwobj_dict[key].wait_ready(timeout)
-        self.update_state()
+        self.update_state(HardwareObjectState.READY)
 
     def get_value_motors(self, motors_list: list | None = None) -> dict:
         """Get the positions of diffractometer motors. If the motors_list is
--- a/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
@@ -71,7 +71,9 @@ class DiffractometerMockup(AbstractDiffractometer):
     def abort(self):
         self.update_state(HardwareObjectState.READY)
 
-    def wait_status_ready(self, timeout=None):
+    def wait_status_ready(self, timeout: float | None = None) -> bool:
+        """Always finish wait as READY."""
+        self.update_state(HardwareObjectState.READY)
         return True
 
     def save_centring_positions(self):
@@ -119,5 +121,5 @@ class DiffractometerMockup(AbstractDiffractometer):
         self.current_phase = value
         self.update_state(self.STATES.READY)
 
-    def _set_constraint(self, value):
+    def _set_constraint(self, value: DiffractometerConstraint):
         self.current_constraint = value
--- a/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
@@ -123,3 +123,12 @@ class DiffractometerMockup(AbstractDiffractometer):
 
     def _set_constraint(self, value: DiffractometerConstraint):
         self.current_constraint = value
+
+    def set_value_motors(self,
+                         motors_positions_dict: dict,
+                         simultaneous: bool = True,
+                         timeout: float | None = None,
+                         ):
+        """Move specified motors to the requested positions."""
+        super().set_value_motors(motors_positions_dict, simultaneous, timeout)
+        self.update_state(HardwareObjectState.READY)
--- a/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
@@ -124,11 +124,12 @@ class DiffractometerMockup(AbstractDiffractometer):
     def _set_constraint(self, value: DiffractometerConstraint):
         self.current_constraint = value
 
-    def set_value_motors(self,
-                         motors_positions_dict: dict,
-                         simultaneous: bool = True,
-                         timeout: float | None = None,
-                         ):
+    def set_value_motors(
+        self,
+        motors_positions_dict: dict,
+        simultaneous: bool = True,
+        timeout: float | None = None,
+    ):
         """Move specified motors to the requested positions."""
         super().set_value_motors(motors_positions_dict, simultaneous, timeout)
         self.update_state(HardwareObjectState.READY)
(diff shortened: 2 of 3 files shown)


mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
```diff
@@ -315,7 +315,7 @@ def set_value_motors(
         if simultaneous:
             for key in motors_positions_dict:
                 mot_hwobj_dict[key].wait_ready(timeout)
-        self.update_state()
+        self.update_state(HardwareObjectState.READY)
```
beteva: I believe this is not quite right. We need to get the state from the hardware, not to force it to READY. This should be rather implemented in the mock diffractometer. Maybe  `self.update_state(HardwareObjectState.READY)`
can be added in the wait_status_ready of the DiffractometerMockup class.

rhfogh: That could be - but how is this supposed to work? If you look at the set_value_motors function, the diffractometer state is explicitly set to BUSY at the top of the function.  Then if the move is simultaneous, the diffractometer waits till all motors have finished moving, and then it does update_state. I can see that it can make sense that the diffractometer itself keeps track of when it is ready and when it is not. I cannot see why you would set it explicitly to BUSY in a function like set_value_motors but rely on the diffractometer to set it back again - that seems asymmetrical. Why does it not rely on signals from the motors to decide when to start being busy as well?

rhfogh: @beteva Also, I was looking in MicroDiffractometer - what is global_state, and where is it defined or configured? I could find no reference to it in mxcubecore at all, except for it being used ('self.global_state') in MicroDiffractometer.

beteva: > I cannot see why you would set it explicitly to BUSY in a function like set_value_motors but rely on the diffractometer to set it back again

We set it to busy simply to be sure that the state has changed to a known state. But than update_state does get_state (see BasehardwareObject) if the input value is None, so we get the appropriate state from the diffractometer.

> what is global_state, and where is it defined or configured?

This and also the phase_channel are part of the custom_commands and custom_channels. Indeed I should have put a bit more of an explanation on this. In principle you can add any command or channel, available from the MD2 application. This is how it is configured in the microdiff xml file:
```
  <commands>
    <run_script>runScript</run_script>
    <save_centring_positions>saveCentringPositions</save_centring_positions>
    <scan_limits>getOmegaMotorDynamicScanLimits</scan_limits>
    <set_plate_vertical>setPlateVertical</set_plate_vertical>
  </commands>
  <channels>
    <beam_position_horizontal>BeamPositionHorizontal</beam_position_horizontal>
    <beam_position_vertical>BeamPositionVertical</beam_position_vertical>
    <global_state>State</global_state>
    <phase_channel>CurrentPhase</phase_channel>
  </channels>
```

rhfogh: @beteva Well, what happens in mock mode is that the diffractometer state is explicitly set to BUSY in set_value_motors - and never set back again, because there is no functionality anywhere to do that. AT least that is how it works for me - I do not know if somebody has tested the mock mode and found a different result. Nor is there any documentation that says what functionality subclasses must implement in order for AbstractDiffractometer to work as it should. Or if there is, at least I have been unable to find it. Would it be possible to make a specification here, as part of the AbstractDiffractometer interface? It would be easier for me to make a proper fix in the DiffractometerMockup if I understood better how the system is expected to function.

As for the channels, I did what I normally do when I am looking for something: I made sure that I was fully up to date with mxcubeweb and mxcubecore, and then I did a global search for 'global_state'. I did not find the microdiff xml file, or anything else that showed me where global_state was configured. For future reference, since this seems to be a core part of the functioning of the diffractometer, what is the location where one should look to find this kind of thing?

rhfogh:
This looks like a promising approach - and it is clearly better than my original try. But I have a problem with wait_status_ready. First, it is not defined in AbstractDiffractometer, so there is no way of knowing that this is (supposedly) the correct way of checking for readiness. Second, wait_status_ready apparently must be called in order to guarantee that he difffractometer resets to READY state - which side effect is not at all obvious from the name. Third, the obvious way to handle readiness would be to use is_ready, and wait_ready, which _are_ defined, in HardwareObject, and are also available in AbstractDiffractometer. 

Can't we have a single, functioning way of dealing with objects being ready, instead of relying on a special-case alternative method for diffractometers? One way would be to override is_ready and wait_ready for the diffractometer, possibly with an abstract method in AbstractDiffractometer?

Or is wait_status_ready doing something different from wait_ready, and if so how does one know when to use what?

mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
beteva: Dear @rhfogh, you've spotted a problem with the current implementation of the mock diffractometer. I'm suggesting the solution. I've also added a test. Unfortunately, due to the implementation of the MotorMockup, I can only test for the state, not for the values of the motors.Hope this will work with your setup.

beteva:
> I have a problem with wait_status_ready

Indeed there is a problem. The MD2 wait_ready is completely different from the BaseHardwareObject wait_ready - it reads a state from the application, not waiting for an event. It was called wait_ready in the beginning and also in the previous diffractometer implementation, but, if I remember well,  it was you who insisted of having it under different name.  I did not want to pollute the AbstractDiffractometer with methods, which are hardware specific, so that's why it is not there.

rhfogh:
@beteva Sounds like this is my fault, then. I have forgotten that, but I do apologise. I wonder why I did that.
For the immediate purpose we may just need to find a fix.

In the slightly longer term, though, we ought to agree on where we want to end up. Currently I notice that e.g. SampleView calls wait_status_ready. If that is a hardware-specific method that should not be in the AbstractDiffractometer, then surely it should also not be called outside its own class or maybe site-specific code? Whichever way we do it, there needs to be a way that we can check whether the diffractometer is ready (and wait till it becomes so), that can be called from anywhere, that is part of the diffractometer API, and that we make sure works correctly. If I am the one who messed this up to start with I again apologise - but is there functionality that answers to those requirements at the moment? And if not, how are we going to make sure that there will be in the future?

beteva:
> Sounds like this is my fault, then. I have forgotten that, but I do apologise.

There is no fault, only considerations. You must have had some good reasons.

> In the slightly longer term, though, we ought to agree on where we want to end up.

Totally agree with you.

rhfogh:
This looks like a good solution - for now. I cannot review it - since it is under my name - but you can approve and merge it if you think so.

For the longer term, what should we do? Should we make an issue raising the problem? And if so, can you do it (understanding the diffractometer part better) or do  you want me to have a go?

beteva:
After a second though, maybe it is a good idea to put a place holder for wait_status_ready in the AbstractDiffractometer. This will allow to overload it with a specific routine if the hardware needs it or do nothing in the rest of the time. This than allow to use wait_ready with the events as well. @rhfogh What do you think?

rhfogh:
Again, that sounds like a good idea. But while we are talking about this would you be able to tell me what the difference is between wait_ready and wait_status_ready? As in when should you use te eone and when should you use the other (and how would you know)? As long as we are keeping both that would help me understand what is happening.

beteva:
wait_ready, the generic wait for an object to report ready method should be used whenever it is possible. Than there is this former wait_ready, which now became wait_status_ready, which is a very specific for the Microdiff. The application would not allow you to do any action until the previous has not reported ready, even if the previous action uses hardware, not used in the consequent one. You need to poll the application (no events) to check if it reports ready (even more vicious, there are two calls to be sent - global and hardware state). As in the generic wait_ready there is nothing foreseen to ask the hardware, we either have a different wait routine or overload it. I do not remember why in this case we did not overload the wait_ready one.

rhfogh:
OK, unless someone can come up with a good reason it sounds like wait_status_ready is a different, more complex implementation of what is essentailly the same task. And that we basicaly agree that it ought to be overloaded, in the long run.

In the short run, we can do something simple and practical. As a proposal, would it not make more sense to rename wait_status_ready to \_wait_status_ready, have a wait_ready function in AbstractDiffractometer thet just called self._wait_status_ready, and put \_wait_status_ready in AbstractDiffractometer as an abstract method? That should get us to having an overridden wait_ready with a minimum of fuss.

marcus-oscarsson:
I think that we should only have one "wait" method on `AbstractDiffractometer`, `wait_ready`,  and in case of `MicroDiffractometer` that should be the logic that is in `wait_status_ready`. If we need something else internally of `MicroDiffractometer`, sure we could add `_wait_....`

rhfogh:
Withdrawn, as the problem is dealt with (better) elsewhere. Thanks @beteva

PR #1618, opened 2026-06-01
Add unit tests and improve PyISPyB REST client token handling

mohsen-dehesh:


mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -13,7 +14,12 @@ class NoTokenException(Exception):
 
 
 class PyISPyBUnsuccessfulResponse(Exception):
-    """Exception raised when a response from the server is unsuccessful (not 200)."""
+    """Exception raised when a
+    response from the server is unsuccessful (not 200)."""
```
dominikatrojanowska:
```suggestion
    """Exception raised when a response from the server is unsuccessful (not 200)."""
```
no need for a new line, line length is 86 here

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
             response.status_code,
             response.text,
         )
+
         if response.status_code not in (200, 201):
             msg = (
-                f"Request to {response.url} failed with code: {response.status_code} "
+                f"Request to {response.url} "
+                f"failed with code: "
+                f"{response.status_code}. "
```
dominikatrojanowska, line 119: Is there any reason for adding new lines? I'd keep it as it was

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -51,58 +148,45 @@ def _decode_json_response(self, response):
...
-        response = self.post(
-            "auth/login",
-            json={
-                "plugin": "keycloak",
-                "login": user_name,
-                "token": token,
-            },
+        if not access_token:
+            raise NoTokenException(
+                "Authentication failed." " No access token received."
```
dominikatrojanowska:
```suggestion
                "Authentication failed. No access token received."
```
or if new line was desired:

```suggestion
                "Authentication failed.\nNo access token received."
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
     It handles authentication and communication with PyISPyB REST API.
     """
 
+    REFRESH_MARGIN_SECONDS = 60
+
     def __init__(self, rest_root: str, timeout: int = 5):
         self._rest_root = rest_root
         self._session = Session()
         self._timeout = timeout
```
dominikatrojanowska: Remove all redundant new lines, please. In python we are adding them at the end of files, before methods and new classes (in that case two).  In this repo code is usually more compact unless there is logical need of separation. But there is no need of it here and most of the cases in below methods.

```suggestion
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
+            raise AuthenticationExpired("Failed to refresh token")
+
+        self._store_tokens(response.json())
+
+    def post(self, endpoint, **kwargs):
+        return self._request(
+            self._session.post,
+            endpoint,
+            **kwargs,
+        )
```
dominikatrojanowska: this would fit in one line 

```suggestion
        return self._request(self._session.post, endpoint, **kwargs)
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
+
+    def get(self, endpoint, **kwargs):
+        return self._request(
+            self._session.get,
+            endpoint,
+            **kwargs,
+        )
```
dominikatrojanowska: this would fit in one line
```suggestion
        return self._request(self._session.post, endpoint, **kwargs)
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -51,58 +148,45 @@ def _decode_json_response(self, response):
...
+        if not access_token:
+            raise NoTokenException(
+                "Authentication failed." " No access token received."
+            )
+
+        self._access_token = access_token
+        self._refresh_token = refresh_token
+
+        self._token_expiry = datetime.now(timezone.utc)
+        +timedelta(seconds=expires_in)
```
dominikatrojanowska:
```suggestion
        self._token_expiry = datetime.now(timezone.utc) + timedelta(seconds=expires_in)
```
88 sign line length

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
+                "plugin": "keycloak",
+                "login": user_name,
+                "token": token,
+            },
+            skip_refresh=True,
+        )
+
+        self._store_tokens(response)
+
+    def refresh_access_token(self):
```
dominikatrojanowska: I think this should be private, no external entities should force token refresh, only expiration of current one should trigger it.

Btw, how long is the expiration time set in pyispyb token? I think that there should be some mechanism to refresh token automatically before the current one expires as long as user has active session:
- it should not expire if user is logged to mxcube but does not interact with py-ipsyb, it seems it may happen with current implementation,
- it should not expire when the collection is on going even if the user was logged out of mxcube / SSO

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
+        log.info("Refreshing keycloak access token")
+
+        url = urljoin(self._rest_root, "auth/refresh")
+
+        response = self._session.post(
+            url,
+            timeout=self._timeout,
+            json={
+                "refreshToken": self._refresh_token,
+            },
```
dominikatrojanowska:
```suggestion
            json={"refreshToken": self._refresh_token},
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
+
+        timeout = kwargs.pop("timeout", self._timeout)
+
+        url = urljoin(self._rest_root, endpoint)
+
+        response = method(
+            url,
+            timeout=timeout,
+            **kwargs,
+        )
```
dominikatrojanowska:
```suggestion
        response = method(url, timeout=timeout, **kwargs)
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -22,24 +28,115 @@ class PyISPyBRestClient:
...
+        if response.status_code == 401 and not skip_refresh:
+            log.warning("Received 401. Attempting token refresh.")
+
+            self.refresh_access_token()
+
+            response = method(
+                url,
+                timeout=timeout,
+                **kwargs,
+            )
```
dominikatrojanowska:
```suggestion
            response = method(url, timeout=timeout, **kwargs)
```

test/factories/pyispyb_factory.py
dominikatrojanowska, line 1: I have an impression this file should be transformed into fixtures or `@pytest.mark.parametrize`  or `client.get = Mock(return_value=<dict>)` instead of importing these methods and asserting against hardcoded values. That's pytonic way of introducing programmable context for tests.

test/test_pyispyb_data_adapter.py
```diff
@@ -0,0 +1,151 @@
+from unittest.mock import Mock
+
+from mxcubecore.HardwareObjects.abstract.PyISPyBDataAdapter import PyISPyBDataAdapter
+from mxcubecore.HardwareObjects.abstract.PyISPyBRestClient import PyISPyBRestClient
+import pytest
```
dominikatrojanowska, line 4: order of imports:  
built-in  
installed  
local  
take a look: https://peps.python.org/pep-0008/#imports

test/test_pyispyb_data_adapter.py
```diff
@@ -0,0 +1,151 @@
...
+
+    assert len(proposals) == 2
+    assert proposals[0].proposal_id == "10"
+    assert proposals[1].proposal_id == "11"
+
+
+def test_find_proposal(
+    adapter,
+    client,
+):
```
dominikatrojanowska: applicable in other places as well:
```suggestion
def test_find_proposal(adapter, client):
```

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
mohsen-dehesh: the flake8 that I have on my local, oblige me to reformat them, I will take your suggestion, thank you

beteva:
Hi there,
completely unrelated, but I think there is a rebase problem with you pull request. There are few file (e.g. AbstractDiffractometer.py) which should not be a part of this PR.

mohsen-dehesh:
Hello @beteva 

I will rebase again, they wont be a part of this PR,

test/factories/pyispyb_factory.py
mohsen-dehesh: Hello @dominikatrojanowska 

Thanks for the suggestion.

I agree that @pytest.mark.parametrize is useful when the same test logic is executed with different input/output values.

In this case, I chose factories because the mocked API responses are relatively large and are reused across multiple tests. The factories help reduce duplication, keep the test data readable, and make the intent of each test clearer by exposing only the fields that are relevant to the scenario being tested.

mxcubecore/HardwareObjects/MAXIV/MAXIVPyISPYyBDataAdapter.py
dominikatrojanowska, line 1: hm, I'm not sure what happened with this PR, but this file seems exactly the same as at the parent branch, so shouldn't appear in the diff 🤔 probably refers to some other changes

test/factories/pyispyb_factory.py
dominikatrojanowska: parametrization was only one of the options but if it's best solution in you opinion I will not argue :)

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
mohsen-dehesh: - Yes, you are right!. I've made refresh_access_token() private. Thanks for pointing it out.

- The access token lifetime is provided by PyISPyB via the expiresIn field. The client automatically refreshes it before expiration (60-second margin) and retries once after a 401. As long as PyISPyB communication continues, refresh is transparent. The main consideration is the refresh token lifetime, which is managed by **PyISPyB/Keycloak.**

- If a user is logged into MXCuBE but inactive with PyISPyB, the access token may expire. However, the next PyISPyB request will automatically refresh it using the stored refresh token.

- During an ongoing collection, the client refreshes the access token as needed, so access token expiration should not interrupt the collection.

test/test_pyispyb_data_adapter.py
mohsen-dehesh: the order is fine, now, thanks for mentioning it :)

mohsen-dehesh:
I am closing this PR and opening a new one that includes the updated changes on top of maxiv-soleil-pyispyb-integration.

PR #1619, opened 2026-06-05
[WIP] DoseRate

beteva:
Create DoseRate  HO, 
Add dose_rate to beamline for easy access.
Create DoseRateMockup class and its unit test,

### DIFF
--- a/mxcubecore/HardwareObjects/DoseRate.py
+++ b/mxcubecore/HardwareObjects/DoseRate.py
@@ -1,0 +1,82 @@
+# encoding: utf-8
+#
+#  Project name: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
+#  (at your option) any later version.
+#
+#  MXCuBE is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public License
+#  along with MXCuBE. If not, see <http://www.gnu.org/licenses/>.
+"""Radiation dose for a sample as function of the flux. Uses the beamline.flux
+The dose calculation routine has to be provided externally.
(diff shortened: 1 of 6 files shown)


rhfogh:
This looks very interesting, but all the really important bits would be in the dose_calculator - which is not included. Hopefully it will be, eventually? How do you see this being organised?

marcus-oscarsson:
Closing this for now

beteva:
Reopening as dose rate and dose estimation are two different calculations

mxcubecore/HardwareObjects/DoseRate.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def get_value(self) -> float:
+        """Get the dose rate value as function of the photon flux.
+        Returns:
+            dose rate [MGy/s]
+        """
+        self._nominal_value = self.calculate_dose(HWR.beamline.flux.get_value())
+        return self._nominal_value
+
+    def calculate_dose(self, flux: float | None = None) -> float:
```
rhfogh, line 71: Do you mean calculate_dose_rate?

beteva: Yes.
Tghis is still WIP iand it has been just left as it was from the exercise we did during the code camp. Now that we have both dose rate and dose calculation, I guess it will be somehow merged or at least cleaned up. So here everything should be named  dose_rate 😄

rhfogh:
The existing code already allows you to calculate dose rates. The underlying functions are in AbstractFlux.
Before we rewrite this, should we not discuss why the existing functions are wrong and why the alternative is better? 

Specifically, we have the functions
AbstractFlux.get_dose_rate_per_photon_per_mmsq(energy)
and
AbstractFlux.get_average_flux_density(transmission=None)
Respectively these give the conversion between flux density and dose rate, and the average flux density over the beam.

You can then calculate the dose rate simply as (see e.g. GphlWorkflow.maximum_dose_rate)
dose_rate = (
    flux.get_average_flux_density()
    \* flux.get_dose_rate_per_photon_per_mmsq(energy)
    \* 1.0e-6
)
in MGy/s.

This would then need to be corrected for crystal radiation sensitivity and/or for different ways of calculating the effective flux density at the sample (GphlWorkflow.maximum_dose_rate has an example, of giving an effective dose rate for the case where the beam is much narrower than the crystal). 

Personally I like the conceptual clarity of separating the conversion between flux density and dose rate from the calculation of the flux density and the resulting dose rate. Also I think we could make a default implementation in AbstractDoseRate along the lines shown here (probably bringing in the relative sensitivity as well), which would serve as a default implementation. But can we talk about it?

PR #1620, opened 2026-06-05
Maxiv soleil pyispyb integration

mohsen-dehesh:
fix the git conflic

### DIFF
--- a/mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
+++ b/mxcubecore/HardwareObjects/MAXIV/ISPyBLims.py
@@ -1,79 +1,15 @@
 # ruff: noqa: TD003, FIX002, ERA001
-import logging
-
-try:
-    # duo is a part of sdm package that lives at MaxIV private repository
-    from duo.UO import RestDuo
-    from sdm.config import DUOPASSWORD, DUOUSER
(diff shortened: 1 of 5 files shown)


mxcubecore/HardwareObjects/MicroDiffractometer.py
```diff
@@ -426,10 +435,16 @@ def do_mesh_scan(
         self.set_value_motors(grid_centre, simultaneous=True, timeout=timeout)
 
         scan_params = f"{(end - start):0.3f}\t"
-        scan_params += f"{-mesh_range['horizontal_range']:0.3f}\t"
-        scan_params += f"{mesh_range['vertical_range']:0.3f}\t"
+        if self.head_otientation == "vertical":
```
rhfogh: This (and others) looks like a typo. That should be 'head_orientation'

beteva: This has already been corrected and merged to develop. This PR should only contain the PyIspyBDataAdapter, The other files do not belong to it.
Anyhow it is marked as draft, so we will wait until the request is done. @mohsen-dehesh I believe you still have some work to do on the cleaning and re-basing.

PR #1621, opened 2026-06-08
Implement PyISPyB token refresh and implement integration tests

mohsen-dehesh:
Implement PyISPyB token refresh and implement integration tests
- Add refresh token support to REST client
- Add mocked API tests for PyISPyB client and adapter methods
- Align adapter endpoints with PyISPyB API

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
+++ b/mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
@@ -48,7 +48,7 @@ class PyISPyBDataAdapter:
         return Proposal(
             code=proposal.get("proposalCode").upper(),
             number=proposal.get("proposalNumber"),
-            proposal_id=proposal.get("proposalId"),
+            proposal_id=str(proposal.get("proposalId")),
             title=proposal.get("title"),
             type=proposal.get("type", ""),
             name=proposal.get("proposal"),
@@ -59,7 +59,7 @@ class PyISPyBDataAdapter:
         """Converts session data received from PyISPyB REST API to a Session object."""
         proposal_name = session.get("proposal")
         proposal_code = "".join([c for c in proposal_name if not c.isdigit()])
-        proposal_number = proposal_name[len(proposal_code) :]
+        proposal_number = proposal_name[len(proposal_code):]
         title = session.get("title", "")
         start_datetime = datetime.fromisoformat(session.get("startDate"))
         end_datetime = datetime.fromisoformat(session.get("endDate"))
@@ -67,18 +67,21 @@ class PyISPyBDataAdapter:
             code=proposal_code,
             number=proposal_number,
             proposal_name=proposal_name,
-            proposal_id=session.get("proposalId"),
-            session_id=session.get("sessionId"),
+            proposal_id=str(session.get("proposalId")),
+            session_id=str(session.get("sessionId")),
             beamline_name=session.get("beamLineName", self.beamline_name),
             title=title,
             comments=session.get("comments"),
             start_datetime=start_datetime,
             end_datetime=end_datetime,
-            nb_shifts=session.get("nbShifts", "") or "",
-            scheduled=session.get("scheduled", "False"),
+            nb_shifts=str(session.get("nbShifts", "")),
+            scheduled=str(session.get("scheduled", "False")),
             is_scheduled_time=self.__is_time_between(start_datetime, end_datetime),
             is_scheduled_beamline=True,  # MAX IV does not care about this value
         )
+
+    def __find_proposal_by_id(self, proposal_id: int) -> Proposal:
+        return self.__to_proposal(self.client.get("proposals?proposalId=%s" % (proposal_id))[0])
 
     def get_current_user_data(self) -> dict:
         """Fetches current user details."""
@@ -93,7 +96,7 @@ class PyISPyBDataAdapter:
     def find_proposal(self, code: str, number: str) -> Proposal:
         """Finds a proposal by its code and number."""
         return self.__to_proposal(
-            self.client.get("proposals?proposal=%s%s" % (code, number))
+            self.client.get("proposals?search=%s%s" % (code, number))[0]
         )
 
     def get_sessions_by_code_and_number(
@@ -113,15 +116,13 @@ class PyISPyBDataAdapter:
         self, code: str, number: str, beamline: str
     ) -> list[Session]:
         """Finds todays sessions by proposal code, number and beamline name."""
-        # TODO@dominikatrojanowska: add ``day`` to the url when implemented in PyISPyB,
-        # Until then fetch all sessions for month and year, next filter by day in mxcube
         today = datetime.today()  # noqa: DTZ002
-        month, year = today.month, today.year
+        day, month, year = today.day, today.month, today.year
         return [
             self.__to_session(session)
             for session in self.client.get(
-                "sessions?proposal=%s%s&beamLineName=%s&year=%s&month=%s"
-                % (code, number, beamline, year, month)
+                "sessions?proposal=%s%s&beamLineName=%s&year=%s&month=%s&day=%s"
+                % (code, number, beamline, year, month, day)
             )
             if self.__is_time_between(
                 datetime.fromisoformat(session.get("startDate")),
@@ -165,12 +166,7 @@ class PyISPyBDataAdapter:
 
     def create_session(self, proposal: Proposal) -> Session:
         """Creates new session via PyISPyB REST API for the given proposal."""
-        # TODO@SOLEIL: This is the closest implementation to current PyISPyBAdapter.
-        # Ensure session creation is correct (session data, posting to PyISPyB).
-        # Which part can be common for SOLEIL and MAX IV?
-        # TODO@dominikatrojanowska: check if response is ok and handle errors
-        #  NOT TESTED
-        start_time = datetime.today().replace(hour=0, minute=0, second=0, microsecond=0)  # noqa: DTZ002
+        start_time = datetime.today().replace(hour=0, minute=0, second=0, microsecond=0)
         end_time = start_time + timedelta(
             days=self.new_session_duration_days, hours=7, minutes=59, seconds=59
         )
@@ -243,7 +239,9 @@ class PyISPyBDataAdapter:
         image_id = 0
         if "dataCollectionId" in image_dict:
             try:
-                image_id = self.client.post("images/image", json=image_dict)
+                response = self.client.post("images/image", json=image_dict)
+                self.logger.debug("PyISPYB store image response: %s", response)
+                image_id = response["imageId"]
             except PyISPyBUnsuccessfulResponse:
                 self.logger.exception("Exception in store_image")
             else:
@@ -258,16 +256,16 @@ class PyISPyBDataAdapter:
     def get_samples(self, proposal_id: int) -> list[dict]:
         """Fetches samples for the given proposal id from PyISPyB."""
         try:
-            samples = self.client.get(
-                "samples?proposal=%s&beamLineName=%s"
-                # TODO@dominikatrojanowska: when available at PyISPyB API change to:
-                #  "samples?proposalId=%s&beamLineName=%s"
-                % (proposal_id, self.beamline_name),
+            proposal = self.__find_proposal_by_id(proposal_id)
+
+            return self.client.get(
+                "samples?proposal=%s%s&beamLineName=%s"
+                % (proposal.code, proposal.number, self.beamline_name),
                 timeout=10,
             )
         except PyISPyBUnsuccessfulResponse:
             self.logger.exception("Error in get_samples")
-        return samples
+            return []
 
     def store_robot_action(self, robot_action: dict) -> int:
         """Stores robot action.
@@ -279,8 +277,6 @@ class PyISPyBDataAdapter:
         Returns:
             The id of the robot action. 0 if failed or required keys are missing.
         """
-        # TODO@dominikatrojanowska: ensure correct data type, maybe: datetime.strptime(
-        #  endTime, "%Y-%m-%d %H:%M:%S").strftime("%Y-%m-%dT%H:%M:%S.%fZ")
         self.logger.info("Storing robot actions in LIMS")
         robot_action_id = 0
         try:
@@ -315,9 +311,8 @@ class PyISPyBDataAdapter:
             "blSampleId": entry_dict["blSampleId"],
         }
         try:
-            # TODO@dominikatrojanowska: endpoint will be created
-            assoc_id = self.client.post(
-                "events/energyscan/scanId/BlSample/blSampleId", json=payload
+            assoc_id = self.client.patch(
+                "events/energyscan/associate-bl-sample", json=payload
             )
         except Exception:
             self.logger.exception(
@@ -336,11 +331,13 @@ class PyISPyBDataAdapter:
             Dictionary containing data collection details or empty if fetching failed.
         """
         try:
-            response = self.client.get("events/datacollection/%s" % data_collection_id)
+            response = self.client.get("events?dataCollectionId=%s" % data_collection_id)
         except Exception:
             self.logger.exception("Failed to get data collection from PyISPyB")
             return {}
-        dc = asdict(response)
+        if not response:
+            return {}
+        dc = response[0]
         dc["startTime"] = datetime.strftime(dc["startTime"], "%Y-%m-%d %H:%M:%S")
         dc["endTime"] = datetime.strftime(dc["endTime"], "%Y-%m-%d %H:%M:%S")
         return dc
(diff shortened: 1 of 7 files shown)


mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -59,27 +59,30 @@ def __to_session(self, session: dict) -> Session:
...
         start_datetime = datetime.fromisoformat(session.get("startDate"))
         end_datetime = datetime.fromisoformat(session.get("endDate"))
         return Session(
             code=proposal_code,
             number=proposal_number,
             proposal_name=proposal_name,
-            proposal_id=session.get("proposalId"),
-            session_id=session.get("sessionId"),
+            proposal_id=str(session.get("proposalId")),
+            session_id=str(session.get("sessionId")),
```
dominikatrojanowska, line 68: pydantic coerces the type conversion automatically, so `str()` here is redundant. This is completely correct code:
```
class Sth(BaseModel):
    field: str = ""

sth = Sth(field=10)
sth
# Sth(field='10')
```

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -243,7 +239,9 @@ def store_image(self, image_dict: dict) -> int:
         image_id = 0
         if "dataCollectionId" in image_dict:
             try:
-                image_id = self.client.post("images/image", json=image_dict)
+                response = self.client.post("images/image", json=image_dict)
+                self.logger.debug("PyISPYB store image response: %s", response)
```
dominikatrojanowska, line 242: some leftover after debugging?

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -16,27 +17,106 @@ class PyISPyBUnsuccessfulResponse(Exception):
...
             response.url,
             response.status_code,
             response.text,
         )
         if response.status_code not in (200, 201):
             msg = (
-                f"Request to {response.url} failed with code: {response.status_code} "
+                f"Request to {response.url} "
+                f"failed with code: "
+                f"{response.status_code}. "
```
dominikatrojanowska: why so many separate lines?

```suggestion
                f"Request to {response.url} failed with code: {response.status_code}. "
```
or 
```suggestion
                f"Request to {response.url} failed with code: "
                f"{response.status_code}. "
```
if there are arguments for separation, or

f-strin is not needed
```suggestion
                f"Request to {response.url} "
                "failed with code: "
                f"{response.status_code}. "
```

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -336,13 +331,14 @@ def get_data_collection(self, data_collection_id: int) -> dict:
...
             self.logger.exception("Failed to get data collection from PyISPyB")
             return {}
-        dc = asdict(response)
-        dc["startTime"] = datetime.strftime(dc["startTime"], "%Y-%m-%d %H:%M:%S")
-        dc["endTime"] = datetime.strftime(dc["endTime"], "%Y-%m-%d %H:%M:%S")
+        if not dc:
+            return {}
+        dc["startTime"] = datetime.fromisoformat(dc["startTime"]).strftime("%Y-%m-%d %H:%M:%S")
+        dc["endTime"] = datetime.fromisoformat(dc["endTime"]).strftime("%Y-%m-%d %H:%M:%S")
         return dc
```
dominikatrojanowska, line 345: According to the document you sent I'm getting:
```
ValueError: Invalid isoformat string: '2026-04-09T16:15:53.366Z'
```

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -466,12 +466,14 @@ def _store_data_collection(self, mx_collection: dict, bl_config: dict | None):
                 bl_config.get("detector_type", ""),
             )
             if detector:
-                mx_collection["detectorId"] = detector.get("detectorId", 0)
+                detector_id = detector.get("detectorId", 0)
+                mx_collection["detectorId"] = detector_id
```
dominikatrojanowska: any reason for that change?

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -16,27 +17,106 @@ class PyISPyBUnsuccessfulResponse(Exception):
...
         self._session = Session()
         self._timeout = timeout
+        self._username = None
+        self._access_token = None
+        self._refresh_token = None
+        self._token_expiry = None
+
+    def authenticate(self, user_name: str, token: str):
+        self._username = user_name
+
```
dominikatrojanowska: Remove all redundant new lines, please. In python we are adding them at the end of files, before methods and new classes (in that case two). In this repo code is usually more compact unless there is logical need of separation. But in this file there is no need of it here and most of the cases in below methods.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
mohsen-dehesh: Hello Dominika, Thanks for your comment, I agree with your commment but: 
when I tried removing the explicit str() conversion, the existing tests fail with a Pydantic validation error:
`pydantic_core._pydantic_core.ValidationError: 2 validation errors for Session E session_id E Input should be a valid string [type=string_type, input_value=1, input_type=int] E For further information visit https://errors.pydantic.dev/2.8/v/string_type E proposal_id E Input should be a valid string [type=string_type, input_value=10, input_type=int] E For further information visit https://errors.pydantic.dev/2.8/v/string_type`

In our current Pydantic configuration, Session.session_id and Session.proposal_id are declared as str and integer values from the pyispyb API are not automatically coerced to strings. The explicit conversion is therefore required to construct a valid Session object.

Maybe this non-strict coercion has been validated in pydantic v1, but here we use pydantic 2.8.2 maybe that is why we should explicitly convert it.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
mohsen-dehesh: the response of this api /ispyb/api/v1/images/image  is 
```
{
    "imageId": 1,
    "dataCollectionId": 2
}
```

the logger should be after the response for double checking of dataCollectionId and imageId then we extract the imageId as response of the method

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
mohsen-dehesh: Thank you for this comment, it is interesting that you mentioned  it. The startTime and endTime fields returned by the pyispyb API are already strings in ISO format without timezone information:
```
"startTime": "2026-06-11T11:55:20",
"endTime": "2026-06-11T11:55:20",
```
In Swagger, (in pdf document also) the same fields are displayed with a Z suffix because of OpenAPI/serialization behavior, but this does not reflect a change in the underlying stored data.

Since MXCuBE only consumes the API response, we only reformat the string for display purposes and do not modify timezone information.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
mohsen-dehesh: Thanks for attention to this refractoring, we should keep the same behavior as existed in the def _store_data_collection of  **ISPyBDataAdapter:**  LEGACY CODE and there is a reason for that!

This change is mainly to make the logic more explicit and avoid relying on side effects from mutating mx_collection.

Keeping detector_id in a local variable (as it is in legacy code) ensures we always return a deterministic value even when no detector is found, and avoids depending on whether detectorId is present in the input dict or added during processing.

It also keeps the return value independent from the input mutation, which makes the method easier to reason about and test.

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
mohsen-dehesh: Thanks, I removed the redundant blank lines.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
rhfogh: Marking this discussion as 'resolved' straight away does not conform to established practice in MXCuBE. These things are supposed to be discussions between developers - if Dominika (or anyone else) has a critical question to a piece of code, the discussion should be left up for others to see it, and for the commenter to have the option of saying 'yes, OK', or to raise further points or counterarguments.

As it happens, I agree with Dominika on the specific point, but  the important issue here is developer practice.

mohsen-dehesh: Hello
Thanks for the clarification, I resolved the discussion because I had addressed the comment, I will follow the MXCuBE convention and leave discussion open in the future.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
dominikatrojanowska: I see what the response is and quite agree that logging (if needed) should show the results. 

But my comment refers to line 243, I see that in the `Conversation` tab  it's not that clear as in the `Files changed` tab: 
image

My question is: should this logging be kept? Actually, non other method logs anything after communication with `pyipsyb` server.  If this logging is added to follow the `ISPyBDataAdapter` implementation then the [opening log message](https://github.com/mxcube/mxcubecore/blob/4b487182e5a28747918d7b05b39afd1f4d9aa8b7/mxcubecore/HardwareObjects/abstract/ISPyBDataAdapter.py#L370) is missed. :thinking:

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
dominikatrojanowska: Thanks Rasmus :) 

Again I was asking explicitly about lines 469-470. 

But explaining everything around you draw my attention to mutation of the input `mx_collection` dictionary. Indeed, this code does this, while the old implementation seems to be introducing completely new object relying only on the input dict but not changing it. [The `mx_colection["detectorId"]` assignment ](https://github.com/mxcube/mxcubecore/blob/4b487182e5a28747918d7b05b39afd1f4d9aa8b7/mxcubecore/HardwareObjects/abstract/AbstractCollect.py#L601) happens anyway but maybe this code should send deep copy of input dict to `pyipsyb` avoiding premature mutation? :thinking:

rhfogh: @dominikatrojanowska Good point.

On the original change I would maintain that the new lines 469-470 have _exactly_ the same effect as the old line 469 - down to byte code - except that the new version introduces an additional variable that is not used anywhere else (which is arguably a minor defect in itself). 

But as to the input dictionary I would agree that mutating the input dictionary is  an obscure and unexpected side effect, and a bad practice that can lead to hard-to-discover bugs. It is something that should be avoided unless there is a particularly strong reason to do it here - and if there is such a strong reason, it should be made _very_ clear in the function documentation that this is happening and why. So, what might that strong reason be?

test/test_pyispyb_data_adapter.py
dominikatrojanowska, line 1: Generally given, when, then for python tests is not required and not used by convention here. Also please remove redundant new lines.

test/factories/pyispyb_factory.py
dominikatrojanowska, line 1: 1. I would use pytest fixtures instead of this factory as it's more pytonic way of writing tests. Additionally it would remove boilerplate imports from test file (if fixtures stored in conftest). And handles scopes by definition but maybe all of them should use default one. 

If parametrization is needed: 
```
@pytest.fixture
def make_proposal_response():
    def _make(
        proposal_id: str = "10",
        proposal_code: str = "MX",
        proposal_number: str = "20090662",
    ):
        return {
            "proposalCode": proposal_code,
            "proposalNumber": proposal_number,
            "proposalId": proposal_id,
            "title": "Test Proposal",
            "proposal": f"{proposal_code.lower()}{proposal_number}",
            "state": "open",
        }
    return _make
# ...
# then in tests:
    client.get.return_value = [
        make_proposal_response(proposal_id="10"),
        make_proposal_response(proposal_id="11"),
    ]
```

2. Additionally, to avoid scattering of the same values across fixtures (factories) and test code; and potential mismatches when changing in one place, please define them as constants in relevant place.

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
```diff
@@ -55,54 +118,32 @@ def _decode_json_response(self, response):
...
+            refresh_token = response.get("refreshToken")
+            expires_in = response.get("expiresIn", 300)
+        except Exception as ex:
             log.exception(
-                "Authentication failed. Status code: %s, Response: %s",
-                response.status_code,
-                response.text,
+                "Authentication response malformed: %s",
+                ex,
             )
```
dominikatrojanowska: This is a dead code. `get` called on dictionary never rises exception. Is there anything what is expected to raise?

dominikatrojanowska: I meant the whole except clause

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
mohsen-dehesh: we added a comment that this logger will be removed later

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
rhfogh: @dominikatrojanowska I suppose it could fail if response was not a dictionary but e.g. None. It does not sound like that would be worth a separate except clause, though.

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
```diff
@@ -466,12 +466,14 @@ def _store_data_collection(self, mx_collection: dict, bl_config: dict | None):
...
 
-        dc_id, *_ = self.client.post(
+        response = self.client.post(
             "datacollections/datacollection", json=mx_collection
         )
```
dominikatrojanowska, line 479: according to [discussion](https://github.com/mxcube/mxcubecore/pull/1621/changes#r3389744544), this is my proposition:
```
            if detector:
                detector_id = detector.get("detectorId", 0)

        response = self.client.post(
            "datacollections/datacollection",
            json={**mx_collection, "detectorId": detector_id}
            if detector_id
            else mx_collection,
        )
```
And there are two propositions of changes in relevant tests.

test/test_pyispyb_data_adapter.py
```diff
@@ -0,0 +1,978 @@
...
+           }
+
+    adapter.store_beamline_setup = Mock()
+    adapter.find_detector = Mock(return_value=None)
+
+    result = adapter.store_data_collection(mx_collection, bl_config)
+
+    assert result == (10, 0)
+
+    assert mx_collection["detectorId"] is None
```
dominikatrojanowska:
```suggestion
    client.post.assert_called_once_with(
        "datacollections/datacollection", json=mx_collection
    )
    assert mx_collection["detectorId"] is None
```

test/test_pyispyb_data_adapter.py
```diff
@@ -0,0 +1,978 @@
...
+    )
+
+    adapter.find_detector.assert_called_once_with(
+        "ExampleManufacturer",
+        "ExampleModel",
+        "1x1",
+        "PixelDetector",
+    )
+
+    assert mx_collection["detectorId"] == 42
```
dominikatrojanowska:
```suggestion
    client.post.assert_called_once_with(
        "datacollections/datacollection",
        json={**mx_collection, "detectorId": 42},
    )
    assert mx_collection["detectorId"] is None
```
the last line ensures we are not modifying the input dictionary within `store_data_collection` method

dominikatrojanowska (changes requested):

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
dominikatrojanowska: I'm closing this thread, as started [one](https://github.com/mxcube/mxcubecore/pull/1621#discussion_r3427799851) referring to more lines of code :)

mxcubecore/HardwareObjects/abstract/PyISPyBDataAdapter.py
rhfogh: Looks good

mxcubecore/HardwareObjects/abstract/PyISPyBRestClient.py
dominikatrojanowska: provided here: https://github.com/mxcube/mxcubecore/pull/1621/changes/d5396b3fd610038d0f2af0cfc99f5bd443282316

PR #1628, opened 2026-06-22
Implement LNLSXRF class

pedrobenetton:


beteva (commented):
How does it work without signals and states handling?

beteva:
Hi @pedro-benetton, it is fine what you've done for running the spectrum. What I do not see is how do you handle the state and the signals, needed by the **xrf_state_handler** in the queue entry.
If I were you, though, I would inherit from AbstractXRF, as it handles the states for the queue and the signals.
It is fine to overload the start_spectrum, as it is called by the queue_model, but maybe the best would be to implement your bluesky call in **_execute_spectrum**.
If needed, spectrum_command_finished allows to start a  fitting or whatever analyse program you have (at ESRF we use PyMca) and most of all it sets the state to READY.
The handling of the directories of AbstractXRF is only optional, so you will not be using it if no data and archive directories configured.
You may want to check how is the ESRFXRFSpectrum done - our procedure is like your - in bliss and not in mxcube

mxcubecore/HardwareObjects/LNLS/LNLSXRF.py
```diff
@@ -0,0 +1,52 @@
...
+            "new_sample": False,
+        }
+
+        self._bluesky_api.execute_plan(
+            plan_name="xrf",
+            kwargs=plan_kwargs,
+        )
+
+        self._ready_event.set()
+        return
```
beteva, line 52: Is this enough to see the state changing in the GUI?

mxcubecore/HardwareObjects/LNLS/LNLSXRF.py
```diff
@@ -0,0 +1,52 @@
...
+
+    %YAML 1.2
+    ---
+    class: LNLS.LNLSXRF.LNLSXRF
+    configuration: {}
+    """
+
+    def __init__(self, name):
+        super().__init__(name)
+        self.energy = HWR.beamline.get_object_by_role("energy")
```
beteva, line 23: Do you still need this one?

marcus-oscarsson: Simply `HWR.beamline.energy`, should work

mxcubecore/HardwareObjects/LNLS/LNLSXRF.py
```diff
@@ -0,0 +1,52 @@
...
+        self._bluesky_api = HWR.beamline.get_object_by_role("bluesky")
```
marcus-oscarsson, line 24: Perhaps subclass Beamline and create something like `LNLSBeanmline` and add `bluesky`, like this you could simply do `HWR.beamline.bluesky` its perhaps nicer to type ?

Otherwise fine :)

marcus-oscarsson (approved):

PR #1629, opened 2026-06-22
Implement LNLSEnergyScan class

pedrobenetton:


mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
```diff
@@ -0,0 +1,96 @@
...
+          element:
+            - energy: K
+              symbol: Mn
+            - energy: K
+              symbol: Fe
+            - energy: K
+    """
+
+    def __init__(self, name):
+        AbstractEnergyScan.__init__(self, name)
```
beteva:
```suggestion
        super().__init__(name)
```

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
```diff
@@ -0,0 +1,96 @@
...
+        self._bluesky_api = HWR.beamline.get_object_by_role("bluesky")
```
beteva:
```suggestion
        self._bluesky_api = None
```

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def init(self):
+        self.ready_event = gevent.event.Event()
+        self.energy_scan_parameters = {}
```
beteva:
```suggestion
        self._bluesky_api = self.get_object_by_role("bluesky")
```
Normally get_object_by_role is available because from the inheriting classes. I do not know, though, you config file, maybe you need to specify HWR.beamline.
        self.energy_scan_parameters = {} is already defined in the AbstractEmergyScan,
       self.ready_event = gevent.event.Event() is already defined in the HardwareObject, so no need to redefine it.

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
```diff
@@ -0,0 +1,96 @@
...
+        edge,
+        directory,
+        prefix,
+        session_id=None,
+        blsample_id=None,
+        cpos=None,
+    ):
+        """Do the scan"""
+        if self._egyscan_task and not self._egyscan_task.ready():
+            raise RuntimeError("Scan already started.")
```
beteva: This you only need if you spawn the scan like this
`self._egyscan_task = gevent.spawn(self.do_energy_scan)` , 
which is not your case.

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self._bluesky_api.execute_plan(
+            plan_name="energy_scan",
+            kwargs=plan_kwargs,
+        )
+
+        self.emit("energyScanFinished", (self.energy_scan_parameters,))
+        self.ready_event.set()
+
+    def do_chooch(self, elt, edge, scan_directory, archive_directory, prefix):
```
beteva, line 86: Simple curiosity - if all in None, how do you get the results from the scan to be passed for a MAD data collection?

beteva (commented):
As the bluesky plan is doing everything, you do not need more hat what you've done infeed.
Few suggestion for a cleaner code. Question -  does you energy/wavelength are in BUSY state when you do the scan?

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
pedrobenetton: Hi Antonia, about the ready_event, I had to define it because if I don't, `mxcubecore/queue_entry/energy_scan.py` calls `HWR.beamline.energy_scan.ready_event.wait()` and I get `AttributeError: 'LNLSEnergyScan' object has no attribute 'ready_event'`. From what I saw at HardwareObject class, it seems to have `_ready_event` but not `ready_event`.

Perhaps we could check that because the classes PX1EnergyScan SOLEILEnergyScan and EMBLEnergyScan are doing the same

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
pedrobenetton: I'm not entirely sure, but I guess we don't plan on doing MAD experiments yet. But if ever needed I can get the results from Prefect and return the values properly at do_chooch.
Currently, Blueksy triggers a Prefect workflow that processes the data and saves it in our storage.

mxcubecore/HardwareObjects/LNLS/LNLSEnergyScan.py
pedrobenetton: Thanks, I added the Bluesky equivalent of this :)

marcus-oscarsson (approved):

marcus-oscarsson:
Looks good over all, as I mentioned for the other PR. Its perhaps a good idea to standardize the access to `bluesky` either via a `LNLSBeamline` object or by introducing something generic on `Beamline` that we all can use. I.e `control` or `control_proxy` or similar.

PR #1631, opened 2026-06-30
Handle session at LNLS

pedrobenetton:
This PR:
- Updates our icat URL
- Adjusts the functions `finalize_data_collection`, `get_proposal` and `get_base_image_directory` to work for our cases
- Every time user A logins and selects a proposal, logging out would not clear that selected proposal. If user B tries to login, either that user also has access to the proposal  previously selected by user A, or user B login would not work. What I did at `clear_session` and `login` fixes this, but maybe there is a better way to handle this.

### DIFF
--- a/mxcubecore/HardwareObjects/LNLS/LNLSLIMS.py
+++ b/mxcubecore/HardwareObjects/LNLS/LNLSLIMS.py
@@ -1,7 +1,3 @@
-"""
-A client for ISPyB Webservices.
-"""
-
 from pyicat_plus.client.main import IcatClient
 
 from mxcubecore.HardwareObjects.ICATLIMS import ICATLIMS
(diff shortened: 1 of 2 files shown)


mxcubecore/HardwareObjects/LNLS/LNLSLIMS.py
```diff
@@ -15,7 +11,7 @@ def init(self):
         self.samples = []
 
         self.icatClient = IcatClient(
-            icatplus_restricted_url="https://icat-plus.cnpem.br"
+            icatplus_restricted_url="https://icat-plus2.cnpem.br"
```
marcus-oscarsson: We should maybe make these configurable, we have done it the same way ;) but we should probably avoid having these values "hard coded"

marcus-oscarsson (approved):

marcus-oscarsson:
You can rebase and merge, think about adding a configuration for the host/url or follow up on that in another PR

PR #1632, opened 2026-07-01
[Harvester] Send data collection info to CRIMS after successful DC

jbflo:
- `Crims.send_data_collection_info_to_crims` rewritten to match the
  New CRIMS endpoint.
  
 - Harvester and HarvesterMockup both read crims_upload_url /
  crims_upload_key from config (full endpoint URL, no default)

- (existing deployments unaffected).

## Coupled PR
Requires mxcubeweb PR [link](https://github.com/mxcube/mxcubeweb/pull/2137) which wires the signal handler that calls this
function after a successful data collection.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
+++ b/mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
@@ -159,27 +159,51 @@ def get_processing_plan(
 def send_data_collection_info_to_crims(
     crims_url: str,
     crystaluuid: str,
-    datacollectiongroupid: str,
-    dcid: str,
-    proposal: str,
-    rest_token: str,
+    sample_name: str,
+    instrument_name: str,
+    facility_name: str,
+    investigation_id: str,
     crims_key: str,
 ) -> bool:
-    """Send Data collected to CRIMS
+    """Send data collection info to CRIMS.
+
     Return (bool): Whether the request failed (false) or proceed (true)
     """
-    url = (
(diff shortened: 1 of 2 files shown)


mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -159,27 +159,51 @@ def get_processing_plan(
...
+    sample_name: str,
+    instrument_name: str,
+    facility_name: str,
+    investigation_id: str,
     crims_key: str,
 ) -> bool:
-    """Send Data collected to CRIMS
+    """Send data collection info to CRIMS.
+
+    POST /ispyb_checker/api/v2/informationsforicat
```
marcus-oscarsson: I don't think you need this comment

mxcubecore/HardwareObjects/abstract/sample_changer/Crims.py
```diff
@@ -159,27 +159,51 @@ def get_processing_plan(
...
+            record_uuid,
+        )
+        logging.getLogger("HWR").debug(
+            "[Crims.send_data_collection_info_to_crims] url=%s response=%s",
+            crims_url,
+            response.text,
         )
         return True
     except requests.RequestException as e:
-        logging.getLogger("user_level_log").warning("Request to %s failed: %s", url, e)
```
marcus-oscarsson, line 182: I would add a `self.log.exception('')` here so that we have the trace of what went wrong, basically keeping it as it was.

You can add a `user_level_log.warning` but don't add `e`

marcus-oscarsson (approved):
Approved with suggested changes

PR #1639, opened 2026-07-16
[Harvester] Wait for diffractometer between centring focus and vertic…

jbflo:
diffractometer has not settled before moving the next motor
Add a wait_status_ready(60) call between the sample_focus and sample_vertical relative moves in start_harvester_centring.

### DIFF
--- a/mxcubecore/HardwareObjects/SampleView.py
+++ b/mxcubecore/HardwareObjects/SampleView.py
@@ -463,7 +463,9 @@ class SampleView(AbstractSampleView):
 
         # next two motors are not part of the centring motors
         # we move them separately
-        diffr.motors_hwobj_dict["sample_focus"].set_value_relative(_offsets[1])
+        diffr.motors_hwobj_dict["sample_focus"].set_value_relative(
+            _offsets[1], timeout=60
+        )
         diffr.motors_hwobj_dict["sample_vertical"].set_value_relative(_offsets[2])
 
         diffr.wait_status_ready(10)


mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -464,6 +464,7 @@ def start_harvester_centring(self):
         # next two motors are not part of the centring motors
         # we move them separately
         diffr.motors_hwobj_dict["sample_focus"].set_value_relative(_offsets[1])
+        diffr.wait_status_ready(60)
```
beteva: Is there a reason to do this. Normally the set_value_relative waits until the diffractometer becomes "Ready"

jbflo: without it we get :
| MD : Cannot execute command: motor is moving
2026-07-16 12:20:48,832 |HWR.ExporterMotor|ERROR  | Error when moving CentringTableVertical, re-trying
Exception: Cannot execute command: motor is moving

rhfogh: Would it not be better to pass the timeout as a parameter to set_value_relative?

@beteva 
Not sure I agree with you. In ABstractMotor the timeout defaults to 0, which according tot eh documentation means "return at once and do not wait". Maybe this is oncorrect and should eb changed (i.e. maybe the default should be None), but it looks like ther moght eb space for some clean-uip here?

marcus-oscarsson: I think that it make sense that `timeout=0` returns immediately and `timeout=None` means wait forever/no timeout. Defaulting timeout to None could lead to unexpected "blocking" situations so I don't think its a good idea to change the default value from 0 to None. The default value for `set_value` should be consistent though, but I think it is ?

marcus-oscarsson: In this case it would be nicer to pass timeout to `set_value_relative`

marcus-oscarsson:
```suggestion
```

mxcubecore/HardwareObjects/SampleView.py
```diff
@@ -464,6 +464,7 @@ def start_harvester_centring(self):
         # next two motors are not part of the centring motors
         # we move them separately
         diffr.motors_hwobj_dict["sample_focus"].set_value_relative(_offsets[1])
```
marcus-oscarsson:
```suggestion
        diffr.motors_hwobj_dict["sample_focus"].set_value_relative(_offsets[1], timeout=60)
```

marcus-oscarsson: But pay attention that this is not exactly the same as doing `diffr.wait_status_ready(60)`. It depends on what you are waiting for, just the `sample_focus` or the entire diffractometer to finish

mxcubecore/HardwareObjects/SampleView.py
rhfogh: @marcus-oscarsson I agree with you. There was one case of set)value)relative defaulting to timeout=-None, but at a second look I notice that that was in EMBL-specific code, so we can ignore it. 

Searching for e.g. 'def set_value' it does not look like every set_value function has timeout=0, or even a timeout parameter, but this does not look like something tha would eb worth investivgating right now.

mxcubecore/HardwareObjects/SampleView.py
jbflo: > But pay attention that this is not exactly the same as doing `diffr.wait_status_ready(60)`. It depends on what you are waiting for, just the `sample_focus` or the entire diffractometer to finish

I think any motor move requires the whole diffractometer to be ready first, and when one motor finishes, the diffractometer is ready again. so agreed — I'll pass timeout=60 directly to set_value_relative on sample_focus.

marcus-oscarsson (approved):

PR #1640, opened 2026-07-16
[Harvester] Set crystal UUID via set_id so get_id() returns it correctly

jbflo:
sample.id = x_tal_uuid was setting a dynamic attribute, causing get_id() to return None since it reads self._id

Fix: Add public set_id() to Component
 use sample.set_id(x_tal_uuid) so get_id() returns the crystal UUID consistently

### DIFF
--- a/mxcubecore/HardwareObjects/EMBLFlexHarvester.py
+++ b/mxcubecore/HardwareObjects/EMBLFlexHarvester.py
@@ -114,7 +114,7 @@ class EMBLFlexHarvester(EMBLFlexHCD):
             img_target_y = self._harvester_hwo.get_image_target_y(x_tal_uuid)
 
             # Enrich sample object in MXCuBE
-            sample.id = x_tal_uuid
+            sample.set_id(x_tal_uuid)
             sample.state = ha_sample_states[index]
 
             sample._set_image_url(img_url)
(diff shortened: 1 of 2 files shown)


mxcubecore/HardwareObjects/EMBLFlexHarvester.py
```diff
@@ -114,7 +114,7 @@ def get_sample_list(self) -> List[Any]:
             img_target_y = self._harvester_hwo.get_image_target_y(x_tal_uuid)
 
             # Enrich sample object in MXCuBE
-            sample.id = x_tal_uuid
+            sample._id = x_tal_uuid
```
marcus-oscarsson: Its a bit odd to set `_id` from outside of sample, is there no `set_id` or similar that can be used ?

jbflo: > Its a bit odd to set `_id` from outside of sample, is there no `set_id` or similar that can be used ?

 pragmatic but accessing a protected attribute from outside the class is odd i aggreed.
 
 But I checked  — there's no set_id(). The options are:

Keep the pragmatic fix or 
Add set_id() to Component.py — cleanest, fixes the root cause:

def set_id(self, id):
    self._id = id

marcus-oscarsson: Ok, go AI go, do the fix ;)

jbflo: > Ok, go AI go, do the fix ;)

 fixed ! 🤖

mxcubecore/HardwareObjects/abstract/sample_changer/Component.py
```diff
@@ -32,6 +32,10 @@ def get_id(self):
         """
         return self._id
 
+    def set_id(self, id):
+        """Sets the unique ID of an element - if was not scanned from the real object"""
```
marcus-oscarsson:
```suggestion
        """Sets the ID of the component"""
```

marcus-oscarsson: An ID is unique by definition and we don't know what code that sets this ID, it could have been scanned from somewhere.

PR #1643, opened 2026-08-03
Add AbstractDoseEstimation

mockoocy:
This is part of the dose estimation capability that we have presented at the developer's meeting in Berlin

This HardwareObject is meant to provide a way to estimate absorbed
dose by crystal during oscillation scans.

Includes abstract definition along implementations at BioMAX and a mockup one.

Note on the BioMAX implementation:
Calculations inside the module have been based on the old dose estimation
code used at BioMAX and is a subject to change in the future.
It's correctness is not guaranteed, nonetheless it serves as a reference
implementation to iterate on afterwards.

### DIFF
--- a/mxcubecore/HardwareObjects/MAXIV/BioMAX/dose_estimator.py
+++ b/mxcubecore/HardwareObjects/MAXIV/BioMAX/dose_estimator.py
@@ -1,0 +1,172 @@
+from typing import TYPE_CHECKING
+
+from pydantic.v1 import BaseModel
+
+from mxcubecore import HardwareRepository as HWR
+from mxcubecore.HardwareObjects.abstract.AbstractDoseEstimator import (
+    AbstractDoseEstimator,
+    DoseEstimateParameters,
+    DoseEstimation,
+    DoseEstimationError,
+    DoseEstimationOk,
+    ExperimentalGoal,
+)
+
+if TYPE_CHECKING:
+    from mxcubecore.HardwareObjects.MAXIV.BioMAX.BIOMAXFlux import BIOMAXFlux
+    from mxcubecore.HardwareObjects.MAXIV.Energy import Energy
+
+
+class _PolynomialSegment(BaseModel):
+    """
+    Represents the coefficients of piecewise cubic regression.
+
+    Flux value for given energy is being estimated using
+    parameters gathered during beamline operation at BioMAX.
+    Estimated energy is calculated differently for specific
+    thresholds.
+    """
+
+    max_energy_ev: int
+    coeffs: list[float]
+
+
+class DoseEstimator(AbstractDoseEstimator):
+    """Implementation of dose estimation for BioMAX.
+
+    This is an example YAML configuration
+    .. code-block:: yaml
+        class: MAXIV.BioMAX.dose_estimator.DoseEstimator
+        configuration:
+            flux_coefficients:
+                - max_energy_ev: 9060
+                    coeffs: [1.40919e14, -6.95909e10, 1.10638e7, -553.774]
+                - max_energy_ev: 12610
+                    coeffs: [1.40919e14, -6.95909e10, 1.10638e7, -553.774]
+                - max_energy_ev: 18750
+                    coeffs: [1.59363e14, -2.78613e10, 1.69424e6, -35.0571]
+                - max_energy_ev: 20400
+                    coeffs: [1.00206e12, 2.65477e9, -240059, 5.44578]
+                - max_energy_ev: 24000
+                    coeffs: [5.06378e14, -6.52503e10, 2.81565e6, -40.6289]
+            experimental_goals:
+                cryo_highres:
+                    label: Cryo High Resolution
+                    type: resolution_dependent
+                    mgy_per_angstrom: 10
+                room:
+                    label: Room Temperature
+                    type: static
+                    mgy: 0.2
+                cys_cys:
+                    label: Cys-Cys
+                    type: static
+                    mgy: 2
+                s_sad:
+                    label: S-SAD
+                    type: static
+                    mgy: 5
+                mad_sad:
+                    label: MAD/SAD
+                    type: static
+                    mgy: 6
+    """
+
+    class HOConfig(BaseModel):
+        flux_coefficients: list[_PolynomialSegment]
+        experimental_goals: dict[str, ExperimentalGoal]
+
+    def __init__(self, name: str) -> None:
+        self._energy_hwo: Energy | None = None
+        self._flux_hwo: BIOMAXFlux | None = None
+        super().__init__(name)
+
+    def init(self) -> None:
+        super().init()
+
+        self._energy_hwo = HWR.beamline.energy
+        self._flux_hwo = HWR.beamline.flux
+
+    def _get_flux_t_current(self, energy_ev: float) -> float:
+        buckets = self._config.flux_coefficients
+        for bucket in buckets:
+            # in here flux_coefficients is assumed to always have
+            # the max_energy_ev thresholds in ascending order.
+            if energy_ev <= bucket.max_energy_ev:
+                return sum(
+                    coeff * energy_ev**exponent
+                    for exponent, coeff in enumerate(bucket.coeffs)
+                )
+        max_energy_ev = max(bucket.max_energy_ev for bucket in buckets)
+        err_msg = f"Energy out of bounds. Max available value is {max_energy_ev} eV."
+        raise ValueError(err_msg)
+
+    def estimate_dose(  # noqa: PLR0911  many returns make it clearer :)
+        self,
+        params: DoseEstimateParameters,
+    ) -> DoseEstimation:
+        flux = self._flux_hwo.get_value()  # ph/s
+
+        if flux <= 0:
+            return DoseEstimationError(msg="No flux detected, can't estimate dose.")
+
+        beamline_flux_density = self._flux_hwo.flux_density
+
+        if beamline_flux_density == -1:
+            return DoseEstimationError(
+                msg="Flux has not been measured. Please measure it first"
+            )
+        if beamline_flux_density == 0:
+            return DoseEstimationError(
+                msg="No beam has been detected during flux measurements!"
+            )
+
+        flux_density_energy = self._flux_hwo.flux_density_energy
+
+        if flux_density_energy <= 0:
+            return DoseEstimationError(msg="Energy at the time of measurement unknown.")
+
+        beamline_energy = self._energy_hwo.get_value()  # keV
+
+        try:
+            flux_at_beamline_energy = self._get_flux_t_current(
+                energy_ev=beamline_energy * 1_000
+            )
+        except ValueError as ex:
+            return DoseEstimationError(msg=str(ex))
+
+        if flux_at_beamline_energy <= 0:
+            return DoseEstimationError(msg="Flux model predicts value <= 0")
+
+        # This is a relative error term for the flux measurements.
+        flux_scale = flux_at_beamline_energy / flux
+        try:
+            user_flux = (
+                self._get_flux_t_current(energy_ev=params.energy_kev * 1_000)
+                / flux_scale
+            )
+        except ValueError as ex:
+            return DoseEstimationError(msg=str(ex))
+
+        beam_size_x, beam_size_y = HWR.beamline.beam.get_beam_size()  # mm
+
+        user_flux_density = (user_flux * params.transmission_pct / 100) / (
+            beam_size_x * beam_size_y * 1e6
+        )
+        wavelength = self._energy_hwo.calculate_wavelength(energy=params.energy_kev)
+        # The 2000 constant comes from old code.
+        dose_rate = user_flux_density / (2000 / wavelength / wavelength)
+        estimated_dose_gy = dose_rate * params.exp_time_s * params.num_images
+        max_images = (
+            (params.dose_limit_mgy * 1_000_000) / dose_rate / params.exp_time_s
+            if params.dose_limit_mgy
+            else None
+        )
+        return DoseEstimationOk(
+            dose_mgy=estimated_dose_gy / 1_000_000,
+            max_images=max_images,
+        )
+
+    @property
+    def experimental_goals(self) -> dict[str, ExperimentalGoal]:
+        return self._config.experimental_goals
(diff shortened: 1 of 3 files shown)


rhfogh (dismissed):
Hmmm. This may be mostly a question of style, but I would not be that happy having to write an implementation to this abstract class. Could we discuss this?

Mainly I think that this very much exaggerates using Pydantic. First of all DoseEstimateParameters. The normal Python mechanism is that function parameters are defined in the function definition. Here they are defined in a Pydantic class, in a different file. Does that not obfuscate more than it helps? There are cases where data objects make sense, e.g. for the data needed to define an acquisition, that have some kind of independent objecthood and that are used in multiple contexts. But these are (just look at the name) defined simply by being the input parameters to a function. Why the extra packaging?  

Similarly the various classes for DoseEstimation results and goals. What is gained by this rather cumbersome modelling just to group a few attributes together? 

Are we aiming for a situation where the input and output of every function in MXCuBE is separately defined as a Pydantic object in a separate file? Is that really a good idea? And if that is not what we want, how do you decide which contexts can make do  with standard Python, and which need the all-Pydantic treatment? I am generally in favour of data modelling, but part of the cost is that it takes more thought and work to set up a comprehensive model up front, and leads to more rigid code. What is the gain, here?

mockoocy:
> Why the extra packaging?

The Pydantic model is used in a few places, one of them being the backend adapter, which essentially calls the `estimate_dose` function with the same `params` object. It is necessary there because it validates the arguments coming from the form.

Using the same object in `mxcubecore` was mainly a matter of convenience, although I agree that this may introduce too much coupling to Pydantic. The same argument applies to the return types: these Pydantic models make the results very convenient for the frontend to consume, although `mxcubecore` should not really care about what the frontend prefers…

It also occurs to me now that the name `DoseEstimationParameters` may be too generic. The object specifically carries parameters supplied by the user. We may need to distinguish between what the user wants to set and the current beamline state. For example, the BioMAX implementation uses both the user-supplied energy and the current beamline energy to calculate an error-correction term for flux interpolation.

The Pydantic models could be removed from `mxcubecore` fairly easily, for example by using keyword-only arguments for `estimate_dose`, raising an exception such as `DoseEstimationUnavailableError(...)` for expected failures, and keeping the Pydantic models in `mxcubeweb`.

rhfogh:
@mockoocy OK, that is a good explanation. We can still discuss how much the front end should influence the back end (and I still have my opinions, no surprise there) but there are certainly good reasons on both sides. Can we see what the others say, before this is merged? Meanwhile, I will remove my 'modifications required', so as not to be technically blocking.

mxcubecore/HardwareObjects/abstract/AbstractDoseEstimator.py
```diff
@@ -0,0 +1,84 @@
+from abc import abstractmethod
+from typing import Annotated, Literal, Union
+
+from pydantic import BaseModel, Field
+
+from mxcubecore.BaseHardwareObjects import HardwareObject
+
+
+class DoseEstimationOk(BaseModel):
```
marcus-oscarsson, line 9: Would it be possible to merge `DoseEstimationOk` and `DoseEstimationError` to simply `DoseEstimation` containing i.e `status`, `msg`, `dose_mgy`, `max_images`, the later two being None incase of error ? Or containing them within a field called `result` ?

mxcubecore/HardwareObjects/MAXIV/BioMAX/dose_estimator.py
```diff
@@ -0,0 +1,172 @@
...
+                    coeff * energy_ev**exponent
+                    for exponent, coeff in enumerate(bucket.coeffs)
+                )
+        max_energy_ev = max(bucket.max_energy_ev for bucket in buckets)
+        err_msg = f"Energy out of bounds. Max available value is {max_energy_ev} eV."
+        raise ValueError(err_msg)
+
+    def estimate_dose(  # noqa: PLR0911  many returns make it clearer :)
+        self,
+        params: DoseEstimateParameters,
```
marcus-oscarsson, line 106: I would agree with @rhfogh for `params`, is there a reason for not simply defining the method with all the arguments needed ?

mxcubecore/HardwareObjects/MAXIV/BioMAX/dose_estimator.py
```diff
@@ -0,0 +1,172 @@
...
+    ) -> DoseEstimation:
+        flux = self._flux_hwo.get_value()  # ph/s
+
+        if flux <= 0:
+            return DoseEstimationError(msg="No flux detected, can't estimate dose.")
+
+        beamline_flux_density = self._flux_hwo.flux_density
+
+        if beamline_flux_density == -1:
+            return DoseEstimationError(
```
marcus-oscarsson, line 116: I'm not against using `DoeseEsitmationError` in this way, it can be an elegant way to handle an expected error condition. We tend to use exceptions for both "unexpected" and expected error states. Why did you chose to with dose `DoeseEsitmationError` ? 

If you would like to distinguish between exception and an "expected" error. I would merge, as I mentioned above, `DoseEstimationOk` and `DoseEstimationError`.

Otherwise the easiest would perhaps be to raise and let the calling code handle the error ?

marcus-oscarsson:
I understand the idea of using the `PyDantic` models like this to make the implementation in `mxcubeweb` easier/lighter. But as you both point out after the discussion above I also think that moving this part to `mxcubeweb` is maybe more appropriate. The method `estimate_dose` should take all the parameters needed for the estimation, any extra argument can be passed as `kwarg` or handled in some site specific override. 

This otherwise looks great to me :), @rhfogh, is there otherwise anything else in the structure that you would like to discuss.

From what I see, I think we can merge this if the `PyDantic` bits we mentioned above an be moved to `mxcubeweb` ?

rhfogh:
If the pydantic stuff is moved to mxcubeweb Iit is all fine with me. Not that I really like the style, but the web front end has its own legitimate requirements.

beteva:
Sorry, coming a bit late in the discussion (long holidays). In order to agree on the parameters and what comes from the form, hence the  pydantic model, could we narrow down the number of parameters. It seems to me that we can estimate the dose, based on the dose rate and the total exposure time. Is this too simplistic?

rhfogh:
Could we integrate this with the dose rate PR, #1619. I particularly would point to the conceptual organisation proposed in recent comments.

mxcubecore/HardwareObjects/abstract/AbstractDoseEstimator.py
mockoocy: I think I would just raise some named exception (e.g. `DoseEstimationError`) here and have a proper field with no `None` values there. Thus having just one return type (I'd make it a `dataclass` probably or a `TypedDict` for the sake of readability).

mxcubecore/HardwareObjects/MAXIV/BioMAX/dose_estimator.py
mockoocy: Just convenience when writing code for the web frontend, I will just use bunch of kwargs there, someting like
```diff
- def estimate_dose(self, params: DoseEstimateParameters): ...
+ def estimate_dose(self, *, exp_time_s: float, num_images: int, ...): ...
```

mxcubecore/HardwareObjects/MAXIV/BioMAX/dose_estimator.py
mockoocy: I will just go with something proposed in https://github.com/mxcube/mxcubecore/pull/1643#issuecomment-5191860993 and raise indeed. It is a bit more "python-like" :), I will move the web niceties to mxcubeweb.

mockoocy:
Hi all, 
sorry for late reply, but I was holiday.

> @rhfogh:  Could we integrate this with the dose rate PR, https://github.com/mxcube/mxcubecore/pull/1619. I particularly would point to the conceptual organisation proposed in recent comments.

I suppose it could be up to facility to use `HWR.beamline.dose_rate` in their `estimate_dose` implementation? I think we could just leave `AbstractDoseEstimator` oblivious to other hardware objects and probably have some default implementation that would return `HWR.beamline.dose_rate * exp_time_s * num_images`.
Could be part of separate PR / bundled with #1619.

> @beteva: Sorry, coming a bit late in the discussion (long holidays). In order to agree on the parameters and what comes from the form, hence the pydantic model, could we narrow down the number of parameters. It seems to me that we can estimate the dose, based on the dose rate and the total exposure time. Is this too simplistic?

Yes, in principle, given dose rate it is quite simple to get the proper dose estimation later on.
The parameters that are passed into the function that does the estimation would correspond to what user set in the "standard" Data Collection form on the frontend. That is -- they may want to change some parameters (e.g. transmission, energy) and different dose rate (and subsequently dose estimate) could be then expected. In other words, the dose estimate refers to a "requested" beamline state rather than the current one (thus the need to tell what would be expected).

For example, BioMAX implementation tries to estimate flux at energy level provided by the user. To do so it calculates a relative error term between a real flux value and one coming from some cubic regression. This is then used to infer what would be the flux quantity given user-supplied energy.

I see a reason that we could perhaps just operate on beamline values and just prohibit dose estimation when user requests different energy than the one flux has been measured for. It could probably save us some pain, and we could safely just send `transmission, num_images, exp_time_s` to `mxcubecore`.

Sorry for this long wall of text, I hope it is digestible :-)

rhfogh:
I think the approach here is not ideal because it tries to put the entire calculation into a single function/object. The problem with that is that if at some later date you want to make a change to some of the parts or use different assumptions it all gets very complicated, as you would have to decompose it. I would propose the following:

The top formula is 
`dose = get_dose_rate_per_photon_per_mmsq * effective_flux_density * relative_crystal_sensitivity * time`
That should be simple enough that you could inline it wherever you want and would not need a separate function for it. We could then have separate, overridable functions to calculate each of the four terms in the equation, and put those functions into abstract objects as we find best.


We would need functions:
`get_dose_rate_per_photon_per_mmsq(energy)`
Which is a conversion factor that depends only on the energy. 

`get_effective_flux_density(transmission=None)`
The effective flux density could in principle vary from point to point in the crystal. If you use the _average_ flux density, it depends on transmission, flux, and beam area, and if there is a need for it all of these could optionally be passed in (defaulting to the current values) as in 
`get_effective_flux_density(transmission=None, flux=None, beam_area=None)`

See the current AbstractFlux class for the current implementation of these two functions.

If we ever get to using the real (non-top-hat) beam profile and the actual crystal shape to calculate dose, the effective flux density would get a lot more complicated, and there would have to be operational decisions as to which value to use, or whether to calculate some kind of weighted integral. For a simple case (large crystal, narrow beam, approximation) see GphlWorkflow.maximum_dose_rate(). Anyway, this being so dependent on assumptions it might be better not to put it into a top-level combined function (I am quoting Gleb Bourenkov here)

The relative crystal sensitivity is in principle wavelength dependent. For now it can be assumed to be a constant passed in with the diffraction plan (I am not aware of any better implementation).

The relevant time is easy to calculate, and can be left to be done locally.

I think this would give a nice, clean separation between the various terms and make it easier to understand and maintain. Note that we might have not only calculations from crystal shape but also multi-sweep, multi-wavelength experiments, so it is better not to bundle all the calculations up.

mockoocy:
> I think the approach here is not ideal because it tries to put the entire calculation into a single function/object. The problem with that is that if at some later date you want to make a change to some of the parts or use different assumptions it all gets very complicated, as you would have to decompose it. I would propose the following:
> 
> The top formula is `dose = get_dose_rate_per_photon_per_mmsq * effective_flux_density * relative_crystal_sensitivity * time` That should be simple enough that you could inline it wherever you want and would not need a separate function for it. We could then have separate, overridable functions to calculate each of the four terms in the equation, and put those functions into abstract objects as we find best.
> 
> We would need functions: `get_dose_rate_per_photon_per_mmsq(energy)` Which is a conversion factor that depends only on the energy.
> 
> `get_effective_flux_density(transmission=None)` The effective flux density could in principle vary from point to point in the crystal. If you use the _average_ flux density, it depends on transmission, flux, and beam area, and if there is a need for it all of these could optionally be passed in (defaulting to the current values) as in `get_effective_flux_density(transmission=None, flux=None, beam_area=None)`
> 
> See the current AbstractFlux class for the current implementation of these two functions.
> 
> If we ever get to using the real (non-top-hat) beam profile and the actual crystal shape to calculate dose, the effective flux density would get a lot more complicated, and there would have to be operational decisions as to which value to use, or whether to calculate some kind of weighted integral. For a simple case (large crystal, narrow beam, approximation) see GphlWorkflow.maximum_dose_rate(). Anyway, this being so dependent on assumptions it might be better not to put it into a top-level combined function (I am quoting Gleb Bourenkov here)
> 
> The relative crystal sensitivity is in principle wavelength dependent. For now it can be assumed to be a constant passed in with the diffraction plan (I am not aware of any better implementation).
> 
> The relevant time is easy to calculate, and can be left to be done locally.
> 
> I think this would give a nice, clean separation between the various terms and make it easier to understand and maintain. Note that we might have not only calculations from crystal shape but also multi-sweep, multi-wavelength experiments, so it is better not to bundle all the calculations up.

Thank you, for the detailed answer, I feel convinced to the points you've made there :-).
Since the formula will be the same for all the dose estimates, we can use the existing method for retrieving the average flux density. I'd probably defer the question of getting the dose rate to a separate Hardware Object, something akin to the proposal from #1619. 
I guess the calculation itself could be moved to the client-side, with a way to supply the dose rate and average flux density. For the crystal sensitivity perhaps we could go with some kind of default constant and let user provide their own value -- if this is a realistic scenario.

It would probably be nicer to do things this way, at least here at MAX IV most of the work in the dose estimator implementation was there for the calculation of the dose rate itself.

marcus-oscarsson:
Hm, I don't really see an issue with all of this being in one HardwareObject with specializations if needed. I think that that object, after some thought, could be `AbstractFlux`. We then have to agree on the basic functions/functionality it provides. We have to take a closer look at what those functions are but what's mentioned above

```
dose = get_dose_rate_per_photon_per_mmsq * effective_flux_density * relative_crystal_sensitivity * time
get_dose_rate_per_photon_per_mmsq(energy)
```
Seems like a good start. The method `get_average_flux_density` already exists but you also mention `get_effective_flux_density(transmission=None, flux=None, beam_area=None)`. We have to look at the details but I'm not entirely certain about the naming "effective".

We can then have an additional method that is called something like `dose_estimate`, that for MAXIV then would be more or less what's implemented here and each site/beamline implements their `dose_estimate` variant. I don't mind splitting this into two HardwareObjects AbstractFlux and AbstractDoesEstimation but I don't really see the benefit of doing it ?

PR #1644, opened 2026-08-04
Changes in HOs to adapt to yaml configuration files

beteva:
Make the properties of type list, dict, tupple compatible for both xml and yaml configuratuion


rhfogh (approved):
I cannot really evaluate the details, but it looks like a good step in the right direction.

PR #1647, opened 2026-08-10
1646 improve synchronization of samples

### LINKED ISSUE #1646: Improve synchronization of samples
Currently, the samples and their attached files are downloaded from the server on a per-sample basis. It takes approximately 1 second to synchronize each sample. 
When a large number of samples are loaded into the SC, the resulting synchronization time becomes unacceptably long.

antolinos:


pyproject.toml
```diff
@@ -63,6 +63,7 @@ python-ldap = "^3.4.5"
 requests = "^2.33.0"
 colorama = "^0.4.6"
 pyicat-plus = {version = "^1.0.0", optional = true}
+icat-plus-client = {version = "4.93.3", optional = true}
```
bodinm, line 66: I'll be honest, I was a bit disappointed to see `icat-plus-client` show up here as a new dependency, we've put real effort into standardizing on `pyicat-plus` as the one client for ICAT across BLISS, Daiquiri, and MXCuBE, etc. and this quietly moves us away from that.

A few reasons that standardization matters to me:
- Config (URL, auth, timeouts) lives in one place instead of being duplicated and potentially drifting out of sync
- Error handling, retries, and logging stay consistent across every tool, instead of depending on which client happens to be used
- No extra generated package to track for releases and dependencies
- Anyone new only has to learn one client, not decide between two

All the endpoints needed here are already in pyicat-plus, or easy to add/adapt. Happy to help add whatever's missing.

antolinos: Hello,

This PR aims to improve the current implementation, which was heavily based on the ISPyB developments and relies on dictionaries for data exchange, as pyicat-plus currently does. I think we should move towards using proper models wherever possible.

The main drawback preventing us from using pyicat-plus today is precisely the lack of models.

icat-plus-client provides generated models and API methods based on the ICAT+ OpenAPI specification, and is automatically documented and published to PyPI whenever the specification changes. This is much more convenient and less error-prone than implementing the methods manually in pyicat-plus.

If we want to expose ICAT+ access through pyicat-plus, I would recommend making icat-plus-client a dependency rather than reimplementing everything.

That said, if pyicat-plus provides the required models and methods, regardless of how they are implemented, I have no strong objection to using them instead.

As with almost everything, there are pros and cons that need to be carefully evaluated, both from the ESRF perspective and in the context of the wider collaboration. In particular, I’m thinking about partners who use ICAT but cannot use pyicat-plus because it is too ESRF-centric.

It’s something we can discuss in person with everyone involved.

Cheers,
A.

bodinm: I think integrating` icat-plus-client` into `pyicat-plus` is the right direction (https://gitlab.esrf.fr/icat/pyicat-plus/-/work_items/105), and I want to move forward with it. But given how much this PR already replaces (login, investigations, parcels, samples, downloads all now go through` icat_plus_client `directly), I don't think it's safe to assume that integration happens 'later'. My preference would be to sequence it the other way: do the pyicat-plus/icat-plus-client integration first , then rebase this on top of it, so ICATLIMS ends up talking to a single client. I just wanted my concern on record before this goes in.

bodinm: Also worth noting, separately from the client discussion: this PR bundles several distinct changes: the icat-plus-client/model migration, the start/end date fix, the sample synchronization improvement.
Would it be possible to split these into separate PRs, especially the model/client migration? It would make each change easier to review and revert independently, and would let the sync fix move forward without being tied to the client architecture discussion.

marcus-oscarsson:
This has a conflict from @bodinm's last PR. Coudl you please have a look, @antolinos or @bodinm ?

marcus-oscarsson (approved):

PR #1648, opened 2026-08-13
ICATLIMS: use the pyicat-plus client and use models for ICAT fields

woutdenolf:
This is a PR on another PR:

- Use `icat-plus-client` for REST calls through `pyicat-plus`.
- Use `esrf-icat-definitions` models for created ICAT dataset metadata.

Needs (edit: deployed to pypi)

- https://gitlab.esrf.fr/icat/pyicat-plus/-/merge_requests/253
- https://gitlab.esrf.fr/icat/hdf5-master-config/-/merge_requests/233

### DIFF
--- a/mxcubecore/HardwareObjects/ICATLIMS.py
+++ b/mxcubecore/HardwareObjects/ICATLIMS.py
@@ -7,16 +7,9 @@ from pathlib import Path
 from typing import Any, List, Optional
 from zoneinfo import ZoneInfo
 
-from icat_esrf_definitions.models import IcatDatasetParameters
-from icat_plus_client.exceptions import ApiException, ForbiddenException
-from icat_plus_client.models.investigation import Investigation
-from icat_plus_client.models.item import Item
-from icat_plus_client.models.item_experiment_plan_inner import ItemExperimentPlanInner
-from icat_plus_client.models.resourceinformation import Resourceinformation
-from icat_plus_client.models.sample import Sample
-from icat_plus_client.models.sampleinformation import Sampleinformation
-from icat_plus_client.models.session import Session as IcatSession
 from pydantic import ValidationError
+from pyicat_plus import errors as icat_errors
+from pyicat_plus.client import models as icat_models
 from pyicat_plus.client.main import IcatClient
 
 from mxcubecore import HardwareRepository as HWR
@@ -93,7 +86,9 @@ class ICATLIMS(AbstractLims):
             ),
         ]
 
-    def _create_icat_session(self, user_name: str, password: str) -> IcatSession:
+    def _create_icat_session(
+        self, user_name: str, password: str
+    ) -> icat_models.AuthSession:
         try:
             logger.debug(f"Authenticating {user_name}")
             icat_session = self.icatClient.do_log_in(
@@ -101,10 +96,10 @@ class ICATLIMS(AbstractLims):
                 username=user_name,
                 plugin=self.authentication_icat_plugin,
             )
-        except ForbiddenException as e:
+        except icat_errors.ForbiddenException as e:
             logger.error(f" Error occurred while authenticating. Access forbidden {e}")
             raise
-        except ApiException as e:
+        except icat_errors.ApiException as e:
             logger.error(f"Error occurred while authenticating {user_name}: {e}")
             raise
         return icat_session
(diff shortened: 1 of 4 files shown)


mxcubecore/HardwareObjects/ICATLIMS.py
```diff
@@ -244,14 +171,12 @@ def get_proposals_by_user(self, user_name):
...
-    def _get_loaded_pucks(
-        self, icat_token: str, investigation_id: str
-    ) -> List[LoadedPuck]:
+    def _get_loaded_pucks(self, investigation_id: str) -> List[LoadedPuck]:
         """Return pucks with a defined sample changer location."""
         self.parcels = []
         try:
-            self.parcels = self.tracking_api_instance.tracking_session_id_parcel_get(
-                icat_token, investigation_id=investigation_id
+            self.parcels = self.icatClient.get_parcels_by(
```
marcus-oscarsson, line 173: Just curious, is the token not needed at all for this operation or is the token kept internally in the `IcatClient` ?

marcus-oscarsson:
Ok, we just need to have a look at / test the use of IcatClient. We have implemented something where each user has a IcatClient on `esrf-develop` as we have had some issues with icat session synchronization and the handling of multiple users.

mxcubecore/HardwareObjects/ICATLIMS.py
woutdenolf: The token is kept by `IcatClient` and replaced on every `do_log_in` call.

marcus-oscarsson:
Is there any requirement for this to be used with the currently deployed version of ICAT or can the actual deployment to the beamlines wait ?

woutdenolf:
This is a PR on @antolinos PR. Not sure how urgent his is.

Note that this uses `pyicat-plus >=3.0.0rc1`

marcus-oscarsson:
I, see I missed that detail. Perfect, Ill check with @antolinos and Ill let him merge. Otherwise, looks good to me.

mxcubecore/HardwareObjects/ICATLIMS.py
marcus-oscarsson: Great !

marcus-oscarsson:
Thanks

antolinos:
I do not see any major problems with this PR, but it needs to be thoroughly tested. 

Regarding the libraries, it was made clear to me that the fewer libraries we add as dependencies, the better. We now have not only icat-plus-client and pyicat-plus, but also icat-esrf-definitions.

Do not get me wrong: I am of the opinion that libraries are meant to be used, and that the smaller and more clearly scoped they are, the better.

This PR replaces the constants that were refactored into IcatDatasetParameters. I thought it would be preferable to use constants instead of adding icat-esrf-definitions, because if the goal is to delegate the ingestion to BLISS, it makes less sense to me to have both pyicat-plus and icat-esrf-definitions in MXCuBE, since this functionality will eventually be handled by BLISS.

Having said that IcatDatasetParameters is much more elegant and less error-prone, but both the constants and IcatDatasetParameters might just be temporary solutions.

Nevertheless, I would again suggest discussing it all together just to check that we are all in the same page but **the PR looks good to me because it improves its parent PR.**

woutdenolf:
> it makes less sense to me to have both pyicat-plus and icat-esrf-definitions in MXCuBE

You have it anyway since it is a dependency of `pyicat-plus`, in the same way `icat-plus-client` is a dependency of `pyicat-plus`. The question is, do we need to import them in addition to importing `pyicat-plus`.

You would need to import them directly for

1. output type annotations (Pydantic models)
2. simpler input definitions (Pydantic models)

Every library we import also goes in `pyproject.toml`.

- `pyicat-plus`: yes for communication with the Data Portal services
- `icat-esrf-definitions`: yes for the models
- `icat-plus-client`: yes for the models
- `esrf-ontologies`: for defining technique metadata, I thought we had this as well but apparently not

woutdenolf:
In other words

- `pyicat-plus` is for the operational logic
- the other libraries are for the parametrization (input/output Pydantic models)

woutdenolf:
As discussed with @antolinos we should import the models from `pyicat-plus` and not directly from `icat-plus-client`. I will make those modifications.

woutdenolf:
@antolinos we now only depend on `pyicat-plus` as discussed. Ready for review.

pyproject.toml
```diff
@@ -62,8 +62,7 @@ PyTango = "^9.3.6"
 python-ldap = "^3.4.5"
 requests = "^2.33.0"
 colorama = "^0.4.6"
-pyicat-plus = {version = "^1.0.0", optional = true}
-icat-plus-client = {version = "4.93.3", optional = true}
+pyicat-plus = {version = ">=3.0.0rc3", optional = true, allow-prereleases = true}
```
woutdenolf, line 65: Note: will stay as pre-release until properly tested in production.

PR #1650, opened 2026-08-14
Rename actual metadata parameters

bodinm:


marcus-oscarsson (approved):

marcus-oscarsson:
Sure, if you say so :). Is this dependent on some other change or update on the ICAT side ?

bodinm:
yes I'll rename the metadata in the ICAT DB, please just let met know when this can be deployed, and I'll do the changes in ICAT. Thanks!

marcus-oscarsson:
Ok, we will deploy for the startup so, sometime the next two weeks

PR #1652, opened 2026-08-18
Update BlissNState

Phantominumm:
- Register the actuator via bliss_proxy.hardware.register() before fetching it
- Remove the _pending_target optimistic-UI logic and the multi-step value/state fallback heuristics
- Rely on the non-blocking move()/move_in()/move_out() callback instead of raising on error
- Add wait_ready() to block on the stored callback, mirroring BlissMotor

### DIFF
--- a/mxcubecore/HardwareObjects/BlissNState.py
+++ b/mxcubecore/HardwareObjects/BlissNState.py
@@ -202,9 +202,8 @@ class BlissNState(AbstractNState):
         Returns:
             (Enum): "ValueEnum" with predefined values.
         """
-        # Always load values from config first so VALUES is never empty
-        # (this is also called from AbstractNState.init before device_type is set)
-        super().initialise_values()
+        if self.device_type == "actuator":
+            super().initialise_values()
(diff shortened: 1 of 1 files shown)


mxcubecore/HardwareObjects/BlissNState.py
```diff
@@ -129,63 +145,77 @@ def _set_value(self, value):
...
     def initialise_values(self):
         """Get the predefined values. Create the VALUES Enum
         Returns:
             (Enum): "ValueEnum" with predefined values.
         """
-        if self.device_type == "actuator":
-            super().initialise_values()
+        # Always load values from config first so VALUES is never empty
+        # (this is also called from AbstractNState.init before device_type is set)
+        super().initialise_values()
```
beteva: VALUES is never empty. It has at least UNKNOWN member, defined in AbstractNState.
```suggestion
        if self.device_type == "actuator":
            super().initialise_values()
```
Revert as it was before/

marcus-oscarsson (approved):

PR #1654, opened 2026-08-18
Update BlissRontecMCA

Phantominumm:
- Call bliss_proxy.hardware.register(actuator_name) before get_object(), matching the other Bliss hardware objects


beteva:
This is a deprecated HardwareObject. In principle it works, but is not used any more.

marcus-oscarsson:
Closing as BlissRontecMCA is now deprecated, @beteva we should perhaps then also follow up with a PR that removes the object ?

beteva:
Close as HO is deprecated.

PR #1656, opened 2026-08-18
Update BlissProxy

Phantominumm:
BlissProxy wraps blissclient (BlissClient/Hardware/Session) to give the other
Bliss* hardware objects (BlissMotor, BlissNState, BlissShutter, BlissRontecMCA)
a single access point to BLISS session devices over the REST API.

- init() waits for the BLISS REST service to report ready (polls /api/object),
  creates the BlissClient, registers connect/disconnect callbacks, and eagerly
  loads every known session object into a local cache (_load_known_objects)
- get_object(name) serves objects from that cache; if a name isn't cached
  (e.g. its type isn't pre-registered, as for BlissRontecMCA) it falls back to
  a direct blissclient lookup, and raises KeyError listing available objects
  if the name truly doesn't exist
- get_object() blocks callers until init() has finished (or times out) so
  hardware objects created before BlissProxy is ready still get served
- hardware/session properties expose the underlying blissclient Hardware and
  Session objects directly, used by ESRFBeamlineActions to run commands in
  the BLISS session
- refresh()/list_objects()/objects_by_type() are utility accessors over the
  cached objects

### DIFF
--- a/mxcubecore/HardwareObjects/BlissProxy.py
+++ b/mxcubecore/HardwareObjects/BlissProxy.py
@@ -1,0 +1,264 @@
+#  Project name: MXCuBE
+#  https://github.com/mxcube
+#
+#  This file is part of MXCuBE software.
+#
+#  MXCuBE is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


marcus-oscarsson:
Very nice, wonderful contribution :fireworks: :smiley:

marcus-oscarsson (approved):

PR #1665, opened 2026-09-09
Improve the phase handling for the mock diffractometer.

beteva:
Set the phase to COLLECT in the mock collect loop.

PR #1666, opened 2026-09-09
Fix for mockup phase and readiness problem (tested)

rhfogh:
This expands on PR #1665, expending the fix to more mockup classes and adding some additional fixes. It should replace #1665. It has been tested and works in mock mode with the GPhL CollectEmulator and GPhL workflows. Changes that are not simply expansion of @beteva''s fix in #1665 are commented. With one exception all changes are limited to mock mode.

### DIFF
--- a/mxcubecore/HardwareObjects/Gphl/CollectEmulator.py
+++ b/mxcubecore/HardwareObjects/Gphl/CollectEmulator.py
@@ -309,8 +309,13 @@ class CollectEmulator(CollectMockup):
                 resource.RLIMIT_STACK, (resource.RLIM_INFINITY, resource.RLIM_INFINITY)
             )
 
+        diffr = HWR.beamline.diffractometer
         try:
             self.emit("collectStarted", (None, 1))
+
+            diffr.set_phase(diffr.get_phase_enum.COLLECT)
+            diffr.update_state(diffr.STATES.BUSY)
+
             running_process = subprocess.Popen(
                 command_list, stdout=fp1, stderr=fp2, env=envs, preexec_fn=set_ulimit
             )
@@ -326,6 +331,7 @@ class CollectEmulator(CollectMockup):
             self.log.error("Error in GPhL collection emulation")
             raise
         finally:
+            diffr.update_state(diffr.STATES.READY)
             fp1.close()
         process = gphl_connection.collect_emulator_process
         gphl_connection.collect_emulator_process = None
--- a/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
+++ b/mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
@@ -100,6 +100,28 @@ class DiffractometerMockup(AbstractDiffractometer):
         )
         self.set_value_motors(motors_positions_dict)
 
+    def set_value_motors(
+        self,
+        motors_positions_dict: dict,
+        simultaneous: bool = True,
+        timeout: float | None = None,
+    ):
+        """ Wrap function to ensure diffractometer always ends up ready
+        NB the signals from the motors signalling 'ready' are absent in mock mode
+
+        Args:
+            motors_positions_dict:
+            simultaneous:
+            timeout:
+
+        Returns:
+
+        """
+        super().set_value_motors(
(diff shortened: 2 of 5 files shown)


mxcubecore/HardwareObjects/Gphl/CollectEmulator.py
```diff
@@ -309,8 +309,13 @@ def set_ulimit():
                 resource.RLIMIT_STACK, (resource.RLIM_INFINITY, resource.RLIM_INFINITY)
             )
 
+        diffr = HWR.beamline.diffractometer
         try:
             self.emit("collectStarted", (None, 1))
+
+            diffr.set_phase(diffr.get_phase_enum.COLLECT)
+            diffr.update_state(diffr.STATES.BUSY)
```
rhfogh, line 317: This class is GPhL-only. In addition to tehfix from #1665 it sets the diffractometer to be BUSY while acquiring. The same might be nice for non-GPhL mnockup mode, but that would have to be done by someone who can use and test it

mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
```diff
@@ -100,6 +100,28 @@ def move_motors(self, motors_positions_dict):
         )
         self.set_value_motors(motors_positions_dict)
 
+    def set_value_motors(
```
rhfogh, line 103: It turned out that in mock mode the diffractometer did not return to READY after set_value_motors, which blocked operation. As I understand this is done in real operation by signals from the diffractometer motors, that are not available in mock mode. With this change the diffractometer will always be ready after set_value_motors in any subclass of DiffractometerMockup

mxcubecore/queue_entry/sample_centring.py
```diff
@@ -108,6 +108,8 @@ def execute(self):
 
     def pre_execute(self):
         BaseQueueEntry.pre_execute(self)
+        diffr = HWR.beamline.diffractometer
```
rhfogh: This automatically puts you in CENTRE model when you start a sample centring. As the only change in this PR it is also effective in normal operation - and should be evaluated accordingly.

mxcubecore/queue_entry/sample_centring.py
```diff
@@ -108,6 +108,8 @@ def execute(self):
...
+        diffr.set_phase(diffr.get_phase_enum.CENTRE)
```
beteva: There are beamlines which do not use the centring phase to centre the crystal. We should find a way to customise this.

mxcubecore/HardwareObjects/mockup/DiffractometerMockup.py
beteva: Indeed the abstract class sets the state to busy and does update_state at the end of the movement, but there is nothing in the mockup to set it back to ready. Well spotted 👍

beteva (changes requested):
Well done @rhfogh. The only problem I have is setting to CENTRE phase in the sample_centring.py. I would suggest that you remove this file for time being, until we find a way to customise and I will merge the rest.

mxcubecore/queue_entry/sample_centring.py
rhfogh: I have removed it, and tested the result. It makes no difference to my mock run, except that the phase control button does not change from collect to centre during those centrings. We can leave it like this just fine.

mxcubecore/HardwareObjects/Gphl/CollectEmulator.py
beteva: The whole business with the diffractometer state comes from the fact that the MD accepts one command at a time. Other diffractometers may not need this. This is why it has been not treated in the abstract class, but left for the specific implementation to handle it. This is also why we have almost systematically the timeout parameter when it comes to do some action. Again it will be the specific implementation which would know whether actions ca be executed in parallel (the timeout set 0) or should wait for previous action to finish. The parallel execution is done when possible when looking for performance. As you need this in your mock class, you did the right implementation. Not sure other mock classes need it, but it can be discusses.

beteva (approved):

PR #1669, opened 2026-09-11
Make AbstractDiffractometer.get_phase_list operate on enum values.

mockoocy:
For now get_phase_list is used to send list of setable diffractometer phases. What exactly is sent now is member names from a DiffractometerPhase enum. The front-end is supposed to operate on the member names, rather than values. It also leads to them being displayed to the users. Member names should be internal representation and we should send values instead, e.g. BeamLocation rather than `SEE_BEAM`,

This includes an extension of `value_to_enum` method, so that the enum object can be coerced from a string value.
A small bug has been also fixed, where the assumed exception type there was a wrong one.

### DIFF
--- a/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
+++ b/mxcubecore/HardwareObjects/abstract/AbstractDiffractometer.py
@@ -504,9 +504,8 @@ class AbstractDiffractometer(HardwareObject):
         """Return a list of all the defined phases."""
         phase_list = []
         for member in DiffractometerPhase:
-            _nam = member.name
-            if _nam not in ["IN", "OUT", "UNKNOWN"]:
-                phase_list.append(_nam)
+            if member.name not in ["IN", "OUT", "UNKNOWN"]:
+                phase_list.append(member.value)
         return phase_list
 
     @property
@@ -526,7 +525,7 @@ class AbstractDiffractometer(HardwareObject):
             value = DiffractometerPhase(value)
         if self.current_phase != value:
             self.current_phase = value
-            self.emit("phaseChanged", (value.name,))
+            self.emit("phaseChanged", (value.value,))
 
     # -------- Constraints --------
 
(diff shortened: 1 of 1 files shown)


beteva:
> Member names should be internal representation and we should send values instead, e.g. BeamLocation rather than `SEE_BEAM`,

It is just the opposite. The values are hardware specific, They differ from the Arinax MD to Mini Diffractometer and some other models. That's why the Enum concept has been introduced - the name stays unchanged whatever the model of the diffractometer is. This is part if the API, where we guarantee that it is independent of the specific hardware implementation.
Our users did not fond it difficult to change to the new phase labels.

mockoocy:
Hi, we've got a complaint from our BL Staff about these, so I think it is something that is worth looking into.

This essentially changes what is sent between the client and mxcubecore. The value is then changed into a proper enum member in `value_to_enum`. The implementations can still rely on `DiffractometerPhase.SomePhase` and send whatever they want to the diffractometer. It does not change the fact that implementations using different diffractometer would still need to provide a mapping from the enum member names/values to some different phase names.

I suppose that the point is that we could rely too much on Arinax terminology? Then we could perhaps just add some kind of labels to be sent along the identifiers (`SEE_BEAM, COLLECT, ...`), e.g. "Beam Location", "Data Collection",  that would be then displayed in the UI rather than sending the enum values.

beteva:
> Hi, we've got a complaint from our BL Staff about these, so I think it is something that is worth looking into.

Our staff and users did not complain about it, but if it really is a big problem (not just question of habits), we can look into the names and change them to something we agree on. 

> This essentially changes what is sent between the client and mxcubecore.

No, the essintial change is what is displayed to the users. The values  can differ from one diffarctometer to another (not all diffarctometers have the MD strings as value), I think we should leave them specific and not display them to users.

The thing is that the look and feel should be the same in all the synchrotrons, so what is displayed  should be the same everywhere. Behind the phases is not always a command sent to Arinax MD - it can be a whole bunch of commands sent to a diffracometer (which is already the case for at least 2 non arinax diffarctometers)

mockoocy:
I've assumed that we will always have the same phase names and corresponding enum values since they seem hardcoded in the `AbstractDiffractometer`.
I suppose I will close this one and come back to this topic later. I don't think it is a big deal anyway, just an UI oddity.

beteva:
> I've assumed that we will always have the same phase names and corresponding enum values since they seem hardcoded in the `AbstractDiffractometer`.

The Enum values are changed, where needed. The AbstractDiffractometer contains just the most common set of values.

marcus-oscarsson:
Hm, I think that we should keep using the names of the Enum so that the phase is always called the same regardless of what diffractometer is used. I think that will make it easier for the user to understand what the phase means conceptually. Then I also think that we can look at the current naming and perhaps improve on what has been suggested.

Perhaps just simply listing the names that you expect is a good start ? We can then see how to go from there ?

PR #1671, opened 2026-09-11
Removed ISPyB from list of LIMS names

marcus-oscarsson:
We are no longer retrieving sample information from ISPyB so we have removed it from the list of available LIMS systems (at ESRF).


beteva (approved):
Thanks

PR #1673, opened 2026-09-11
Separated gathering/assembly and writing to file and uploading to ICAT

marcus-oscarsson:
This is a first attempt to separate the metadata gathering/assembly from upload the actual saving to disk or upload LIMS (in this case ICAT). The longer term goal would be to make this as LIMS agnostic as possible but its for the moment quite tightly coupled to the format used by ICAT.


mxcubecore/HardwareObjects/ESRF/SSXICATLIMS.py
```diff
@@ -42,83 +49,145 @@ def finalize_data_collection(self, parameters):
...
+            "Sample_name": path_parameters.prefix,
+            "scanNumber": 0,
+            "InstrumentMonochromator_wavelength": beamline_parameters.wavelength,
+            "chipModel": extra_lims_values.chip_model,
+            "monoStripe": extra_lims_values.mono_stripe,
+            "energyBandwidth": beamline_parameters.energy_bandwidth,
+            "detector_id": HWR.beamline.detector.get_property("detector_id"),
+            "experimentType": extra_lims_values.experiment_type,
+            "scanType": extra_lims_values.experiment_type,
+            "Experiment_name": path_parameters.experiment_name,
```
beteva, line 99: Some of these parameters can be retrieved using store_common_data or  gather common_metadata. Maybe it is worth using it here as well.

beteva (approved):
Good initiative to split gathering, writing and uploading. Apart from the data dictionary for the SSX, the rest is fine.

PR #1674, opened 2026-09-14
Added utility to fetch the configuration files from mxcube_configuration repo

### LINKED ISSUE #944: Configuration files setup

marcus-oscarsson:
Added utility to fetch the configuration files from a git repository for example `mxcube_configuration` repo https://github.com/mxcube/mxcube_configuration.

So basically running  `mxcubeweb-server -r "$(mxcube-fetch-config --for [DIR])" --static-folder ui/build` would get the configuration from the configured repository which by default is https://github.com/mxcube/mxcube_configuration. The later can be changed by setting the `MXCUBE_CONFIG_URL` environment variable. 

--for takes one or more directories, each given relative to the mxcube configuration checkout, and joins them into the lookup path in the order given

```
    mxcube-fetch-config --for demo.yaml
    mxcube-fetch-config --for [beamlinename]
    mxcube-fetch-config --for [beamlinename] [beamlinename]/gphl
```

Closes #944 

@rhfogh, better later than never ;). What do do you think ?


rhfogh:
@marcus-oscarsson  This looks excellent, and I want it,  but I am not quite sure how to work it (or, therefore, how I should test it).

First, are the contents of https://github.com/mxcube/mxcube_configuration, develop branch, currently up to date with the mock web version and/or the ESRF production version - so I can use that for testing?

Second, where would you actually use mxcube-fetch-config? That string is not to be found  in either mxcubeweb or mxcubecore. Is it used in running, or only in setup, for downloading the directory? And if downloading, which branch does it get you? 

Third, my current set-up has a permanent installation of mxcube_configuration on disk, which I update separately, and I run mxcubeweb as 
`mxcubeweb-server -r /home/rhfogh/pycharm/mxcube_configuration/demo.yaml/gphl/:/home/rhfogh/pycharm/mxcube_configuration/demo.yaml/mxcube-web/:/home/rhfogh/pycharm/mxcube_configuration/demo.yaml/ --static-folder /home/rhfogh/pycharm/mxcubeweb/mxcubeweb/ui/build/ -L debug`
For the same manner of working how should I set up my command line? 

Sorry for the confusion.

marcus-oscarsson:
For the configuration files in `mxcube_configuration` I made a PR with a few updates https://github.com/mxcube/mxcube_configuration/pull/3. 

The script takes a repo and a tag/branch so you can pass it both, thus I tested with my branch without it being merged. The script `mxcube-fetch-config` comes with `mxcubecore` and you should get it when you install `mxcubecore`. I usually install it with `pip install -e . --no-deps`. You probably wont get the utility script if you install it as a dependency of `mxcubeweb` or `mxcubeqt`. I need to have a second look at if we can find a better way to install it.

The utility `mxcube-fetch-config` is not meant to replace  whatever you pass after -r today. Its meant as a convenience utility to work with configuration files that are located in a git repository, like the ones in  `mxcube_configuration`. The only thing the script does is to checkout the branch/tag at the given repo and echo the paths passed (the paths on the local file system). By default the script will fetch what's in `mxcube_configuration` demo.yaml.

I thought this could be handy as most of us probably have the configuration files in a separate git repository, each beamline having its subfolder. It at the same time solved our other issue with having an easy way to get the common `mxcube_configuration` config files.

beteva:
Do we still maintain GenericDiffarctometer?

rhfogh:
Ideally I would like to get the AbstractDiffractometer upgrade fully integrated in Qt, just in case,  before dropping what was before. It should hopefully not take too long, but I need to get hold of Martin before I can do more there. But the other people on Qt might have stronger opinions (either way) on this than I have.

PR #1676, opened 2026-09-15
bump pyicat-plus to 3.0.0 stable release

bodinm:
https://pypi.org/project/pyicat-plus/ 

pyicat-plus 3.0.0 is now published as a stable release


bodinm:
ping @woutdenolf @marcus-oscarsson  @antolinos

woutdenolf:
Can we use `3.0.1` so we avoid the `pint` upper-bound? Thanks!

bodinm:
yes sorry, done!

marcus-oscarsson:
Could we please pin this with `==`, it quickly becomes quite confusing and very inconvenient to rollback changes otherwise. We had first time experience with this the other day with rc3, rc5, etc ...

woutdenolf:
> We had first time experience with this the other day with rc3, rc5, etc ...

We care about stable API, deprecation cycles and semantic versioning. Only in between RC releases we don't consider ourselves beholden to this. That's what RC is for.

Regardless you can pin of course, up to you. This means you won't be getting bug fix releases without modifying your pyproject.toml file. But maybe in your ops workflow that is perfectly fine.

woutdenolf:
Or maybe you were thinking more about patch updates only: 3.0.1 → 3.0.2, 3.0.3, ...

```
pyicat-plus = {version = "^3.0.1", optional = true}
```

bodinm:
exact pin done

marcus-oscarsson:
I think our last issue was there where some incompatibilities with the models the library provided and what was actually used in MXCuBE. In the mean time the `pyicat-plus` was updated and as it was pinned to use the latest version we got some issue there. Anyhow, patch version would be fine as long as thats respected. But for these things an exact pin perhaps even better. I assume/hope that we will be notified about any serious bugs that requires an update.

marcus-oscarsson (approved):

PR #1678, opened 2026-09-22
Moving the queue serilization from `mxcubeweb` to `mxcubecore`

marcus-oscarsson:
This PR moves the queue serilization from `mxcubeweb` to `mxcubecore` and formalizes the current JSON format used in the web client so that it can be re-used in a easier way.


rhfogh:
This is a very good initiative, and much appreciated. One question, though: How will it integrate with the existing mxcubecore queue structure, and thus, crucially, with the Qt branch? It looks like it might be called only from mxcubeweb for now (there is a mo-queue-refactor branch one notices).

If we can move over to a new, shared queue structure that would be great - but it might require a fair amount of work of clean-up and on the Qt side, and that ought to be taken into consideration up front. If we are proceeding with a web-only queue structure in parallel with a de facto qt-only queue structure we would be storing up cruft towards a future reckoning. One can see potential problems piling up as the web version moves towards a great future and the two versions need to be somehow reconciled. Maybe we need to talk about this?

beteva:
Great initiative.

marcus-oscarsson:
This is simply an attempt to formalize the JSON format that is used to exchange the queue between the web client and the backend/`mxcubecore`. The format was kept as it was with the addition of version handling. The queue structure is the same, a Queue with `QueueNodes` from `QueueModelObjects` and `QueueEntry` objects from the `queue_entry` submodule is constructed by the `QueueBuilder` from the JSON data.

Basically, there is no real structural difference between the queue used by `mxcubeweb` or `mxcubeqt`. The queue is executed by the `QueueManager` and a such a queue should be usable in both web and qt versions. There is maybe some subtle differences in the nesting of different nodes but I have not looked into the details on how exactly the qt queue is constructed.

Our hope is that this can be used a s "wire" format for exchanging the queue and that a future replacement of `XMLRPC` could use it. 

Yes we would need to purify/crystalize ;) this format a bit but we know that its working as it is today and the idea is to elaborate this format further.

rhfogh:
@marcus-oscarsson That all makes sense. And, as AB said, it is a great initiative. Presumably long-term we would want to replace the current queue/model structure with the JSON equivalent?? Maybe a first step would be to make code that could generate the JSON from the Queue (etc.) objects (or is it there already?), just like the QueueBuilder can generate the objects from the JSON?

Since this is merely porting already functional code from web to core, then 1) it should be OK as long as the web version keeps working. 2) I may not be the best person to evaluate it, at least until there is a parallel PR for mxcubeweb. But by all means go ahead.

marcus-oscarsson:
This currently converts in both directions from `mxcubecore` queue to JSON format, and from JSON format to `mxcubecore` queue.

rhfogh:
@marcus-oscarsson Cool!

